PR #1220, opened 2014-07-11
Improvements to FlxCamera#followLerp

MSGhero:
The elapsed's get cancelled in the numerator and denominator anyway. Probably doesn't jitter anymore when followLerp = 0, but I didn't check for it.

### DIFF
--- a/flixel/FlxCamera.hx
+++ b/flixel/FlxCamera.hx
@@ -60,11 +60,11 @@ class FlxCamera extends FlxBasic
 	 */
 	public var targetOffset(default, null):FlxPoint;
 	/**
-	 * Used to smoothly track the camera as it follows.
-	 * Valid values between 0.0 and 1.0.
-	 * A value of 1 means no camera easing. A value of 0 means the camera does not move.
+	 * Used to smoothly track the camera as it follows: The percent of the distance to the follow target the camera moves per 1/60 sec.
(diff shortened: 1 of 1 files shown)


xaedes:
I have previously done the math behind this very case: Lerp with variable time step.
The code before this merge was on the right way, but it was not quite complete. 

TLDR: The mathematical correct solution for your kind of lerp with variable timestep is:
`scroll += (scrollTarget - scroll) * (lerp * elapsedTime) / (lerp*elapsedTime - lerp*(1/60)+(1/60))`

Where does the (1/60) come from? 
A lerp like `scroll += (scrollTarget - scroll) * lerp` only makes sense with constant timestep. So there is some assumption about the constant timestep baked in the single lerp parameter. From the code I guess, it was meant to be at 60 FPS, hence (1/60).

For details see: https://xaedes.github.io/tech-gain-filter.html#h2_applications
It starts from an equation corresponding to `scroll += (scrollTarget - scroll) * lerp`.
Then this is transformed into a gain filter formulation. See website for what that is, basically just `x_new = y * gain + (1-gain) * x_old`.
Now the on the website derived adaption of the gain filter to variable timestep is used to compute `lerp` for a variable timestep. The result needs a reference `lerp` with a reference timestep as input. I think this should be the `lerp` in code and (1/60).

PR #1316, opened 2014-10-04
Added FlxSoundGroup for managing volume of music/sfx/speech etc separately, closes #362

### LINKED ISSUE #362: Split volume control into sfx and music
Like already suggested in [flixel-community](https://github.com/FlixelCommunity/flixel/issues/132), I think that the way flixel handles volume is too simple and that there should be two seperate volume values, `FlxG.volumeMusic` and `FlxG.volumeSFX` (or `FlxG.volumeSound`) instead of `FlxG.volume`.

While this would be a really simple change, I'm not sure what to do about the `soundTray`. It would be rather complex to allow it to manipulate both volumes individually. Plus, there would be more keys needed to make that happen (and the volume control already uses a lot), at least one to switch between the two volume bars. Of course, you could just make the `soundTray` control both volumes at once as well.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/1022039-split-volume-control-into-sfx-and-music?utm_campaign=plugin&utm_content=tracker%2F254189&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F254189&utm_medium=issues&utm_source=github).
</bountysource-plugin>

JoeCreates:
I snuck another small modification in here, too. Just making a for loop more efficient.

`FlxSound` now has a group. You can set a sound's group either by adding the sound to the group, setting the group of the sound, or by passing the name of a group to `play()`, `load()` or `playMusic()`.

Presently groups must be added to the `FlxSoundFrontend` by calling `addGroup`.

Example use:
```
// Create a group named "sfx"
FlxG.sound.addGroup("sfx");
// Play a sound effect
FlxG.sound.play("mysound", 1, false, "sfx");
// Set sfx volume
FlxG.sound.getGroup("sfx").volume = 0.5;
```


gamedevsam:
Any opposition to this change? It's been lingering for a while, seems like we should merge it and improve it as necessary.

Issue #1307, opened 2014-09-28
Rethinking the collision/separation system


danim1130:
Since there have been a few issues popping up with the collision system flixel uses (see #1301, #1270, #1078), and seeing how the next version of Flixel will be 4.0.0, this seems like the perfect time to rewrite the collision system. I would be glad to try my best in doing this after we have decided what the most optimal solution should be. Here's what I think now:

I think that the base of flixel shouldn't contain any separation logic whatsoever, only overlap checking. For that, I think integrating hxcollisions would be the best soulution. Haxepunk uses the same (or at least, it uses the same theorem), it's already well-written, and probably bug-free aswell. For the separation/physics, I think Phaser has an interesting, and maybe one of the best approach to this: they provide three different engine. One basic, something like flixel has now, one more advanced, with Circle hitboxes included, and one full-blown physics engine. Following this, we could have two different engines for flixel: one basic, something like we have now, and one with nape. For them to work with flixel, we could either provide an interface (hard to do, but more OOP, easier to integrate custom engines after), have different Flx classes (easier, but more classes, like FlxState and FlxNapeState, FlxSprite and FlxNapeSprite, what happens if we add a NapeSprite to a basic state?), or rewrite the classes so that their functionality depends on what engine is currently in use (so if you add an FlxSprite to a state, either an FlxBasicSprite, or an FlxNapeSprite gets added, depending on the engine the state uses). Nape integration is somewhat done, and the basic engine could use the same separation flixel has now, but it would have continuous collision detection (you wouldn't have to call FlxG.collide()), and masks. I personally like the third option, but the first could be manageable as well.

So what do you guys think?
##

gamedevsam:
I wanted to contribute something related to this discussion, and other physics related issues. Recently, the development team of Lazer (Phaser's successor) have converted their physics engine to be data based (all object data is contained in contiguous typed arrays, with getters and setters on object references). You can see the beginnings of their refactor on this commit: photonstorm/lazer@219b17d

They created a benchmark comparing the two approaches (object-oriented, vs data-oriented) with dramatic results: http://perftestlazer.azurewebsites.net/

I believe the core physics engine of HaxeFlixel should be as fast and light as possible (since Nape already provides a more complex and feature filled alternative), and any major revisions to the physics API / implementation should move towards a data-oriented approach.

PR #1323, opened 2014-10-05
Added polygon and circle hitboxes, raycasting.

danim1130:
As discussed in #1307, I have added a much more capable overlap checking system. Currently, I haven't changed anything in the base code, so there shouldn't be any compatibility issue (maybe if someone has extended FlxSprite, and defined a hitbox variable for it, but that's it). There are still a few things I want to add (like handling concave polygons better), but don't know when I can get around it. Also, there are a few things that aren't in use as of now. They will be used when I refractor the collision system, when I remove the old overlap system and integrate this one. For this, I would like it if there would be a separate branch for this on Github (altough I'm not exactly sure how this works, who creates it etc.).

The features:
-Added circle and convex polygon hitboxes (present in hxCollision).
-Added FlxHitboxList, which can be used to bundle multiple hitboxes into a bigger hitbox (not present in hxCollision)
-Raycasting (present in hxCollision, but it was buggy, so I had to do some tinkering)
-Added an overlap function so that using this overlap is almost the same as using FlxG.overlap.

How to use:
-Add a hitbox to a sprite (every parameter for the hitbox is in the sprites local coordinate system).
-Call flixel.hitbox.FlxOverlap.overlap() with the sprite.
-Use flixel.hitbox.FlxOverlap.castRay() to cast a ray against a sprite or an FlxGroup.

Here is a demo of it: http://www.fastswf.com/z77dAng
And the source code : http://pastebin.com/bRYY8VP3

### DIFF
--- a/flixel/FlxCamera.hx
+++ b/flixel/FlxCamera.hx
@@ -8,14 +8,14 @@ import flash.geom.ColorTransform;
 import flash.geom.Point;
 import flash.geom.Rectangle;
 import flixel.FlxCamera.FlxCameraShakeDirection;
+import flixel.graphics.FlxGraphic;
+import flixel.graphics.tile.FlxDrawStackItem;
+import flixel.graphics.tile.FlxTilesheet;
 import flixel.math.FlxMath;
 import flixel.math.FlxPoint;
 import flixel.math.FlxRect;
-import flixel.system.layer.DrawStackItem;
-import flixel.system.layer.TileSheetExt;
 import flixel.util.FlxColor;
 import flixel.util.FlxDestroyUtil;
-import flixel.util.loaders.CachedGraphics;
 import openfl.display.Tilesheet;
 
 /**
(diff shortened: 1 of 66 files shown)


gamedevsam:
This is a good change, please make the PR to the correct branch so it can be merged and battle tested!

qntf:
@gamedevsam @Gama11  I know this is over a year later than this conversation, but what happened to this in the end? As far as I can see, it wasn't merged, though it looks like a great change. Any chance that someone else can submit the patch in his place? To me it seems unproductive to just wait for the original author if he hasn't been available for ~2.5 years.

danim1130:
I'm still here, sorry that I haven't done anything about this PR.

The reason why I didn't make the PR was for multiple reasons, which I will
summarize here.

This code was written when I was only starting to learn more seriously
about programming. As a result, this PR is probably ridden with bugs and
performance issues (part by the fact that it is 3-4 years old), and don't
really think it should be merged.

The other reason is this was part one of a bigger change list, which
would've refactored the whole physics part of HF. The current system
doesn't really separate the concerns of each object, where IMHO FlxSprite
has became the ultimate god object, containing code for drawing, velocity
and acceleration computing, and a few other things. I wanted these things
to become separated, which would've allowed a much more flexible system
with less bugs. ALAS this proved to be too big of a refactor, where
basically any class thet can be added to a state had to be reworked (like
tilemaps as well). Also, there wasn't a clear support about this idea among
the core developers, and I didn't want to waste my time on something that
might be rejected ultimately because of back-compatibility issues.

Also, the code presented here is basically a copy and paste of one of the
open source haxe project, which might not be ultimately legal to do (I'm
not sure about the licenses).

Ultimately, I think I will be closing this PR. Maybe this could be reworked
as an flx extension, but ultimately, I think the right thing to do would be
to refractor the base of Flixel and separate physics from graphics. I would
gladly work on this, but only if everyone on the core group could reach a
mutual agreement about how the system should look like.

Sorry for bringing this a little off-topic.


On Jun 22, 2017 17:12, "qntf" <notifications@github.com> wrote:

@gamedevsam <https://github.com/gamedevsam> @Gama11
<https://github.com/gama11> I know this is over a year later than this
conversation, but what happened to this in the end? As far as I can see, it
wasn't merged, though it looks like a great change. Any chance that someone
else can submit the patch in his place? To me it seems unproductive to just
wait for the original author if he hasn't been available for ~2.5 years.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/HaxeFlixel/flixel/pull/1323#issuecomment-310410205>,
or mute
the thread

PR #1414, opened 2015-01-04
Allow TexturePacker's JSON-Hash

MSGhero:
Export options in TP for json include JSON-Array and JSON-Hash, with hash being more useful imo.

No breaking changes.

### DIFF
--- a/flixel/graphics/frames/FlxAtlasFrames.hx
+++ b/flixel/graphics/frames/FlxAtlasFrames.hx
@@ -53,36 +53,53 @@ class FlxAtlasFrames extends FlxFramesCollection
 		
 		var data:Dynamic = Json.parse(Description);
 		
-		var rotated:Bool;
-		var name:String;
-		var sourceSize:FlxPoint;
-		var offset:FlxPoint;
-		var angle:FlxFrameAngle;
-		var frameRect:FlxRect;
-		
-		for (frame in Lambda.array(data.frames))
+		// JSON-Array
+		if (Std.is(data.frames, Array))
 		{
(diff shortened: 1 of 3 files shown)


flixel/graphics/frames/FlxAtlasFrames.hx
```diff
@@ -53,39 +53,56 @@ class FlxAtlasFrames extends FlxFramesCollection
...
-				angle = FlxFrameAngle.ANGLE_NEG_90;
+				texturePackerHelper(frame.filename, frame, frames);
 			}
-			else
+		}
+		
+		// JSON-Hash
+		else
+		{
+			for (frameName in Reflect.fields(data.frames))
```
Gama11: Is Reflection necessary here? I.e., couldn't we parse this with a typedef?

Gama11:
Would be nice to have a small unit test for this, espeically since you probably have some test for this already anyway.

flixel/graphics/frames/FlxAtlasFrames.hx
Gama11: On second thought, I guess we're already working with Dynamic anyway... (which isn't ideal to begin with).

MSGhero:
Idk what the deal is with the test package, but it's not recognizing the files I put into the assets folder even if I add an assets path in project.xml. I got around the json files by pasting their data directly, and I just set the bitmapdata to a non-null one so it doesn't error from being null.

Gama11:
Yeah, there are several issues with using assets in the unit tests... I usually used workarounds similar to yours.

Thanks for the test!

PR #1420, opened 2015-01-15
Fix iteration bug in FlxSignal

kevinresol:
There would be array iteration bug if the listener is removed while the signal is processing the array of listeners inside `dispatch()` (no matter the remove is invoked by `dispatchOnce` property or by user code through the `remove()` function)

### DIFF
--- a/flixel/util/FlxSignal.hx
+++ b/flixel/util/FlxSignal.hx
@@ -277,12 +277,18 @@ private class Macro
 	{
 		return macro
 		{ 
+			var pendingRemove = [];
 			for (handler in _handlers)
 			{
 				handler.listener($a{exprs});
 				
 				if (handler.dispatchOnce)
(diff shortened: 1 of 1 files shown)


Gama11:
What exactly is the bug / got an example that reproduces it?

kevinresol:
In short, the bug appears when you `remove()` a listener while the signal is `dispatch()ing`. It is just the classic "remove item while iterating" bug.

```haxe
var signal = new FlxSignal();
var listener1 = function() trace("L1");
var listener2 = function() trace("L2");
signal.addOnce(listener1);
signal.add(listener2);
signal.dispatch(); // trace: L1 (L1 triggered but not L2 due to the bug)
signal.dispatch(); // trace: L2 (only trigger L2 as L1 already removed)
```

PR #1437, opened 2015-02-09
Added overlap condition for button touch release. 

jefvel:
Previously it fired a click event if you had touched it, and released outside of it. Not anymore


Gama11:
Thanks!

PR #1441, opened 2015-02-12
Killing an FlxObject shouldn't set exists to false

danim1130:


Gama11:
This seems like a very bad idea, *a lot* of code depends on this behavior. Why do you think this is necessary?

danim1130:
Yeah, I thought exists would show if destroy() has been called on the object or not. I was thinking that if for example I add an object to a group, and then later destroy that group, but still keep a reference to that object, I couldn't decide if the object is alive or not, and so if it still has it's resources allocated.

Gama11:
You don't really need to destroy groups (or any added objects for that matter) manually anyway, as it happens automatically on state switches.

PR #1445, opened 2015-02-25
Don't automatically install dependencies

steverichey:
Haxelib's dependency management is problematic. Even if I have a particular version of OpenFL and/or Lime installed, installing Flixel from Haxelib automatically starts installing these libraries, which can break often fragile Haxelib configurations.

Speaking as someone who has had to cancel these installations many times, I recommend that Flixel not force users to download them.


goshki:
Great point, should be also changed in flixel-demos, flixel-addons and filxel-ui (as I have recently experienced on my haxelib).

Gama11:
Admittedly, this can be kind of annoying, especially considering the

PR #1448, opened 2015-03-02
Fix behavior of fadeIn and fadeOut methods

Tiago-Ling:
`fadeIn` and `fadeOut` do not play and stop the music - it is only logical that when the user calls `fadeIn` it is expected for the sound to start playing. 

The same applies to `fadeOut`: when the fading tween completes the sound should call stop instead of continue playing with the volume down.

### DIFF
--- a/flixel/system/FlxSound.hx
+++ b/flixel/system/FlxSound.hx
@@ -440,7 +440,7 @@ class FlxSound extends FlxBasic
 	 * @param	Duration	The amount of time the fade-out operation should take.
 	 * @param	To			The volume to tween to, 0 by default.
 	 */
-	public inline function fadeOut(Duration:Float = 1, ?To:Float = 0, ?onComplete:Void->Void = null):FlxSound
+	public inline function fadeOut(Duration:Float = 1, ?To:Float = 0, ?onComplete:FlxTween->Void = null):FlxSound
 	{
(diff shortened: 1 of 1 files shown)


Tiago-Ling:
I think either one would be a good choice. Letting the user pass a callback would give him/her more control, while using a boolean would make it simpler. If a callback is preferred i think it makes sense to let the user pass one also in `fadeIn`.

What do you think @Beeblerox, @gamedevsam, @Gama11, @larsiusprime?

flixel/system/FlxSound.hx
```diff
@@ -440,9 +440,13 @@ class FlxSound extends FlxBasic
 	 * @param	Duration	The amount of time the fade-out operation should take.
 	 * @param	To			The volume to tween to, 0 by default.
 	 */
-	public inline function fadeOut(Duration:Float = 1, ?To:Float = 0):FlxSound
+	public inline function fadeOut(Duration:Float = 1, ?To:Float = 0, ?onComplete = null):FlxSound
 	{
-		FlxTween.num(volume, To, Duration, null, volumeTween);
+		if (onComplete != null)
```
Beeblerox: why add this check? You can replace these 4 lines with just one:
```
FlxTween.num(volume, To, Duration, onComplete, volumeTween);
```

Tiago-Ling:
There, i changed it.

PR #1457, opened 2015-03-19
Fixes for openfl-next

larsiusprime:


include.xml
```diff
@@ -4,7 +4,7 @@
 	
 	<assets path="assets/sounds" include="*.ogg" if="html5" unless="unit-test" />
 	<assets path="assets/sounds" include="*.mp3" if="flash" embed="true" />
-	<assets path="assets/sounds" include="*.ogg" unless="flash || html5" embed="true" />
```
Gama11: Was this causing issues / does lime legacy not support embedding .oggs?

If so, we should at least embed them when `lime_legacy` is defined.

flixel/system/debug/BitmapLog.hx
```diff
@@ -272,7 +272,10 @@ class BitmapLog extends Window
 			_entries[i] = null;
 		}
 		_entries = [];
-		_canvas.fillRect(_canvas.rect, FlxColor.TRANSPARENT);
+		if (_canvas != null && _canvas.rect != null)
```
Gama11: Under what circumstances did this cause issues?

Gama11:
@larsiusprime Are these changes still relevant?

Gama11:
Closed due to lack of followup.

larsiusprime:
Hey sorry about the delay on this one; I did this to get next up and running right around when the transition started, but if you've gotten next to run without it then it shouldn't be an issue.

PR #1463, opened 2015-03-31
Collision refactor - First step in separating physics

danim1130:
So this is the first big update about collision refactoring. This update shouldn't break anything written in flixel, since it doesn't remove anything from the base code. The main features added are the generalization of physics and the new Classic collision engine (heavily based on the old FlxG.collide function). Note that this is still a very incomplete version, but it should give you an idea about how I imagined this.

(Note : I'm also including the CollisionAndGrouping demo rewritten for this update)
https://dl.dropboxusercontent.com/u/6870579/CollisionAndGrouping.rar

The concept:
=====================
Everything movement, collision etc. related things are handled by spaces (every space has a common interface called IFlxSpace). A space operates on a group of bodies (which have a common interface IFlxBody). The body is the physical appearance of an FlxObject. Every FlxObject can only have one body, which can only be assigned to one space. Spaces can be added to FlxState.spaces, meaning that there could be multiple spaces working parallel in a single state.

How to use:
=====================
The first step is to create a space (currently there's only FlxClassicSpace and FlxNapeSpace), and add it to FlxState.spaces. After that, you can create a body for an object with space.createBody(object), which will automatically create the right body (either FlxClassicBody or FlxNapeBody), assign the space to the created body, assign the body to the object, and return the body. This body has most of the property used in the earlier version (like velocity, acceleration, mass, elasticity etc.)
Collision filtering is very similar to Nape : two object only collide if object1.collisionGroup & object2.collisionMask != 0, and object2.collisionGroup & object1.collisionMask != 0.
For the classic system, function callbacks are handled with the FlxClassicBody.overlap property. If two object collide, and they both have an overlap function set, then both of them will be called.

Progress:
=====================
FlxClassicSpace is somewhat completed, though there might be some features that I missed in implementing. The separation algorithm haven't been changed, only the overlap finding process. According to my tests, under the same circumstances, FlxClassicSpace was a little faster than FlxG.collide. Also, every callback function is now only called once. (With FlxG.collide, the callback functions were called multiple times for the same collision.)
FlxNapeSpace is in a very basic state. My main focus for the first release was the classic space, but it is in a workable state.
One of the biggest problems still is FlxTilemaps. I'm still not sure how to make them work with the new structure. I think I'll leave this to the very end, when everything else is working as it's supposed to be.

Benefits:
======================
One of my main goal for me is to make switching engines as easy as possible. I have somewhat achieved this with this update. For example, in the provided demo, all you have to do to use Nape is change var space = new FlxClassicSpace() to new FlxNapeSpace().

Also, currently there's no documentation. I hope that the variable names will help you understand what they do, and i will write a complete documentation by the time everything is finished.

Most of the time I'm online at slack, feel free to PM me there if something's not clear.


MSGhero:
1) The commits omg
2) You're adding bodies and spaces to Flixel when those already exist in Nape, but why? High level, it sounds like you're rewriting Nape to fit inside Flixel but also letting Flixel use Nape stuff. Why not just use Nape?

Gama11:
Yeah, the diff isn't very useful here as he had to merge 400ish commits from dev into the collision-refactor branch I guess.

danim1130:
@Gama11 Sorry, should I just delete this pull, make the pull from dev, and then do this pull?

@MSGhero My goal was to abstract everything movement and physics related from FlxObject into it's own separate class, and I thought Nape's structure would be a good fit. I also wrote a new Space around FlxG.collide. I'm open to suggestions though about how should we do the physics/collison abstraction.

Gama11:
@danim1130 Might be cleaner, but I guess we can [manually compare flixel dev against your branch](https://github.com/HaxeFlixel/flixel/compare/dev...danim1130:collision-refactor).

Why did you add the CollisionAndGrouping demo? That doesn't belong in this repo.

danim1130:
@Gama11 Sorry, I meant to upload the demo separately, but I forgot to remove that folder. (It was easier for me when I was moving to have them in the same repo). Going to remove it now.

gamedevsam:
Prize money is available as a reward for the integration of physics refactoring into the dev branch. It seems this pull request is out of date, please close this PR and restart the review process.

Please read over the (Coding Style)[http://haxeflixel.com/documentation/code-style/] once more, and double check your code conforms to it. We won't merge a pull request that does not adhere to the coding style.

Thanks for your contributions, every PR is deeply appreciated, every core developer of started out making and revising their first Pull Request. And don't forget, I reward good work with $$$.

PR #1465, opened 2015-04-06
Add pitch support for FlxSound

kevinresol:
this is a hidden gem...:smirk:
reference: http://community.openfl.org/t/change-pitch-or-speed-of-audio/1118/2

### DIFF
--- a/flixel/system/FlxSound.hx
+++ b/flixel/system/FlxSound.hx
@@ -78,10 +78,12 @@ class FlxSound extends FlxBasic
 	 * Set volume to a value between 0 and 1 to change how this sound is.
 	 */
 	public var volume(get, set):Float;
+	#if (cpp || neko)
 	/**
 	 * Set pitch, which also alters the playback speed. Normal is 1.
 	 */
 	public var pitch(get, set):Float;
+	#end
 	/**
 	 * The position in runtime of the music playback.
 	 */
@@ -111,10 +113,12 @@ class FlxSound extends FlxBasic
 	 * Internal tracker for volume.
 	 */
 	private var _volume:Float;
+	#if (cpp || neko)
 	/**
 	 * Internal tracker for pitch.
 	 */
 	private var _pitch:Float = 1.0;
+	#end
 	/**
 	 * Internal tracker for total volume adjustment.
 	 */
@@ -641,6 +645,7 @@ class FlxSound extends FlxBasic
 	}
 	
 	
+	#if (cpp || neko)
 	private inline function get_pitch():Float
 	{
 		return _pitch;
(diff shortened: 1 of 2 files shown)


Gama11:
In what OpenFL version was this added (i.e., would we be dropping support for any older OpenFL version that's still working fine with flixel by adding this without any openfl version conditionals)?.

What about openfl next? In that thread Joshua seems to suggest that this is only a thing for openfl legacy.

kevinresol:
from [this commit](https://github.com/openfl/openfl/pull/322)
we can add the legacy compiler flag maybe

flixel/system/FlxSound.hx
```diff
@@ -629,6 +640,23 @@ class FlxSound extends FlxBasic
...
+	{
+		return _pitch;
+	}
+	
+	private function set_pitch(v:Float):Float
+	{
+		#if (cpp || neko)
+		_channel.pitch = v;
+		#else
+		FlxG.log.warn("Pitch is only supported on OpenAL targets (i.e. cpp & neko)");
```
Gama11: Wondering if it wouldn't be better to only have this property exist  for cpp and neko in the first place. That's how we handle most of the input stuff in `FlxG` for example. Then again. other features like post processing use this approach (except for the warn() call), so I guess we're not really consistent about it...

flixel/system/FlxSound.hx
```diff
@@ -79,6 +79,10 @@ class FlxSound extends FlxBasic
 	 */
 	public var volume(get, set):Float;
 	/**
+	 * Set pitch, which also alters the playback speed. Normal is 1.
+	 */
+	public var pitch(get, set):Float;
```
Gama11: "Default" would probably describe this better than "Normal".

flixel/system/FlxSound.hx
```diff
@@ -629,6 +644,20 @@ class FlxSound extends FlxBasic
...
+	
+	#if ((cpp || neko) && openfl_legacy)
+	private inline function get_pitch():Float
+	{
+		return _pitch;
+	}
+	
+	private function set_pitch(v:Float):Float
+	{
+		return _pitch = _channel.pitch = v;
```
Gama11: This can produce null exceptions since `_channel` might be null.

Gama11:
GitHub doesn't send notifications for new commits on pull requests, so it's a good idea to comment whenever you add one.

I just manually merged this: 415f410ebc9c55 Thanks for the implementation, this is a really cool feature!

kevinresol:
@Gama11 roger that, will comment next time.

ashes999:
Sorry to necro. @Gama11 is this in a released version of HaxeFlixel? I didn't see any mention of pitch when I looked at the api.haxeflixel.org entry for FlxSound.

Gama11:
Yes. The api docs are currently only generated for flash.

ashes999:
Thanks for clarifying. (We should probably have an issue to address the docs ...)

Gama11:
We do. https://github.com/HaxeFlixel/flixel-docs/issues/132

PR #1474, opened 2015-04-14
FlxText: adding proper FlxTextAlign <----> TextFormatAlign conversion fu...

larsiusprime:
...nctions that work for both legacy and next

### DIFF
--- a/flixel/text/FlxText.hx
+++ b/flixel/text/FlxText.hx
@@ -639,12 +639,12 @@ class FlxText extends FlxSprite
 	
 	private inline function get_alignment():FlxTextAlign
 	{
-		return cast(_defaultFormat.align, String);
+		return FlxTextAlign.fromOpenFL(_defaultFormat.align);
 	}
 	
 	private function set_alignment(Alignment:FlxTextAlign):FlxTextAlign
 	{
-		_defaultFormat.align = convertTextAlignmentFromString(Alignment);
+		_defaultFormat.align = FlxTextAlign.toOpenFL(Alignment);
 		updateDefaultFormat();
 		return Alignment;
 	}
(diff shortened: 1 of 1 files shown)


larsiusprime:
Note this depends on this PR as well:
https://github.com/HaxeFlixel/flixel/pull/1474

flixel/text/FlxText.hx
```diff
@@ -1066,4 +1048,32 @@ abstract FlxTextAlign(String) from String
...
 	var RIGHT = "right";
 	var JUSTIFY = "justify";
+	
+	#if !openfl_legacy
+	public static function fromOpenFL(tf:TextFormatAlign):FlxTextAlign
+	#else
+	public static function fromOpenFL(tf:String):FlxTextAlign
+	#end
+	{
+		//ifs instead of a switch b/c you can only switch on lower-case variable names and TextFormatAlign are all caps constants in legacy
```
Gama11: !?

larsiusprime: See here:
https://github.com/HaxeFlixel/flixel-ui/issues/99#issuecomment-92924244

I tried it initially with a switch function, but when compiling in legacy mode it wouldn't work.

MSGhero:
I was wondering when this issue would come up in flixel. I don't like that openfl uses a straight enum, as even haxe's flash.text.TFA is an (extern) abstract enum. But oh well, nice change.

larsiusprime:
So what's the consensus? Pull this in or convince Joshua to change TextFormatAlign to an abstract string enum?

Beeblerox:
let's merge it

flixel/text/FlxText.hx
Gama11: Define "trouble" (error message?), seems to be working here.

larsiusprime: Hmm, you're right, now it's suddenly working again. 

The error *was* "Capture variables must be lower-case"

PR #1477, opened 2015-04-15
Method to snap camera to target

JoeCreates:
Added a method to snap the camera to the current target. I also made updateFollow public, seeing as this is useful when trying to have some special custom following behavior without subclassing FlxCamera.


Gama11:
What's the use case for `snapToTarget()`?

JoeCreates:
Its for when you have a lerp, but want to jump straight to the sprite the camera is following. If you have a lerp and spawn a player at 500, 500, for example, at your camera is centered on 0, 0, you don't necessarily want to see the camera pan over to the player.

Workarounds are to temporarily set the lerp to 1 then call update() on the camera (or just updateFollow() with this PR), or manually center the scroll on the player.

Gama11:
I see, makes sense. I included that use case in the docs to make it clearer when the function is useful: 10544644f45f7e30a6e126c91d7f36cf0077caee

I cherry picked the relevant commit here (no need to include those 4 merge commits): 1455a47f5256008be5314538d1c324f5d412db59

sruloart:
Shouldn't it be inlined like focusOn()?

Gama11:
I guess you could inline it, then again, this is hardly a function that's gonna be called often.

sruloart:
I'm picky because of the coding convention, not really because of the benefits...

JoeCreates:
@Gama11 Wouldn't git have ignored those when merging into the branch that they were merges from, anyway?

Gama11:
@JoeCreates Good question, not sure...

PR #1478, opened 2015-04-21
Compile fix for Flash

larsiusprime:


Gama11:
Are we sure it's safe to cast here?

I think the underlying issue here are the `equal()` and `notEqual()` operators in `FlxColor`, which are being excluded in haxe 3.2 via conditionals because they didn't compile anymore.

Also, what's with all those commits?

larsiusprime:
The issue is a Uint / Int thing. One of the parameters is being treated as a Uint in flash and the compiler complains.

As for the commits -- whenever my local repo and flixel get out of synch at all, I do my best to rebase/merge, etc, but in the end it just winds up making things worse. 

Is there a way to get a "clean history" without having to clone the repo from scratch?

Gama11:
I typically just create a fresh branch from the main branch (dev in this case) whenever I open a pull request. Rebase would also work I think. Anyway, there's definitely no need to clone the repo again, you can always Checkout -> New Branch -> origin/dev (assuming you cloned it from the official repo / that's your origin). Might need to rename the local branch to something else since you already have a "dev".

larsiusprime:
I mean at that point do I just have to kill the dev branch if it gets out of whack?

Anyways, I'm not actually executing this code; it's just preventing me from compiling, so I guess I'm not 100% sure how safe it is. What's the plan for moving forward?

Gama11:
I just cherry-picked the relevant commit: https://github.com/HaxeFlixel/flixel/commit/2ffbf456fd492a764ccdc065972857436f043538

I guess we'll notice if this causes any issues, for now compiling is better than not compiling (this is only an issue on Haxe 3.2 btw).

larsiusprime:
Ah, cool, works for me! Thanks for the tips on PR's, I'll see if I can get cleaner history in the future. Any idea why haxe 3.2 was no longer compiling those equal/notEqual operators?

Gama11:
Something about "recurisve operator method" - IIRC, the operator methods were originally added to deal with a Haxe bug related to `Null<FlxColor>` (@mrcdk / #1224), which should be fixed now. My guess is that this fix also leads to the compiler error.

PR #1479, opened 2015-04-21
Set the label text (through FlxButton).

LaughingLeader:
Should be pretty self-explanatory.


Gama11:
Thanks - guess this only worked once...

Issue #923, opened 2014-03-06
Consider a different implemenation of origin for sprites


Gama11:
The issue is described in this pull request: #365

@Beeblerox
##

JoeCreates:
Note frame width and frame height become wrong when you start scaling. Everything with hitboxes/origin breaks when you start scaling. :P I've wasted hours trying to get things to work with the current implementation. It's not even that it's unintuitive, it's that when you start scaling, things simply aren't what they say they are any more. For example, the offsets are measured from where the top left of the graphic *would* be had it not been scaled. I'm currently having a nightmare with FlxNapeSprites, because apparently the origin is also not behaving properly when I've scaled.

Tiago-Ling:
Also, anything you get from a `FlxAtlasFrames` comes already with non-default origin. I'm using this code to load a frame from the atlas into a `FlxSprite`:
```haxe
sprite = new FlxSprite(pos.x, pos.y);
sprite.frame = atlas.getByName('sprite_frame.png');
```
Trying to rotate an image loaded like this will always result in different behavior than using rotation in a sprite loaded through `loadGraphic`.

JoeCreates:
Any advances on this one, lately? It's quite a blocking issue for me, especially when it comes to using nape, too.

Beeblerox:
@JoeCreates many of the people could be upset when this change will be done. 
i know how to do it, but will wait for some time before next big release.
In the meantime I can only offer to make changes in FlxNape classes, so they will support scaled sprites aswell

larsiusprime:
Right now I'm constantly doing updateHitBox() every time I scale a sprite and that seems to work, though it is annoying.

Breaking changes are fine for the next big release, I would just like to know what the side effects will be so I can plan for them.

Beeblerox:
@larsiusprime Exactly this change will be resulted in different rendering position of sprites
but i'm planning to make additional properties, such as `top`, `bottom`, `left` and `right`, which will help user with posiotioning sprites. and user will use say `left` property (in the future version) instead of `x` (in current version) to change sprite's position. (but `x` and `y` will point to sprite's origin)

larsiusprime:
oh that's a good change I think. Breaking for sure, but if it's for a major point release and there's a simple and clear upgrade guide then I say do it.

MSGhero:
There should be a tshirt that says "I <3 breaking changes" but the heart is the github octocat

Beeblerox:
[...]

PR #1482, opened 2015-04-27
Add FlxObject#setPositionUsingCenter()

Aleph0Null:
Added setPositionUsingCenter(), which is more convenient than setPosition() for some situations.

This is my first ever pull request. If this is ok, I have some ideas about introducing 'pivot' points on FlxObjects which generalize the origin and can be used for e.g. setting position, rotating around, connecting with other pivot points.

### DIFF
--- a/flixel/FlxObject.hx
+++ b/flixel/FlxObject.hx
@@ -857,6 +857,19 @@ class FlxObject extends FlxBasic
 	}
 	
 	/**
+	 * Helper function to set the coordinates of this object 
+	 * using the center instead of the top-left corner.
+	 * 
+	 * @param	X	The new x position of the center
+	 * @param	Y	The new y position of the center
+	 */	
+	public function setPositionUsingCenter(X:Float = 0, Y:Float = 0):Void
+	{
+		x = X - width / 2;
+		y = Y - height / 2;
+	}
+	
+	/**
 	 * Shortcut for setting both width and Height.
 	 * 
 	 * @param	Width	The new sprite width.


Gama11:
Hm... I see how this can be useful, but it seems like a bit of a workaround for a proper pivot point / origin.

What do you think @Beeblerox? You had some plans to refactor origin right?

sruloart:
I think https://github.com/HaxeFlixel/flixel/issues/923 these issues should be answered by something like this proposal: https://github.com/GoodBoyDigital/pixi.js/issues/997

Everything is relative :)

Tiago-Ling:
+1000 to using the exact same properties that @sruloart linked! Having `origin`, `pivot` and `anchor` working as described would be awesome!

larsiusprime:
Yeah I agree with @Tiago-Ling  and @sruloart, that would be amazing.

Aleph0Null:
Regarding pivots and anchors, I was thinking more generally in fact. I wanted to be able to designate a set of named 'reference points' (Pivot may not be the best choice of name) on sprites (FlxObjects). Not just two for positioning and one for rotation, but any number the user wishes. You could then do stuff like:

setCurrentPivot(name:String)    //should be used when no pivot is specified. Maybe have two 'defaults': setCurrentPositioningPivot(), setCurrentRotationPivot()
addPivot(name:String, x:Float, y:Float)
addRelativePivot(name:String, referencePivots:Array<String>, weights:Array<Int>, offsetX:Float, offsetY:Float)  //adds a pivot at a weighted barycenter of other pivots. Useful for e.g. defining midpoints
setPositionUsingPivot(?name:String, x:Float, y:Float)
rotateAroundPivot(?name:String, angleChange:Float)
positionOnPivot(?name:String, target:FlxObject, ?targetPivot:String)
linkToPivot(?name:String, target:FlxObject, ?targetPivot:String)   //this not just changes the current position of the sprite, but should also preserve relative position and angle during future frames, until unlinked

When constructed, FlxObjects should come with default pivots "Center", "Top Left", "Bottom Left", "Top Right", "Bottom Right" (and maybe suitable aliases, and/or make them case insensitive.) These should change accordingly when the sprite is resized. User-defined pivots can be defined relative to these pivots.

The screen could also support pivots, so you can do things like anchor UI elements to window corners that stay at the corners when the window is resized.

-----

What bothers me is that setPosition() uses top left corner, which is less natural to me than using center). In particular, I have sets of sprites which share the same centers but have different top left corners. I'm currently using setPosition(...) { setPositionUsingCenter(...); }, setPositionUsingCenter(), setPositionUsingTopLeft(), in my local copy, thought I'd share the simplest addition that doesn't break stuff for others and see how it goes.

PR #1483, opened 2015-04-27
Micro-fix to method description

Tiago-Ling:
For the border style to work, at least in Windows and Neko the user has to pass a color in the ARGB format instead of what is in the method description (RGB).


Tiago-Ling:
I thought it was possible to select especific commits to create a pull request - but it appears that all commits get included. Damn.

sruloart:
I think you need to create a special branch for each one, but maybe with Source Tree it is possible, never tried it though.

Issue #1490, opened 2015-05-01
BitmapData.colorTransform() acts simultaneously on different FlxSprites with the same BitmapData pixels 


Aleph0Null:
This is for windows and neko (BitmapData.colorTransform() doesn't work on html5 and flash, and I can't test other platforms). Consider the following code:

		var image1 = new FlxSprite(100, 100);
		image1.makeGraphic(64, 32, FlxColor.BLUE);
		add(image1);
		
		var image2 = new FlxSprite(300, 100);
		image2.makeGraphic(64, 32, FlxColor.BLUE);
		add(image2);
				
		var changedRect = new Rectangle(32, 0, 32, 32);
		var toTransparent = new ColorTransform(0, 0, 0, 0, 0, 0, 0, -0xFF);
		image1.pixels.colorTransform(changedRect, toTransparent);

Only image1 has pixels.colorTransform() called, but the transformation is applied to both image1 and image2. If you change image2's color to something else, or load different graphics onto both of them, then the transformation is applied to only image1.

If you add
		
		FlxG.debugger.visible = true;
		FlxG.log.add(image1.pixels == image2.pixels);

then the debugger log displays "true". The Haxe == operator on objects only returns true when both  references are identical (the values of the member variables are irrelevant). So what seems to be happening is HaxeFlixel (or one of the libraries it uses?) is copying a reference when identical BitmapData is generated, instead of creating a new instance of BitmapData with the same pixels.

MSGhero:
The next parameter in makeGraphic is Unique which is false by default. You'll have to set that to true so they don't share bmds.

Aleph0Null:
Thanks, I hadn't paid attention to that parameter.

PR #1491, opened 2015-05-02
Update FlxSprite.hx class description

Aleph0Null:
Expanded the description of the class to explain a bit about how HaxeFlixel handles the BitmapData. 

I had a frustrating evening with #1490. I didn't expect the default behaviour, and after realizing the BitmapData from both sprites had the same reference, I didn't find anything that explained it via Ctrl-F 'pixels', 'BitmapData'. Hopefully an explicit explanation can save other people the frustration.

### DIFF
--- a/flixel/FlxSprite.hx
+++ b/flixel/FlxSprite.hx
@@ -37,6 +37,17 @@ private class GraphicDefault extends BitmapData {}
 /**
  * The main "game object" class, the sprite is a FlxObject
  * with a bunch of graphics options and abilities, like animation and stamping.
+ *
+ * Load an image onto a sprite using the loadGraphic*() functions, 
+ * or create a base monochromatic rectangle using makeGraphic().
+ * The image bitmap is stored in the pixels field.
+ *
+ * HaxeFlixel's graphic caching system keeps track of loaded image data. 
+ * When you load or make an identical copy of a previously used image, by default
+ * HaxeFlixel copies the previous reference onto pixels instead of creating
+ * another copy of the image data, to save memory. If you need to work with
+ * the image data independently, you can disable this by setting the 'Unique' 
+ * parameter in the loadGraphic*() or makeGraphic() functions to true. 
  */
 class FlxSprite extends FlxObject
 {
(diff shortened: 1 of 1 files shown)


Gama11:
Not sure the class description is the right place for this. Also, using a wildcard for `loadGraphic()` to indidcate that it works for all those functions is incorrect, only `loadGraphic()` has a `Unique` param.

Maybe it would be better to mention how the cache key is generated in `makeGraphic()` (combination of width, height and color) and if you want to use the graphic as a "canvas", `Unique` should be true?

I can definitely understand your frustration, this needs to be better documented at the least - I actually ran into the same issue back in AS3 Flixel a few years ago.

Aleph0Null:
Both `loadGraphic()` and `loadGraphicFromTexture()` have the `Unique` param, but `loadGraphicFromSprite()` doesn't. So yeah, using `loadGraphic*()` isn't technically precise, but it's a succinct way of showing people where the param can be used. I can make the description explicit.

I don't yet feel confident explaining the key. What about other situations like `loadGraphic("<file>")`? (Loading the same file in different sprites results in the same `BitmapData` reference. That leads to questions of whether the key 'hash' is collision-resistant/checks are made against collisions and another key assigned.) Also, can `BitmapData` be retrieved from keys by the user? These are the natural questions that come to me when I think about the keys, if I can understand more about this, maybe I can explain something about them.

I like having some mention of it in `FlxSprite.hx` itself, since that's the natural place to go to right now to learn more about `FlxSprite` behaviour, given the lack of detailed documentation. What would be other good places for this description?

Gama11:
I'm not sure what the best place to document this is tbh, but at least in Flixel, the docs for the actual class are typically kept short and give a brief overview over the functionalities of a class, whereas this is more of a detail.

There is no `loadGraphicFromTexture()` on the dev branch anymore btw.

Aleph0Null:
I've moved the explanation of the caching behaviour to the `loadGraphic()` and `makeGraphic()` descriptions, plus some other minor changes.

PR #1492, opened 2015-05-02
Update FlxBitmapTextDocs

Tiago-Ling:
Just small additions / corrections to make the docs clearer.

### DIFF
--- a/flixel/text/FlxBitmapText.hx
+++ b/flixel/text/FlxBitmapText.hx
@@ -52,7 +52,7 @@ class FlxBitmapText extends FlxSprite
 	/**
 	 * Specifies how the text field should align text.
 	 * JUSTIFY alignment isn't supported.
-	 * Note: For alignment to work, 'autoSize' must be set to false.
+	 * Note: 'autoSize' must be set to false or alignment won't show any visual differences.
 	 */
 	public var alignment(default, set):FlxTextAlign = FlxTextAlign.LEFT;
 	
--- a/flixel/text/FlxBitmapText.hx
+++ b/flixel/text/FlxBitmapText.hx
@@ -52,6 +52,7 @@ class FlxBitmapText extends FlxSprite
 	/**
 	 * Specifies how the text field should align text.
 	 * JUSTIFY alignment isn't supported.
+	 * Note: For alignment to work, 'autoSize' must be set to false.
 	 */
 	public var alignment(default, set):FlxTextAlign = FlxTextAlign.LEFT;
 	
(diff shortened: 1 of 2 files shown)


flixel/text/FlxBitmapText.hx
```diff
@@ -52,6 +52,7 @@ class FlxBitmapText extends FlxSprite
 	/**
 	 * Specifies how the text field should align text.
 	 * JUSTIFY alignment isn't supported.
+	 * Note: For alignment to work, 'autoSize' must be set to false.
```
Gama11: Hm.. Technically, alignment probably works, it's just that it doesn't look any different with the width being as small as possible?

I guess a regular FlxText behaves the same way?

flixel/text/FlxBitmapText.hx
```diff
@@ -1461,7 +1462,7 @@ class FlxBitmapText extends FlxSprite
 	 * Set border's style (shadow, outline, etc), color, and size all in one go!
 	 * 
 	 * @param	Style outline style
-	 * @param	Color outline color in flash 0xRRGGBB format
+	 * @param	Color outline color in flash 0xAARRGGBB format
```
Gama11: The method description here was copied from `FlxText`, where the border color does indeed ignore alpha values. Not a big fan of this behavior, the two text classes should be as consistent as possible I think... I wonder if transparency is possible in FlxText's borderColor.

Beeblerox: color outline should be in 0xRRGGBBAA format?

Gama11: Hm, I guess is a conscious limitation to not have alpha for FlxText's borderColor because of the way it's drawn (copying the original BitmapData to slightly offsetted positions). Would look weird with the overlaps.

Might still be possible to do if the outline is first copied entirely with alpha = 1, and then the transparency is applied in a separate draw call. Would be more expensive though?

Gama11: Then again, border style (outline at least) already uses so many draw calls that an additional one wouldn't matter much.

Gama11: I think I got it mostly working...

flixel/text/FlxBitmapText.hx
Tiago-Ling: Since autoSize is the default i think this note is a good way to save the user from some tries until he/she figures it out. I can put this note in FlxText also if people agree.

Gama11: Yeah, we should be consistent about that and mention it in both places.

I would it a little differently though: "For alignment to have an effect [...]"

It's not that it doesn't work, you just wouldn't see it.

PR #1493, opened 2015-05-02
Added `pauseAll` to FlxTimer

Tiago-Ling:
It makes easier to control timer behavior across different classes.

### DIFF
--- a/flixel/util/FlxTimer.hx
+++ b/flixel/util/FlxTimer.hx
@@ -144,6 +144,18 @@ class FlxTimer implements IFlxDestroyable
 	}
 	
 	/**
+	 * Toggles pause for all timers.
+	 * @param	value	True to pause and false to resume
+	 */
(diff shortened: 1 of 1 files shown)


Gama11:
Doesn't setting `FlxTimer.manager.active = false;` do the trick?

MSGhero:
I prefer `pauseAll()` and `resumeAll()` as inlined ^, though idk if there are consequences of doing it that way.

Gama11:
I don't think inlining would make any noticable difference here, this isn't a function you'd typically call very often.

MSGhero:
I guess, but two functions over one imo.

Tiago-Ling:
@Gama11 You're right, you can set active to false. Not very intuitive
actually, only tried it after you pointed it out. So you think there's no
need for this method? If so i can close the pull request.

Gama11:
@Tiago-Ling Yes, seems redundant.

Issue #1087, opened 2014-05-13
Rethink the concept of plugins / get rid of them


Gama11:
Plugins are not a very useful concept. They're not different from FlxBasics, yet they behave differently - using a kind of singleton pattern where only one instance of the plugin can exist at a time. This creates some interesting issues with substates, as plugins have no concept of belonging to a state like everything else in flixel does. This means tweens in a parent state won't stop if a substate is opened, even if everything else pauses.

It would be much more flexible, intuitive and powerful if plugins were add()ed to states like everything else.

Gama11:
I should probably mention that @JoeCreates is working on a solution for this (just to two people working on the same thing and duplicating efforts).

PR #1495, opened 2015-05-10
FlxTween#cancelTweensOf(), closes #1458

### LINKED ISSUE #1458: Kill all tweens of
TweenLite has a static function like this, and it would be good to have to avoid null object refs when something gets destroyed during its tween.

My question is, since this only affects VarTweens but needs access to the Tween Manager's private list of tweens AND VarTween's private `object`, which one should be exposed? Should it be `FlxTween.killTweensOf(obj)` or should it go into `FlxDestroyUtil`?

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/9786623-kill-all-tweens-of?utm_campaign=plugin&utm_content=tracker%2F254189&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F254189&utm_medium=issues&utm_source=github).
</bountysource-plugin>

MSGhero:
#1458

I chose to make it manual adding and removing/canceling tweens instead of automatic so performance wouldn't be affected.

Usage would be like
```haxe
var tw = FlxTween.num(..., { onComplete : FlxTween.unregisterTween.bind(_, this) }, ...);
FlxTween.registerTween(tw, this);

override public function destroy():Void {
     super.destroy();
     FlxTween.cancelTweensOf(this);
}
```

### DIFF
--- a/flixel/tweens/FlxTween.hx
+++ b/flixel/tweens/FlxTween.hx
@@ -3,6 +3,7 @@ package flixel.tweens;
 import flixel.FlxG;
 import flixel.FlxObject;
 import flixel.FlxSprite;
+import flixel.math.FlxPoint;
 import flixel.tweens.FlxEase.EaseFunction;
 import flixel.tweens.misc.AngleTween;
 import flixel.tweens.misc.ColorTween;
@@ -17,7 +18,6 @@ import flixel.tweens.motion.QuadPath;
 import flixel.util.FlxArrayUtil;
 import flixel.util.FlxColor;
 import flixel.util.FlxDestroyUtil.IFlxDestroyable;
-import flixel.math.FlxPoint;
 
 class FlxTween implements IFlxDestroyable
 {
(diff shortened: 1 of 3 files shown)


Gama11:
Looks useful, but at the same time it seems like a somewhat awkward workaround for tweens not being added to `FlxBasic` / `FlxObject` in the first place (which would also solve issues regarding substate pausing etc discusses in #1087). That would be quite a big refactor though...

MSGhero:
Yeah @Beeblerox said this would be part of a refactor later on. I need the functionality now, and I figured I'd post how I did it.

JoeCreates:
I use something along these lines to ensure I don't have multiple tweens fighting over the same object. The implementation of mapping objects to a tween in the same, but I also have an `exclusiveTween` method, which is identical to `FlxTween.tween` except you also pass it the dynamic key. When you `exclusiveTween` something, it automatically cancels the tweens with the same key.

This prevents things jittering and breaking when you, for example, have a tween pan the camera from one scene to another, but the user presses a button to tween back to the first scene before the first tween has finished.

Could this perhaps be worked into flixel, too?

MSGhero:
How do you remove the key-val of the tween ends normally? Do you search through all the pairs looking or is it manual?

I had something automatic like that, but it didn't seem very performant.

JoeCreates:
@MSGhero I didn't automate the removal. For the most past I just left the tweens in the map even after they were finished, seeing as it's fine to cancel an already finished tween. I haven't had any need to remove them, yet, seeing as I have mainly used this for UI with a finite number of elements, and so the map is generally small. It's useful to be able to manually remove them, of course. I just haven't needed to remove them at all , yet.

MSGhero:
I was thinking in terms of garbage collection. You have the tween and the object it references chilling in the map.

JoeCreates:
@MSGhero It's not something I've needed to consider yet, due to how frequently I reuse those objects.

MSGhero:
I don't like this pull anymore. I'll leave #1458 up so it gets remembered, but I don't think anyone wants to *have* to bind a function and waste an onComplete to use it.

A proper way to do this should be a lazily created "tween manager" inside of each FlxObject and maybe an optional `AddTweenTo:FlxObject` parameter in the tween methods. Idk. 4.0.0 stuff.

PR #1500, opened 2015-05-19
Fix xbox 360 gamepad inputs and ensure consistent behavior with flash

larsiusprime:


flixel/input/gamepad/XboxButtonID.hx
```diff
@@ -125,16 +125,16 @@ class XboxButtonID
...
+	public static inline var X:Int = 2;
+	public static inline var Y:Int = 3;
+	public static inline var LB:Int = 4;
+	public static inline var RB:Int = 5;
+	public static inline var BACK:Int = 6;
+	public static inline var START:Int = 7;
+	public static inline var LEFT_ANALOG:Int = 8;
+	public static inline var RIGHT_ANALOG:Int = 9;
 	
 	public static inline var XBOX:Int = 14;
```
Gama11: The DPAD constants aren't working (correct?) so they should probably be removed.

Gama11:
Somebody is gonna need to check the IDs for Mac and Linux as well.

flixel/input/gamepad/XboxButtonID.hx
larsiusprime: The DPAD constants *do* work, but not as button presses. In the joystick api there are no true button events that are generated. I have two options:

1. when the hat moves, generate "fake" dpad button press events
2. create a FlxGamepadDpad object to wrap the current hat object, and add up/down/left/right buttons inside so we can get pressed/justpressed/justreleased/ etc for each dpad direction

Gama11: How would you use them with the FlxGamepad API though? gamepad.hat is a FlxPoint, and there already are getters for the dpad directions for which you don't need any indices.

Is 1. even an option? Does the hat event come with one of these indices?

larsiusprime: So the getters for the dpad directions tell you if something is currently pressed, but they're kind of ugly:

1. They only work on cpp, not flash. So inconsistent.
2. They don't give you justpressed/justreleased, only "is pressed now" or not.

The above "1." is an option if you generate fake dpad button press event ids. Basically you would create a constant like "DPAD_DOWN = 10" and so on, and then just assign those by hand in the function based on whether the hat was moved x = -1/1, or y= -1/1, etc. And make sure you use an unused button constant for what you set DPAD_DOWN/LEFT/RIGHT/UP to. 

This could be wrapped up in a way that is consistent with flash's behavior at the API level.

Gama11: Ah, I hadn't even considered that with them not being buttons you only have the "pressed" information. Yeah, that's inconvenient...

If you can make that hack work consistently, I think we should go for that. Of course it would be even better if it was consistent with flash at lime / openfl level (which it used to be until that mentioned SDL change). I guess in that case we should remove the getters instead?

larsiusprime: Yeah I think ultimately we should do that. There's other things to reconcile too... on some controller models the analog triggers have digital buttons at the end that "click", on some controllers (ps3/ps4) the face buttons have analog pressure... it's a lot to wrap, but the basics should be doable without exploding the scope of this PR.

Just for now I think generating fake button events based on hat stuff in legacy will work for now, and then we can map more cleanly when targeting next.

MSGhero:
What's the deal with [openfl's GameInput](https://github.com/openfl/openfl/blob/master/openfl/ui/GameInput.hx) and [lime's GamepadButton](https://github.com/openfl/lime/blob/master/lime/ui/GamepadButton.hx) then? Is lime's setup just the 360 controller?

Also there's a bit of overlap with openfl's new controller support and flixel's existing gamepad support, but I'll look into it more later.

larsiusprime:
@MSGhero so those are definitely for next, what I'm submitting is to patch things up for legacy. Currently nothing at all works in flixel on next, I'll definitely be using the available openfl/lime API's for that. One thing that may be going on in lime is that it's using SDL, which handles the "wrapping" of all the different devices into one X360-like standard.

MSGhero:
@larsiusprime Gotcha, I've only been testing in flash where `-D openfl-next` is defined by default. didn't realize it wasn't for cpp yet.

larsiusprime:
I've got a better branch of this now, gonna close this one.

PR #1502, opened 2015-05-21
Refactor the gamepad API, closes #1476

### LINKED ISSUE #1476: XboxButtonID keycodes incorrect for windows target
I'm on 8.1, with 360 driver version 6.3.9600.17054 ( 17/03/2014 ) - not sure error persists for other drivers / windows versions.

To reproduce, try this for neko / windows target : 

function getButtonPressed():String 
{
	switch(FlxG.gamepads.lastActive.firstPressedButtonID()) {
		case XboxButtonID.A : return 'A pressed';
		case XboxButtonID.B : return 'B pressed';
		case XboxButtonID.X : return 'X pressed';
		case XboxButtonID.Y : return 'Y pressed';
		case XboxButtonID.LB : return 'LB pressed';
		case XboxButtonID.RB : return 'RB pressed';
		case XboxButtonID.START : return 'START pressed';
		case XboxButtonID.BACK : return 'BACK pressed';
		case XboxButtonID.XBOX : return 'XBOX pressed';
		case XboxButtonID.LEFT_ANALOGUE : return 'L3 pressed';
		case XboxButtonID.RIGHT_ANALOGUE: return 'R3 pressed';
	}
	return 'none';
}

The correct codes are 

public static inline var A:Int = 0;
public static inline var B:Int = 1;
public static inline var X:Int = 2;
public static inline var Y:Int = 3;
public static inline var LB:Int = 4;
public static inline var RB:Int = 5;
public static inline var BACK:Int = 6;
public static inline var START:Int = 7;
public static inline var LEFT_ANALOGUE:Int = 8;
public static inline var RIGHT_ANALOGUE:Int = 9;
public static inline var XBOX:Int = 10;

The axis array indices are also a bit shuffled - these are correct : 
	
public static inline var LEFT_ANALOGUE_X:Int = 0;
public static inline var LEFT_ANALOGUE_Y:Int = 1;
public static inline var RIGHT_ANALOGUE_X:Int = 4;
public static inline var RIGHT_ANALOGUE_Y:Int = 2;
public static inline var LEFT_TRIGGER:Int = 3;
public static inline var RIGHT_TRIGGER:Int = 5;

I'm not getting anything for the d pad - although i've only tried button indexes that are consecutive to the ones already used ( ie, starting at 11 )

It works as expected for the flash target.

lib list is : 

actuate: 1.8.3 [1.8.4]
air3: [0.0.1]
box2d: [1.2.3]
flambe: [4.1.0]
flixel-addons: [1.1.0]
flixel-demos: [1.1.1]
flixel-templates: [1.0.2]
flixel-tools: [1.0.5]
flixel-ui: [1.0.2]
flixel: 3.0.1 [3.3.8]
format: [3.1.2]
hxcpp: 3.2.37 [3.2.81]
hxsl: [2.0.5]
layout: [1.2.0]
lime-samples: 2.2.0 [2.3.0]
lime-tools: [1.5.7]
lime: 2.2.1 2.3.0 [2.3.1]
nape: [2.0.16]
openfl-bitfive: [3.0.0]
openfl-html5: [1.4.2-beta]
openfl-samples: [2.2.2]
openfl: 2.0.0 2.2.1 3.0.0-beta.3 [3.0.1]
swf: 1.8.2 [1.8.4]

Can someone confirm that this is happening on all versions of windows? If so ill do a PR

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/11159581-xboxbuttonid-keycodes-incorrect-for-windows-target?utm_campaign=plugin&utm_content=tracker%2F254189&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F254189&utm_medium=issues&utm_source=github).
</bountysource-plugin>

larsiusprime:
Updates the Gamepad API so that it works consistently on legacy, flash, and next. Only tested on windows with a single XBOX 360 pad, but should be a solid foundation. Comments/critique welcome!

### DIFF
--- a/flixel/input/gamepad/ButtonID.hx
+++ b/flixel/input/gamepad/ButtonID.hx
@@ -1,0 +1,50 @@
+package flixel.input.gamepad;
+
+import flixel.system.macros.FlxMacroUtil;
+
+/**
+ * Maps enum values and strings to integer keycodes.
+ */
+@:enum
+abstract ButtonID(Int) from Int to Int
+{
+	public static var fromStringMap(default, null):Map<String, ButtonID>
+		= FlxMacroUtil.buildMap("flixel.input.gamepad.ButtonID");
+		
+	public static var toStringMap(default, null):Map<ButtonID, String>
+		= FlxMacroUtil.buildMap("flixel.input.gamepad.ButtonID", true);
+	
+	// Key Indicies
+	var ANY             = -2;
+	var NONE            = -1;
+	var A               =  0;
+	var B               =  1;
+	var X               =  2;
+	var Y               =  3;
+	var LEFT_SHOULDER   =  4;
+	var RIGHT_SHOULDER  =  5;
+	var BACK            =  6;
+	var START           =  7;
+	var LEFT_STICK_BTN  =  8;
+	var RIGHT_STICK_BTN =  9;
+	var GUIDE           = 10;
+	var DPAD_UP         = 11;
+	var DPAD_DOWN       = 12;
+	var DPAD_LEFT       = 13;
+	var DPAD_RIGHT      = 14;
+	var LEFT_TRIGGER    = 15;
+	var RIGHT_TRIGGER   = 16;
+	
+	@:from
+	public static inline function fromString(s:String)
+	{
+		s = s.toUpperCase();
+		return fromStringMap.exists(s) ? fromStringMap.get(s) : NONE;
+	}
+	
+	@:to
+	public inline function toString():String
+	{
+		return toStringMap.get(this);
+	}
+}
--- a/flixel/input/gamepad/ButtonID.hx
+++ b/flixel/input/gamepad/ButtonID.hx
@@ -14,26 +14,33 @@ abstract ButtonID(Int) from Int to Int
 	public static var toStringMap(default, null):Map<ButtonID, String>
 		= FlxMacroUtil.buildMap("flixel.input.gamepad.ButtonID", true);
 	
-	// Key Indicies
-	var ANY             = -2;
-	var NONE            = -1;
-	var A               =  0;
-	var B               =  1;
-	var X               =  2;
-	var Y               =  3;
-	var LEFT_SHOULDER   =  4;
-	var RIGHT_SHOULDER  =  5;
-	var BACK            =  6;
-	var START           =  7;
-	var LEFT_STICK_BTN  =  8;
-	var RIGHT_STICK_BTN =  9;
-	var GUIDE           = 10;
-	var DPAD_UP         = 11;
-	var DPAD_DOWN       = 12;
-	var DPAD_LEFT       = 13;
-	var DPAD_RIGHT      = 14;
-	var LEFT_TRIGGER    = 15;
-	var RIGHT_TRIGGER   = 16;
+	// Button Indices
+	var ANY               = -2;		
+	var NONE              = -1;		
+	var A                 =  0;		//BOTTOM face button
+	var B                 =  1;		//RIGHT  face button
+	var X                 =  2;		//LEFT   face button
+	var Y                 =  3;		//TOP    face button
+	var LEFT_SHOULDER     =  4;		//left "bumper"
+	var RIGHT_SHOULDER    =  5;		//right "bumper"
+	var BACK              =  6;		//a.k.a "select", left center button
+	var START             =  7;		//right center button
+	var LEFT_STICK_BTN    =  8;		//Click left stick
+	var RIGHT_STICK_BTN   =  9;		//Click right stick
+	var GUIDE             = 10;		//a.k.a "xbox" or "PS" or "home" button in center
+	var DPAD_UP           = 11;
+	var DPAD_DOWN         = 12;
+	var DPAD_LEFT         = 13;
+	var DPAD_RIGHT        = 14;
+	var LEFT_TRIGGER_BTN  = 15;		//digital button at end of left analog trigger full squeeze (if available)
+	var RIGHT_TRIGGER_BTN = 16;		//digital button at end of right analog trigger full squeeze (if available)
+	
+	var LEFT_TRIGGER       = 17;
+	var RIGHT_TRIGGER      = 18;
+	
+	// FlxGamepadAnalogSticks (wraps up the _X/_Y into a structure)
+	var LEFT_ANALOG_STICK  = 19;
+	var RIGHT_ANALOG_STICK = 20;
 	
 	@:from
 	public static inline function fromString(s:String)
--- a/flixel/input/gamepad/ButtonID.hx
+++ b/flixel/input/gamepad/ButtonID.hx
@@ -15,30 +15,43 @@ abstract ButtonID(Int) from Int to Int
 		= FlxMacroUtil.buildMap("flixel.input.gamepad.ButtonID", true);
 	
 	// Button Indices
-	var ANY               = -2;		
-	var NONE              = -1;		
-	var A                 =  0;		//BOTTOM face button
-	var B                 =  1;		//RIGHT  face button
-	var X                 =  2;		//LEFT   face button
-	var Y                 =  3;		//TOP    face button
-	var LEFT_SHOULDER     =  4;		//left "bumper"
-	var RIGHT_SHOULDER    =  5;		//right "bumper"
-	var BACK              =  6;		//a.k.a "select", left center button
-	var START             =  7;		//right center button
-	var LEFT_STICK_BTN    =  8;		//Click left stick
-	var RIGHT_STICK_BTN   =  9;		//Click right stick
-	var GUIDE             = 10;		//a.k.a "xbox" or "PS" or "home" button in center
+	var ANY               = -2;
+	var NONE              = -1;
+	
+	/**BOTTOM face button*/ 
+	var A                 =  0;
+	/**RIGHT face button*/  
+	var B                 =  1;
+	/**LEFT face button*/
+	var X                 =  2;
+	/**TOP face button*/
+	var Y                 =  3;
+	/**left digital "bumper"*/
+	var LEFT_SHOULDER     =  4;
+	/**right digital "bumper"*/
+	var RIGHT_SHOULDER    =  5;
+	/**also known as "select", the leftmost center button*/
+	var BACK              =  6;
+	/**the rightmost center button*/
+	var START             =  7;
+	/**digital click of the left analog stick*/
+	var LEFT_STICK_BTN    =  8;
+	/**digital click of the right analog stick*/
+	var RIGHT_STICK_BTN   =  9;
+	/**this is the "XBox" or "PS" or "home" button in the center*/
+	var GUIDE             = 10;
 	var DPAD_UP           = 11;
 	var DPAD_DOWN         = 12;
 	var DPAD_LEFT         = 13;
 	var DPAD_RIGHT        = 14;
-	var LEFT_TRIGGER_BTN  = 15;		//digital button at end of left analog trigger full squeeze (if available)
-	var RIGHT_TRIGGER_BTN = 16;		//digital button at end of right analog trigger full squeeze (if available)
+	/**digital click at end of left analog trigger's squeeze (if available)*/
+	var LEFT_TRIGGER_BTN  = 15;
+	/**digital click at end of right analog trigger's squeeze (if available)*/
+	var RIGHT_TRIGGER_BTN = 16;
 	
 	var LEFT_TRIGGER       = 17;
 	var RIGHT_TRIGGER      = 18;
 	
-	// FlxGamepadAnalogSticks (wraps up the _X/_Y into a structure)
 	var LEFT_ANALOG_STICK  = 19;
 	var RIGHT_ANALOG_STICK = 20;
 	
--- a/flixel/input/gamepad/ButtonIndex.hx
+++ b/flixel/input/gamepad/ButtonIndex.hx
@@ -67,10 +67,10 @@ class ButtonIndex
 			case RIGHT_SHOULDER: LogitechButtonID.SIX;
 			case LEFT_TRIGGER: LogitechButtonID.SEVEN;
 			case RIGHT_TRIGGER: LogitechButtonID.EIGHT;
-			case DPAD_UP: LogitechButtonID.DPAD_UP;
-			case DPAD_DOWN: LogitechButtonID.DPAD_DOWN;
-			case DPAD_LEFT: LogitechButtonID.DPAD_LEFT;
-			case DPAD_RIGHT: LogitechButtonID.DPAD_RIGHT;
+			//case DPAD_UP: 
+			//case DPAD_DOWN: 
+			//case DPAD_LEFT: 
+			//case DPAD_RIGHT: 
 			//case GUIDE:
 			default: -1;
 		}
--- a/flixel/input/gamepad/ButtonIndex.hx
+++ b/flixel/input/gamepad/ButtonIndex.hx
@@ -1,8 +1,6 @@
 package flixel.input.gamepad;
-import flixel.input.gamepad.FlxGamepad.GamepadAxisID;
-import flixel.input.gamepad.FlxGamepad.GamepadButtonID;
+import flixel.input.gamepad.ButtonID;
 import flixel.input.gamepad.FlxGamepad.GamepadModel;
-import flixel.input.gamepad.FlxGamepad.FlxAxes;
 
 /**
  * ...
@@ -17,61 +15,55 @@ class ButtonIndex
 		model = Model;
 	}
 	
-	public function get(button:GamepadButtonID):Int
+	/**
+	 * Given a ButtonID, return the raw hardware code
+	 * @param	buttonID the "universal" ButtonID
+	 * @return	the raw hardware code
+	 */
+	
+	public inline function getRaw(buttonID:ButtonID):Int
 	{
 		return switch(model)
 		{
-			case Logitech: getLogitech(button);
-			case OUYA: getOUYA(button);
-			case PS3: getPS3(button);
-			case PS4: getPS4(button);
-			case Xbox: getXbox(button);
-			case XInput: getXInput(button);
+			case Logitech: getRawLogitech(buttonID);
+			case OUYA: getRawOUYA(buttonID);
+			case PS3: getRawPS3(buttonID);
+			case PS4: getRawPS4(buttonID);
+			case Xbox: getRawXbox(buttonID);
 			default: -1;
 		}
 	}
 	
-	public function toButtonID(i:Int):GamepadButtonID
+	/**
+	 * Given a raw hardware code, return the "universal" ButtonID
+	 * @param	RawID	the raw hardware code
+	 * @return	the "universal" ButtonID
+	 */
+	
+	public inline function getBtn(RawID:Int):ButtonID
 	{
 		return switch(model)
 		{
-			case Logitech: toLogitech(i);
-			case OUYA: toOUYA(i);
-			case PS3: toPS3(i);
-			case PS4: toPS4(i);
-			case Xbox: toXbox(i);
-			case XInput: toXInput(i);
-			default: UNKNOWN;
+			case Logitech: getBtnLogitech(RawID);
+			case OUYA: getBtnOUYA(RawID);
+			case PS3: getBtnPS3(RawID);
+			case PS4: getBtnPS4(RawID);
+			case Xbox: getBtnXbox(RawID);
+			default: NONE;
 		}
 	}
 	
-	public function getAxis(button:GamepadAxisID):Int
+	public inline function getRawOUYA(buttonID:ButtonID):Int
 	{
-		return switch(model)
-		{
-			case Logitech: getAxisLogitech(button);
-			case OUYA: getAxisOUYA(button);
-			case PS3: getAxisPS3(button);
-			case PS4: getAxisPS4(button);
-			case Xbox: getAxisXbox(button);
-			case XInput: getAxisXInput(button);
-			default: -1;
-		}
-	}
-	
-	/****give GamepadButtonID, get int****/
-	
-	public inline function getOUYA(button:GamepadButtonID):Int
-	{
-		return switch(button)
+		return switch(buttonID)
 		{
 			case A: OUYAButtonID.O;
 			case B: OUYAButtonID.A;
 			case X: OUYAButtonID.U;
 			case Y: OUYAButtonID.Y;
 			case GUIDE: OUYAButtonID.HOME;
-			case LEFT_STICK: OUYAButtonID.LEFT_ANALOG;
-			case RIGHT_STICK: OUYAButtonID.RIGHT_ANALOG;
+			case LEFT_STICK_BTN: OUYAButtonID.LEFT_STICK_BTN;
+			case RIGHT_STICK_BTN: OUYAButtonID.RIGHT_STICK_BTN;
 			case LEFT_SHOULDER: OUYAButtonID.LB;
 			case RIGHT_SHOULDER: OUYAButtonID.RB;
 			//case BACK:
@@ -84,9 +76,9 @@ class ButtonIndex
 		}
 	}
 	
-	public inline function getLogitech(button:GamepadButtonID):Int
+	public inline function getRawLogitech(buttonID:ButtonID):Int
 	{
-		return switch(button)
+		return switch(buttonID)
 		{
 			case A: LogitechButtonID.TWO;
 			case B: LogitechButtonID.THREE;
@@ -94,8 +86,8 @@ class ButtonIndex
 			case Y: LogitechButtonID.FOUR;
 			case BACK: LogitechButtonID.NINE;
 			case START: LogitechButtonID.TEN;
-			case LEFT_STICK: LogitechButtonID.LEFT_ANALOG;
-			case RIGHT_STICK: LogitechButtonID.RIGHT_ANALOG;
+			case LEFT_STICK_BTN: LogitechButtonID.LEFT_STICK_BTN;
+			case RIGHT_STICK_BTN: LogitechButtonID.RIGHT_STICK_BTN;
 			case LEFT_SHOULDER: LogitechButtonID.FIVE;
 			case RIGHT_SHOULDER: LogitechButtonID.SIX;
 			case LEFT_TRIGGER: LogitechButtonID.SEVEN;
@@ -109,9 +101,9 @@ class ButtonIndex
 		}
 	}
 	
-	public inline function getPS4(button:GamepadButtonID):Int
+	public inline function getRawPS4(buttonID:ButtonID):Int
 	{
-		return switch(button)
+		return switch(buttonID)
 		{
 			case A: PS4ButtonID.X;
 			case B: PS4ButtonID.CIRCLE;
@@ -120,8 +112,8 @@ class ButtonIndex
 			case BACK: PS4ButtonID.SELECT;
 			case GUIDE: PS4ButtonID.PS;
 			case START: PS4ButtonID.START;
-			case LEFT_STICK: PS4ButtonID.LEFT_ANALOG;
-			case RIGHT_STICK: PS4ButtonID.RIGHT_ANALOG;
+			case LEFT_STICK_BTN: PS4ButtonID.LEFT_STICK_BTN;
+			case RIGHT_STICK_BTN: PS4ButtonID.RIGHT_STICK_BTN;
 			case LEFT_SHOULDER: PS4ButtonID.L1;
 			case RIGHT_SHOULDER: PS4ButtonID.R1;
 			case LEFT_TRIGGER: PS4ButtonID.L2;
@@ -134,9 +126,9 @@ class ButtonIndex
 		}
 	}
 	
-	public inline function getPS3(button:GamepadButtonID):Int
+	public inline function getRawPS3(buttonID:ButtonID):Int
 	{
-		return switch(button)
+		return switch(buttonID)
 		{
 			case A: PS3ButtonID.X;
 			case B: PS3ButtonID.CIRCLE;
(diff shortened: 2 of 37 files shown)


flixel/input/gamepad/ButtonID.hx
```diff
@@ -0,0 +1,57 @@
+package flixel.input.gamepad;
+
+import flixel.system.macros.FlxMacroUtil;
+
+/**
+ * Maps enum values and strings to integer keycodes.
+ */
+@:enum
+abstract ButtonID(Int) from Int to Int
```
Gama11: Not sure I like this name... guess `FlxGamepadButton` already exists... `FlxGamepadButtonID` is rather long.

flixel/input/gamepad/ButtonID.hx
```diff
@@ -0,0 +1,57 @@
...
+	var NONE              = -1;		
+	var A                 =  0;		//BOTTOM face button
+	var B                 =  1;		//RIGHT  face button
+	var X                 =  2;		//LEFT   face button
+	var Y                 =  3;		//TOP    face button
+	var LEFT_SHOULDER     =  4;		//left "bumper"
+	var RIGHT_SHOULDER    =  5;		//right "bumper"
+	var BACK              =  6;		//a.k.a "select", left center button
+	var START             =  7;		//right center button
+	var LEFT_STICK_BTN    =  8;		//Click left stick
```
Gama11: I don't like the `_BTN` here... could we just lose that part / aren't these all buttons anyway? Why are there `_TRIGGER` and `_TRIGGER_BTN, the former should be axes right?

flixel/input/gamepad/ButtonID.hx
larsiusprime: Sure, I wouldn't mind something shorter. For reference this corresponds to FlxKey in its implementation and use. Got any suggestions?

flixel/input/gamepad/ButtonID.hx
```diff
@@ -0,0 +1,57 @@
...
+	public static var fromStringMap(default, null):Map<String, ButtonID>
+		= FlxMacroUtil.buildMap("flixel.input.gamepad.ButtonID");
+		
+	public static var toStringMap(default, null):Map<ButtonID, String>
+		= FlxMacroUtil.buildMap("flixel.input.gamepad.ButtonID", true);
+	
+	// Button Indices
+	var ANY               = -2;		
+	var NONE              = -1;		
+	var A                 =  0;		//BOTTOM face button
```
Gama11: If you're gonna add comments for these, use `/** */`-style comments so it shows up in completion.

flixel/input/gamepad/ButtonIndex.hx
```diff
@@ -0,0 +1,405 @@
...
+				case XInput: XInputButtonID.RIGHT_ANALOG_STICK;
+				default: null;
+			}
+		}
+		return null;
+	}
+	
+	/*****************************************/
+	
+	public inline function getRawOUYA(buttonID:ButtonID):Int
```
Gama11: What's with all these commented out functions?

flixel/input/gamepad/ButtonIndex.hx
```diff
@@ -0,0 +1,405 @@
...
+			
+			//case LEFT_TRIGGER: 
+			//case RIGHT_TRIGGER: 
+			default: -1;
+		}
+	}
+	
+	/**************/
+	
+	public inline function getBtnOUYA(rawID:Int):ButtonID
```
Gama11: Not sure inlining these is a good idea, they're quite large...

flixel/input/gamepad/ButtonID.hx
larsiusprime: Glad you asked. So, there's a lot of ambiguity involved if we're not careful.

Analog sticks can be clicked, which is a separate digital button input from the movement of the sticks themselves.  Furthermore, on certain gamepads there are additional digital button inputs at the end of the left and right analog shoulder triggers, which means although there's 4 physical shoulder inputs (left/right digital bumpers, left/right analog triggers), there's 6 actual inputs - 4 digital (bumpers + hidden button clicks at end of trigger squeeze), 2 analog (triggers).

So you need to distinguish these parts of the stick:
LEFT_ANALOG_STICK (the stick itself)
LEFT_STICK_BTN (click the stick)

LEFT_TRIGGER (the analog shoulder trigger)
LEFT_TRIGGER_BTN (the digital button at the end of the squeeze on some models)

flixel/input/gamepad/ButtonID.hx
larsiusprime: Do those need to go before or after the line?

flixel/input/gamepad/FlxGamepad.hx
```diff
@@ -319,40 +425,77 @@ class FlxGamepad implements IFlxDestroyable
...
+	 * use this only for things like <code>XboxButtonID.LEFT_TRIGGER</code>,
+	 * use getXAxis() / getYAxis() for analog sticks!
+	 */
+	public inline function getAxisRaw(RawAxisID:Int):Float
+	{
+		var axisValue = getAxisValue(RawAxisID);
 		if (Math.abs(axisValue) > deadZone)
 		{
+			#if (!flash && !next)
+				//(-1,1) range, normalize to (0,1) for legacy target only
```
Gama11: Why?

flixel/input/gamepad/ButtonIndex.hx
larsiusprime: They're not commented out, I think it's getting confused because of the /****/ separator

flixel/input/gamepad/ButtonIndex.hx
larsiusprime: Yeah I don't know what the best practice is there. Whatever you think is best.

flixel/input/gamepad/FlxGamepadManager.hx
```diff
@@ -89,12 +89,12 @@ class FlxGamepadManager implements IFlxInputManager
 			firstActive = null;
 	}
 	
-	private function createByID(GamepadID:Int):FlxGamepad
+	private function createByID(GamepadID:Int,?Model:GamepadModel):FlxGamepad
```
Gama11: Minor issue, but flixel source code has a space after commas.

flixel/input/gamepad/FlxGamepad.hx
larsiusprime: To get consistent results? On legacy I was getting a range of (-1,1) but on flash/next I was getting (0,1), from the same hardware.

flixel/input/gamepad/FlxGamepadManager.hx
```diff
@@ -421,12 +423,35 @@ class FlxGamepadManager implements IFlxInputManager
...
+		var strip = ["-", "_"];
+		for (s in strip)
+		{
+			while (str.indexOf(s) != -1)
+			{
+				str = StringTools.replace(str, s, "");
+			}
+		}
+		
+		if (str.indexOf("xbox") != -1) return Xbox;					//"Microsoft X-Box 360 pad"
```
Gama11: This is why it's recommended to use spaces instead of tabs for formatting - as soon as the tab width changes, like here on GitHub, it looks horrible. ;)

flixel/input/gamepad/FlxGamepadManager.hx
```diff
@@ -421,12 +423,35 @@ class FlxGamepadManager implements IFlxInputManager
...
+		if (str.indexOf("playstation") != -1) return PS3;			//"Sony PLAYSTATION(R)3 Controller"
+		if (str.indexOf("ouya") != -1) return OUYA;					//"OUYA Game Controller"
+		if (str.indexOf("wireless controller") != -1) return PS4;	//"Wireless Controller"
+		if (str.indexOf("logitech") != -1) return Logitech;
+		if (str.indexOf("xinput") != -1) return XInput;				
+		trace(str);
```
Gama11: Leftover debug trace?

flixel/input/gamepad/FlxGamepadManager.hx
```diff
@@ -482,8 +507,92 @@ class FlxGamepadManager implements IFlxInputManager
...
+			var change = false;
+			
+			//We see if there's been a change so we can properly set "justPressed"/"justReleased", etc.
+			if (oldx != newx)
+			{
+				change = true;
+				
+				if (oldx == -1)
+				{
+					newType = "buttonUp";
```
Gama11: Might be better to declare these strings as constants somewhere?

flixel/input/gamepad/OUYAButtonID.hx
```diff
@@ -16,8 +16,8 @@ class OUYAButtonID
 	public static inline var A:Int = 1;
 	public static inline var LB:Int = 6;
 	public static inline var RB:Int = 7;
-	public static inline var LEFT_ANALOG:Int = 10;
-	public static inline var RIGHT_ANALOG:Int = 11;
+	public static inline var LEFT_STICK_BTN:Int = 10;
```
Gama11: Again, not a fan of the abbreviated `BTN` here - you don't really find that sort of naming in the flixel core source.

flixel/input/gamepad/XInputButtonID.hx
```diff
@@ -0,0 +1,40 @@
...
+	public static inline var X:Int = 8;
+	public static inline var Y:Int = 9;
+	public static inline var LB:Int = 15;
+	public static inline var RB:Int = 16;
+	public static inline var BACK:Int = 10;
+	public static inline var START:Int = 12;
+	public static inline var LEFT_STICK_BTN:Int = 13;
+	public static inline var RIGHT_STICK_BTN:Int = 14;
+	
+	public static inline var GUIDE:Int = 11;
```
Gama11: What's "guide"?

larsiusprime: In the LIME gamepad API, it seems to stand in for the "Xbox", "PS", and/or "Home" button consoles stick in the middle these days.

Gama11:
Since we're not gonna get rid of it, I think we should be consistent about using the `Flx` prefix - there's inconsistencies even within the types you added (some use it, some don't). Is there any particular reason?

larsiusprime:
Not really, brevity perhaps. FlxGamepadButtonID is pretty long. Open to suggestions.

flixel/input/gamepad/ButtonID.hx
Gama11: Before.

flixel/input/gamepad/ButtonID.hx
Gama11: Okay, but what's the analog `LEFT_TRIGGER` doing here, it's an axis and not a button, right?

flixel/input/gamepad/ButtonIndex.hx
```diff
@@ -0,0 +1,405 @@
...
+		}
+	}
+	
+	/**
+	 * Given a raw hardware code, return the "universal" ButtonID
+	 * @param	RawID	the raw hardware code
+	 * @return	the "universal" ButtonID
+	 */
+	
+	public function getBtn(RawID:Int):ButtonID
```
Gama11: Should be `getButton()`.

flixel/input/gamepad/ButtonIndex.hx
```diff
@@ -0,0 +1,405 @@
...
+import flixel.input.gamepad.FlxGamepad.FlxGamepadAnalogStick;
+import flixel.input.gamepad.FlxGamepad.GamepadModel;
+
+/**
+ * ...
+ * @author larsiusprime
+ */
+class ButtonIndex
+{
+	public var model:GamepadModel;
```
Gama11: Should probably be `(default, null)`

flixel/input/gamepad/ButtonIndex.hx
```diff
@@ -0,0 +1,405 @@
...
+	{
+		model = Model;
+	}
+	
+	/**
+	 * Given a ButtonID, return the raw hardware code
+	 * @param	buttonID the "universal" ButtonID
+	 * @return	the raw hardware code
+	 */
+
```
Gama11: There's isn't usually a newline here in flixel source code.

flixel/input/gamepad/ButtonID.hx
larsiusprime: Well... so I suppose this is a mark against the name "ButtonID." 

More properly, this is a list of numerically unique identifiers for various different control inputs on the device. You pass in one of these IDs to the general function you want, and then it uses the controllers model to map that to the propery raw hardware ID necessary for the final lookup. Internally it maps these universal ids to buttons in justpressed/pressed/justreleased lookups, but it's also used to fetch the axis for analog trigger lookups, and it is also used to fetch the analogStick object (with both axes) for the analog stick x/y axis lookup.

flixel/input/gamepad/ButtonIndex.hx
Gama11: I think it's better to lose the inline here. Inline makes sense for very short functions without complex control structures (loops, switch-case, even if-else I think) that are called often. None of that is the case here I think?

flixel/input/gamepad/ButtonIndex.hx
```diff
@@ -0,0 +1,405 @@
...
+			case LogitechButtonID.FOUR: Y;
+			case LogitechButtonID.NINE: BACK;
+			case LogitechButtonID.TEN: START;
+			case LogitechButtonID.LEFT_STICK_BTN: LEFT_STICK_BTN;
+			case LogitechButtonID.RIGHT_STICK_BTN: RIGHT_STICK_BTN;
+			case LogitechButtonID.FIVE: LEFT_SHOULDER;
+			case LogitechButtonID.SIX: RIGHT_SHOULDER;
+			case LogitechButtonID.SEVEN: LEFT_TRIGGER;
+			case LogitechButtonID.EIGHT: RIGHT_TRIGGER;
+			//case DPAD_UP:
```
Gama11: Why are these commented out?

flixel/input/gamepad/FlxGamepad.hx
```diff
@@ -15,12 +16,14 @@ import flash.system.Capabilities;
...
 	public var id(default, null):Int;
+	public var model(default, null):GamepadModel;
+	public var buttonIndex:ButtonIndex;
 	public var buttons(default, null):Array<FlxGamepadButton> = [];
 	public var connected(default, null):Bool = true;
 	
 	/**
 	 * Gamepad deadzone. Sets the sensibility. 
-	 * Less this number the more gamepad is sensible. Should be between 0.0 and 1.0.
+	 * Less this number the more sensitive the gamepad is. Should be between 0.0 and 1.0.
```
Gama11: :+1:

flixel/input/gamepad/ButtonIndex.hx
larsiusprime: Reminders -- I think I hadn't finished implementing these because I don't have a Logitech controller to test with.

flixel/input/gamepad/FlxGamepad.hx
```diff
@@ -43,18 +41,67 @@ class FlxGamepad implements IFlxDestroyable
...
 			deadZone = GlobalDeadZone;
 		}
+		
+		pressed = new FlxGamepadButtonList(FlxInputState.PRESSED, this);
+		justPressed = new FlxGamepadButtonList(FlxInputState.JUST_PRESSED, this);
+		justReleased = new FlxGamepadButtonList(FlxInputState.JUST_RELEASED, this);
+		analog = new FlxGamepadAnalogList(this);
+	}
+	
+	public inline function btnID(RawID:Int):ButtonID
```
Gama11: buttonID

Gama11: Any reason why these functions don't have a `get` prefix? Seems inconsistent.

larsiusprime: They're modeled exactly after what we have in FlxG.keys.

You do:
```FlxG.keys.justPressed.A```

Likewise here you do:

```gamepad.justPressed.A```

File: flixel/input/gamepad/FlxGamepad.hx (hunk 6 above)
Comment (Gama11) [inline flixel/input/gamepad/FlxGamepad.hx]: Ah, the comment confused me. Might just be me, but the way it's worded it could be taken as "the range is (always) (-1, 1), but we're only normalizing in legacy for some reason. :)

File: flixel/input/gamepad/FlxGamepad.hx (hunk 18 above)
Comment (larsiusprime) [inline flixel/input/gamepad/FlxGamepad.hx]: Oh wait, were you talking about "btnID" and not pressed/justPressed/justReleased ?

File: flixel/input/gamepad/FlxGamepadManager.hx [hunk 19]
```diff
@@ -549,4 +657,11 @@ class FlxGamepadManager implements IFlxInputManager
 		}
 		return globalDeadZone;
 	}
+	
+	#if (!flash && !next)
+	
+	private static inline var JOYSTICK_BUTTON_UP: "buttonUp";
```
Gama11: Nice, this even added some self-documentation because of the added `JOYSTICK`. :+1:

flixel/input/gamepad/XInputButtonID.hx
Gama11: Gotcha.

flixel/input/gamepad/FlxGamepad.hx
larsiusprime: I'll update to clarify.

flixel/input/gamepad/FlxGamepad.hx
Gama11: That and `rawID` / `rawAnalogStick`.

flixel/input/gamepad/ButtonIndex.hx
larsiusprime: Actually, it makes sense to leave this settable, because users might want to change their mapping to handle various edge cases (the computer thinks it's "Generic XInput" but I KNOW it's XYZ!), and updating this directly should totally work.

IE:
```
gamepad.model = Logitech;
```

Right after you do that you've effectively remapped it to Logitech controls.

Gama11:
The gamepad package has a lot of classes now... Might make sense to have a `flixel.input.gamepad.buttonID` package for all the ButtonID related classes? Thoughts?

Gama11:
Btw, I'd prefer `FlxGamepadButtonID` just for consistency, even it it's a little on the long side. Can't really think of anything better right now.

Adding the `Flx` prefix to the other types shouldn't be as much of an issue, for example `GamepadModel` -> `FlxGamepadModel`.

flixel/input/gamepad/OUYAButtonID.hx
larsiusprime: Can this be resolved by just expanding to "BUTTON" or would you prefer if it said "CLICK" ? The purpose being simply to disambiguate the analog from the digital input that occupy the same physical control.

Gama11:
Sorry for my nitpicking regarding code style, can't help it. ;)

larsiusprime:
Luckily I'm in a good mood today! :P

flixel/input/gamepad/ButtonIndex.hx
Gama11: Should probably have a short comment explaining that here then?

flixel/input/gamepad/ButtonIndex.hx
```diff
@@ -0,0 +1,398 @@
...
+			//case START:
+			//case DPAD_UP:
+			//case DPAD_DOWN:
+			//case DPAD_LEFT:
+			//case DPAD_RIGHT:
+			default: NONE;
+		}
+	}
+	
+	public function getBtnLogitech(rawID:Int):ButtonID
```
Gama11: These function also use the `Btn` abbreviation.

flixel/input/gamepad/ButtonIndex.hx
Gama11: I have one if you need someone to test something.

flixel/input/gamepad/ButtonIndex.hx
larsiusprime: Sure.

flixel/input/gamepad/ButtonIndex.hx
larsiusprime: Cool, I'll fill out the logic and if you could run the gamepad demo with your logitech controller (perhaps putting some traces in FlxInput to trace out what hardware ID's are being pressed) and deduce what each button input should be, that'd be swell.

flixel/input/gamepad/ButtonIndex.hx
larsiusprime: Oh wait, you're right this is the model variable in ButtonIndex and not Gamepad! This should not be publicly settable. It's Gamepad's that can be settable, I gotta check that logic real quick.

flixel/input/gamepad/OUYAButtonID.hx
Gama11: I'm just not a fan of the abbreviation, hadn't even thought of the click vs button thing. Is it correct to say that these are buttons / is there some official name for it? They certainly aren't "buttons" in the traditional sense.

larsiusprime: CLICK is probably our best bet. They're only buttons in the abstract electronic switch sense, not in terms of physical affordance.

flixel/input/gamepad/ButtonIndex.hx
```diff
@@ -0,0 +1,399 @@
+package flixel.input.gamepad;
+import flixel.input.gamepad.FlxGamepadButtonID;
+import flixel.input.gamepad.FlxGamepad.FlxGamepadAnalogStick;
+import flixel.input.gamepad.FlxGamepad.GamepadModel;
+
+/**
+ * ...
+ * @author larsiusprime
+ */
+class ButtonIndex
```
Gama11: Would `FlxGamepadMapping` / `Mapper` be a better name for this?

flixel/input/gamepad/FlxGamepad.hx
```diff
@@ -486,4 +630,14 @@ enum FlxAxes
...
 	Y;
+}
+
+enum GamepadModel
+{
+	Logitech;
+	OUYA;
+	PS3;
+	PS4;
+	Xbox;
```
Gama11: Shouldn't this be Xbox360 instead? Or rather, is Xbox360 identical to other xbox controllers?

flixel/input/gamepad/ButtonIndex.hx
larsiusprime: Yeah works for me. FlxGamepadMap is probably no good b/c that suggests it's an actual Map type, I like "Mapping"

flixel/input/gamepad/FlxGamepad.hx
larsiusprime: It probably should be. There's an XBox ONE controller too which we shouldn't assume is the same. I was just following what we had already, which just had "XBoxInput" -- I'll change all xbox classes to XBox360

flixel/input/gamepad/OUYAButtonID.hx
Gama11: It's also shorter. :)

flixel/input/gamepad/XInputButtonID.hx
```diff
@@ -0,0 +1,40 @@
...
+	public static inline var DPAD_UP:Int = 17;
+	public static inline var DPAD_DOWN:Int = 18;
+	public static inline var DPAD_LEFT:Int = 19;
+	public static inline var DPAD_RIGHT:Int = 20;
+	
+	/**
+	* Axis array indicies
+	*/
+	public static var LEFT_ANALOG_STICK(default, null):FlxGamepadAnalogStick = [FlxAxes.X => 0, FlxAxes.Y => 1];
+	public static var RIGHT_ANALOG_STICK(default, null):FlxGamepadAnalogStick = [FlxAxes.X => 2, FlxAxes.Y => 3];
```
Gama11: In hindsight, it seems like a weird design decision on my part to make this a map - it would probably better as a `FlxGamepadAnalogStick` class with `x` and `y` fields. Thoughts?

flixel/input/gamepad/ButtonIndex.hx
Gama11: :+1:

flixel/input/gamepad/FlxGamepad.hx
Gama11: Also, I'm not sure how `XInput` fits into this - my understanding is that controllers *are* not really "XInput", they only *use* it (or don't). Right?

larsiusprime: I'm just going strictly off of what I'm getting from the device.name from the Flash API. 

On flash target, I get nice useful strings that tell me what it is, "Microsoft X-Box 360 pad"
On legacy target, I get nothing, and just default to 360 b/c I have no choice.
On next target, I get "XInput device" and the numerical mappings are different than what worked for the XBox 360 in windows on legacy.

So maybe this is a conversation to have with @singmajesty about what the openfl-next gamepad API is returning for the device string names.

flixel/input/gamepad/FlxGamepad.hx
```diff
@@ -43,18 +44,75 @@ class FlxGamepad implements IFlxDestroyable
...
 	
-	public function new(ID:Int, GlobalDeadZone:Float = 0) 
+	#if (flash)
+	private var _isChrome:Bool = false;
+	#end
+	
+	/**
+	 * Helper class to check if a button is pressed.
+	 */
+	public var pressed:FlxGamepadButtonList;
```
Gama11: These should all be `(default, null)` - just realized that isn't even the case in `FlxKeyManager` where you copy-pasted them from. :o

flixel/input/gamepad/FlxGamepad.hx
```diff
@@ -319,40 +436,77 @@ class FlxGamepad implements IFlxDestroyable
...
 	{
 		var axisValue = getAnalogueAxisValue(FlxAxes.Y, Axes);
 		
 		// the y axis is inverted on the Xbox gamepad in flash for some reason - but not in Chrome!
-		// WARNING: this causes unnecessary string allocations - we should remove this hack when possible.
-		#if flash
-		if ((_device != null) && _device.enabled && (_device.name.indexOf("Xbox") != -1) && 
-		   (Capabilities.manufacturer != "Google Pepper"))
+		#if (flash)
+		if (model == Xbox && !_isChrome)
```
Gama11: Still a hack, but at least it reads much more nicely now. ;)

flixel/input/gamepad/buttons/XBox360ButtonID.hx
```diff
@@ -1,11 +1,11 @@
-package flixel.input.gamepad;
+package flixel.input.gamepad.buttons;
 
 import flixel.input.gamepad.FlxGamepad;
 
 /**
  * Button IDs for Xbox 360 controllers
  */
-class XboxButtonID
+class XBox360ButtonID
```
Gama11: Technically, these classes having `Button` in the name is incorrect since they also contain axis IDs (it's also longer). So, if you're not tired of find-and-replace yet... :D Due to the package change the API for these classes isn't backwards compatible anyway.

Gama11: I guess that also makes the package name incorrect, maybe that one should be `id` instead of `buttons`?

larsiusprime: Sure, so XBox360ID in package id, and so forth?

Gama11: Yeah. I think that makes more sense?

FlxBaseGamepadButtonList -> FlxBaseGamepadList too

larsiusprime:
@Gama11: I believe I have now responded to each and every comment you've made so far. Have I passed the "gama gauntlet" ? :D

flixel/input/gamepad/FlxGamepad.hx
```diff
@@ -486,4 +633,14 @@ enum FlxAxes
 {
 	X;
 	Y;
+}
+
+enum GamepadModel
```
Gama11: This still needs the `Flx`.

larsiusprime:
One last thought -- since we removed "ButtonID" most places should I change "FlxGamepadButtonID" to just "FlxGamepadInputID" ? (FlxGamepadID would be ambiguous, ie, ID of the gamepad itself)

Gama11:
Yeah, I think that makes sense.

Have you seen my comment about changing `FlxGamepadAnalogStick` from a Map to a class? I can do that myself after this is merged though.

larsiusprime:
Ah, forgot about that one. Yeah, that's a good idea, feel free to fix that after this is merged.

flixel/input/gamepad/FlxGamepadButtonID.hx
```diff
@@ -0,0 +1,70 @@
...
+	/**digital click of the right analog stick*/
+	var RIGHT_STICK_CLICK =  9;
+	/**this is the "XBox" or "PS" or "home" button in the center*/
+	var GUIDE             = 10;
+	var DPAD_UP           = 11;
+	var DPAD_DOWN         = 12;
+	var DPAD_LEFT         = 13;
+	var DPAD_RIGHT        = 14;
+	/**digital click at end of left analog trigger's squeeze (if available)*/
+	var LEFT_TRIGGER_BTN  = 15;
```
Gama11: Still two `BTN`s here.

flixel/input/gamepad/FlxGamepadMapping.hx
```diff
@@ -0,0 +1,391 @@
...
+import flixel.input.gamepad.id.XInputID;
+import flixel.input.gamepad.FlxGamepadButtonID;
+import flixel.input.gamepad.FlxGamepad.FlxGamepadAnalogStick;
+import flixel.input.gamepad.FlxGamepad.GamepadModel;
+
+/**
+ * ...
+ * @author larsiusprime
+ */
+class FlxGamepadMapping
```
Gama11: This class also still uses the inprecise "button" / "btn" terminology.

Gama11:
I thought you wanted to rename to `FlxGamepadInputID` since `FlxGamepadID` is ambigious?

larsiusprime:
D'oh. It's been a long day.

jgranick:
I think we should be able to expose support for adding additional gamepad bindings (on the SDL side) as well if we want, also, please feel free to provide feedback regarding how the Lime API works here

larsiusprime:
Anything else @Gama11  ?

Gama11:
Thanks for this amazing contribution!

PR #1505, opened 2015-05-22
Fix for flash

larsiusprime:
You have to explicitly check for "xbox 360" before you check for "xinput" because "xinput" is included in the string name of the xbox 360 controller identifier on flash target only.


larsiusprime:
Argh, history got a bit trashed here. Anyways, it's a simple change and you can cherry pick it.

Gama11:
Ah, now that code makes more sense. :/ I should probably also add a comment explaining why this is important...

PR #1516, opened 2015-05-31
Stop FlxButton's animations from over triggering. Closes #1365.

### LINKED ISSUE #1365: FlxButton: Animations are over-triggering
For more information and @01010111's code snippet go here: http://haxeflixel.com/forum/?place=msg%2Fhaxeflixel%2F3aIzye_Uldw%2FKfDSQYCkC_YJ

The gist of it is that on every update the animation is being replayed, over and over again, even if it's the current animation:

https://github.com/HaxeFlixel/flixel/blob/29c5827107c1a81e0cd9ab68c8e6e2d0a77cad17/flixel/ui/FlxButton.hx#L264-284

Now, you can just add a safety check, something like: 
```haxe
//Making sure animations won't trigger more than once per input
private var _status = -1;
...
if (_status != status) 
  {
    updateStatusAnimation();
    _status = status;
   }
```
The problem, maybe, is that people may intuitively assume continuity, i.e. if their "normal" animation is not looped, and was already finished playing once, it means that when the button is in "NORMAL" status again, the animation should be where it was (as they left it), not replay all over again. 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/5779175-flxbutton-animations-are-over-triggering?utm_campaign=plugin&utm_content=tracker%2F254189&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F254189&utm_medium=issues&utm_source=github).
</bountysource-plugin>

sruloart:


Beeblerox:
looks fine to me

flixel/ui/FlxButton.hx
```diff
@@ -178,13 +178,20 @@ class FlxTypedButton<T:FlxSprite> extends FlxSprite implements IFlxInput
...
 	
 	/**
 	 * The input currently pressing this button, if none, it's null. Needed to check for its release.
 	 */
 	private var currentInput:IFlxInput;
+
+	/**
+	 * A helper for making sure animations won't trigger more than once per input.
+	 */
+	private var _status = -1;
```
Gama11: `lastStatus` would be a more descriptive name, then you could get rid of the comment too. No underscore.

Gama11:
I think this behavior makes more sense, otherwise you can't really have non-looped button animations.

Would be nice to have a unit-test for this.

flixel/ui/FlxButton.hx
sruloart: I was thinking about the far future, when someone who isn't aware of all the history would want to change it to a public var, but I guess we'll have to risk it :)

sruloart:
Yeah, I'll be right on it once I'll figure out why SourceTree is so slow, and also read a little about munit :D

flixel/ui/FlxButton.hx
Gama11: That wouldn't really be an accurate description anymore if it becomes public API, though. ;)

Gama11:
SourceTree being slow is pretty normal in my experience.. at least on Windows. :/

sruloart:
A unit test contribution document would sure be very useful... 
And on a completely unrelated note I think I'm going to need some help with this unit test :P

Gama11:
Nevermind, I added a test myself.

Some docs might be nice, yeah.. Anything in particular you struggled with?

sruloart:
Thanks! I'm not really used to unit testing so there's that :) I was doing something very close but I didn't want to make you work twice (fixing my horrible test). I just need more details on how and when the different metadatas and `Assert`-s should be used, along side some general description of HaxeFlixel's test stuff like `step()`; examples, basically. The current unit tests aren't very descriptive in that sense.

Issue #1030, opened 2014-04-24
More control over camera shake


JoeCreates:
It would be nice to have a bit more control over camera shaking, for example, having different types of shake, and being able to tween the intensity.
##

MSGhero:
I'll do this too

JoeCreates:
I'm having another stab at this. Some notes:

* I want Camera scroll shaking so that parallax is maintained. This is really hard at the moment sue to scroll being a FlxPoint used in multiple places for rendering and assumed to give the render position in many others. I think I have a way that would leave the scroll interface the same, though.
* I also want angular shake
* While I'll make it so that tweening can be used, it is not an ideal solution, as the best camera shakes are cumulative, as if the camera is on a spring. I'll be implementing it to be more like a spring.

PR #1522, opened 2015-06-09
Update FlxMath.hx

Tiago-Ling:
The minimum value for floats in BlackBerry should be the same used for Flash, js and iOS.

This fixes problems where FlxMath.MIN_VALUE_FLOAT is used on BlackBerry (e.g. FlxG.camera.fade)

### DIFF
--- a/flixel/math/FlxMath.hx
+++ b/flixel/math/FlxMath.hx
@@ -12,7 +12,7 @@ import flixel.input.touch.FlxTouch;
  */
 class FlxMath
 {	
-	#if (flash || js || ios)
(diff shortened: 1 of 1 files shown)


Gama11:
Hm... this is somewhat ridicolous, I think this has caused similar issues before on iOS. I don't think there's any good reason to use `FlxMath.MIN_VALUE_FLOAT` in `FlxCamera` here in the first place, another sufficiently small value should do the trick (`0.000001` or whatever, that would make it more consistent with the `0.999999` value used for `_fxFadeIn` too).

Eventually these should probably be replaced by tweens anyway (see #1030).

PR #1526, opened 2015-06-10
Update FlxSprite.hx

SeiferTim:
Made `X` and `Y` arguments for `new` skippable.
Allows for simpler usage, if desired: `sprite = FlxSprite.new('images/image.png');`

### DIFF
--- a/flixel/FlxSprite.hx
+++ b/flixel/FlxSprite.hx
@@ -209,7 +209,7 @@ class FlxSprite extends FlxObject
 	 * @param	Y				The initial Y position of the sprite.
 	 * @param	SimpleGraphic	The graphic you want to display (OPTIONAL - for simple stuff only, do NOT use for animated images!).
 	 */
-	public function new(X:Float = 0, Y:Float = 0, ?SimpleGraphic:FlxGraphicAsset)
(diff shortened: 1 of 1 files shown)


Gama11:
Not sure if that's a good idea, that leads to them becoming `Null<Float>`. Somewhat inconsistent with the rest of the API too..

SeiferTim:
Well, what's the point in being able to have skippable, optional attributes, if we don't use them? 
It's really ugly to have a bunch of `object = new FlxSprite(0, 0, "assets/image.png");`
Also, since the constructor is setup as `new(?X:Float=0...`, does it default to 0 whenever it's null?

Gama11:
Hm, seems like it... Wouldn't have expected that behavior. http://try.haxe.org/#c633d

SeiferTim:
Yeah, I think if it was `new(?X:Float, ?Y:Float...` then it would be `Null<Float>` and, I agree, we don't want that, but the `= 0` is preventing that.
I think this helps when someone wants to do something quick and simple with FlxSprites, and positon it after the fact...

Gama11:
Oh, it's definitely still `Null<Float>` as you can see in the output of `$type`:

`Warning : ?i : Null<Int> -> Void`

It's just that calling it with `null` explicitly doesn't seem to lead to the value being `null`.

PR #1528, opened 2015-06-12
Fix for next: Keyboard listeners would not work after the window lost…

larsiusprime:
… focus. This seems to fix it and doesn't seem to break the console's focus logic.

### DIFF
--- a/flixel/system/debug/Console.hx
+++ b/flixel/system/debug/Console.hx
@@ -63,7 +63,7 @@ class Console extends Window
 	 */
 	private var _input:TextField;
 	
-	#if (cpp || neko)
+	#if (!next && sys)
 	private var inputMouseDown:Bool = false;
 	private var stageMouseDown:Bool = false;
 	#end
@@ -129,7 +129,7 @@ class Console extends Window
 		#end
 	}
 	
-	#if (cpp || neko)
+	#if (!next && sys)
 	override public function update():Void
 	{
 		super.update();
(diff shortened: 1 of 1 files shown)


flixel/system/debug/Console.hx
```diff
@@ -112,7 +112,7 @@ class Console extends Window
 		_input.addEventListener(FocusEvent.FOCUS_OUT, onFocusLost);
 		_input.addEventListener(KeyboardEvent.KEY_DOWN, onKeyPress);
 		
-		#if (cpp || neko) // workaround for broken TextField focus on native
```
Gama11: There are two more code sections that use these conditionals, those should probably use `!next && sys` too?

larsiusprime: Yeah probably!

larsiusprime: Hmm... I wonder if the legacy fix in that custom update function is the culprit. If we add the new conditional I wonder if it will "just work" in next.

Gama11: By "just work" you mean keyboard listeners not being broken, rather than the console working (there's no input text field support in next yet is there?).

larsiusprime: Nope, it definitely needs the (sys && next) exit-early guards.

larsiusprime: Yes, by "just work" I meant that I suspected that perhaps it was the old (cpp/neko legacy) hacks that were causing the weird behavior in next, but it appears simplying guarding them off with ifdefs doesn't result in correct behavior.

larsiusprime: And yes, this does not fix input text field support in any case. It just gets the key events back to where they should be. (Though as a bonus, the focus in/out from clicking on and off the debugger console seems to have the right behavior now with this fix, even if you can't type stuff in yet)

PR #1529, opened 2015-06-15
BitmapFrontEnd: fix edge case where rarely-misaligned object keys cau…

larsiusprime:
…se images to remain in the cache forever

### DIFF
--- a/flixel/system/frontEnds/BitmapFrontEnd.hx
+++ b/flixel/system/frontEnds/BitmapFrontEnd.hx
@@ -397,7 +397,7 @@ class BitmapFrontEnd
 			obj = _cache.get(key);
 			if (obj != null && !obj.persist)
 			{
-				removeByKey(obj.key);
+				removeByKey(key);
 			}
 		}
 	}
@@ -417,7 +417,7 @@ class BitmapFrontEnd
 				obj = _cache.get(key);
 				if (obj != null && obj.useCount <= 0 && !obj.persist && obj.destroyOnNoUse)
 				{
-					removeByKey(obj.key);
+					removeByKey(key);
 				}
 			}
 		}


larsiusprime:
The clearCache functions were assuming that the ".key" member variable of each object in the cache would always be the same as the key that is used to fetch it. I found a few edge cases in my own code that violated this -- it seems safest to always use the same key that fetches an object to remove it if the goal is to clear the entire cache, yes? Unless there was some subtle reason for using "obj.key" instead of just "key" that I'm overlooking.

Gama11:
> The clearCache functions were assuming that the ".key" member variable of each object in the cache would always be the same as the key that is used to fetch it.

Isn't that (or rather, shouldn't that be) the case? The docs on `FlxGraphic#key` literally say: `Key in BitmapFrontEnd cache`

> I found a few edge cases in my own code that violated this

Got any concrete example? Would be useful for a unit test.

larsiusprime:
It's a little hard to post the concrete example as it's deep in the guts of Defender's Quest and highly contextual, but the gist is that I have a methodology where I scale an asset using this utility function:

https://github.com/HaxeFlixel/flixel-ui/blob/dev/flixel/addons/ui/U.hx#L1224

Then I store it with a generated key. So if I'm loading "assets/foo.png" and scaling it by 0.5x, I store the scaled version in the cache as "assets/foo.png_scale0.5x"

Somewhere along the line, although the keys are *supposed* to be the same, they get mis-aligned, and so the entry for "assets/foo.png_scale0.5x" has a cachedGraphics object whose .key member is "assets/foo.png".

So one could argue "fix your code" -- which is certainly true, I should find out why this misalignment happens. However, my argument here is that the user expects the clearCache function to do just that -- clear the cache of everything that's not marked persist. However, cachedGraphic has a public variable ("key") that could be modified at any time -- and clearCache is treating that as a guaranteed state when it clearly isn't. 

So even if the object.key variable is used elsewhere, in clearCache I can't see why it should accessed an object with the map key during the clear loop, only to use the member variable to do the actual clear command, when the former is *guaranteed* to match, but the second isn't.

That's my spiel :P

Gama11:
I agree, the code definitely makes more sense with your fix. I was mainly wondering if there might be  other issues in flixel that caused this misalignment in the first place.

Gama11:
Some of those properties in FlxGraphic should probably be read-only...

larsiusprime:
The misalignments are probably only happening if the user misbehaves. Likely I would say that's only going to be happening in flixel-ui's scaling functions, but we could do a quick audit of flixel itself and see where the member variable "key" is being used.

PR #1532, opened 2015-06-17
FlxText: bring in line with new Openfl changes.

larsiusprime:


flixel/text/FlxText.hx
```diff
@@ -532,12 +532,16 @@ class FlxText extends FlxSprite
...
 	}
 	
-	private inline function get_size():Float
+	private inline function get_size():Int
 	{
-		return _defaultFormat.size;
+		#if openfl_legacy
+			return Std.int(_defaultFormat.size);
+		#else
+			return _defaultFormat.size;
```
Gama11: This code seems to trigger even when compiling with legacy (leading to a compiler error: `Null<Float> should be Int`)

larsiusprime:
I did away with some of the #ifs, just using an Std.int(). Does that get optimized away if the underyling type is already an int?

MSGhero:
@larsiusprime yes in js without analyzer, probably the rest too

Gama11:
Probably not that big of a deal...

The flixel-ui counterpart of this (HaxeFlixel/flixel-ui#107) still uses conditionals btw.

PR #1534, opened 2015-06-20
fastSin/Cos, closes #1255

### LINKED ISSUE #1255: Optimize math
The maths need a bit of optimization. I realize it's mostly small stuff, but it could make a difference.

isEven / isOdd could be inlined and replaced with a simple return, instead of an if-else construct, which could cause the processor to mispredict the branch.

Fast math could be introduced, and should replace the math used in the engine itself, while also being open to users.

You can see some examples on what kind of fastmath i'm talking about here:

This is the Away3D project. They have a MathUtils class which introduces fast math available thanks to haXe. They differentiate int and float functions, which is great. Also very interesting to note is the sqrt and invSqrt functions. Those are way faster.
https://github.com/away3d/away3dlite-core-haxe/blob/master/src/away3dlite/haxeutils/MathUtils.hx

Here's polygonal's  Mathematics class, which has some additional functions.
https://github.com/polygonal/core/blob/master/src/de/polygonal/core/math/Mathematics.hx

Actually, the whole math package from polygonal would be helfpul, cause it has vectors and other stuff: https://github.com/polygonal/core/tree/master/src/de/polygonal/core/math - notice how much stuff is inlined. This would help HaxeFlixel improve its performance a bit on the most intensive stuff.

Thanks for the attention!

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/3453519-optimize-math?utm_campaign=plugin&utm_content=tracker%2F254189&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F254189&utm_medium=issues&utm_source=github).
</bountysource-plugin>

MSGhero:
The formula is a mathematical simplification of the original formula in `FlxPoint.rotate()`, except it's faster and more accurate.

It's 2-6 times faster within ~0.05% error versus Math.sin and cos, so it could probably be used elsewhere in flixel.

### DIFF
--- a/flixel/math/FlxMath.hx
+++ b/flixel/math/FlxMath.hx
@@ -110,16 +110,9 @@ class FlxMath
 	 * @param	n	The number to check 
 	 * @return	True if the given number is odd. False if the given number is even.
 	 */
-	public static function isOdd(n:Float):Bool
+	public static inline function isOdd(n:Float):Bool
 	{
-		if ((Std.int(n) & 1) != 0)
-		{
-			return true;
-		}
-		else
-		{
-			return false;
-		}
+		return (Std.int(n) & 1) != 0;
 	}
 	
 	/**
@@ -130,14 +123,7 @@ class FlxMath
 	 */
 	public static function isEven(n:Float):Bool
 	{
-		if ((Std.int(n) & 1) != 0)
-		{
-			return false;
-		}
-		else
-		{
-			return true;
-		}
+		return (Std.int(n) & 1) == 0;
 	}
 	
 	/**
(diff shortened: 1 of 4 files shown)


flixel/math/FlxMath.hx
```diff
@@ -527,6 +513,49 @@ class FlxMath
...
+		{
+			f += (Math.ceil( -f) >> 1) << 1;
+		}
+		
+		// this approx only works for -pi <= rads <= pi, but it's quite accurate in this region
+		if (f > 0)
+		{
+			return f * (3.1 + f * (0.5 + f * ( -7.2 + f * 3.6)));
+		}
+
```
Gama11: Weird newline here? :)

flixel/math/FlxPoint.hx
```diff
@@ -319,72 +319,9 @@ class FlxPoint implements IFlxPooled
...
-			if (cos < 0)
-			{
-				cos = .225 * (cos *-cos - cos) + cos;
-			}
-			else
-			{
-				cos = .225 * (cos * cos - cos) + cos;
-			}
-		}
+		var sin:Float = FlxMath.fastSin(radians);
```
Gama11: Wow... I always disliked how long this function was, amazing cleanup! :+1:

Gama11:
Have you done any tests to make sure these work correctly? Would be nice to have some unit tests for this.

flixel/math/FlxMath.hx
MSGhero: Did some algebra and that, plus `f / pi` at the top equals the old one :+1:

MSGhero:
Yeah, http://try.haxe.org/#C620a

I tested the original, this one, a taylor series approx, and the actual sin/cos for speed (not shown in try) and accuracy. This one ended up being fastest and the most accurate alternative. Some of those weird constants in the original are actually multiples or powers of pi that I factored out, and that ended up simplifying the math.

I also plotted them in matlab to be sure, and everything looked fine.

MSGhero:
This might close #1255 unless you want to go farther with it. I'm fairly sure `sqrt` is faster than `1/invSqrt` on modern systems, but I could be wrong.

Gama11:
Thanks! :+1:

MSGhero:
`isEven` should have been inlined, whoops.

Gama11:
Fixed: 306bdfcdf

PR #1537, opened 2015-06-23
Fix potential null error in FlxRandom#color(), closes #1536

### LINKED ISSUE #1536: FlxRandom.color has broken null checks.
```FlxG.random.color(FlxColor.GRAY)```, for example, breaks because of trying to access properties of the max color which is null.

In FlxRandom.color() there are checks like ```if (Max == null) {red = int(0, Max.red)}```. I'm not sure what the intended behaviour is.

@steverichey

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/22782967-flxrandom-color-has-broken-null-checks?utm_campaign=plugin&utm_content=tracker%2F254189&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F254189&utm_medium=issues&utm_source=github).
</bountysource-plugin>

steverichey:
The `Min` and `Max` null checks were swapped.

### DIFF
--- a/flixel/math/FlxRandom.hx
+++ b/flixel/math/FlxRandom.hx
@@ -351,14 +351,14 @@ class FlxRandom
 			blue = int(0, 255);
 			alpha = Alpha == null ? int(0, 255) : Alpha;
 		}
-		else if (Min == null)
+		else if (Max == null)
 		{
 			red = int(Min.red, 255);
 			green = GreyScale ? red : int(Min.green, 255);
 			blue = GreyScale ? red : int(Min.blue, 255);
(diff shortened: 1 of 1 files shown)


Gama11:
Thanks!

Gama11:
This sort of thing is very easy to unit-test, and it's good practice to include a unit test for bugs like this to prevent them from happening again (plus, they also act as a specifiction telling you how a method should behave).

I added one (aef7e06b19), but consider writing one yourself in the future. :)

PR #1542, opened 2015-06-25
Update FlxColor.hx

sruloart:
spelling...
No spell-check in FD for some reason :)


Gama11:
Spell-checking would be pretty annoying / hard to get right. You'd have to have a way to separate English text from code. Not to mention that the docs might not even be written in English...

MSGhero:
You could run nick.ml aka make me proofread them ;)

PR #1555, opened 2015-07-05
- Adding method updateTouchingFlags to FlxObject, closes #1556

### LINKED ISSUE #1556: Performing overlap call that also updates the "touching" flags of objects
Hello,
I used to use flixel a lot in the past, mostly for AS3 projects, and recently I am toying with some other ideas for which I am trying haxeFlixel: I am really liking it, it is very cool to have an actively developed, cross-platform version of flixel.
I recently found myself into wanting to call FlxG.overlap() but also wanting to have the "touching" variables of the objects involved update. This did not seem to be possible with current flixel, am I right?
Therefore, I have changed it: https://github.com/HaxeFlixel/flixel/pull/1555
I do not know how is contribution to HaxeFlixel organized, but if it looks interesting, I can polish (like, adding a unit test).
Hope it helps!

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/23736707-performing-overlap-call-that-also-updates-the-touching-flags-of-objects?utm_campaign=plugin&utm_content=tracker%2F254189&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F254189&utm_medium=issues&utm_source=github).
</bountysource-plugin>

winterismute:
When a "collide" call is performed, flixel updates the flags of the objects to signal where they touched something in the 4 directions. This does not happen when an overlap call is issued, because updating the flags is performed inside the separate methods.
This pull request changes this, and allows to write

FlxG.overlap(obj1, obj2, hittingObjects, FlxObject.updateTouchingFlags);

and having the "touching" variables of obj1 and obj2 updated when reaching the function "hittingObjects".

### DIFF
--- a/flixel/FlxObject.hx
+++ b/flixel/FlxObject.hx
@@ -83,6 +83,78 @@ class FlxObject extends FlxBasic
 	}
 	
 	/**
+	 * Similar to "separate", but only checks whether any overlap is found and updates 
+	 * the "touching" flags of the input objects, but no separation is performed.
+	 * 
+	 * @param	Object1		Any FlxObject.
+	 * @param	Object2		Any other FlxObject.
+	 * @return	Whether the objects in fact touched.
+	 */
+	public static function updateTouchingFlags(Object1:FlxObject, Object2:FlxObject):Bool
+	{
+		var touchingX:Bool = updateTouchingFlagsX(Object1, Object2);
+		var touchingY:Bool = updateTouchingFlagsY(Object1, Object2);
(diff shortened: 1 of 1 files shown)


Gama11:
Thanks for this awesome change! Not only does it allow the logic to be used in other scenarios like you described, it also makes it a lot more readable since it's now split up into smaller functions. :+1:

PR #1557, opened 2015-07-07
New signal (preStateCreate), dispatched with FlxState as parameter

j1mmie:
I found this useful while working on a game with dependency injection and would like to propose it for the main line. I wanted a single listener to detect every time the state has changed, and automatically inject dependencies into that state.

The solution was to create a new signal, `preStateCreate`, that includes the newly-change-to `FlxState` as the payload of the signal. This new signal is dispatched after the stateChange signal, and before the `.create()` method is called on the new state (hence, "pre state create").

### DIFF
--- a/flixel/FlxGame.hx
+++ b/flixel/FlxGame.hx
@@ -611,6 +611,8 @@ class FlxGame extends Sprite
 		// Finally assign and create the new state
 		_state = _requestedState;
 		
+		FlxG.signals.preStateCreate.dispatch(_state);
+		
 		_state.create();
 		
 		if (_gameJustStarted)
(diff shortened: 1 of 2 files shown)


Gama11:
Thanks!

That's actually a very elegant solution to this problem (without any breaking API changes!). Some people simply add arguments to the FlxState child class constructors, but that breaks any usage of `FlxG.resetState` since that attempts to create a new instance using Reflection without any arguments.

There's also the option to use static variables, but that's kind of a hack as well.

j1mmie:
Yeah I was originally adding an accessor to FlxGame for _newState, and using the old stateChanged signal. That seemed a bit hacky.

Thanks for the review. I'll check to see if there are some docs I can modify as a result of this addition.

Gama11:
This also makes `FlxSignal#addOnce()` seem a lot more useful. :)

PR #1562, opened 2015-07-14
Frame Draw Changes

SeiferTim:
This is a change that allows you to use `framePixels` on native targets in order to change the current frame of a `FlxSprite` non-destructively.
This allows behavior that works the same between targets.

Before this change, doing something like this:
```
override function draw():Void
{
  if (putColor)
  {
    dirty = true;
    getFlxFrameBitmapData();
    framePixels.copyPixels(anotherBitmap, anotherBitmap.rect, _flashPointZero, framePixels, _flashPointZero, true);
  }  
  super.draw();
}
```
Would work on flash, but not on native targets.

This also includes a minor fix that caused a crash with this change, and would have been causing memory leakages before to `FlxFrame` - when `FlxSprite.destroy` is called, it says: `frame = null`, which calls `FlxSprite.set_frame` which was not checking if it was being passed a null or not, and would try to make `frame` be a new `FlxFrame`, etc.

### DIFF
--- a/flixel/FlxCamera.hx
+++ b/flixel/FlxCamera.hx
@@ -517,6 +517,11 @@ class FlxCamera extends FlxBasic
 	
 	public function drawPixels(?frame:FlxFrame, ?pixels:BitmapData, matrix:FlxMatrix, cr:Float = 1.0, cg:Float = 1.0, cb:Float = 1.0, ca:Float = 1.0, blend:BlendMode = null, smoothing:Bool = false):Void
 	{
+		if (pixels != null)
+		{
+			frame = frame.copyTo();
+			frame.parent.bitmap = pixels.clone();
+		}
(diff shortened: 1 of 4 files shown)


gamedevsam:
This seems fine to me, any opposition?

Gama11:
I'd like @Beeblerox to review this one since it's rendering-related.

Beeblerox:
i think it might create a lot of garbage

Beeblerox:
need to think about better solution. but this is a good start

Beeblerox:
made my version of it: https://github.com/HaxeFlixel/flixel/commit/970c7ca9021dd0946a151a9c5b04b5117b10456c

gamedevsam:
Should we close this then?

Beeblerox:
Yes, i think so

Beeblerox:
i don't like my solution now. 
Probably will add `useFramePixels` property which will handle this case. But need to consoder other options

Issue #1462, opened 2015-03-28
Game not started on iOS and mac safari when targeting to openfl-html5 backend


fysx:
when I target any haxeflixel project to openfl-html5 backend? game not loading on iOS and Mac Safari.
Error in console:
TypeError: undefined is not an object (evaluating 'a.apply') in sound.js.min:17:894

Project is clean (Hello world), all haxe libraries are up to date. Any projects on simple openFl works well.
##

fysx:
if i removed following lines:

	urls.push("assets/sounds/beep.mp3");
	types.push("MUSIC");
	urls.push("assets/sounds/flixel.mp3");
	types.push("MUSIC");
	urls.push("assets/sounds/beep.ogg");
	types.push("SOUND");
	urls.push("assets/sounds/flixel.ogg");
	types.push("SOUND");

 in compiled js file, evereything works well

fysx:
I converted beep.mp3 and flixel.mp3. It helps.

Masadow:
This issue should not be closed

AGulev:
This issue should not be closed

sruloart:
Safari can't play .ogg files, so you better use .wav files, .m4a files, or disregard the legal implications and use .mp3 files. 

You should also make sure that .ogg files aren't being included with the build `unless="IE | Safari"` (not sure if there's a Safari flag actually, https://github.com/HaxeFlixel/flixel/blob/3.3.11/include.xml) or just delete the default files directly from your assets folder.

MSGhero:
@sruloart You can't reasonably haxeflag for each browser, it's not like you can control which version gets loaded per user unless it's running on your own site. It looks like there are some plugins for Safari, and you might link the player to one of those, but I have no idea if they work.

sruloart:
@MSGhero I'm not great (or modestly good) with words, I can promise you there's no "IE" flag either :D I just meant, "don't include .ogg files if you intend to use IE or Safari" and forgot to add a happy face to express cynicism :) 

It is possible to play other audio formats on Safari (as well as on IE btw), it's just not possible to include .ogg files without a fall-back to some supported formats (.mp3 ahem ahem) because OpenFL (SoundJS) would refuse to load the game (a known bug that was fixed but not integrated into OpenFL). That is the actual problem (also, **you should delete the default Ariel.ttf font for now if you want your game to load on Firefox and IE**, until OpenFL issue a fix.  

[...]

PR #1563, opened 2015-07-14
Add Wiimote support

larsiusprime:
This adds wiimote support using either the MayFlash Dolphin Bar or the [HID-Wiimote driver](https://github.com/jloehr/HID-Wiimote).

Currently works and tested in flash and legacy. Does not currently work in next for some reason, but doesn't break anything, either. I think it just isn't detecting the gamepad ID for some reason.

Legacy support needs this (simple) pull request to lime to work: 
https://github.com/openfl/lime/pull/499


flixel/input/gamepad/FlxGamepad.hx
```diff
@@ -86,40 +104,93 @@ class FlxGamepad implements IFlxDestroyable
...
 		_isChrome = (Capabilities.manufacturer == "Google Pepper");
 		#end
-		
-		pressed = new FlxGamepadButtonList(FlxInputState.PRESSED, this);
-		justPressed = new FlxGamepadButtonList(FlxInputState.JUST_PRESSED, this);
-		justReleased = new FlxGamepadButtonList(FlxInputState.JUST_RELEASED, this);
-		analog = new FlxGamepadAnalogList(this);
+	}
+	
+	public function traceAxes(f:FlxTimer):Void
```
Gama11: These trace functions look like debug code?

flixel/input/gamepad/FlxGamepad.hx
```diff
@@ -63,6 +73,14 @@ class FlxGamepad implements IFlxDestroyable
 	 * Helper class to get the justMoved, justReleased, and float values of analog input.
 	 */
 	public var analog(default, null):FlxGamepadAnalogList;
+	/**
+	 * Helper class to get the float values of motion-sensing input, if it is available
+	 */
+	public var motion(default, null):FlxGamepadMotionValueList;
+	/**
+	 * Helper class to get the float values of mouse-like pointer nput, if it is available
```
Gama11: `nput`? :)

flixel/input/gamepad/FlxGamepad.hx
```diff
@@ -33,6 +37,12 @@ class FlxGamepad implements IFlxDestroyable
 	public var model(default, set):FlxGamepadModel;
 	
 	/**
+	 * For gamepads that can have things plugged into them (the Wii Remote, basically).
+	 * Making the user set this helps
```
Gama11: "helps..." what?

flixel/input/gamepad/FlxGamepad.hx
```diff
@@ -719,13 +856,55 @@ class FlxGamepadAnalogStick
...
+	}
+	
+	public function toString():String
+	{
+		return("stick(" + x + "," + y + ",(" + rawUp + "," + rawDown + "," + rawLeft + "," + rawRight + ") @" + digitalThreshold + ":" + mode+")");
 	}
 }
 
+typedef FlxGamepadAnalogStickSettings = {
+	@:optional var up:Int;
```
Gama11: Same optional vs ? thing.

flixel/input/gamepad/FlxGamepad.hx
larsiusprime: haha, I did forget to finish that didn't I.

flixel/input/gamepad/FlxGamepad.hx
larsiusprime: Does the ? go before the "var" or before the variable name?

Gama11: There's no `var` with that syntax.

flixel/input/gamepad/FlxGamepadInputID.hx
```diff
@@ -70,6 +70,25 @@ abstract FlxGamepadInputID(Int) from Int to Int
...
+	var TILT_PITCH           = 26;
+	/**tilting side-to-side (think "twisting", or "do a barrel roll!")**/
+	var TILT_ROLL            = 27;
+	
+	/**for a mouse-like input such as touch or IR camera. Horizontal axis.**/
+	var POINTER_X            = 28;
+	/**for a mouse-like input such as touch or IR camera. Vertical axis.**/
+	var POINTER_Y            = 29;
+	
+	/**an extra digital button that don't fit cleanly into the universal template**/
```
Gama11: `don't fit` -> `doesn't fit`

flixel/input/gamepad/FlxGamepadManager.hx
```diff
@@ -427,6 +427,8 @@ class FlxGamepadManager implements IFlxInputManager
 	
 	private function getModelFromFlashDeviceName(str:String):FlxGamepadModel
 	{
+		trace("str = " + str);
```
Gama11: Debug code.

flixel/input/gamepad/FlxGamepadManager.hx
```diff
@@ -437,13 +439,20 @@ class FlxGamepadManager implements IFlxInputManager
 			}
 		}
 		
+		trace("str = " + str);
```
Gama11: Debug code.

flixel/input/gamepad/id/FlxGamepadMotionValueList.hx
```diff
@@ -0,0 +1,29 @@
...
+ * A helper class for gamepad input.
+ * Provides optimized gamepad button checking using direct array access.
+ */
+@:keep
+class FlxGamepadMotionValueList
+{
+	private var gamepad:FlxGamepad;
+	
+	@:allow(flixel.input.gamepad.FlxGamepad)
+	public var available(default, null):Bool = true;
```
Gama11: Isn't this always true?

flixel/input/gamepad/id/LogitechID.hx
```diff
@@ -7,6 +7,12 @@ import flixel.input.gamepad.FlxGamepad;
  */
 class LogitechID
 {
+	public static inline var SUPPORTS_MOTION = false;
```
Gama11: "Motion" means pointer input?

flixel/input/gamepad/id/FlxGamepadMotionValueList.hx
larsiusprime: No, it gets set to false if the gamepad model is switched to something that doesn't support motion.

flixel/input/gamepad/FlxGamepadManager.hx
```diff
@@ -500,16 +511,80 @@ class FlxGamepadManager implements IFlxInputManager
...
+					}
+				}
 			}
 		}
+		
+		gamepad.axis = newAxis;
+		
+		if (anyAbove)
+		{
+			//trace("axis = " + newAxis);
```
Gama11: Debug code.

flixel/input/gamepad/FlxGamepadManager.hx
```diff
@@ -500,16 +511,80 @@ class FlxGamepadManager implements IFlxInputManager
...
+					}
+					else if (newVal >= neg && oldVal < neg)
+					{
+						     if (i == stick.x) digitalButton = stick.rawLeft;
+						else if (i == stick.y) digitalButton = stick.rawUp;
+						handleButtonUp(new JoystickEvent(JoystickEvent.BUTTON_UP, FlashEvent.bubbles, FlashEvent.cancelable, FlashEvent.device, digitalButton));
+					}
+					
+					//pressed/released for digital RIGHT/DOWN
+					if(newVal > pos && oldVal <= pos)
```
Gama11: `if(` -> `if (`

flixel/input/gamepad/FlxGamepadManager.hx
```diff
@@ -500,16 +511,80 @@ class FlxGamepadManager implements IFlxInputManager
...
+				{
+					var newVal = newAxis[i];
+					var oldVal = oldAxis[i];
+					
+					var neg = stick.digitalThreshold * -1;
+					var pos = stick.digitalThreshold;
+					var digitalButton = -1;
+					
+					//pressed/released for digital LEFT/UP
+					if(newVal < neg && oldVal >= neg)
```
Gama11: `if(` -> `if (`

flixel/input/gamepad/FlxGamepadMapping.hx
```diff
@@ -88,16 +130,48 @@ class FlxGamepadMapping
...
 	}
 	
+	/**
+	 * Returns 1 or -1 depending on whether this axis needs to be flipped
+	 * @param	AxisID
+	 * @return
+	 */
+	public function getFlipAxis(AxisID:Int):Int
+	{
+		return switch(model)
```
Gama11: `switch(` -> `switch (`

flixel/input/gamepad/FlxGamepadMapping.hx
```diff
@@ -88,16 +130,48 @@ class FlxGamepadMapping
...
+					}
 				default: null;
 			}
 		}
 		return null;
 	}
 	
+	/**
+	 * Returns 1 or -1 depending on whether this axis needs to be flipped
+	 * @param	AxisID
```
Gama11: Might as well remove the `@param` and `@return` here, there's no description for them.

flixel/input/gamepad/id/LogitechID.hx
larsiusprime: Not quite. Motion and Pointer are two different things, at least based on how I've got it set up currently.

Motion is pitch/roll/yaw detection -- So far in all my tests the WiiMote supports pitch/roll, and the PS4/PS3 also support pitch/roll, but only the WiiMote reports them in a way I can reliably detect without crazy custom driver mappings.

The WiiMote also supports an Infrared camera which is what it uses to drive the on-screen pointer on the Wii/WiiU. Basically the infrared camera and the sensor bar combine to give you X/Y coordinates, and then you can multiply the (0.0-1.0),(0.01-1.0) X/Y values to get a pointer value.

The PS4 also has a touchpad, as does the Vita, the 3DS, the WiiU gamepad, and the Ouya, and so the way I thought of this, was that the WiiMote IR camera pointer has more in common with these touchpads than it does with "motion" as it just spits out analog X/Y coordinates, so we should have a separate "pointer" API for all of those.

As of right now I have only been able to get the WiiMote to be able to actually use the pointer API on the PC, but on the native console side all of the above could fill out the pointer API.

flixel/input/gamepad/FlxGamepad.hx
```diff
@@ -719,13 +845,55 @@ class FlxGamepadAnalogStick
...
+	@:optional var down:Int;
+	@:optional var left:Int;
+	@:optional var right:Int;
+	@:optional var threshold:Float;
+	@:optional var mode:AnalogToDigitalMode;
+}
+
+enum AnalogToDigitalMode
```
Gama11: Missing Flx prefix.

flixel/input/gamepad/FlxGamepad.hx
```diff
@@ -719,13 +845,55 @@ class FlxGamepadAnalogStick
...
+{
+	SendBoth;
```
Gama11: Does the "Send" add anything here?

flixel/input/gamepad/FlxGamepadManager.hx
larsiusprime: On it.

flixel/input/gamepad/FlxGamepad.hx
```diff
@@ -63,6 +73,14 @@ class FlxGamepad implements IFlxDestroyable
...
 	 */
 	public var analog(default, null):FlxGamepadAnalogList;
+	/**
+	 * Helper class to get the float values of motion-sensing input, if it is available
+	 */
+	public var motion(default, null):FlxGamepadMotionValueList;
+	/**
+	 * Helper class to get the float values of mouse-like pointer nput, if it is available
+	 */
+	public var pointer(default, null):FlxGamepadPointerValueList;
```
Gama11: Would it make more sense to use a `FlxPointer` for this instead? That's the base class for `FlxMouse` and `FlxTouch` -> consistent API.

flixel/input/gamepad/FlxGamepad.hx
larsiusprime: so just like this?

```haxe
typedef FlxGamepadAnalogStickSettings = {
	?up:Int;
	?down:Int;
	?left:Int;
	?right:Int;
	?threshold:Float;
	?mode:AnalogToDigitalMode;
}
```

flixel/input/gamepad/id/FlxGamepadPointerValueList.hx
```diff
@@ -0,0 +1,29 @@
...
+ * A helper class for gamepad input.
+ * Provides optimized gamepad button checking using direct array access.
+ */
+@:keep
+class FlxGamepadPointerValueList
+{
+	private var gamepad:FlxGamepad;
+	
+	@:allow(flixel.input.gamepad.FlxGamepad)
+	public var available(default, null):Bool = true;
```
Gama11: Might make sense to change `available` to `isSupported`? That's what `FlxAccelerometer` uses -> consistent API.

flixel/input/gamepad/FlxGamepad.hx
larsiusprime: they are, but could be useful to keep around in the API if they're not actually called... maybe? I use them all the time.

flixel/input/gamepad/id/FlxGamepadPointerValueList.hx
larsiusprime: sounds good to me.

flixel/input/gamepad/FlxGamepad.hx
larsiusprime: Guess not.

flixel/input/gamepad/FlxGamepad.hx
larsiusprime: Yeah probably! 

As a quick aside -- is it possible to override Flixel's mouse input with another source, so you could use a gamepad analog stick, or wiimote pointer input, or even keyboard presses, etc, to drive the mouse, rather than native mouse input? Would be super swell for accessibility.

larsiusprime: Hmm, actually looking at it I'm not sure. As for right now I'm just spitting out the raw analog output from the controller, and there's a whole bunch of uses for it. I think ONE of those uses is definitely to map that to the screen coordinates (which seems to be FlxPointer's purpose), but the game designer / player could also just want the raw analog input for something else. 

I'd love the option to map it to the screen but I don't want to over-process the data before the user has access to it.

flixel/input/gamepad/FlxGamepad.hx
larsiusprime: Fixed among other things.

larsiusprime:
What else?

flixel/input/gamepad/FlxGamepad.hx
Gama11: Wouldn't the values already be screen coordinates, and be mapped to world coordinates? It's not like the controller knows about flixel's coordinate system.

Gama11: Didn't you add setGlobalScreenPositionUnsafe() for having a different source?

larsiusprime: I did? Is that what the blame history shows? Hmmm.

In any case, the values are not actually screen coordinates, they come off of the controller as two analog values between 0 and 1. Not only do they have no knowledge of flixel's coordinate system, they have no knowledge of the world coordinate systems either. It's just two values between the minimum and maximum values of the x/y axes of the detection range, whether it's from the sensor bar or the touchpad.

I mean I suppose I could map it world (OpenFL stage) coordinates using FlxPointer. 

Basically what I'm asking is, do you think there's any value in also reporting the 0-1 values, or will users just always multiply it times screen coordinates and be really annoyed that it doesn't do it for them.

Gama11: The blame history might not show that because FlxPointer didn't exist back then, but I'm sure you added it specifically for flixel-ui's custom cursor stuff.

That seems a bit misleading then? The usage of "pointer" led me to think that these coords are comparable to the x/y of the mouse / a touch. This at least needs to be documented properly - I don't think the [0, 1] range is mentioned anywhere either?

flixel/input/gamepad/FlxGamepad.hx
Gama11: Meh, not sure I'm a fan of having that be part of the API... Perhaps we could have a  `FlxGamepad` [tracker profile](https://github.com/HaxeFlixel/flixel/blob/dev/flixel/system/debug/Tracker.hx#L84) instead?

flixel/input/gamepad/FlxGamepad.hx
```diff
@@ -719,13 +852,55 @@ class FlxGamepadAnalogStick
...
+			rawDown  = (settings.down  != null ? settings.down  : -1);
+			rawLeft  = (settings.left  != null ? settings.left  : -1);
+			rawRight = (settings.right != null ? settings.right : -1);
+			digitalThreshold = (settings.threshold != null ? settings.threshold : -1);
+		}
+	}
+	
+	public function toString():String
+	{
+		return("stick(" + x + "," + y + ",(" + rawUp + "," + rawDown + "," + rawLeft + "," + rawRight + ") @" + digitalThreshold + ":" + mode+")");
```
Gama11: Might make sense to use [Flixel's standard toString() logic](https://github.com/HaxeFlixel/flixel/blob/dev/flixel/FlxObject.hx#L1061-L1070)?

flixel/input/gamepad/FlxGamepad.hx
larsiusprime: Hmmm, this will take some thinking.

So, we don't have the consoles hooked up yet, but here's every device that might use the "gamepad.pointer" API:

- Nintendo 3DS
- Nintendo WiiU
- OUYA touchpad on controller
- PS4 touchpad on controller
- Steam controller (multiple touchpads)
- Wii remote (IR camera)

So the obvious super easy thing to do on the native console side is to rig up the input from the touchscreen to Lime/OpenFL's mouse/touch events, report absolute world coordinates, and call it a day. Flixel doesn't even need to "know" about this.

The problem with this solution is that you might have more than one device that has this kind of "pointery" input, and the mouse/touch API is built around the paradigm that there is ONE mouse/touch surface on the device/computer. So if we do that with the API, then you can't make a PS4 game where everyone is using their own touchpad.

On the 3DS obviously we should just use the same touch API we use for mobile. On the WiiU -- can you have more than one gamepad active on the same console?

So there's this issue to work out.

Other than that, what do you propose besides "gamead.pointer" as the name?

flixel/input/gamepad/FlxGamepad.hx
larsiusprime: It takes five seconds to add them in when I need them, I'll just trash em.

flixel/input/gamepad/FlxGamepad.hx
larsiusprime: didn't know we had such a thing, thx.

flixel/input/gamepad/FlxGamepad.hx
Gama11: Pointer might be fine (unless you have a better name, I don't right know), as long as it's documented what the values mean.

larsiusprime: Sure, let's go with that for now. Where should I put the documentation, in FlxGamepadPointerValueList or in FlxGamepad itself, right next to the pointer object declaration?

Gama11: Would make sense to have some general info on the field in FlxGamepad, and more concrete info on the x/y fields (value range, etc).

flixel/input/gamepad/FlxGamepad.hx
```diff
@@ -86,40 +111,63 @@ class FlxGamepad implements IFlxDestroyable
...
+		
 		#if FLX_JOYSTICK_API
 		leftStick = getRawAnalogStick(FlxGamepadInputID.LEFT_ANALOG_STICK);
 		rightStick = getRawAnalogStick(FlxGamepadInputID.RIGHT_ANALOG_STICK);
 		#end
+		
 		return model;
 	}
 	
+	private var t:FlxTimer = null;
```
Gama11: Don't think this is used anywhere (anymore)?

flixel/input/gamepad/FlxGamepad.hx
```diff
@@ -33,6 +37,18 @@ class FlxGamepad implements IFlxDestroyable
...
+	 * For gamepads that can have things plugged into them (the Wii Remote, basically).
+	 * Making the user set this helps Flixel properly interpret inputs properly. 
+	 * EX: if you plug a nunchuk into the Wii Remote, you will get different values for 
+	 * certain buttons than with the Wii Remote alone.
+	 * (This is probably why Wii games ask the player what control scheme they are using.)
+	 * 
+	 * In the future, this could also be used for any attachment that exposes new API features
+	 * to the controller, e.g. a microphone or headset
+	 */
+	public var attachment(default, set):FlxGamepadModelAttachment;
```
Gama11: Hm... in theory, a gamepad could have multiple attachements?

flixel/input/gamepad/FlxGamepad.hx
```diff
@@ -719,11 +822,61 @@ class FlxGamepadAnalogStick
...
 	public var y(default, null):Int;
 	
-	public function new(x:Int, y:Int)
+	//these values let the analog stick to send digital inputs to, say, the dpad
+	public var rawUp(default, null):Int = -1;
+	public var rawDown(default, null):Int = -1;
+	public var rawLeft(default, null):Int = -1;
+	public var rawRight(default, null):Int = -1;
+	
+	//the value the dpad must be above before digital inputs are sent
```
Gama11: If you want this to appear in completion, it needs to be a `/**` comment I think.

flixel/input/gamepad/FlxGamepad.hx
```diff
@@ -719,11 +822,61 @@ class FlxGamepadAnalogStick
...
+	public var digitalThreshold(default, null):Float = 0.5;
+	
+	//when analog inputs are received, how to process them digitally
```
Gama11: Same here.

flixel/input/gamepad/FlxGamepad.hx
```diff
@@ -154,6 +203,11 @@ class FlxGamepad implements IFlxDestroyable
 		return gamepadButton;
 	}
 	
+	public inline function getFlipAxis(AxisID:Int):Int
```
Gama11: Could this be used to invert the y axis of the Xbox 360 controller in chrome, which is currently hardcoded in getYAxis?

flixel/input/gamepad/FlxGamepad.hx
larsiusprime: In theory, yes; but let's wait until we actually have a use case for that ... [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) and all that.

Unless you can think of something super obvious that we could test with.

Gama11: Yeah, this was more of a general question rather than "this needs to be done differently".

flixel/input/gamepad/FlxGamepad.hx
larsiusprime: I originally added it to fix a bug that has since been fixed by other means, but in short, YES. The idea was to move as much weird hardware-and-platform specific fixup code and mapping logic to the SomeControllerID.hx classes and out of the FlxGamepadMapping class.

flixel/input/gamepad/FlxGamepad.hx
larsiusprime: Sure. But I'm also super interested to know if you have a game controller like that in mind; I'm already thinking about other stuff we could do with attachments -- X360 has the microphone/headset for instance, but not sure if we have any chance of exposing API's for that.

Gama11: No, wasn't thinking about any controller in particular.

flixel/input/gamepad/FlxGamepad.hx
larsiusprime: Good catch on this, as it was also broken on "Adobe Windows" Capabilities.manufacturer value (ie, the standalone flash player on windows).

larsiusprime: This fix is in now.

larsiusprime: Okay maybe I'm crazy, but on my computer chrome doesn't flip the Y axis in flash (so manually inverting it messes it up)... only on the standalone windows player do I have to flip it. 

It even says manufacturer = "Google Pepper" in the log trace in flash/chrome, so I dunno...

I'll commit what works here and you tell me what's happening for you.

larsiusprime:
How bout now?

Gama11:
HTML5 builds are failing (see travis).

flixel/input/gamepad/FlxGamepad.hx
Gama11: Ah, right, I didn't remember that correctly. That's what the old code did as well actually.

```haxe
if (model == XBox360 && !_isChrome)
	axisValue = -axisValue;
```

Gama11:
I can't test if the whole Wiimote stuff actually works, but the API looks ok to me now (and the Xbox 360 stuff still works). Thanks! :+1:

larsiusprime:
Does the fact that we test HTML5 in travis imply that flixel works in HTML5? If so I can hook up the gamepad ID's for that too... are there any special instructions? (I just tried and got a black screen).

Gama11:
That alone just implies that it compiles.. but, yes, in general it should also work. Mode works fine for example.

Getting any errors in the js console? What browser? Browsers other than Chrome have some issues atm I think.

larsiusprime:
Cool, that's a good starting point, I'll go from there. Are we using canvas or GL, and/or what custom backend/add'l dependencies, if any?

Gama11:
You shouldn't need to add anything. I think OpenFL ises canvas by default, haven't tested GL.

I'm not sure Lime's gamepad API is even implemented on HTML5 right now?

sruloart:
BTW, If the black screen issue continues see [here](https://github.com/openfl/lime/issues/493#issuecomment-120238001) and [here](https://github.com/HaxeFlixel/flixel/issues/1462#issuecomment-120654596). 

(and AFAIK for dev it's canvas + Openfl (legacy) html5 backend).

larsiusprime:
This should help with setting up gamepads for HTML5, I think:
http://www.html5rocks.com/en/tutorials/doodles/gamepad/gamepad-tester/tester.html

sruloart:
It's only implemented in bitfive, looks like, not openfl...

PR #1565, opened 2015-07-15
FlxButton: pull default graphic load out into a function so it can be…

larsiusprime:
… overridden (and suppressed) in fancier button sub-classes

This helps with flixel-ui stuff in particular, since all those button sub-classes completely override the default button loading logic, this lets us skip that step entirely.

This should be a small in-place change that has zero effect on flixel itself but gives more flexibility down the line.


larsiusprime:
I don't see any complaints, so I'm going to go ahead and merge this?

PR #1568, opened 2015-07-18
Added test for regression bug, closes #1564

### LINKED ISSUE #1564: Add a unit test for #1561
<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/24379270-add-a-unit-test-for-1561?utm_campaign=plugin&utm_content=tracker%2F254189&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F254189&utm_medium=issues&utm_source=github).
</bountysource-plugin>

winterismute:
Addresses https://github.com/HaxeFlixel/flixel/issues/1564

### DIFF
--- a/tests/unit/src/flixel/FlxObjectTest.hx
+++ b/tests/unit/src/flixel/FlxObjectTest.hx
@@ -4,6 +4,7 @@ import flixel.FlxObject;
 import flixel.graphics.FlxGraphic;
 import flixel.math.FlxPoint;
 import flixel.tile.FlxTilemap;
+import flixel.math.FlxMath;
 import massive.munit.Assert;
 
 class FlxObjectTest extends FlxTest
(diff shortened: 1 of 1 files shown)


tests/unit/src/flixel/FlxObjectTest.hx
```diff
@@ -74,58 +74,72 @@ class FlxObjectTest extends FlxTest
...
+		Assert.isTrue(FlxG.collide(object1, bar));
+		Assert.areEqual(bar.y, object1.y + object1.height);
+	}
+
 	@Test
 	function testUpdateTouchingFlagsHorizontal():Void
 	{
-		var object1 = new FlxObject(0, 0, 10, 10);
-		var object2 = new FlxObject(20, 2, 10, 6);
+		var object1:FlxObject = new FlxObject(0, 0, 10, 10);
```
Gama11: Not a fan of adding the `:FlxObject` explicitly here. That's where Haxe's type inference comes in handy and helps to avoid redundancy (see in the "Leveraging Type Inference" section [here](http://haxeflixel.com/documentation/code-style/)).

Gama11:
If I go back to 1eaf0fedc6d3c4e9e58249d259ca8d6f1b9dc391 (just before my fix), this test is not failing as it should.

winterismute:
Strange, I did check that before pushing... I will look into this soon.

winterismute:
I have updated the branch, you were right, I made a mistake when double-checking (put AND instead of OR). I have changed the test with a different one which involves a tilemap, which seems to be the main (if not only) case in which it could go wrong (which was btw also the reported case).
I have also changed those styling things. Hope now it's fine.

Gama11:
Thanks!

Issue #1466, opened 2015-04-08
Request: "Simple" Function that returns the first object in a group that overlaps


SeiferTim:
Ideally, there should be a function that you call like this:

 getFirstOverlap(Object:FlxObject, Group:FlxObject):FlxObject

You pass it an object and it returns the first item in the group that overlaps that object, or null if there are none.

winterismute:
I implemented it: https://github.com/HaxeFlixel/flixel/pull/1574
However, it seemed to me that having an Array instead of a Group would make this more intuitive since the "order" in a FlxGroup sometimes gets lost, but it is just my opinion.

Gama11:
@winterismute Where does the order "get lost"? A `FlxGroup` isn't much more than a fancy array.

@SeiferTim What was your original use case for this? I'm not sure it's useful enough to include...

SeiferTim:
@Gama11 Oh man, this was a while ago... I think it was just that I wanted some simple way to tell what object, if any, I was overlapping while within an `if` statement... something like:
```
if (something)
{
  var obj = group.firstCollision(player);
  if (obj != null)
  {
    // do something with obj
  }
}
```
I don't really remember the specifics...

winterismute:
@Gama11 What I meant is that, if I remember correctly, a flxGroup does a sort of "compaction" on additions while an array does not.
So, if you have

```Haxe
var group = new FlxGroup();
var queryObject = new FlxSprite();
var obj1 = new FlxSprite()
var obj2 = new FlxSprite()
var obj3 = new FlxSprite() // OVERLAPS queryObject
group.add(obj1);
group.add(obj2);
group.add(obj3);
obj2 = null // Trigger GC

var obj4 = new FlxSprite() // OVERLAPS queryObject
group.add(obj4); // Takes the null place that was of object2

var obj = group.firstCollision(queryObject); // returns obj4, user might have expected obj3 since it was added BEFORE object4
```
But maybe it is just a detail.

Gama11:
I'd rather not include this in `FlxG` for now, that class is already big enough...

Tim's example there suggests he wanted to have this in `FlxGroup`. With the implementation in #1574, that would mean having a function in `FlxGroup` depend on `FlxG.overlap()`, which doesn't really feel right either.

PR #1574, opened 2015-07-22
Issue 1466

winterismute:


flixel/FlxG.hx
```diff
@@ -350,6 +351,30 @@ class FlxG
...
+	 * 
+	 * @param	InputObject	First object to test against
+	 * @param	Group		The group to test against
+	 * @return	First member of Group intersecting InputObject, or null
+	 */
+	public static function getFirstOverlap(InputObject:FlxBasic, Group:FlxGroup):FlxObject
+	{
+		var minIndex = -1;
+		var firstObject:FlxObject = null;
+		FlxG.overlap(InputObject, Group, function MyOverlap(Objec1:FlxObject, Object2:FlxObject) {
```
Gama11: Huh, didn't know you could give an anonymous function a name.. I don't think there's any point to doing so.

Gama11:
Sorry for taking a while to review / merge your pull requests. Will probably have time for it in about a week.

flixel/FlxG.hx
winterismute: I was not sure, I thought it was mandatory... Will remove it then later.

winterismute:
Don't worry about taking time, nobody is in a hurry :D I am just having some breaks sometimes and since I am using flixel myself, sometimes I try to be useful, that's it.

flixel/FlxG.hx
MSGhero: I'm not a fan of the anon function, would prefer a private helper `myOverlap()`

winterismute: @MSGhero is it possible without the closure to access "minIndex" and "firstObject"? The only way I see it is to create an inner class with those 2 properties that will be used only in this function, and pass an object of that kind as the third parameter of the  `myOverlap()` function, assuming that the signature allows to get extra params...

MSGhero: I'm not sure, I was thinking in C rather than Haxe. They could be static, and I personally prefer that over anon, but we'll see what Gama thinks.

Gama11:
See https://github.com/HaxeFlixel/flixel/issues/1466#issuecomment-141759847.

PR #1575, opened 2015-07-25
Add Stage3D support, closes #48

### LINKED ISSUE #48: Consider stage3D renderer for flash target
I'll need someone's help with this.
##

Beeblerox:
Adding initial stage3d support


MSGhero:
This is an awesome and terrifying merge. Needs a demo.

Beeblerox:
@MSGhero it should work with flixel demos, only thing you need to do is to add `<haxedef name="FLX_RENDER_TILE" />` to project.xml (i hope so, i've checked only couple of demoes)
There are few things that should be improved, but it's a good start

sruloart:
@Beeblerox Whoa! I didn't know you're so far advanced with this, it's beyond great :D

Beeblerox:
@sruloart this work is heavilly based on TilesheetStage3d project, but i had to significantly rewrite it and add new functionality. And integrating it with flixel helped to find some bugs in it.

Beeblerox:
i see that there are some conflicts exists, and i'll need help with merging it, since i'm not very familiar with git.

MSGhero:
Does this mean the PostProcess API can be used with flash now with some tweaking?

Beeblerox:
@MSGhero probably with some additional work this will be possible, but i can't say exactly

Gama11:
What's with the code in the `com.asliceofcrazypie` package? It looks like you wrote it, at least it has `@author Zaphod`? Why isn't it in the `flixel` package then?

sruloart:
In any case it looks like it's a MIT code, so even if he wasn't the writer you could drop the package structure. Some licenses do not allow you to use the original packages' names if you made any changes.

Also, the #if flash11 flag ([for example, here](https://github.com/HaxeFlixel/flixel/commit/0b1c7d6c99ded9db1793044c888b8466c2d58aef#diff-94b9eaad2ff08af37c45d06e4a05f4a5R3)) means Flash 11 and above, or just Flash 11?

Gama11:
That means Flash 11+.

Beeblerox:
@Gama11 ok, i'll change packages (this weekend maybe)

include.xml
```diff
@@ -9,10 +9,10 @@
 	<assets path="assets/fonts/arial.ttf" embed="true" unless="FLX_NO_DEBUG" />
 	
 	<set name="legacy" unless="next" />
-	
 	<haxelib name="openfl" unless="create" />
 	
 	<haxeflag name="--macro" value="flixel.system.macros.FlxDefines.run()" />
+	<haxeflag name="-swf-lib" value="assets/AGAL.swc" if="flash" />
```
Gama11: Shouldn't this swc only be included if `FLX_RENDER_TILE` is defined?

Gama11:
Just solved the merge conflicts.

Flash doesn't compile right now unless `FLX_RENDER_TILE` is defined.

flixel/text/FlxText.hx
```diff
@@ -775,7 +787,13 @@ class FlxText extends FlxSprite
 			newHeight = oldHeight;
 		}
 		
+		_flashRect.setTo(0, 0, newWidth, newHeight);
+		
+		#if flash11
```
Gama11: Why is this `#if flash11`? Flash Player version 11+ doesn't necessarily mean that you're using Stage3D rendering. I think this should be `#if (flash11 && FLX_RENDER_TILE)`. 
Since you're gonna need that in a number of places, I recommend creating a helper define in `FLX_DEFINES` for this (`FLX_RENDER_STAGE3D` maybe ?).

com/asliceofcrazypie/flash/ContextWrapper.hx
```diff
@@ -0,0 +1,683 @@
...
+		for (n in rawData)
+		{
+			bytes.writeByte(n);
+		}
+		
+		return bytes;
+	}
+	
+	//misc methods
+	public static inline function clearArray<T>(array:Array<T>):Void
```
Gama11: There's a `clearArray()` in `FlxArrayUtil`.

com/asliceofcrazypie/flash/ContextWrapper.hx
```diff
@@ -0,0 +1,683 @@
...
+	}
+	
+	private function getTextureLookupFlags(mipMapping:Bool, smoothing:Bool):String
+	{
+		//var repeat:Bool = true; // making it default for now (to make things simpler)
+		var repeat:Bool = false; // making it default for now (to make things simpler)
+		
+		var options:Array<String> = ["2d", repeat ? "repeat" : "clamp"];
+		
+		if (smoothing == false) {
```
Gama11: `{` should go in new line.

Gama11:
`FlxCamera#shake()` seems to be broken (at least in Mode's main menu).

Gama11:
I'm getting fairly poor performance in FlxBunnyMark with the default amount of bunnies (500), not even reaching 60 fps. Is that expected?

flixel/text/FlxText.hx
Beeblerox: yeah, it's good idea!

com/asliceofcrazypie/flash/ContextWrapper.hx
Beeblerox: sorry, will remove this method

include.xml
Beeblerox: you are right, but how to do it? (i really don't know)

Beeblerox:
@Gama11 thanks for resolving merge issues! I know that this works isn't complete. i'll make it work in blit mode on flash.
Poor perfomance on flash is caused by constant text changes (i believe). If you won't update text then it will work much faster - text updates causes new texture generation which is very costly, so we should try to stick to bitmap fonts there.

com/asliceofcrazypie/flash/ContextWrapper.hx
Gama11: The implementation is a bit different, not sure which one is preferable.

Gama11:
Hm, I see. There's only a sinlge updating text field in the bunny mark though I think (the fps text), so that must be extremely expensive.

Beeblerox:
yes, there is only one changing text, but text texture should be square and should have power of two size, so it means creating quite big texture (1024x1024, i guess) every frame, but it could be optimized by changing the size of the text object

gamedevsam:
What's the status of this PR?

Beeblerox:
it needs a lot of polish

Gama11:
Closing this for now, it's not really in a mergable / reviewable state (lots of merge conflicts due to changes like #1668).

PR #1579, opened 2015-07-30
New Tile Filter logic for `FlxTilemap`

SeiferTim:
Changed `FlxTile`, `FlxTilemap`, and `FlxBaseTilemap` to allow multiple filters to be specified for each tile.

Use `FlxBaseTilemap.addTileFilter()` to add a new filter. You can specify a Class to use as the filter, Collisions, Callback, and ProcessCallback which will all work within `FlxTilemap.overlapWithCallback()` to have your class(es) collide/not collide, and trigger callbacks properly.


Gama11:
If I understood #1615 correctly, this is causing some issues.

Issue #1374, opened 2014-11-15
Rename FlxMouseEventManager to FlxPointerEventManager?


Gama11:
The current name is rather misleading and implies that it doesn't work with touches (which it does), see #1372.

The new name would be more in line with `FlxPointer` as well.

Gama11:
This class has been renamed often enough already, let's just leave it alone for now.

PR #1581, opened 2015-08-07
Added a note for FlxG.overlaps() regarding FlxTilemaps, closes #1376

### LINKED ISSUE #1376: FlxG.overlap() always returns true with tilemaps
I have created a tilemap with Ogmo Editor and placed my player inside it. `FlxG.collide(player,_map) `works as expected, my player only collides with solid tiles.  `FlxG.overlap(player,_map)` however always returns true as long as the player is within the bounds of the map, whether there is a tile in that space or not.

I've heard that Ogmo can cause issues since it uses -1 as indices for empty tiles so I used the map and code from [this](http://haxeflixel.com/demos/TileMap/) tutorial but to no avail. Whenever I check overlap between a FlxObject and a FlxTilemap, it always returns true unless the player leaves the map.

Note: I can force the function to return `false` by using the ProcessCallback but that doesn't really help.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/6213017-flxg-overlap-always-returns-true-with-tilemaps?utm_campaign=plugin&utm_content=tracker%2F254189&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F254189&utm_medium=issues&utm_source=github).
</bountysource-plugin>

nazywam:
[Issue 1376](https://github.com/HaxeFlixel/flixel/issues/1376)

### DIFF
--- a/flixel/FlxG.hx
+++ b/flixel/FlxG.hx
@@ -324,6 +324,7 @@ class FlxG
 	 * whatever floats your boat! For maximum performance try bundling a lot of objects
 	 * together using a FlxGroup (or even bundling groups together!).
 	 * NOTE: does NOT take objects' scrollfactor into account, all overlaps are checked in world space.
+	 * NOTE: for checking overlap with a FlxTileMap use FlxTileMap#overlaps() instead.
 	 * 
 	 * @param	ObjectOrGroup1	The first object or group you want to check.
 	 * @param	ObjectOrGroup2	The second object or group you want to check.  If it is the same as the first, flixel knows to just do a comparison within that group.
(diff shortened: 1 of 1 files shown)


Gama11:
I don't think that's entirely accurate... I think `FlxG.overlaps` doesn't take the collision properties of the tilemap into account - in my code snippet in #1374, the sprite overlaps an emtpy tile of the tilemap (index 0). That's still considered part of the tilemap, it just isn't collidable.

Also, it's `FlxTilemap`, not `FlxTileMap`.

flixel/FlxG.hx
```diff
@@ -324,6 +324,7 @@ class FlxG
 	 * whatever floats your boat! For maximum performance try bundling a lot of objects
 	 * together using a FlxGroup (or even bundling groups together!).
 	 * NOTE: does NOT take objects' scrollfactor into account, all overlaps are checked in world space.
+	 * NOTE: for checking overlap with FlxTilemaps that contain empty tiles use FlxTilemap#overlaps() instead.
```
Gama11: Hm, still not a huge fan of this description.. We should explain the _why_, not just redirect to another part of the API without any explanation.

PR #1587, opened 2015-08-20
Adding enumerations for more blend modes. 

larsiusprime:
Pending this PR to openfl:
https://github.com/openfl/openfl/pull/803


larsiusprime:
OpenFL has merged the dependent PR's, so this is safe now.

Gama11:
Urm.. how safe is "safe"? I guess this is not on haxelib yet? Travis is failing anyway...

larsiusprime:
They just put out a maintenance release today?

Gama11:
Ok, cool (that was after my comment).

Still, this means we're not backwards compatible at all, and #if openfl here probably wouldn't be a bad idea.

Issue #1188, opened 2014-06-22
Pass FlxG.elapsed through update methods


JoeCreates:
Most game engines tend to pass a time delta down through update() methods, whereas HaxeFlixel uses a single static variable `FlxG.elapsed`. The problem with this is that it gives no control to updating classes over how quickly they should update.

HaxeFlixel has `FlxG.timeScale` for if you wish to slow down or speed up the game, but when is it ever useful to slow down or speed up _everything_? Generally if you want to have such an effect, you'll have a HUD updating at a constant rate while the world updates at a different speed.

For this reason, I would suggest `update()` methods are replaced with `update(elapsed:Int)`. It would this be trivial for any updating object to have it's own elapsed coefficient for changing the speed at which its members update.

It might be possible to stop this change from being a breaking change by making `elapsed` optional?

PR #1590, opened 2015-08-26
Added timescale-independent version of FlxG.elapsed

BluShine:
Added independentElapsed to FlxG, which works exactly like FlxG.elapsed,
but is independent from FlxG.timescale. This is very useful for keeping
things like UI and menus independent from slow-motion effects.

### DIFF
--- a/flixel/FlxG.hx
+++ b/flixel/FlxG.hx
@@ -118,6 +118,11 @@ class FlxG
 	 */
 	public static var elapsed(default, null):Float = 0;
 	/**
+	 * Represents the amount of time in seconds that passed since last frame, independent from timescale.
+	 * Useful for keeping track of time during slow-motion effects. If timescale is 1, this will be equal to elapsed. 
+	 */
+	public static var independentElapsed(default, null):Float = 0;
+	/**
 	 * Useful when the timestep is NOT fixed (i.e. variable), to prevent jerky movement or erratic behavior at very low fps.
 	 * Essentially locks the framerate to a minimum value - any slower and you'll get slowdown instead of frameskip; default is 1/10th of a second.
 	 */
--- a/flixel/FlxGame.hx
+++ b/flixel/FlxGame.hx
@@ -713,10 +713,12 @@ class FlxGame extends Sprite
 		if (FlxG.fixedTimestep)
 		{
 			FlxG.elapsed = FlxG.timeScale * _stepSeconds; // fixed timestep
+			FlxG.independentElapsed = _stepSeconds;
 		}
 		else
 		{
 			FlxG.elapsed = FlxG.timeScale * (_elapsedMS / 1000); // variable timestep
+			FlxG.independentElapsed = (_elapsedMS / 1000);
 			
 			var max = FlxG.maxElapsed * FlxG.timeScale;
 			if (FlxG.elapsed > max) 


MSGhero:
There was a discussion about this a while ago. Someone suggested removing `timescale` from `FlxG` and add it to `FlxGroup` which I kinda like. You could alter the state's timescale or just your main group's to keep UI going normally.

Gama11:
The dev branch passes `elapsed` to each `update()` method. You can modify it that way. Read the full discussion here: #1188

BluShine:
If this is the case, is there any reason to keep FlxG.timescale?

sruloart:
@BluShine, yeah, in case you want to slow down the entire game :)

BluShine:
Idk, it just seemed kinda useless to me when there's no easy way to keep track of "real" time independent of timescale. If the "right" way to slow down time is by passing a different elapsed value to objects/groups, keeping the timescale variable seems like it would just encourage people to slow down time the "wrong" way. But that's just my personal use-case, which apparently is uncommon.

Gama11:
Dunno if there's really too much harm in keeping timescale.. It can be kinda convenient, and removing it would be yet another breaking change.

We could definitely add a note to its docs mentioning elapsed.

BluShine:
Searching in flixel-demos, it looks like Mode and FlxNape are the only two flixel-demos that change timeScale. Oddly, MinimalistTD has two occurances of "FlxG.timeScale = 1", but never actually sets it to something else. And IMHO, the convenience factor is somewhat offset by the lack of a convenient way to keep track of real time. 

Documentation improvements are never a bad idea either.

Gama11:
MinmalistTD has a "speed up" button in the lower right corner (which most TD games have). In `speedButtonCallback()` you'll find this line:

```haxe
FlxG.timeScale = _speed;
```

Apparently, GitHub's repo search doesn't find that for some reason - I was quite confused at first as well. The repo search is quite poor really...

Anyway, technically that's not ideal, since the game has a UI that will be sped up as well.

I'm not *totally* against removing `timeScale`, especially since breaking changing are a no-go after 4.0.0. We could also mark it as `@:deprecated` (Haxe built-in meta meaning that any usage of it causes a compiler warning - not sure how practical this is considering core classes need to know about it and use it).

@JoeCreates I'm interested in your opinion on this issue.

sruloart:
@Gama11 Just to clear out some understandable (but a colossal) confusion:
* `_elapsedMS` gives you the "real" time that has passed between **frames** (and thus between draws).
* `_stepMS` gives you the estimated time between **updates** (well, `step()`, but close enough). So, if you're using a fixed time step then it's the "real" time, and if not then you can always compare it to the frame time above.
* `FlxG.elapsed` is not a "real" loop value, it's simply a mechanism for a device called `timeScale`, a variable that its only role is to give you the ability to manipulate your **velocity** (but also several minor things, like the camera's `flash`, `fade` and `shake` for example) **without actually changing** the time that has passed between updates (oh the irony). That's right, it doesn't affect it at all. 


* I blame the descriptions in FlxGame and FlxG, which are horribly old, and this is one of the many reasons why I'm refractoring FlxGame for my smoothing everything's out project. Pretty nasty stuff actually.

Gama11:
>why I'm refractoring FlxGame for my smoothing everything's out project

@sruloart That means you're working on a pull request?

sruloart:
I just don't want a giant PR that would get washed away like the collisions PR did. So, I rather have a fully working system first, with all the changes, and then slowly integrate features with post 4.0.0 release dev branch.

Gama11:
As long as there are no breaking changes.

PR #1591, opened 2015-08-27
Add support for flipped FlxFrames

larsiusprime:
Adding support for flipped FlxFrames. This lets you get your atlases packed even tighter by re-using mirror images.

### DIFF
--- a/flixel/graphics/frames/FlxAtlasFrames.hx
+++ b/flixel/graphics/frames/FlxAtlasFrames.hx
@@ -243,6 +243,8 @@ class FlxAtlasFrames extends FlxFramesCollection
 		var data:Fast = new haxe.xml.Fast(Xml.parse(Description).firstElement());
 		
 		var angle:Int;
+		var flipX:Bool;
+		var flipY:Bool;
 		var name:String;
 		var trimmed:Bool;
 		var rotated:Bool;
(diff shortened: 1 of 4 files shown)


sruloart:
Good thinking! talking about tighter packing, do you think your [MaxRects](https://github.com/larsiusprime/tdrpg-tools/blob/34c32fda9908d904f747d215a8e66d8760b5f9eb/dqsprite/src/MaxRectsBinPack.hx) implementation can work with FlxAtlas? currently it's missing a proper packing algorithm.

larsiusprime:
Yeah!

I just added a readme:
https://github.com/larsiusprime/tdrpg-tools/tree/master/dqsprite

The tool itself is a bit hacky and designed specifically for defender's quest, but you can probably just look at what it's doing and get most of what you need right from there. I set it up to export a sparrow-compatible xml file and then feed it straight into FlxAtlas. It's working in my game right now.

Gama11:
I'll let @Beeblerox review this one.

larsiusprime:
@sruloart:

My EntitySprite class in flixel-editors (more like flixel-lars'-weird-experiments, lol) makes use of this if you want more examples

https://github.com/HaxeFlixel/flixel-editors/blob/dev/flixel/editors/EntitySprite.hx

flixel/graphics/frames/FlxAtlasFrames.hx
```diff
@@ -256,6 +258,8 @@ class FlxAtlasFrames extends FlxFramesCollection
 			name = texture.att.name;
 			trimmed = texture.has.frameX;
 			rotated = (texture.has.rotated && texture.att.rotated == "true") ? true : false;
+			flipX = (texture.has.flipX && texture.att.flipX == "true") ? true : false;
```
Gama11: No need for the ternaries here, these are already boolean expressions. :)

PR #1595, opened 2015-08-30
Make #1587 work in legacy mode.

FryDay:
[Pull request 1587](https://github.com/HaxeFlixel/flixel/pull/1587) creates issues with lime legacy. This will allow games to be built without have to use `<set name="next" />` in Project.xml.

Fix was suggested [here](https://github.com/HaxeFlixel/flixel/commit/fd7b5da6f33646346c563fbbe63b148139b14438).


Gama11:
:+1: 

I think we need to take the OpenFL version into account as well here, as this will cause isssues even on older next versions.

FryDay:
I agree, but I wanted people to be able to continue development in the short term. :)

Gama11:
Oh, of course - it's definitely appreciated. :)

We should now be backwards compatible again (although I'm not really sure down to what OpenFL+Lime version tbh).

Issue #1601, opened 2015-09-05
Changing the arial font


sruloart:
The arial font isn't free, and there are some smaller size alternatives that are (like Noto). However, I'm not really seeing the point in including any such font to begin with, so...

My overall suggestion here is to use different system fonts on different targets instead of bundling a default font.

larsiusprime:
Consistent cross-platform rendering might still be important? I think Liberation Sans is a good free alternative to Arial (it's a Helvetica lookalike, just like Arial is, IIRC)

sruloart:
Liberation Sans would work too, sure, there are some other options as well. 

The problem here is that you have a (relatively) gigantic unicode (most users use plain English) font included as default in size-sensitive targets like Flash/Html5/Android. I mean, what default languages does HaxeFlixel even support? by the look of things: English.

I think that by using the correct system fonts, rendering would not only be consistent but also improved (by using non-hinted fonts on Android/Mac for example, html5 would look better in almost any case), but that requires some testing...

For consistent "HaxeFlixel-ish" look a user can always use the nokiafc22 font (or his own choice). You don't generally look at Arial and say, hey, that's a Flixel font! :)

larsiusprime:
Ah, that's a good point :)

Gama11:
Arial is basically just included for the debugger.

sruloart:
So I'll be disposing of it in my html5 PR :) question is should we replace it or just let whatever default system font available to to take over?

Gama11:
I don't know. At the time there was some inconsistency between the font in the debugger on windows vs flash, which Arial fixed.

sruloart:
I did see some logic in having a multiple-languages supported font, but if that's not really used to begin with it seems kind of redundant to add a new one. 

I'll test with and without, and if that makes sense I'll add a debug default font that looks like Arial but is much much smaller and free to use.

Gama11:
This was addressed by #1602.

Ohmnivore:
[posted earlier in the pull request because was confused]

What do you think of using DejaVu Sans? It's public domain. Same size as monsterrat, practically the same look but a lot crisper. I find it a lot easier to read the debugger console with DejaVu.

Here's a side by side comparison of both (DejaVu Sans on the right):
![fonts](https://cloud.githubusercontent.com/assets/3769354/12209496/588305c4-b621-11e5-89d2-6eb0898017c1.png)

Gama11:
[...]

Issue #1597, opened 2015-09-01
HTML5: Fixing the fonts/audio problems on Firefox and IE11 and Safari


sruloart:
* Fonts: AFAI can tell, the fonts became an issue with this commit: https://github.com/HaxeFlixel/flixel/commit/7c0a0dc66f315ced29f9203ff26b045151e6b397, so maybe some html5 flags can fix at least the fonts problems (which are relevant anywhere but on Chrome). Not embedding fonts (and using Assets.load... instead of Assets.get... obviously) seems to do the trick.

* Audio: The audio files were always a problem on IE11/Edge and Safari because they don't support the .ogg format, and you can't legally include .mp3 with your build (unless you have less than 5000 total downloads, which is ridiculous). A probable fix would be not using the same name for the different files (.mp3/.ogg), and adding another "safe" format (.m4a) version (and .wav files do work anywhere but on IE11, so that's also worth considering, as long as they are kept small enough).

sruloart:
###### O.K, I'm half way done. There are some necessary consequences though:
* Sound: 
  * Use only `.m4a` files (or `.wavs`, that's fine too if you don't mind the size) for html5 (`.ogg` isn't supported anywhere but Chrome, so don't use it for html5 at all).
 * `FlxSplash` doesn't auto-play the `flixel.m4a` file on html5, but that's not because of the file format, it's something to do with `FlxSplash` itself (there's a bug somewhere preventing it to load/play, I dunno). `.m4a` files work very well otherwise, on all major browsers, so I think that's a reasonable outcome for now.

* Fonts
 * Create the missing web fonts on your own (check your console, use anything available) and you should be fine.
 * BUT We need to replace `arial.ttf` because it's a huge unnecessary beast of a font (760kb! wat) that causes problem on html5. 
 * So, suggestions / preferences would be welcomed. Just mind the license. A monospace font scales better on the other platforms, but would require a specific solution on html5 to make it scale like it should, so we actually need two fonts: one free monospace to replace `arial`, and something different for html5. 

Now I guess some .m4a files / web fonts should be added to the demos section as well, I'll get there soon enough.

Gama11:
There's not really much to be done here [as .m4a files aren't working consistently either](https://github.com/HaxeFlixel/flixel/pull/1602#issuecomment-149040724).

PR #1602, opened 2015-09-06
Different debugger font, replace .ogg by .m4a on HTML5, improved LogStyles

sruloart:
…ead.

Made the debugger's LogStyle messages more readable.
Added .m4a files for the html5 target.


sruloart:
Whoops, let me send the branch instead...let me see

sruloart:
I'm not sure what I'm doing anymore :) I think this is alright no? anyhow, now I'm going to add some .m4a files to the flixel-demos :D

sruloart:
Closes #1597 
Closes #1601

include.xml
```diff
@@ -1,12 +1,13 @@
 <?xml version="1.0" encoding="utf-8"?>
 <project>
-	<icon path="assets/images/logo/HaxeFlixel.svg" />
+	<!--<icon path="assets/images/logo/HaxeFlixel.svg" />-->
```
Gama11: ?

sruloart: Right, my bad :)

Gama11:
Not a huge fan of using nokiafc22 in the debugger... I prefer the "thinner" arial, looked cleaner to me:



vs.



Definitely takes uses a bit more space that messes with spacing in some places:



The second screenshot just looks very... "crowded". Open to more opinions on it though.

sruloart:
Something like Open Sans could work, but it doesn't scale well. I'm going to try out some options...

Gama11:
Why were you worried about Arial in the first place? File size really shouldn't matter in debug mode anyway.

sruloart:
* First I worried because it caused a nasty breaking bug on the OpenFL/Lime html5 backend.
* Then, I've realized this font is incredibly big relative to the average project size, so it can easily affect performance, even if that's just in debug.  
* Then I've wondered why is it so big and assumed it was meant so you can support more languages.
* Then I've looked around and found out it's infringing copyrights as we speak...
* At that point I just wanted to get rid of it, the sooner the better.

sruloart:
Finished, hopefully. 

Also, it turns out `unless="FLX_NO_DEBUG"` doesn't actually exclude the debug font from being added to the release build (tested with all platforms). 

I just hope that doesn't mean this flag is broken, as it is used wide and often. 

Also found another bug in the html5 target. Good times :)

Gama11:
Huh. That probably means that broke when I changed `@:font` embeds to embeds in the `include.xml` to fix that Android text bug (which still isn't fixed on the OpenFL side to my knowledge). The exclusion definitely worked at some point.

include.xml
```diff
@@ -2,11 +2,12 @@
...
 	
-	<assets path="assets/sounds" include="*.ogg" if="html5" unless="unit-test" />
+	<assets path="assets/sounds" include="*.m4a" if="html5" unless="unit-test" />
 	<assets path="assets/sounds" include="*.mp3" if="flash" embed="true" />
-	<assets path="assets/sounds" include="*.ogg" unless="flash || html5" embed="true" />
+	<assets path="assets/sounds" include="*.ogg" if="native" embed="true" />
+	
 	<assets path="assets/fonts/nokiafc22.ttf" embed="true" />
-	<assets path="assets/fonts/arial.ttf" embed="true" unless="FLX_NO_DEBUG" />
+	<assets path="assets/fonts/montserrat-Regular.ttf" embed="true" if="debug" />
```
Gama11: Sounds like a typo of "monsterrat". :D

larsiusprime: Looks correct to me (unless you meant it as a joke):
http://www.fontsquirrel.com/fonts/montserrat

I think it's meant to be "Mont Serrat" as opposed to "Monster Rat"

Gama11: Intended as a joke - had to google to be sure though. ;)

larsiusprime: Always good to check :) I could have easily seen it go either way.

Gama11: Is the `if=debug` here safe? `!FLX_NO_DEBUG` isn't the same as `debug`, meaning you could have a flixel debugger in release mode if you wanted - in which case the font would be missing.

Gama11:
Hm, this montserrat font doesn't look bad (interestingly enough, it looks better on native than it does on flash) - however, I really dislike how similar `C` and `G` are, you can barely tell the difference..



Compared to Arial:

include.xml
Gama11: Yeah, that breaks the debugger completely - there's no fallback font either:

Gama11: Btw, the `unless="FLX_NO_DEBUG"` wasn't working because it's generally defined after flixel is included, and order matters in xml config files. I don't think there's a way to do this reliably there, which is why I preferred the `@:font` embedding - which is still broken on Android to my knowledge. :(

Gama11:
In #1597 you said `flixel.m4a` doesn't play because of a `FlxSplash` bug, but I can't get flixel sounds to work elsewhere either. I tried

```haxe
FlxG.sound.play("assets/sounds/beep.m4a");
FlxG.sound.play("assets/sounds/flixel.m4a");
```

(`PlayState#create()`)
in IE, Chrome and Firefox - always the same error in the console:

`Uncaught TypeError: Cannot set property 'volume' of null` (in `FlxSound#updateTransform()`)

It outputs the following to the log window: `[ERROR] Could not find a Sound asset with an ID of assets/sounds/flixel.m4a` - however, `openfl.Assets.list()` lists it as an asset. Turns out the `AssetType` of the .m4a files is `BINARY`.

The files not being detected correctly is an OpenFL issue, and in theory we should be able to work around it by adding `type="sound` - that didn't seem to do much other than making  `openfl.Assets.list(AssetType.SOUND)` work as expected though, so I guess it's not purely a detection issue.

Gama11:
Hm, actually, I can't reproduce the `@:font` issue on Android anymore - maybe it's safe to switch back to that.

sruloart:
* Including the font in debug mode is possible as long as it's kept small enough. I rather use the FLX_NO_DEBUG though. 
* The Android problem is bigger than the html5 sound problem. I would keep the android solution for now.
* The G-C problem I can probably solve by making small adjustments to the font. It took a while to find, I hate to see it go.
* Yeah, currently there seem to be a number of issues with OpenFL preventing you from using `FlxG.sound.play` or any other method other than the `FlxG.sound.load(FlxAssets.getSound("path")).play;` method to play the sounds in html5. So, the PR is still safe to integrate, but none of the demos are, so I'm keeping that commit at bay atm. 
* That's something on top the whole "FlxSplash's flixel.m4a won't play automatically" issue AFAIK (but, again, I suspect it's an OpenFL issue, not Flixel's).

* In any case we do need to remove the Arial font, it's not free and it's in-needlessly big.

Gama11:
I do like the font as well. Is adjusting that something that's easy to do?

sruloart:
I'd say yes. It's not like we're changing the letter's width/height. However, it would require a new name since I'm modifying it. How does `monsterrat` sound like? :)

sruloart:
OK, Fixed the G==C issue, renamed the font to Monsterrat (license's terms for modification), happy times.

Gama11:
Fyi, I reverted the .m4a part of this pull in f2cfdfa for now.

sruloart:
I'm fine with it, playing audio in the new OpenFL html5 backend is faulty anyhow, so let's wait for simpler times (maybe a new soundJS) :)

Ohmnivore:
What do you think of using DejaVu Sans? It's public domain, not aware of G-C issues with it (but then again i dont know what to look for). Same size as monsterrat, practically the same look but a lot crisper. I find it a lot easier to read the debugger console with DejaVu.

Here's a side by side comparison of both (DejaVu Sans on the right):
fonts

PR #1603, opened 2015-09-07
Allow negative letterSpacing in FlxBitmapFont

schonstal:
If you set the value to a negative value, it spaces the letters apart based on the positive value instead. Negative spacing is useful if you want your letters to blend together. It doesn't look like there's a technical reason to prohibit the value from being negative.

I implemented this fix for my Ludum Dare entry, and it is working there: https://github.com/schonstal/LudumDare33


Gama11:
@Beeblerox any objections to this?

PR #1604, opened 2015-09-08
FlxPool & FlxTypedButton: optimization that removes dynamic casting in c++

larsiusprime:
AFAIK, this has no external changes to the API or the behavior, and from inspecting the generated CPP and hxscout, removes the dynamic cast entirely.

before:
```cpp
Dynamic tmp = this->label;
```

after:
```cpp
::flixel::FlxSprite tmp = this->_labelAsSprite;
```

UPDATE: I also added the @:generic meta to FlxPool to properly generate the generic variants the avoid the dynamic casting.

### DIFF
--- a/flixel/ui/FlxButton.hx
+++ b/flixel/ui/FlxButton.hx
@@ -313,10 +313,10 @@ class FlxTypedButton<T:FlxSprite> extends FlxSprite implements IFlxInput
 	{
 		super.draw();
 		
-		if (label != null && label.visible)
+		if (_spriteLabel != null && _spriteLabel.visible)
 		{
-			label.cameras = cameras;
-			label.draw();
+			_spriteLabel.cameras = cameras;
+			_spriteLabel.draw();
 		}
 	}
 	
@@ -328,9 +328,9 @@ class FlxTypedButton<T:FlxSprite> extends FlxSprite implements IFlxInput
 	{
 		super.drawDebug();
 		
-		if (label != null) 
+		if (_spriteLabel != null) 
 		{
-			label.drawDebug();
+			_spriteLabel.drawDebug();
 		}
 	}
 #end
--- a/flixel/ui/FlxButton.hx
+++ b/flixel/ui/FlxButton.hx
@@ -177,6 +177,8 @@ class FlxTypedButton<T:FlxSprite> extends FlxSprite implements IFlxInput
 	public var pressed(get, never):Bool;
 	public var justPressed(get, never):Bool;
 	
+	private var _labelAsSprite:FlxSprite;
+	
 	/** 
 	 * We don't need an ID here, so let's just use Int as the type.
 	 */
@@ -251,6 +253,7 @@ class FlxTypedButton<T:FlxSprite> extends FlxSprite implements IFlxInput
 	override public function destroy():Void
 	{
 		label = FlxDestroyUtil.destroy(label);
+		_labelAsSprite = null;
 		
 		onUp = FlxDestroyUtil.destroy(onUp);
 		onDown = FlxDestroyUtil.destroy(onDown);
@@ -456,18 +459,18 @@ class FlxTypedButton<T:FlxSprite> extends FlxSprite implements IFlxInput
 	
 	private function updateLabelPosition()
 	{
-		if (label != null) // Label positioning
+		if (_labelAsSprite != null) // Label positioning
 		{
-			label.x = (pixelPerfectPosition ? Math.floor(x) : x) + labelOffsets[status].x;
-			label.y = (pixelPerfectPosition ? Math.floor(y) : y) + labelOffsets[status].y;
+			_labelAsSprite.x = (pixelPerfectPosition ? Math.floor(x) : x) + labelOffsets[status].x;
+			_labelAsSprite.y = (pixelPerfectPosition ? Math.floor(y) : y) + labelOffsets[status].y;
 		}
 	}
 	
 	private function updateLabelAlpha()
 	{
-		if (label != null && labelAlphas.length > status) 
+		if (_labelAsSprite != null && labelAlphas.length > status) 
 		{
-			label.alpha = alpha * labelAlphas[status];
+			_labelAsSprite.alpha = alpha * labelAlphas[status];
 		}
 	}
 	
@@ -539,6 +542,8 @@ class FlxTypedButton<T:FlxSprite> extends FlxSprite implements IFlxInput
 		}
 		
 		label = Value;
+		_labelAsSprite = label;
+		
 		updateLabelPosition();
 		
 		return Value;
--- a/flixel/ui/FlxButton.hx
+++ b/flixel/ui/FlxButton.hx
@@ -113,6 +113,9 @@ class FlxButton extends FlxTypedButton<FlxText>
 /**
  * A simple button class that calls a function when clicked by the mouse.
  */
+#if !display
+@:generic
+#end
 class FlxTypedButton<T:FlxSprite> extends FlxSprite implements IFlxInput
 {
 	/**
--- a/flixel/ui/FlxButton.hx
+++ b/flixel/ui/FlxButton.hx
@@ -177,7 +177,10 @@ class FlxTypedButton<T:FlxSprite> extends FlxSprite implements IFlxInput
 	public var pressed(get, never):Bool;
 	public var justPressed(get, never):Bool;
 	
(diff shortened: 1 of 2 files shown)


flixel/ui/FlxButton.hx
```diff
@@ -177,6 +177,8 @@ class FlxTypedButton<T:FlxSprite> extends FlxSprite implements IFlxInput
 	public var pressed(get, never):Bool;
 	public var justPressed(get, never):Bool;
 	
+	private var _labelAsSprite:FlxSprite;
```
gamedevsam: I think the name ```_spriteLabel:FlxSprite``` would be nicer looking. You might also want to add a comment above this variable declaration explaining why it's necessary.

larsiusprime: Sure

larsiusprime:
All good?

gamedevsam:
Looks good to me, but I'll give @Beeblerox and @Gama11 a chance to review it.

flixel/ui/FlxButton.hx
```diff
@@ -177,6 +177,11 @@ class FlxTypedButton<T:FlxSprite> extends FlxSprite implements IFlxInput
 	public var pressed(get, never):Bool;
 	public var justPressed(get, never):Bool;
 	
+	/**
+	 * We cast label to a FlxSprite for internal operations to avoid Dynamic casts in C++
+	 */
+	private var _spriteLabel:FlxSprite;
```
Beeblerox: so is this variable still necessary?

larsiusprime: Yes. Unless we want to turn FlxTypedButton into a full @:generic -- there we some concerns about that (see MSGhero below), so just using an underlying _spriteLabel avoids the Dynamic casts in a minimally invasive way. Unfortunately there's no way for me to use the same minimal approach on FlxPool -- it's @:generic or nothing.

Gama11:
@larsiusprime It's not really a FlashDevelop isuse, but one with Haxe compiler completion. Unfortunately I wasn't able to reproduce it consistently and/or isolate it, so reporting it wouldn't really help much - so I decided [to just get rid of the @:generic](https://github.com/HaxeFlixel/flixel/commit/d0c8957226ab9d3e401508a076fb9541703abbd0).

It's a bit of a hack, but it probably helps with completion issues - we could wrap the metadata into `#if !display` again like I did before that commit I linked.

larsiusprime:
Sounds like an excellent compromise -- provided, of course, that you guys all agree that this change is worth doing (I certainly think it is, but you're all more experienced than me)

MSGhero:
It's totally worth doing (and I need to do that in my own generic classes). I didn't get a chance to try the generic tag wrapped in !display, so that might do it.

It's the most reproducible bug for me since it happens every time I hit period when involving certain classes, but I wouldn't know how to boil it down to its simplest form.

Beeblerox:
i like the idea with `#if !display`. we should try it!

larsiusprime:
@Gama11 where's that link to the "!display" thing? I'll insert it here if you show me how it's done.

Beeblerox:
@larsiusprime something like this?
```
#if !display
@:generic
class FlxTypedButton<T:FlxSprite> extends FlxSprite implements IFlxInput
#else
class FlxTypedButton<T:FlxSprite> extends FlxSprite implements IFlxInput
#end
```

or even
```
#if !display
@:generic
#end
class FlxTypedButton<T:FlxSprite> extends FlxSprite implements IFlxInput
```

larsiusprime:
Ah! Interesting, so you present one version to the compiler for actual compiling, and one version for code completion. CLEVER.

Beeblerox:
@larsiusprime have you tried to make `FlxTypedButton` class generic also and remove `_spriteLabel` variable? Will it help to avoid dynamics and using additional var at the same time?

larsiusprime:
I can definitely do that if you like! Wasn't there some issue with doing that though, that you wanted to avoid? Or was it just the completion thing that we can fix with an #if !display ?

Beeblerox:
just curious. i'm fine with this var in this class

Gama11:
The spriteLabel var feels like a hack to me - @:generic seems cleaner.

Also, what about subclasses that continue to use the regular label?

larsiusprime:
The spriteLabel var definitely is a hack. As for subclasses, the T: is already typed to be a FlxSprite, so any subclasses that use, say, a FlxText or other class won't be adversely afffected, because FlxText, FlxBitmapText, etc -- all of them inherit from FlxSprite. You wouldn't be able to use them as T's even before this change because FlxTypedButton requires T to at least be a FlxSprite.

MSGhero:
Not sure if it's applicable here, but I'm about to implement what nape does with arbiters.

A regular arbiter has `isCollisionArbiter`, `isSensorArbiter` methods and `castCollisionArbiter`, etc methods. The casts don't actually `cast` but they return `inner_collarb:CollisionArbiter` which is `this` if it's a collision arb or `null` if it's not.

I'm doing this because I have some interfaces that all "extend from" flxsprite, but I have to `cast` back and forth a lot. Basically emulating a Java abstract class.

larsiusprime:
I just did a test without the ```_spriteLabel``` and adding the generic, and the Dynamics came back.

I think this is because the FlxTypedButtons I'm using are FlxUITypedButtons:
```
class FlxUITypedButton<T:FlxSprite> extends FlxTypedButton<T> implements IFlxUIButton implements IResizable implements IFlxUIWidget implements IFlxUIClickable implements IHasParams implements ICursorPointable
```

So, in this case it will not properly create the underlying FlxTypedButton generics, but it will create the FlxUITypedButton generics if I add the metadata there. However, even then it will be using the dynamic version of FlxTypedButton for the super set_x/set_y, etc, calls, creating dynamics.

So absent a really creative solution the only way I can see to remove the Dynamic allocations in a way that works for both flixel and flixel-ui is to use the ```_spriteLabel``` trick.

Thoughts?

sruloart:
Use _spriteLabel if FLX_UI :)

larsiusprime:
@sruloart I thought the goal was to *avoid* hacks :P

sruloart:
It's not a hack, it's the official Haxe way of avoiding hard problems (aka "there's a flag for that!") :D
Seriously though, it's not much of a hack if the all point is optimization. Heck (pun fully intended), we can have a FLX_OPTIMIZE flag and let the hacks loose if they really save performance.

MSGhero:
Hacks don't make a difference to the user though, no reason not to use FLX_OPTIMIZE ever. We'd need a real-life ifdef, like `#if gama11 is looking, pretty solution #else hack`.

larsiusprime:
In any case, the PR as it stands represents my best attempt at the feature taking everyone's feedback into account.

HOW JUDGE YE

sruloart:
@MSGhero your solution won't work (no #end). but +1 for having a #if gama11#end :)

Gama11:
So if generic doesn't do anything for FlxButton, why does it have both generic meta and the _spriteLabel var now?

larsiusprime:
@Gama11 

If a user extends FlxTypedButton without the @:generic metadata present, they will fall through to the dynamic behavior and get poor performance.

Example 1:
```class SomethingButton extends FlxTypedButton<Something>```

Here the user is just trying to create a typed version of FlxTypedButton, without forwarding the T type to the user to type further. You'll only ever create a ```new SomethingButton()```, never a ```new SomethingButton<SomethingElse>()```.

In this case the @:generic metadata ensures that SomethingButton resolves as compile-time variant of FlxTypedButton in cpp.

However, even with the @:generic metadata present in FlxTypedButton, it's still possible for the user to extend FlxTypedButton in a way that causes it to resolve the T type of label in a dynamic way.

Example 2:
```class FlxFancyButton<T:FlxSprite> extends FlxTypedButton<T>```

Here the user is created an extension of FlxTypedButton and intending to forward the T type to the user to subtype themselves. So here a user would create ```new FlxFancyButton<Something>()```. In this case, all access to label in this class will be Dynamic, regardless of whether there is a @:generic present on FlxTypedButton.

Therefore, both the @:generic metdata and the _spriteLabel trick are necessary to cover all use cases.

In case you're wondering about Example 2, this is exactly the case that FlxUITypedButton falls into ;)

Gama11:
Gotcha, thanks for the in-depth explanation.

larsiusprime:
Cool, I'll push corresponding changes to flixel-ui now.

PR #1606, opened 2015-09-17
Use tile renderer for HTML5/WebGL

jgranick:
Fixes Mode for HTML5 / WebGL :smile:

### DIFF
--- a/flixel/system/macros/FlxDefines.hx
+++ b/flixel/system/macros/FlxDefines.hx
@@ -110,7 +110,7 @@ class FlxDefines
 	{
 		if (!defined(FLX_RENDER_BLIT) && !defined(FLX_RENDER_TILE))
 		{
-			if (defined("flash") || defined("js"))
+			if (defined("flash") || (defined("js") && !defined("webgl")))
 			{
 				define(FLX_RENDER_BLIT);
 			}


sruloart:
* Nice catch, maybe we should switch to webgl as default if we're at it, canvas has a weird issue where the background sprite won't show up (I guess it doesn't scale for some reason) and blitting creates another big bitmap for whatever the cause (build in debug mode, enter `viewCache` to see it, it's called "pixels"). 

* The debugger still has a few problems which are shared across canvas/webgl (it's not able to detect the memory used, and there's something wrong with the symbols placements).   

* I am noticing that setting `hardware="false"` doesn't seem to make any difference if `webgl` is defined, is that on purpose?

larsiusprime:
@sruloart flash has the same behaviour, it's carried over from the old days of flash flixel, it's basically a virtual framebuffer that's placed on the stage.

Gama11:
@jgranick How do you enable webgl? `<set name="webgl"/>` before including flixel does the trick for me, but that isn't a haxedef, which makes your change meaningless (`defined("webgl")` still returns false).

`<haxedef name="webgl"/>` doesn't seem to do anything.

sruloart:
@Gama11 Not true, this works for me after merging this change:

``` xml
<haxedef name="webgl" />
```

It's right in the middle of the haxedefs, and I know as a fact that the output is indeed webgl.

Bastcri:
Hello jgranick! not sure if this is the best way to contact you. We see that you know a lot about Three.js and we are learning. We have a question and were hoping you could help us. Please let me know if this is not the right place for asking. The question is how can we change colour in only certain areas of an imported 3D object. For example if we have an imported object that is a 3D car how can we change the colour of only the windows? Thanks for your help!

sruloart:
"we are learning" spooky. @Bastcri this tutorial is as a good place to start as any: http://solutiondesign.com/blog/-/blogs/webgl-and-three-js-texture-mappi-1/.

Bastcri:
Thanks sruloart! Yes spooky!! thanks for the link! If you know anything more specific to the question please let me know. We are not so rookies so hit us with as technical and code wise as you can!!
Thanks again!!

Tw1ddle:
@Bastcri The three.js developers will refer you to Stackoverflow where you can make these questions: http://stackoverflow.com/questions/tagged/three.js - also check this guide about how to best ask the question: http://stackoverflow.com/help/how-to-ask

Bastcri:
Thanks a million Tw1ddle!!

Gama11:
@sruloart What's the best way to tell the difference?

sruloart:
* First there's DrawTiles in the debugger. 
* With a GPU enabled a very slow moving sprite (I have a test for that) has a relative smooth motion. It's pretty noticeable. 
* And currently there's a bug with canvas (and not on webgl) that causes the bg sprite to disappear, so that's helpful :)

IkonOne:
This actually adversely affects performance on chrome android(and assuming probably other mobile browsers.

This is a screen of demo found in #1627 profiled on a Galaxy S4 using chrome remote debugging.
webgl-128

The same demo without webgl.
default-128

It should be noted also that, in Chrome at least, openfl doesn't render using webgl.  And as far as I know, openfl doesn't even have a webgl renderer yet.  If I am wrong here, please correct me and direct me to how to enable the webgl renderer...

mrcdk:
OpenFL supports webgl, it wouldn't make sense to create this PR if it wasn't :sweat_smile: It fallbacks to canvas if a webgl context can't be created (because the browser doesn't support it or there was a problem creating the context) You should check if your browser supports webgl using this page http://webglreport.com/

IkonOne:
Both Chrome and Firefox on my system support both versions of webgl.

I'm going to continue to play devils advocate here and say that I still don't think that openfl targets webgl when publishing to html5.

Just ctrl+f "webgl" in the changelog and you will see 2 mentions of webgl.  It does look like it was in there at some point, but canvas is the default renderer now.
https://github.com/openfl/openfl/blob/4320e84e9565521e8f240f6e2ce2870c82ad4970/CHANGELOG.md

Additionally, in Graphics.hx you can see that the canvas context used is ```js.html.CanvasRenderingContext2D```.  The context is found here: https://github.com/openfl/openfl/blob/master/openfl/display/Graphics.hx#L91

Finally, you can check what renderer is being used in the javascript console.  Here is the output from a webgl experiment http://labs.fluuu.id/geo/dist/:
fluid

And the output from a haxeflixel demo with the "webgl" flag: 
flixel demo program

It looks like to use webgl, we would have to write either a custom OpenGLView renderer or commit to Openfl.

mrcdk:
It's Lime the one in charge of creating the webgl context https://github.com/openfl/lime/blob/master/lime/_backend/html5/HTML5Renderer.hx#L56-L80 OpenFL uses offscreen canvases when you use vector graphics or image modifications via `BitmapData` methods.

That doesn't look that haxeflixel is using webgl, try this `openfl test html5 -Dwebgl -DFLX_RENDER_TILE` that should force webgl and the non-blitting render path of haxeflixel. 

This is from a few days ago:

IkonOne:
Yes I found that just after I posted that wall of text.  I didn't look at the lime backends because I had read documentation somewhere that lime does not implement rendering.  It may have taken me an entire day to find all this out, but it was a good lesson!

gamedevsam:
What's the status of this PR?

larsiusprime:
Hey everyone! I'm extremely interested in getting this up to speed, any way I can help?

Gama11:
This isn't really relevant anymore after #1668.

PR #1610, opened 2015-09-22
make FlxAngle.wrapAngle actually wrap the angle

l0rb:
old wrapAngle(190) returns 180
new wrapAngle(190) returns -170 (where +190 actually would be)
in cases where you want the old behaviour you can use FlxAngle.angleLimit(190,-180,180)


Gama11:
Guess nobody was really using this (not a single usage of it across all flixel repos at least).. The implementation is even the same in the AS3 version / the flixel power tools.

Thanks!

PR #1611, opened 2015-09-23
Preventing Memory leak and exposing FlxRect and FlxPoint pools to the user

Tiago-Ling:
The FlxPoints couldn't be 'putWeak' inside 'FlxMath.getDistance' because 'get()' was used instead of 'weak()'.


Gama11:
Thanks! :+1:

PR #1612, opened 2015-09-23
Making FlxPoint and FlxRect use their pools

Tiago-Ling:
- Updated code in some classes to get all `FlxPoint`s and `FlxRect`s through their pools instead of using `new`


MSGhero:
Woot!

PR #1613, opened 2015-09-23
disable htmltext in debugger for lime_console

larsiusprime:


Gama11:
What's `lime_console`?

mrcdk:
The console backend

PR #1614, opened 2015-09-24
FlxTween: add .wait(), and .then() for chaining, closes #1064

### LINKED ISSUE #1064: Add better "tween chaining" option
it can be done only with tween onComplete() now, which isn't very good

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/2084992-add-better-tween-chaining-option?utm_campaign=plugin&utm_content=tracker%2F254189&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F254189&utm_medium=issues&utm_source=github).
</bountysource-plugin>

larsiusprime:
This lets you go from this:

```Haxe
FlxTween.tween(this, { y:endY }, time, { onComplete: 
    function(f:FlxTween):Void {
        f.cancel();
        new FlxTimer().start(DISPLAY_TIME, function(t:FlxTimer):Void {
            t.cancel();
            FlxTween.tween(this, { y:startY },time, { onComplete: finishAnimation } );
        });
    }
});
```		

To this:
```Haxe
FlxTween.tween(this, { y:endY }, time).thenWait(DISPLAY_TIME).thenTween(this, { y:startY }, time, {onComplete: finishAnimation } );
```

### DIFF
--- a/flixel/tweens/FlxTween.hx
+++ b/flixel/tweens/FlxTween.hx
@@ -6,7 +6,6 @@ import flixel.FlxSprite;
 import flixel.tweens.FlxEase.EaseFunction;
 import flixel.tweens.FlxTween.ThenCommand;
 import flixel.tweens.FlxTween.TweenOptions;
-import flixel.tweens.FlxTween.TweenMethod;
 import flixel.tweens.misc.AngleTween;
 import flixel.tweens.misc.ColorTween;
 import flixel.tweens.misc.NumTween;
--- a/flixel/tweens/FlxTween.hx
+++ b/flixel/tweens/FlxTween.hx
@@ -4,7 +4,6 @@ import flixel.FlxG;
 import flixel.FlxObject;
 import flixel.FlxSprite;
 import flixel.tweens.FlxEase.EaseFunction;
-import flixel.tweens.FlxTween.ChainedTween;
 import flixel.tweens.FlxTween.TweenOptions;
 import flixel.tweens.misc.AngleTween;
 import flixel.tweens.misc.ColorTween;
@@ -580,6 +579,10 @@ class FlxTween implements IFlxDestroyable
 				{
 					doNextTween(then.tween, _thens);
 				}
+				else
+				{
+					onEnd();
+				}
 			}
 			else
 			{
@@ -794,7 +797,11 @@ class FlxTweenManager extends FlxBasic
 		}
 		
 		Tween.active = false;
-		Tween.destroy();
+		
+		if (Destroy)
+		{
+			Tween.destroy();
+		}
 		
 		FlxArrayUtil.fastSplice(_tweens, Tween);
 		
--- a/flixel/tweens/FlxTween.hx
+++ b/flixel/tweens/FlxTween.hx
@@ -10,7 +10,7 @@ import flixel.tweens.FlxTween.QuadMotionParams;
 import flixel.tweens.FlxTween.QuadPathParams;
 import flixel.tweens.FlxTween.ThenCommand;
 import flixel.tweens.FlxTween.TweenOptions;
-import flixel.tweens.FlxTween.TweenType;
+import flixel.tweens.FlxTween.TweenMethod;
 import flixel.tweens.misc.AngleTween;
 import flixel.tweens.misc.ColorTween;
 import flixel.tweens.misc.NumTween;
@@ -425,7 +425,7 @@ class FlxTween implements IFlxDestroyable
 	 * @return
 	 */
 	
-	public function then(tweenType:TweenType, params:Dynamic):FlxTween
+	public function then(tweenType:TweenMethod, params:Dynamic):FlxTween
 	{
 		if (_thens == null)
 		{
@@ -457,7 +457,7 @@ class FlxTween implements IFlxDestroyable
 	 * @param	params
 	 * @return
 	 */
-	public function waitThen(Delay:Float, tweenType:TweenType, params:Dynamic):FlxTween
+	public function waitThen(Delay:Float, tweenType:TweenMethod, params:Dynamic):FlxTween
 	{
 		if (_thens == null)
 		{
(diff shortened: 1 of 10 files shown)


MSGhero:
Is this sort of a Promise thing? I'm actually trying to implement those in my project right now.

Tiago-Ling:
@larsiusprime Very nice! Chaining callbacks makes the code much more readable - does it change anything performance-wise?

larsiusprime:
I think the two syntaxes I showed here will be about the same. 

Also, tweens have never been particularly optimized with regards to performance -- it's using a big fat dynamic object to hold all your tweenable values, which it accesses with reflection, so even if this syntax added some weight it probably wouldn't be a big deal. 

But the whole point of tweens (as I understand it, I could be wrong) is an easy and simple way to animate, and it just has to be "fast enough" to not be the bottleneck.

Tiago-Ling:
Yeah, i was actually hoping that somehow this could improve performance - even removing it from the call, the anonymous object still exists inside the tween right?

larsiusprime:
Well, we should really profile it first to see if it's actually causing a bottleneck -- usually drawing, not update functions like this, are the culprit. Unless inefficient tweens are one of the top 10 causes of slowdown, or eating lots of memory, it's usually not worth optimizing it no matter how "bad" it looks.

Tiago-Ling:
I know, it's not a bottleneck at all, at least here. If your entire game is made around FlxTweens it could matter though. Anyway i just wondered if there was any additional reason besides ease of use and readability for this change. :+1:

larsiusprime:
Yeah, that makes sense! This is just a syntax upgrade -- it might have the slight benefit of not having to instantiate a new FlxTimer(), but I'm not sure that matters at all.

If we did want to optimize out that dynamic object, I'm not sure it could be done without making the syntax a lot less convenient... I guess we'll just wait to see if anyone is using tweens for *EVERYTHING* 0_0

larsiusprime:
It looks like the Travis CI build that are failing is in RPG Interface, related to FlxUI:

```
/home/travis/haxe/lib/flixel-ui/git/flixel/addons/ui/FontDef.hx:309: characters 3-7 : String should be openfl.text.TextFormatAlign
/home/travis/haxe/lib/flixel-ui/git/flixel/addons/ui/FontDef.hx:309: characters 3-7 : String should be openfl._legacy.text.TextFormatAlign
```

Gama11:
I don't really like how this is `VarTween`-specific, there are a lot more tween types after all. I guess it is the one most commonly used, but it still makes for an inconsistent API.

Interested in @JoeCreates' opinion on this since he seemed to have strong opinions on this in #1064. Also wondering if and how this affects his refactor concerning #1087.

larsiusprime:
It could probably be refactored to not require VarTween

larsiusprime:
@Gama11 Okay, did a quick refactor. What do you think now?

flixel/tweens/FlxTween.hx
```diff
@@ -621,7 +745,120 @@ typedef TweenOptions = {
...
+	Object:FlxObject,
+	Points:Array<FlxPoint>,
+	?DurationOrSpeed:Float,
+	?UseDuration:Bool,
+	?Options:TweenOptions
+}
+
+typedef TweenParams = OneOfNine<VarTweenParams,AngleTweenParams,ColorTweenParams,NumTweenParams,CircularMotionParams,CubicMotionParams,LinearMotionParams,QuadMotionParams,QuadPathParams>;
+
+enum TweenType
```
Gama11: Calling this "tween type" conflicts with `FlxTween.PINGPONG` which are also considered tween types. It could see that causing quite a bit of confusion.

larsiusprime: Alternative?

larsiusprime:
@Gama11 that better?

Gama11:
Definitely less ambigious, yeah.

flixel/tweens/FlxTween.hx
```diff
@@ -621,7 +745,120 @@ typedef TweenOptions = {
...
+	?DurationOrSpeed:Float,
+	?UseDuration:Bool,
+	?Options:TweenOptions
+}
+
+typedef TweenParams = OneOfNine<VarTweenParams,AngleTweenParams,ColorTweenParams,NumTweenParams,CircularMotionParams,CubicMotionParams,LinearMotionParams,QuadMotionParams,QuadPathParams>;
+
+enum TweenMethod
+{
+	None_;
```
Gama11: Is `TweenMethod.None` actually needed for anything?

Gama11: Also, are the `_` suffixes necessary?

larsiusprime: None is needed, yes, at least in this implementation. I use it for implementing the delays.

The suffixes are unfortunately necessary. I couldn't come up with a good naming convention otherwise:

attempt 1: remove "Tween"
```
AngleTween;   //<---- these look fine
ColorTween;
NumTween;
VarTween;
CircularMotionTween;   //<---- these conflict with their corresponding classes
CubicMotionTween;
LinearMotionTween;
QuadMotionTween;
QuadMotionTween;
```

attempt 2: just use one word
```
Angle;   //<----these look fine
Color;
Num;
Var;
Circular; //<----Circular what? Cubic what?
Cubic;
Linear;
Quad;
Quad;
```

attempt 3: end with "tween"
```
AngleTween;   //<---- these conflict with the classes now
ColorTween;
NumTween;
VarTween;
CircularMotionTween;   //<---- these are okay
CubicMotionTween;
LinearMotionTween;
QuadMotionTween;
QuadMotionTween;
```

Gama11:
Do they conflict even if you use the fully qualified name? (`TweenType.AngleTween` instead of just `AngleTween` for example).

larsiusprime:
Maybe not, I haven't checked. Would it be confusing though?

larsiusprime:
It does work fully-qualified. Is it okay if the names are otherwise the same as the classes?

Gama11:
Not sure...

This might not be the best approach to begin with. You're collecting the data needed to be able to act as a "tween factory" later when the tween whose turn it is is needed. That data is the "method" (kind of awkward naming here) and the data (dynamically typed).

What about creating all the tweens in advance, which means `.then()`'s signature would be `FlxTween->FlxTween`? Usage would look something like

```haxe
FlxTween.tween(this, { y:endY }, time).
	thenWait(DISPLAY_TIME).
	thenTween(FlxTween.tween(this, { y:startY }, time, {onComplete: finishAnimation } ));
```

larsiusprime:
Yeah, that could work I guess

larsiusprime:
@Gama11 how's it look now?

Gama11:
All those param typedefs etc can be removed now right?

larsiusprime:
@Gama11 : yup. Just forgot.

flixel/tweens/FlxTween.hx
```diff
@@ -406,8 +410,61 @@ class FlxTween implements IFlxDestroyable
...
+	}
+	
+	/**
+	 * Add a delay and a chained tween in the same command
+	 * @param	Delay
+	 * @param	method
+	 * @param	params
+	 * @return
+	 */
+	public function waitThen(Delay:Float, Tween:FlxTween):FlxTween
```
Gama11: Is there really much of a point to this combined function? It doesn't really add anything, and it's not really shorter either. Compare:

```haxe
wait(1).then(tween)
waitThen(1, tween)
```

The first one seems more readable to me.

flixel/tweens/FlxTween.hx
```diff
@@ -406,8 +410,61 @@ class FlxTween implements IFlxDestroyable
 		onUpdate = null;
 		onComplete = null;
 		ease = null;
+		_thens = null;
 	}
-
+	
+	/**
+	 * After this tween has finished, do this tween next
+	 * @param	method
```
Gama11: Might as well delete these empty `@param` tags, especially considering they're out of sync with the method signature.

flixel/tweens/FlxTween.hx
larsiusprime: Sure, that's fine. One less function to maintain.

flixel/tweens/FlxTween.hx
```diff
@@ -621,7 +718,7 @@ typedef TweenOptions = {
 	?onUpdate:TweenCallback,
 	?onComplete:TweenCallback,
 	?startDelay:Null<Float>,
-	?loopDelay:Null<Float>,
```
Gama11: Interesting, didn't know Haxe allowed a trailing comma here for the last element. :)

flixel/tweens/FlxTween.hx
```diff
@@ -735,3 +853,15 @@ class FlxTweenManager extends FlxBasic
 		}
 	}
 }
+
+class ThenCommand
```
Gama11: Not sure I like the name `ThenCommand`. `ChainedTween` maybe? That doesn't really cover the wait use-case though.. `ChainedCommand`? `ChainedAction`?

Thoughts?

flixel/tweens/FlxTween.hx
```diff
@@ -4,6 +4,9 @@ import flixel.FlxG;
 import flixel.FlxObject;
 import flixel.FlxSprite;
 import flixel.tweens.FlxEase.EaseFunction;
+import flixel.tweens.FlxTween.ThenCommand;
+import flixel.tweens.FlxTween.TweenOptions;
+import flixel.tweens.FlxTween.TweenMethod;
```
Gama11: Kind of doubt this compiles anymore with `TweenMethod` removed. ;)

flixel/tweens/FlxTween.hx
Gama11: Also, there's not really any reason to expose this API to the user, would mean we have to worry about backwards compatibilty etc. Should be a `private class`.

flixel/tweens/FlxTween.hx
```diff
@@ -723,6 +799,28 @@ class FlxTweenManager extends FlxBasic
...
+	
+	/**
+	 * Remove a FlxTween without destroying it
+	 * 
+	 * @param	Tween		The FlxTween to remove.
+	 * @param	Destroy		Whether you want to destroy the FlxTween.
+	 * @return	The added FlxTween object.
+	 */
+	@:allow(flixel.tweens.FlxTween)
+	private function removeSoft(Tween:FlxTween):FlxTween
```
Gama11: Might make sense to bring back the Destroy flag to the original `remove()` considering this is a 1:1 copy of that otherwise.

flixel/tweens/misc/VarTween.hx
```diff
@@ -60,6 +61,12 @@ class VarTween extends FlxTween
 		return this;
 	}
 	
+	@:access(flixel.tweens.FlxTween)
+	override function onEnd():Void
```
Gama11: This override doesn't look like it does anything anymore?

larsiusprime:
Good now?

Gama11:
I've added some more comments, don't think you've addressed those yet?

flixel/tweens/FlxTween.hx
larsiusprime: Kay

flixel/tweens/FlxTween.hx
```diff
@@ -523,6 +567,39 @@ class FlxTween implements IFlxDestroyable
...
 		_running = false;
 		finished = true;
+		
+		if (_thens != null && _thens.length > 0)
+		{
+			var then = _thens[0];
+			
+			_thens.splice(0, 1);
+			
+			if (then.delay <= 0)
```
Gama11: This looks like it might fail with `wait(0)` (which should just be "no wait" - doesn't make sense if written like that but the delay could be dynamically set from some variable that happens to become 0 at some point)? And what's the expected result for negative values?

flixel/tweens/FlxTween.hx
larsiusprime: ChainedTween is fine.

flixel/tweens/FlxTween.hx
larsiusprime: So this instead?
```Haxe
if (then.delay <= 0 && then.tween != null)
{
	doNextTween(then.tween, _thens);
}
else
{
	doNextTween(FlxTween.num(0,0,then.delay), _thens);
}
```

larsiusprime: Working on it...

larsiusprime: Right you were, it needs to call onEnd() again in the case that delay is <= 0 AND the tween is null, signifying it's a delay that should terminate instantly and run the next "then" command.

larsiusprime:
@Gama11 good now? Just tested it and it seems to work fine. Could probably use a demo eventually.

larsiusprime:
@Gama11 Can we merge this today? I think I've addressed everything.

PR #1617, opened 2015-09-26
Fixed bug with diagonally-cast FlxTilemap.ray()

Nihilate:
If your start and end points were perfectly diagonal, FlxTilemap.ray() would fail to detect a collision. This was caused by the ray assuming that whenever it checked a position within a tile it would fall *inside* the bounds, rather than exactly on one of the tile's corners.

### DIFF
--- a/flixel/tile/FlxTilemap.hx
+++ b/flixel/tile/FlxTilemap.hx
@@ -731,7 +731,7 @@ class FlxTilemap extends FlxBaseTilemap<FlxTile>
 				rx = q;
 				ry = ly + stepY * ((q - lx) / stepX);
 				
-				if ((ry > tileY) && (ry < tileY + _scaledTileHeight))
+				if ((ry >= tileY) && (ry <= tileY + _scaledTileHeight))
 				{
 					if (Result == null)
 					{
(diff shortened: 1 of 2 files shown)


Gama11:
Got some code that reproduces the problem (preferably a unit test :))?

Nihilate:
No worries, I'll throw something together.

Nihilate:
Here a re a few unit tests using haxe.unit that show the 1:1 diagonal ray failing to get the correct result. Wasn't sure how you wanted to take a look, so here's a Dropbox link: https://dl.dropboxusercontent.com/u/10259753/FlxTilemapRayTests.zip

Gama11:
Awesome!

There's a munit project on this repo - these should go into [`FlxTilemapTest`](https://github.com/HaxeFlixel/flixel/blob/dev/tests/unit/src/flixel/tile/FlxTilemapTest.hx).

Gama11:
The unit tests now finally have [a Readme](https://github.com/HaxeFlixel/flixel/blob/dev/tests/unit/README.md) btw.

Nihilate:
Great, thanks! I was having some issues with it - turns out my version of Haxe was still on 3.1.3 haha. I've added a few unit tests that test out ray functionality (blank tilemap, horizontal, 1:1 diagonal and slightly-less-than 1:1 diagonal).

Gama11:
Nice, thanks! :)

[You should've gotten a helpful error message for that though?](https://github.com/HaxeFlixel/flixel/blob/2a296be8e422c784acc394dd46d0f7ae9e0b48e0/flixel/system/macros/FlxDefines.hx#L62-L65)

Nihilate:
I did, but it was like 4am and I'm an idiot haha. Took me all of about 3 minutes to figure out once I'd got some sleep and had my morning caffeine.

PR #1618, opened 2015-09-27
Remove FlxAngle#angleLimit()

Gama11:
It does exactly the same as `FlxMath.bound()` / there's nothing angle-specific about it. In that way, it's even misleading because the `@param` docs claim the values must be between -180 and +180.

Pinging @HaxeFlixel/owners since this is a breaking change.

### DIFF
--- a/flixel/math/FlxAngle.hx
+++ b/flixel/math/FlxAngle.hx
@@ -82,29 +82,6 @@ class FlxAngle
 	}
 	
 	/**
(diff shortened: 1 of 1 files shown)


Gama11:
There aren't any usages of the method anywhere, except two commented-out ones in `flixel.addons.plugin.control.FlxControlHandler` (looks like something that's not implemented yet / TODO).

Beeblerox:
I agree that this method could be deleted from the codebase.

PR #1625, opened 2015-10-05
added zeroPad() function

leoncvlt:
A simple function to add leading zeroes to a number - seen and been using this a lot in arcade-style games.


MSGhero:
`StringTools` already has `lpad()` to do this.

PR #1626, opened 2015-10-05
Changing alpha no longer updates the ColorTransform if the value set …

IkonOne:
…is not different than the current value.


Gama11:
Seems sensible.

Any particular reason this came up? Does updating the color transform have a performance impact?

Beeblerox:
`set_alpha()` calls `updateColorTransform()` which sets `dirty` to true which means `framePixels` regeneration on flash target. So yes, this change is useful for flash.

IkonOne:
This came up because I am using the FlxSpine addon and the class set's alpha every frame on update.

Issue #1156, opened 2014-06-06
Consider partially rewriting the console to use hscript


Gama11:
From what I've seen, hscript is quite powerful, and we would support a subset of haxe code from within the console basically. I think this could be done in a way that doesn't make hscript a dependency. (i.e. when trying to use a command that uses hscript, print an error in the console saying it has to be installed, and cover the relevant code parts in `#if hscript`).

PR #1631, opened 2015-10-12
Add get command to console, Closes #1621

### LINKED ISSUE #1621: "get" console command
Playing around with the debugger for the first time. I know there's the watch table already, but sometimes I just want to see the value of something without writing it in the code. "get hero.isDead" or something using `getProperty()`. I'm not familiar with `addCommand()`, otherwise I would have pulled something.

Also, "fields" doesn't do anything for me, should it be listing things in the log window?

Recent dev, not the latest.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27042783-get-console-command?utm_campaign=plugin&utm_content=tracker%2F254189&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F254189&utm_medium=issues&utm_source=github).
</bountysource-plugin>

MSGhero:
#1621 

These work:
"get sprite"
"g sprite.x"
"get sprite.cameras.length"
"g sprite.cameras[0]"
"get arr" <- "arr" is [0, 1, 2, 3]

This doesn't:
"get arr[0]" <- when there is only array access, no properties

I'm not totally sure all the code in that function (or in set) gets used, but it works for the above cases.

Edit: Part of me thinks using hscript here would save a lot of code and be able to handle more cases.

### DIFF
--- a/flixel/system/debug/ConsoleCommands.hx
+++ b/flixel/system/debug/ConsoleCommands.hx
@@ -35,6 +35,7 @@ class ConsoleCommands
 		console.addCommand(["create", "cr"], create, "Creates a new FlxObject and registers it - by default at the mouse position.", 
 							"[FlxObject] (MousePos = true)", 3, 3);
 		console.addCommand(["set", "s"], set, "Sets a variable within a registered object.", "[Path to variable] [Value]", 3);
+		console.addCommand(["get", "g"], get, "Gets a variable within a registered object or the object itself.", "[Path to variable]", 2);
 		console.addCommand(["call", "c"], call, "Calls a registered function / function within a registered object.", 3, 2);
 		console.addCommand(["fields", "f"], fields, "Lists the fields of a class or instance", "[Class or path to instance] [NumSuperClassesToInclude]", 2);
 		
@@ -210,7 +211,7 @@ class ConsoleCommands
 
 		if (isArrayIndex && variable.length <= index)
 		{
-			FlxG.log.error("set: '" + varName + ":" + FlxStringUtil.getClassName(variable, true) + "' cannot access indice " + index);
+			FlxG.log.error("set: '" + varName + ":" + FlxStringUtil.getClassName(variable, true) + "' cannot access index " + index);
 			return;
 		}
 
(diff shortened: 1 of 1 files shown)


Gama11:
I definitely agree that using hscript would be a good idea, see #1156.

MSGhero:
Should the syntax stay the same, "get"/"set"_space_parameters? Hscript works without the get or set, and it's sorta not a breaking change since you manually type these things.

Gama11:
Not sure if I'd consider changing console commands a breaking change, it's not the same as changing API that leads to compiler errors... :) Either way, I think it's fine and even preferable if the syntax changes.

You actually wanna tackle #1156? Seems like a big refactor. ;)

sruloart:
You can just try using dconsole + Mint/HaxeUI for some kind of proper GUI :)

Gama11:
A GUI for a command-line? :s

sruloart:
For a debugger.

Gama11:
>but for some reason people really dislike that strategy

That's a bit of a generalization. There are a lot of people that like it as well.

> I would rather integrate proper existing tools than to keep some old ones

I'd like the keep the dependencies to a minimum. Not really sure what we'd need HaxeUI for.

>like the current debugger's Window

What's wrong with it?

How did we even get here? This issue is only related to the console window.

MSGhero:
Yeah sure I'll take it. Are there any preferences? Or is the way DConsole handles it fine?

Gama11:
I'm not sure how dconsole handles it actually, haven't looked into it too much.

sruloart:
@Gama11 I guess I'm just venting-out some frustration regarding the GUI of debugger, which is just not functional as a GUI imho. 

I was just stating the obvious: Mint is functional, it's available, why should we use a "unique" Window implementation if the only thing you can do with it is to move it around? 

And not adding dependencies is kind of my point. 

@MSGhero I believe that it does. With hscript. (https://github.com/tiagolr/dconsole).

MSGhero:
`thing.x`
`thing.x =50`
`arr[0] = thing'
It's regular haxe syntax, I mainly meant should it have any kind of flixel flair to it, or should it look exactly like DConsole.

sruloart:
@MSGhero What kind of flair? adding `Flx`?

Gama11:
Regular haxe syntax is exactly what I like about hscript. :)

MSGhero:
This gives me a chance to figure out SourceTree since I still haven't used it yet.

MSGhero:
Is there a way to easily toggle between git flixel and my local dev flixel for compiling? I can do `haxelib dev flixel path/to/local/flixel` and `haxelib dev flixel`, but copying that file path is a bit annoying to do each time.

Gama11:
With SourceTree it's juts a double-click on the respective tag / branch.

PR #1634, opened 2015-10-16
PostProcess shader clashing with OpenFL

tanis2000:
The PostProcess shader of Flixel uses a variable called shader. OpenFL's OpenGLView has another variable with the same name and the two clash with each other.
I renamed Flixel's one to overcome the problem.

I also added an initial support for MFi's controllers.


Gama11:
Thanks!

Can't really test the MFi stuff, so I'm just gonna have to assume that it works. Looks reasonable anyway. :)

PR #1635, opened 2015-10-16
Added filter support for FlxCamera and FlxGame

mrcdk:
This commit plus the latest release of OpenFL with custom shader support (so, using `next` not `legacy`) makes possible to have hardware accelerated filters (aka post-processes) per camera (`FlxCamera`) and a global filters support (`FlxGame`)

~~This deprecates the usage of `PostProcess`~~ This makes possible to use post process effects with `next`

### DIFF
--- a/flixel/FlxCamera.hx
+++ b/flixel/FlxCamera.hx
@@ -21,6 +21,7 @@ import flixel.util.FlxColor;
 import flixel.util.FlxDestroyUtil;
 import flixel.util.FlxSpriteUtil;
 import openfl.display.BlendMode;
+import openfl.filters.BitmapFilter;
 import openfl.geom.Matrix;
 import openfl.Vector;
 
@@ -200,6 +201,10 @@ class FlxCamera extends FlxBasic
 	 * Used to force the camera to look ahead of the target.
 	 */
 	public var followLead(default, null):FlxPoint;
+	/**
+	 * Enables or disables the filters set via setFilters()
+	 */
+	public var enableFilters:Bool = true;
 	
 	/**
 	 * Internal, used to render buffer to screen space.
@@ -281,6 +286,10 @@ class FlxCamera extends FlxBasic
 	 * Internal, to help avoid costly allocations.
 	 */
 	private var _point:FlxPoint;
+	/**
+	 * Internal, the filters array to be applied to the camera.
+	 */
+	private var _filters:Array<BitmapFilter>;
 	
 	/**
 	 * Camera's initial zoom value. Used for camera's scale handling.
@@ -786,6 +795,8 @@ class FlxCamera extends FlxBasic
 		updateFade(elapsed);
 		updateShake(elapsed);
 		
+		flashSprite.filters = enableFilters ? _filters : null;
+		
 		updateFlashSpritePosition();
 	}
 	
@@ -1187,6 +1198,16 @@ class FlxCamera extends FlxBasic
 		_fxFadeAlpha = 0.0;
 		_fxShakeDuration = 0;
 		updateFlashSpritePosition();
+	}
+	
+	/**
+	 * Sets the filter array to be applied to the camera.
+	 * 
+	 * @param	filters
+	 */
+	public function setFilters(filters:Array<BitmapFilter>):Void
+	{
+		_filters = filters;
 	}
 	
 	/**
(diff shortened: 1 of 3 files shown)


larsiusprime:
\o/

Mixerman123:
Woot!

sruloart:
I think this is worthy of switching the default `legacy` to `next`. Major kudos :D

flixel/FlxCamera.hx
```diff
@@ -1190,6 +1201,16 @@ class FlxCamera extends FlxBasic
 	}
 	
 	/**
+	 * Sets the filter array to be applied to the camera.
+	 * 
+	 * @param	filters
+	 */
+	public function setFilters(filters:Array<BitmapFilter>):Void
```
Gama11: What's the advantage of having a setter function? Why not just make `_filters` a `public var filters`?

flixel/FlxCamera.hx
```diff
@@ -786,6 +795,8 @@ class FlxCamera extends FlxBasic
 		updateFade(elapsed);
 		updateShake(elapsed);
 		
+		flashSprite.filters = enableFilters ? _filters : null;
```
Gama11: Is this safe performance-wise? I'd imagine this would call `set_filters()` each frame.

flixel/effects/postprocess/PostProcess.hx
```diff
@@ -61,21 +62,21 @@ class PostProcess extends OpenGLView
 
 		buffer = GL.createBuffer();
 		GL.bindBuffer(GL.ARRAY_BUFFER, buffer);
-		GL.bufferData(GL.ARRAY_BUFFER, new Float32Array(cast vertices), GL.STATIC_DRAW);
+		GL.bufferData(GL.ARRAY_BUFFER, new Float32Array(#if !openfl_next cast #end vertices), GL.STATIC_DRAW);
 		GL.bindBuffer(GL.ARRAY_BUFFER, null);
 
-		shader = new Shader([
+		pshader = new PShader([
```
Gama11: Merge conflicts with the changes from #1634.

flixel/FlxGame.hx
```diff
@@ -259,6 +269,16 @@ class FlxGame extends Sprite
 	}
 	
 	/**
+	 * Sets the filter array to be applied to the game.
+	 * 
+	 * @param	filters
+	 */
+	public function setFilters(filters:Array<BitmapFilter>):Void
```
Gama11: This setter seems even more redundant, considering `FlxGame` inherits `filters` from `DisplayObject`.

flixel/effects/postprocess/PostProcess.hx
```diff
@@ -61,21 +62,21 @@ class PostProcess extends OpenGLView
 
 		buffer = GL.createBuffer();
 		GL.bindBuffer(GL.ARRAY_BUFFER, buffer);
-		GL.bufferData(GL.ARRAY_BUFFER, new Float32Array(cast vertices), GL.STATIC_DRAW);
+		GL.bufferData(GL.ARRAY_BUFFER, new Float32Array(#if !openfl_next cast #end vertices), GL.STATIC_DRAW);
```
Gama11: >OpenFL next doesn't need to cast.

Does it really matter to have the `cast` there in next anyway though? Unsafe casts shouldn't have much of a runtime overhead.

mrcdk: The new typed arrays implementation chooses how to construct the array from the type of the argument, having it as an unsafe cast breaks it.

flixel/FlxGame.hx
mrcdk: `filters` will cache the object after applying the filters. By default, every time you modify the object the cache will be updated, but because we are using `drawTiles()` on native the object won't be notified about the change so we need to reapply the filters every time we want it to be updated.

Adding this we can control how and when to reapply the filters without the user having to do the same by himself so...

flixel/FlxCamera.hx
mrcdk: To add a bit of coherence between the `FlxCamera` and `FlxGame` API. Read https://github.com/HaxeFlixel/flixel/pull/1635#discussion_r42324248

flixel/FlxCamera.hx
mrcdk: It should, read https://github.com/HaxeFlixel/flixel/pull/1635#discussion_r42324248 to know why it has to apply the filters each time. And if there are no filters `set_filters` only flags the `DisplayObject` to update which is already true because other modifications.

PR #1636, opened 2015-10-19
fix null colortransform issue

ramsestom:
null colortransform in FlxSprite and extended classes making haxeflixel dev to crash when using openfl next...


Gama11:
The additional checks definitely don't hurt, but I'm not sure how colorTransform would become null (seing how it's initialized in `initVars()` which is called from the constructor) or how that would be related to next.

PR #1637, opened 2015-10-20
Hscript console, closes #1156

### LINKED ISSUE #1156: Consider partially rewriting the console to use hscript

MSGhero:
The console itself works fine now, but *now* I need some confirmation that these changes are ok. There are some breaking changes now.

### DIFF
--- a/flixel/system/debug/ConsoleUtil.hx
+++ b/flixel/system/debug/ConsoleUtil.hx
@@ -14,6 +14,9 @@ class ConsoleUtil
 	// The custom hscript interpreter to run the haxe code from the parser.
 	public static var interp:Interp;
 	
+	/**
+	 * Sets up the hscript parser and interpreter.
+	 */
 	public static function init():Void
 	{
 		parser = new Parser();
@@ -25,18 +28,20 @@ class ConsoleUtil
 	
 	/**
 	 * Converts the input string into its AST form to be executed.
+	 * 
 	 * @param	Input	The user's input command.
 	 * @return	The parsed out AST.
 	 */
 	public static function parseCommand(Input:String):Expr
 	{
 		if (StringTools.endsWith(Input, ";"))
-			Input = Input.substr( -1);
+			Input = Input.substr(0, -1);
 		return parser.parseString(Input);
 	}
 	
 	/**
 	 * Parses and runs the input command.
+	 * 
 	 * @param	Input	The user's input command.
 	 * @return	Whatever the input code evaluates to.
 	 */
@@ -44,12 +49,24 @@ class ConsoleUtil
 		return interp.expr(parseCommand(Input));
 	}
 	
+	/**
+	 * Register a new object to use in any command.
+	 * 
+	 * @param	ObjectAlias	The name with which you want to access the object.
+	 * @param	AnyObject	The object to register.
+	 */
 	public static function registerObject(ObjectAlias:String, AnyObject:Dynamic):Void
 	{
 		if (Reflect.isObject(AnyObject))
 			interp.variables.set(ObjectAlias, AnyObject);
 	}
 	
+	/**
+	 * Register a new function to use in any command.
+	 * 
+	 * @param 	FunctionAlias	The name with which you want to access the function.
+	 * @param 	Function		The function to register.
+	 */
 	public static function registerFunction(FunctionAlias:String, Function:Dynamic):Void
 	{
 		if (Reflect.isFunction(Function))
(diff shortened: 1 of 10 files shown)


larsiusprime:
neato

MSGhero:
I'm mainly done now, but yeah it's a breaking change. Since hscript deals with objects directly instead of strings, I removed all the parsing code and string parameters from functions.

MSGhero:
Pausing and TAB don't work on neko, but they probably never did. Didn't test html5.

MSGhero:
The travis script and unit tests work fine on my computer, and the build does install hscript. Why can't it find hscript.Expr?

Gama11:
As is, every flixel project that doesn't include hscript in the xml manually won't compile - would be better to have that in the include.xml.

In theory it would be nice to only include it if it's installed (and only have the console be usable in that case), but I'm not sure if there's a way to do that with the project xml format...

MSGhero:
I added it to include.xml, but would it go somewhere besides the `setup` block? Should I wrap all console code in `#if hscript`?

Gama11:
`setup` is only used for `lime setup flixel`. If you actually want to include the haxelib during compilation, you need the regular `<haxelib name="hscript" />` like the `include.xml` does with openfl.

MSGhero:
Ugh ok. I see what you mean now about always including it. I guess it would always be included for `release` builds too even when the debugger flxdef is off.

MSGhero:
The unit tests look fine, but it's still breaking coverage on the `import hscript.Expr`

Gama11:
The unit tests use a .hxml file for flash.

https://github.com/HaxeFlixel/flixel/blob/dev/tests/unit/test.hxml

MSGhero:
I'm never adding an external lib to flixel again.

MSGhero:
Oh hey, dev cpp is working again. Haxe recently fixed the multiple inheritance issue which was preventing flixel-ui from compiling.

gamedevsam:
This change looks great! Has anyone been able to test this functionality?

MSGhero:
Me lol. Do you need a rundown of the changes or are you making sure it works for other people?

gamedevsam:
I was mostly wondering if there was someone in the community that opposes this change or has something to say before I merge it in. @HaxeFlixel/owners any opposition or can I merge this?

Beeblerox:
i haven't checked this pull request, but i don't mind replacing current functionality with hscript as far as it works

flixel/system/debug/ConsoleUtil.hx
```diff
@@ -380,9 +85,20 @@ class ConsoleUtil
 }
 
 /**
- * Data necessary to set a variable via Reflection. Used by the set command / resolveObjectAndVariable().
+ * hscript doesn't use property access by default... have to make our own.
  */
-typedef PathToVariable = {
-	object:Dynamic,
-	variableName:String
+private class Interp extends hscript.Interp
```
MSGhero: FlxInterp? lol

MSGhero:
I'm not 100% sure if the Watch tab still works all the time, but the `watch` console command does work.

I'm also not sure why the `Event.RENDER` bit was in there, maybe a workaround for an openfl thing.

Gama11:
The ability to run arbitrary Haxe code from the console is pretty amazing. :)

Also:
>202 additions and 809 deletions

:+1: 

Not too mention, seems like it's not as crashy on neko anymore.

MSGhero:
In the original issue, you said that it'd be cool if we could `#if hscript` to display an error message. Is that possible in retrospect, or is hscript a dependency for this lib now?

Gama11:
It's not possible unless we remove the `<haxelib name="hscript" />` from the `include.xml`, so the user would have to add that manually to his `Project.xml`, since there's no way to tell openfl to "include this lib if it's there". If a `haxelib` isn't there, it errors.

Gama11:
We could try using `Compiler.addClassPath()` and check the output of `haxelib path hscript` in the initialization macro. Unfortunately there's no `Compiler.addHaxelib()` (https://github.com/HaxeFoundation/haxe/issues/4678).
Not sure if that would confuse OpenFL in some way.

PR #1638, opened 2015-10-20
Null checks for FlxSpriteGroup width/height measurements

larsiusprime:
This was causing some weird bugs in some edge cases, this prevents rare null members that somehow make their way into a FlxSpriteGroup from interfering with an accurate width/height bounds calculation.


larsiusprime:
This okay to merge in? Anyone have any objections?

CC @Gama11 @Beeblerox @gamedevsam

gamedevsam:
Good change, :+1:

PR #1641, opened 2015-10-23
Null check for FlxText.set_borderColor

larsiusprime:


larsiusprime:
This fixes a crash on neko (Tooltips weren't working).

PR #1648, opened 2015-10-29
Define dialogs

larsiusprime:
(this makes FlxSaveDialogs muuuuuch cleaner & easier to read)

### DIFF
--- a/flixel/system/macros/FlxDefines.hx
+++ b/flixel/system/macros/FlxDefines.hx
@@ -48,6 +48,9 @@ private enum HelperDefines
 	FLX_JOYSTICK_API;
 	FLX_GAMEINPUT_API;
 	
+	FLX_LINC_DIALOGS;
+	FLX_SYSTOOLS_DIALOGS;
+	
 	/**
 	 * Renders all cameras into a single BitmapData using draw() with
 	 * smoothing. Necessary to achieve non-blurry rendering with bitfive.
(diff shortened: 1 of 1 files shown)


larsiusprime:
See: https://github.com/HaxeFlixel/flixel-addons/pull/207

MSGhero:
Didn't openfl or lime make its own system dialog features? Maybe it was planned for the future.

larsiusprime:
So they did... hmmm.

https://github.com/openfl/lime/blob/b6d4093e6b7c4c1bfaa63af53faf4cc3dd862950/lime/ui/FileDialog.hx

https://github.com/openfl/openfl/blob/28984d5e4145f559c46aea2daa0ec9c9fa009508/openfl/net/FileReference.hx

Perhaps we should just use that :D

MSGhero:
There are issues with systools that go into the c source, and it hasn't been updated in forever last I checked.

MSGhero:
I need a breather from implementing big changes to flixel. I'm still in tears about the hscript Travis thing.

larsiusprime:
hah, sorry about that. 

The previous implementation of FlxScreenGrab still used systools; if we move to openfl's native dialogs we can possibly remove all of these dependencies in one go, linc_dialogs and systools.

larsiusprime:
Closing this as I think it's better to use OpenFL's own dialog system as that makes this totally redundant.

Issue #1558, opened 2015-07-08
Add FlxObject#alignTo()


Gama11:
See discussion here: https://github.com/HaxeFlixel/flixel/issues/1540

Gama11:
Maybe we should just leave this to flixel-ui, the `Anchor` API seems flexible as well as pretty isolated (can be applied to any `FlxObject`, not just flixel-ui-specific ones).

PR #1649, opened 2015-10-30
Added `FlxObject#alignTo()` and `FlxAlign`

adrianulima:
Closes https://github.com/HaxeFlixel/flixel/issues/1558 ?

*curiosity: Why we use `#` instead of `.` here in github?*


gamedevsam:
This looks good to me, any opposition?

Gama11:
This seems a bit inflexible [compared to the anchoring in flixel-ui](https://github.com/HaxeFlixel/flixel-ui/blob/dev/flixel/addons/ui/Anchor.hx).

Not sure what a good solution would look like yet...

@adrianulima I personally use `#` over `.` for instance fields - `FlxObject.alignTo()` would imply "static field" to me, since that's what it would mean in regular Haxe syntax.

adrianulima:
What if we add an offset param?
`alignTo(target:FlxObject, ?anchor:FlxAlign, ?offsetX:Float, ?offsetY:Float)`
my as3 version has it and is very usefull to me.

and thanks

flixel/FlxObject.hx
```diff
@@ -946,6 +947,47 @@ class FlxObject extends FlxBasic
...
 	/**
+	 * Aligns this FlxObject to a target by an anchor point.
+	 * 
+	 * @param	target	To target object to be aligned with.
+	 * @param	anchor	On what anchor point to align the object - default is CENTER.
+	 * @return 	This FlxObject for chaining
+	 */
+	public function alignTo(target:FlxObject, ?anchor:FlxAlign):FlxObject
+	{
+		if (anchor == null)
```
gamedevsam: NOTE: I would prefer not to take a default parameter (otherwise the function should be called CenterOn), and I agree we this function should take in an optional offset.

I don't want to add this to the FlxObject class, seems like we could easily encapsulate all behavior inside the new FlxAlign class (via [static extension](http://haxe.org/manual/lf-static-extension.html)). That way we avoid increasing the memory footprint of every FlxObject, and this logic becomes totally incapsulated in its own FlxAlign class.

Using a [static extension](http://haxe.org/manual/lf-static-extension.html), you would still be able to do the following without bloating FlxObject:
```
import flixel.util.FlxAlign;
var flxObj = new FlxObject();
var target = new FlxObject();
flxObj.alignTo(target, anchor, offset);
```

adrianulima: Nice, you're right! I will try to do this today.

Gama11: Like I mentioned, I don't really like the API as is, moving it to a static extension wouldn't that. flixel-ui's approach seems much more flexible.

adrianulima:
Ok, I'm not familiar to Flixel-UI API yet, and I think it is too different of my approach here. Maybe we should close this PR and start another, and ask @larsiusprime help with the new one.

Issue #1422, opened 2015-01-18
FlxText blurry rendering when align="center" and wordWrap is applied


Ohmnivore:
![](http://snag.gy/Vhyay.jpg)

Here's my code:
```
var tOffset:Float = 8;
text = new FlxText(back.x + tOffset, tOffset, back.width - 2 * tOffset,
getPhrase());
text.color = 0xffffffff;
text.alignment = "center";
add(text);
```
The blurring only seems to happen to the last line, btw.

larsiusprime:
This bug has been around pretty much forever, right? I remember it in the old days of Vanilla AS3 flixel.

Ohmnivore:
Oh well then I guess it's here to stay... I'm gonna try and write a class that will break up a multi-line string into single FlxTexts, so that there's one FlxText instance per line. I hoped it wouldn't come to this but ok.

larsiusprime:
Sorry, I didn't mean to be the final authority on this or anything, let's see if @gama11 or anyone else can confirm :P

Gama11:
Yeah, I don't think there's a simple fix for this... This issue seems related btw: #778

goshki:
Try putting extra spaces around "SUBJECTS (AGAIN)" so it looks like " SUBJECTS (AGAIN) ".

larsiusprime:
Did we actually fix this????

Gama11:
To some degree.. there are still some cases where the workaround doesn't work I think.

adrianulima:
My trick only works when the `space char` in the used font have an odd `width`, the blurry problem happens when the line `width` is odd, so I add an extra space to make it even.

Geokureli:
This issue is back, and CI is failing on openfl 9.5.1
https://github.com/openfl/openfl/issues/2839

PR #1650, opened 2015-10-30
FlxText - Prevent blurry lines

adrianulima:
Added a method in `FlxText` that insert an extra space to a line that is blurry. This is not a beautful solution, but it works.

Not work if:
If the font have a space character that width is not odd like the default font. (3 pixels)
If `wordWrap` is enabled and is applying a linebreak, because it seems to trim lines.

This solves https://github.com/HaxeFlixel/flixel/issues/1422

### DIFF
--- a/flixel/text/FlxText.hx
+++ b/flixel/text/FlxText.hx
@@ -840,11 +840,41 @@ class FlxText extends FlxSprite
 			applyBorderTransparency();
 			applyFormats(_formatAdjusted, false);
 			
+			avoidBlurry();
+			
 			graphic.bitmap.draw(textField, _matrix);
 		}
 		
 		_regen = false;
 		dirty = true;
+	}
+	
+	/**
+	 * Internal function that insert an extra space in a line to prevent blurry lines.
+	 * Only work if wordWrap is not working, beacuse wordWrap seems to trim lines.
+	 */
+	private function avoidBlurry() 
+	{
+		var newText = "";
+		
+		for (i in 0...textField.numLines) 
+		{
+			var lineX = textField.getLineMetrics(i).x;
+			if (Math.floor(lineX) != lineX)
+			{
+				var space = textField.getLineText(i).indexOf(" ");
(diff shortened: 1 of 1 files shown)


flixel/text/FlxText.hx
```diff
@@ -840,6 +840,8 @@ class FlxText extends FlxSprite
 			applyBorderTransparency();
 			applyFormats(_formatAdjusted, false);
 			
+			avoidBlurry();
```
gamedevsam: Check if word wrap is enabled, and skip this logic if it is.

gamedevsam:
This seems like a hacky fix that has the potential to reduce performance on games with many FlxText objects. I would prefer if @jgranick took a look at it to see if we could find a proper fix in OpenFL land.

Gama11:
I think this is actually a flash bug (and flash only).

SeiferTim:
@Gama11 not really? I have a project where I was using really small fonts and on all targets it was equally blurry, to the point where I switched over to bitmap fonts.

larsiusprime:
@SeiferTim Were you using -Dnext or not?

adrianulima:
I agree, it is ugly and hacky, maybe it could be a optional method or a method in some Util class?

On neko it draws correct, but on flash it positions the line in 0.5.. I think worth talk to J. Granick and find the difference between targets.

SeiferTim:
@larsiusprime yes.

larsiusprime:
That's strange, I worked on the text layout code for the -Dnext version of OpenFL, and I was pretty sure we had this issue solved. I can take another look at it I guess.

SeiferTim:
@larsiusprime https://haxeflixel.slack.com/archives/help/p1445530593000185

flixel/text/FlxText.hx
adrianulima: @gamedevsam I didn't checked because in some cases `wordWrap` is enabled but is not affecting the text. And in this cases this method can avoid the blurry lines.

Issue #1655, opened 2015-11-09
'Type not found: JoystickEvent' in neovim with vaxe.


danbruegge:
Hi, i use neovim with the vaxe plugin.

I startet with a HelloWorld template.
```
flixel tpl -n "HelloWorld"
```
If i try to use the the autocompletion i got this error:
![vaxe-error-1](https://cloud.githubusercontent.com/assets/1296438/11041027/a3b7400a-870f-11e5-9db4-f4d67d5ef809.png)

A simple build with neko works. So no erros on this part. I only got this on editor completion. 
I opened an issue here, because i think it is more related to haxeflixel rather than vaxe.

Regards, Dan

sruloart:
http://www.openfl.org/blog/2015/10/06/october-update/

danbruegge:
Ok, thanks. If i install the 'dev' version of flixel, the error message is gone. :)

Tiago-Ling:
This also happens to me with the development version of flixel (using Lime
2.6.8 and OpenFL 3.3.8) in FlashDevelop 5.0.2

Tiago Ling Alexandre
Tel: +55 41 8819-3191


2015-11-09 16:12 GMT-02:00 Daniel Brüggemann <notifications@github.com>:

danbruegge:
@Tiago-Ling try install openfl version 3.3.6.

I've now installed:
```
$ haxelib list
flixel-addons: [1.1.0]
flixel-demos: [1.1.2]
flixel-templates: [1.0.2]
flixel-tools: [1.1.1]
flixel-ui: [1.0.2]
flixel: [3.3.11]
lime: [2.7.0]
openfl: [3.3.6]
```

danbruegge:
Maybe, flixel needs a fixed version of openfl when installing via haxelib?

sruloart:
I think this simply requires a small update to: https://github.com/HaxeFlixel/flixel/blob/00d2b1d68b88abc7d114c929763b6e5dae156670/flixel/input/gamepad/FlxGamepadManager.hx with a proper #if to cover older versions of OpenFL that support `JoystickEvent`. 

I'm not too sure if that's worth another release (3.3.12) though.

PR #1656, opened 2015-11-10
Fix completion problems when using OpenFL next

Tiago-Ling:
When using HaxeFlixel + OpenFL this completion error keeps popping up in FlashDevelop (and apparently other IDEs too):
```
flixel/flixel/input/gamepad/FlxGamepadManager.hx:486: lines 486-494 : Type not found : JoystickEvent
```

This simple check fixes it and probably solves #1655

### DIFF
--- a/flixel/input/gamepad/FlxGamepadManager.hx
+++ b/flixel/input/gamepad/FlxGamepadManager.hx
@@ -469,7 +469,7 @@ class FlxGamepadManager implements IFlxInputManager
 	}
 	#end
 	
-	#if FLX_JOYSTICK_API
+	#if (FLX_JOYSTICK_API && openfl_legacy)
 	private function getModelFromJoystick(f:Float):FlxGamepadModel
 	{
 		//id "1" is PS3, but that is not supported as its PC drivers are terrible, and the most popular tools just turn it into a 360 controller


Gama11:
This is the wrong place for the fix, try adjusting the definition of `FLX_JOYSTICK_API` here:

https://github.com/HaxeFlixel/flixel/blob/dev/flixel/system/macros/FlxDefines.hx#L140

Although `!defined("openfl_next")` should already imply `openfl_legacy`. Are you on a recent enough version to have this commit? https://github.com/HaxeFlixel/flixel/commit/14b40430068f57bba95aa66448cd8693e0e034d3

Tiago-Ling:
Ok, i'm on an older version - updating solves it.

I should have checked it before, closing it.

sruloart:
AFAI understand, most of openfl-legacy releases should still support `JoystickEvent`, so what we really need is an `#if` with a version number.

Tiago-Ling:
I think this should be directly tied to `openfl_legacy` flag only since on next the `JoystickEvent.hx` class is just this:

```haxe
package openfl.events;


#if openfl_legacy
typedef JoystickEvent = openfl._legacy.events.JoystickEvent;
#end
```

r3d9u11:
Just replace conditions "#if (cpp || neko)" with #if (openfl_legacy && (cpp || neko)) at source haxelib/flixel/3,3,11/flixel/input/gamepad/FlxGamepadManager.hx

it work's for my vim autocompletion

PR #1658, opened 2015-11-13
Disable controls if FLX_NO_SOUND_TRAY

Tiago-Ling:
I just noticed that even if you disable the sound tray by using the `FLX_NO_SOUND_TRAY` flag the controls will still work. This confused some users which thought it was a bug in the game.

Imho the controls should only work when you're using the sound tray.

### DIFF
--- a/flixel/system/frontEnds/SoundFrontEnd.hx
+++ b/flixel/system/frontEnds/SoundFrontEnd.hx
@@ -29,7 +29,7 @@ class SoundFrontEnd
 	 */
 	public var volumeHandler:Float->Void;
 	
-	#if !FLX_NO_KEYBOARD
+	#if (!FLX_NO_KEYBOARD && !FLX_NO_SOUND_TRAY)
 	/**
 	 * The key codes used to increase volume (see FlxG.keys for the keys available).
 	 * Default keys: + (and numpad +). Set to null to deactivate.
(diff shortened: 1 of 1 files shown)


Gama11:
This is actually intentional. You should be able to have volume controls without a visual representation for it (the sound tray).

Tiago-Ling:
Really? I don't think i ever saw a use case for this: controlling audio without the visual feedback provided by the tray.

I'd propose more thinking on this, it's actually really strange when a user comes complaining that their game is suddenly without audio.

I for one will always have to comment this to prevent these "ghost controls" from working.

Gama11:
A very simple use case would creating a custom sound tray.

Tiago-Ling:
A custom sound tray would have hardcoded controls?

Gama11:
What do you mean by hardcoded?

Tiago-Ling:
My use case is: i don't want to use the sound tray, so i just go and disable it. Then a user comes and without knowing presses `0` and toggles mute. He/she perceives it as a bug, since the game simply stops playing music and sound effects.

I don't see much sense in being able to disable the tray but still keeping the controls. Also, i don't get how it would be a good idea to divide the functionality of a sound tray somewhere in the user code with its controls inside flixel.

Well, i think i won't be able to convince you otherwise, i just think the way this is set up is bad design. There should be at least the option to disable the controls the same you can disable the tray (e.g. `FLX_NO_SOUND_CONTROLS` or something). As it is, to completely disable this feature you have to use the flag and then go and change the `FlxG.sound.volumeUpKeys` and the other ones to empty arrays (i suppose that's the intended way for this to work).

Anyway, i'm closing it.

PR #1659, opened 2015-11-16
Tilemap DiagonalPolicy

larsiusprime:
I was debugging some pathfinding issues and noticed that FlxTilemap seems to always assume that diagonal movement is possible when it's calculating the distances during the A* / Dijkstra's algorithm step.

There was a parameter to modify this, but only to specify whether the diagonal was "wide" or not, ie, whether it would take an additional point to take the diagonal, or whether it was the same as an orthogonal step. There was no way I could see to disallow diagonals entirely.

So, this change swaps out the `WideDiagonal:Bool` for `DiagonalPolicy:FlxTilemapDiagonalPolicy`, which is an abstract enum.

Why an abstract enum and not a regular enum? Because the previous behavior is to set a default value of Wide Diagonals, and [you can't assign an enum as a default parameter because it's technically not a constant and the compiler complains](http://stackoverflow.com/questions/31307992/haxe-enum-default-parameters).

FlxTilemapDiagonalPolicy has three values -- NONE, NORMAL, and WIDE.

WIDE has the same behavior as previously setting the `WideDiagonal` value to true, and NORMAL has the same behavior as previously setting the `WideDiagonal` value to false.

NONE will make the calculator skip the `up && right`, `down && left`, etc, steps, and only consider orthogonal movement.

### DIFF
--- a/flixel/system/macros/FlxDefines.hx
+++ b/flixel/system/macros/FlxDefines.hx
@@ -48,9 +48,6 @@ private enum HelperDefines
 	FLX_JOYSTICK_API;
 	FLX_GAMEINPUT_API;
 	
-	FLX_LINC_DIALOGS;
-	FLX_SYSTOOLS_DIALOGS;
-	
 	/**
 	 * Renders all cameras into a single BitmapData using draw() with
 	 * smoothing. Necessary to achieve non-blurry rendering with bitfive.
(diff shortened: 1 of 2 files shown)


larsiusprime:
@MSGhero wisely points out that using Abstract Null<Bool> as the underlying type avoids a breaking change.

PR #1668, opened 2015-12-03
Render path set at runtime

larsiusprime:
This changes the FLX_RENDER_BLIT and FLX_RENDER_TILE into runtime flags (FlxG.renderTile and FlxG.renderBlit).

I think this change is necessary because there are situations in which you can't know ahead of time whether the user's system will need software rendering (for which BLIT is optimized) or hardware rendering (for which TILE is optimized).

I've identified two such scenarios already:

**1. Windows  -- ANGLE on Vista/7/8/10, SOFTWARE on XP**

In a recent bit of refactoring for OpenFL, we were able to get ANGLE up and running, and it's now going to be the default rendering system for Windows, this increases compatibility across the board and virtually eliminates issues with bad OpenGL drivers. However, ANGLE does now work with Windows XP. Doing some testing, the *only* way I could get reliable, works close to 100% of the time results with XP was to force software rendering -- this hasn't been pushed through quite yet, but I believe Joshua agrees with me. However, the system cannot know at compile time which version of windows the user will use, so it has to be able to switch between hardware and software at runtime.

**2. HTML5 -- WebGL with CANVAS fallback**

This one's more important IMHO. If we want a really good web compatibility experience, we need to target WebGL with an automatic Canvas fallback. For best results this would require us to detect WebGL compatibility at runtime and then pick the right render method then.

This is a big refactor so it deserves a lot of scrutiny, I've been using it myself and it seems stable but I definitely need input and feedback on this.

Hope this is useful.

### DIFF
--- a/flixel/FlxCamera.hx
+++ b/flixel/FlxCamera.hx
@@ -120,30 +120,30 @@ class FlxCamera extends FlxBasic
 	 */
 	public var scroll:FlxPoint;
 	
-	#if FLX_RENDER_BLIT
+	//start renderBlit
+	
 	/**
-	 * The actual bitmap data of the camera display itself.
-	 */
+	* The actual bitmap data of the camera display itself.
+	*/
 	public var buffer:BitmapData;
 	/**
-	 * Whether checkResize checks if the camera dimensions have changed to update the buffer dimensions.
-	 */
+	* Whether checkResize checks if the camera dimensions have changed to update the buffer dimensions.
+	*/
 	public var regen:Bool = false;
-	#end
 	
 	/**
-	 * The natural background color of the camera, in AARRGGBB format. Defaults to FlxG.cameras.bgColor.
-	 * On flash, transparent backgrounds can be used in conjunction with useBgAlphaBlending.
-	 */ 
+	* The natural background color of the camera, in AARRGGBB format. Defaults to FlxG.cameras.bgColor.
+	* On flash, transparent backgrounds can be used in conjunction with useBgAlphaBlending.
+	*/ 
 	public var bgColor:FlxColor;
 	
-	#if FLX_RENDER_BLIT
 	/**
-	 * Sometimes it's easier to just work with a FlxSprite than it is to work directly with the BitmapData buffer.  
-	 * This sprite reference will allow you to do exactly that.
-	 */
+	* Sometimes it's easier to just work with a FlxSprite than it is to work directly with the BitmapData buffer.  
+	* This sprite reference will allow you to do exactly that.
+	*/
 	public var screen:FlxSprite;
-	#end
+	
+	//end renderBlit
 	
 	/**
 	 * Whether to use alpha blending for camera's background fill or not. 
@@ -160,10 +160,10 @@ class FlxCamera extends FlxBasic
 	/**
 	 * Whether the positions of the objects rendered on this camera are rounded.
 	 * If set on individual objects, they ignore the global camera setting.
-	 * Defaults to false with FLX_RENDER_TILE and to true with FLX_RENDER_BLIT.
+	 * Defaults to false with FlxG.renderTile and to true with FlxG.renderBlit.
 	 * WARNING: setting this to false on blitting targets is very expensive.
 	 */
-	public var pixelPerfectRender:Bool = #if FLX_RENDER_TILE false #else true #end;
+	public var pixelPerfectRender:Bool;
 	
 	/**
 	 * How wide the camera display is, in game pixels.
@@ -296,16 +296,18 @@ class FlxCamera extends FlxBasic
 	 */
 	public var initialZoom(default, null):Float = 1;
 	
-	#if FLX_RENDER_BLIT
-	/**
-	 * Internal helper variable for doing better wipes/fills between renders.
-	 */
-	private var _fill:BitmapData;
-	/**
-	 * Internal, used to render buffer to screen space.
-	 */
-	private var _flashBitmap:Bitmap;
-	#end
+	//start FlxG.renderBlit
+		
+		/**
+		 * Internal helper variable for doing better wipes/fills between renders.
+		 */
+		private var _fill:BitmapData;
+		/**
+		 * Internal, used to render buffer to screen space.
+		 */
+		private var _flashBitmap:Bitmap;
+		
+	//end FlxG.renderBlit
 	
 	/**
 	 * Internal sprite, used for correct trimming of camera viewport.
@@ -317,7 +319,8 @@ class FlxCamera extends FlxBasic
 	 */
 	private var _bounds:FlxRect = FlxRect.get();
 	
-#if FLX_RENDER_TILE
+//Start FlxG.renderTile
+	
 	/**
 	 * Sprite for drawing (instead of _flashBitmap for blitting)
 	 */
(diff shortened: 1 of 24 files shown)


MSGhero:
Travis is complaining about `ByteArrayData` which I know you were messing with. CPP is fine, though.

larsiusprime:
Thanks, I'll fix that.

flixel/FlxCamera.hx
```diff
@@ -120,30 +120,30 @@ class FlxCamera extends FlxBasic
...
+	//start renderBlit
 	
-	/**
-	 * The natural background color of the camera, in AARRGGBB format. Defaults to FlxG.cameras.bgColor.
-	 * On flash, transparent backgrounds can be used in conjunction with useBgAlphaBlending.
-	 */ 
-	public var bgColor:FlxColor;
+		/**
+		* The actual bitmap data of the camera display itself.
+		*/
```
Gama11: Not a fan of the increased indent here.

flixel/FlxCamera.hx
```diff
@@ -120,30 +120,30 @@ class FlxCamera extends FlxBasic
...
-	/**
-	 * The actual bitmap data of the camera display itself.
-	 */
-	public var buffer:BitmapData;
-	/**
-	 * Whether checkResize checks if the camera dimensions have changed to update the buffer dimensions.
-	 */
-	public var regen:Bool = false;
-	#end
+	//start renderBlit
```
Gama11: If you're gonna add a comment for these "regions" (which I'm not sure is a good idea as I doubt it'll be maintained very well), you might as well take advantage of [FD's regions](http://stackoverflow.com/a/2617026/2631715) for code folding etc.

flixel/FlxCamera.hx
```diff
@@ -653,6 +665,8 @@ class FlxCamera extends FlxBasic
 	{
 		super();
 		
+		pixelPerfectRender = (FlxG.renderBlit);
```
Gama11: Weird brackets here.

flixel/FlxCamera.hx
```diff
@@ -526,119 +529,128 @@ class FlxCamera extends FlxBasic
...
+			}
 		}
-		else if (frame != null)
+		else
 		{
-			frame.paint(buffer, destPoint, true);
+			_helperMatrix.identity();
+			_helperMatrix.translate(destPoint.x + frame.offset.x, destPoint.y + frame.offset.y);
+			var isColored:Bool = (cr != 1.0) || (cg != 1.0) || (cb != 1.0) || (ca != 1.0);
+			#if !FLX_RENDER_TRIANGLE
```
Gama11: Seems a little strange to still have `FLX_RENDER_TRIANGLE` as a conditional when the rest isn't anymore.. Wondering what @Beeblerox thinks about that.

flixel/FlxG.hx
```diff
@@ -500,6 +505,43 @@ class FlxG
...
+		#else
+			#if (flash || js)
+				renderMethod = BLIT;
+			#else
+				renderMethod = TILES;
+			#end
+		#end
+		
+		#if FLX_RENDER_BLIT
+			//renderMethod = BLIT;
```
Gama11: ?

flixel/FlxG.hx
```diff
@@ -113,6 +113,11 @@ class FlxG
 	 */
 	public static var drawFramerate(default, set):Int;
 	
+	public static var renderMethod(default, null):FlxRenderMethod;
```
Gama11: What's the use case for `renderMethod` if we have two booleans anyway?

flixel/FlxG.hx
```diff
@@ -500,6 +505,43 @@ class FlxG
...
+			//renderMethod = BLIT;
+		#elseif FLX_RENDER_TILE
+			//renderMethod = TILES;
+		#end
+		
+		renderBlit = renderMethod == BLIT;
+		renderTile = renderMethod == TILES;
+		
+		#if sys
+		Sys.println("renderMethod = " + renderMethod);
```
Gama11: Leftover debug trace?

flixel/system/FlxAssets.hx
```diff
@@ -28,7 +28,7 @@ class GraphicLogo extends BitmapData {}
 class GraphicVirtualInput extends BitmapData {}
 
 @:file("assets/images/ui/virtual-input.txt")
-class VirtualInputData extends ByteArray {}
+class VirtualInputData extends #if lime_legacy ByteArray #else ByteArrayData #end {}
```
Gama11: This breaks compilation with older OpenFL versions, so the only way to have this in here is to add a check for the OpenFL / Lime version (which is somewhat tricky with this change not being released yet).

Also seems strange that OpenFL would need to break compatibility here.

flixel/system/debug/Stats.hx
```diff
@@ -37,7 +37,7 @@ class Stats extends Window
 	/**
 	 * The minimal height of the window.
 	 */
-	private static inline var MIN_HEIGHT:Int = #if !FLX_RENDER_TILE 185 #else 200 #end;
+	private static var MIN_HEIGHT(default, null):Int = 0;
```
Gama11: Not much of a point to a private `default, null` property.

flixel/tile/FlxTilemapBuffer.hx
```diff
@@ -93,23 +92,32 @@ class FlxTilemapBuffer implements IFlxDestroyable
...
 	public function fill(Color:FlxColor = FlxColor.TRANSPARENT):Void
 	{
-		pixels.fillRect(_flashRect, Color);
+		if (FlxG.renderBlit)
+		{
+			pixels.fillRect(_flashRect, Color);
+		}
+		else
+		{
+			FlxG.log.error("Function only available if FlxG.renderBlit=BLIT, (you are using " + FlxG.renderMethod +")");
```
Gama11: Not sure we need that error here, `FlxTilemapBuffer` is very much internal.

flixel/util/FlxGradient.hx
```diff
@@ -1,14 +1,14 @@
 package flixel.util;
 
-import flash.display.Bitmap;
```
Gama11: Any particular reason for changing the imports to `openfl.*` here`

flixel/FlxCamera.hx
Gama11: Also, there's no reason to move this to the constructor. Instance variable initializations already happen there.

flixel/FlxObject.hx
```diff
@@ -572,6 +572,10 @@ class FlxObject extends FlxBasic
 		flixelType = OBJECT;
 		last = FlxPoint.get(x, y);
 		scrollFactor = FlxPoint.get(1, 1);
+		if (FlxObject.defaultPixelPerfectPosition == null)
```
Gama11: Initializing this static var from an instance method seems hacky. We could just do that when we set the render flags within `FlxG`. No need to turn it into a `Null<Bool>` that way either.

flixel/FlxCamera.hx
larsiusprime: Those aren't intended to be final. Those are just placeholders to remind myself where the original sections were in case we need to add metadata or whatever to hide some of these from completion.

flixel/FlxCamera.hx
larsiusprime: I decided to refactor one thing at a time. FLX_RENDER_BLIT and FLX_RENDER_TILE moving to FlxG.renderBlit & FlxG.renderTile is a big enough change already, I didn't want to break existing functionality and figured any refactoring for render triangle could wait for a future PR.

flixel/system/FlxAssets.hx
larsiusprime: This is old cruft from managing too many branches. I'll revert it.

flixel/FlxCamera.hx
larsiusprime: I don't understand, should I just not initialize it? 

I did a search and FlxCamera.pixelPerfectRender is not set anywhere that I can find.

In the previous code pixelPerfectRender is initialized based on the compiler constant, but if we replace those with runtime constants, in order to maintain the same logic I have to move this to the constructor. Otherwise, we'll get different behavior as it's not set anywhere else.

Am I missing something?

flixel/FlxG.hx
larsiusprime: Previously the compiler constants would be set by looking at your targets: BLIT for Flash or JS, TILES for cpp, and all the logic seemed to be assuming you were using legacy.

This gives it a better way to guess at initial settings. In next mode, OpenGL and Console rendering should use tiles by default, and HTML5 with canvas, Flash, or CPP with cairo, should use blitting. In legacy mode, it reproduces the same logic as before.

Or maybe you're just asking about the last line? This block?
```
#if FLX_RENDER_BLIT
	//renderMethod = BLIT;
#elseif FLX_RENDER_TILE
	//renderMethod = TILES;
#end
```

In that case that's some old cruft, the first thing I did was just set runtime flags triggered by the compile constant as an early step in the refactor... let me delete that.

flixel/FlxG.hx
larsiusprime: We might add more render methods in the future (e.g. DRAW_TRIANGLES), and it's possible some of those might act as subsets of one of the other two. This seems to be how DRAW_TRIANGLES was set up under the old constants -- *both* DRAW_TILES and DRAW_TRIANGLES compiler constants would be set in the draw_triangles render path, and draw_triangles shared a lot of logic with draw_tiles, just with a few additional logic blocks triggered by it's own constant.

flixel/FlxG.hx
larsiusprime: Yup.

flixel/util/FlxGradient.hx
larsiusprime: I was getting compiler errors. Might have just been because I was using some WIP commits from OpenFL at the time.

flixel/system/FlxAssets.hx
larsiusprime: Okay, so this issue is caused by the recent ByteArray refactor in OpenFL. I do believe it's a breaking change -- but apparently there was a good reason for it.

In any case, ByteArray is now an abstract (over ByteArrayData) and so you can't extend it directly. I'm asking Joshua if there's a better way to handle this, but for now this is all I can think of. What version should we check against, or should we just wait until the next OpenFL release before making this particular change?

larsiusprime: See this PR to OpenFL that fixes a few things. Also if I bump the internal version number I can do a proper check to maintain compatibility.

https://github.com/openfl/openfl/pull/952

larsiusprime:
Note that the version check in the FlxAssets embed line won't work without this PR to OpenFL:
https://github.com/openfl/openfl/pull/952

larsiusprime:
Anything else needed here?

@gamedevsam @Beeblerox @Gama11

larsiusprime:
@Beeblerox -- what do you think of this pull? It should greatly improve our ability to deal with HTML5 properly (I'd like to see a true, well-supported HTML5 target that handles both WebGL and CANVAS fallback, for 2016), as well as give us more flexibility overall.

Beeblerox:
i think it's good enough to be merged in. but as @Gama11 mentioned we'll need to update addons classes aswell.

larsiusprime:
Great, I'll work on that next then?

Gama11:
`FlxDefines` will also need to be updated.

larsiusprime:
@Gama11  Sure, what needs to be done there? Just remove the defines?

Gama11:
Yeah.

larsiusprime:
Alrighty, in doing that I had to move FLX_RENDER_CRISP to FlxG.renderCrisp, I believe I've maintained the same logic, but let me know if I need to fix anything.

larsiusprime:
flixel-addons PR is here:
https://github.com/HaxeFlixel/flixel-addons/pull/215

Gama11:
Hm.... I don't like having `FlxG.renderCrisp`.. it's not really public API, just some internal hack..

larsiusprime:
Want to make it private? Do we still use it?

Gama11:
Why is it necessary to remove that one from FlxDefines? It's not like you can switch from bitfive to a backend target at runtime.

larsiusprime:
@Gama11 for some reason I thought it was looking at FLX_RENDER_BLIT/TILE ... seems you're right, let's revert it.

larsiusprime:
Wait no, it does look at FLX_RENDER_BLIT.

Do we want to just leave it as it was, but add an extra check for FlxG.renderBlit in every #if CRISP check?

Gama11:
Add an extra check.

larsiusprime:
@Gama11 Okay, I think I properly reverted it.

As far as I can tell, there aren't any else checks on the #if FLX_RENDER_CRISP, so it's safe to stack a second runtime if (FlxG.renderBlit) underneath each one, and just making sure to match whether there's a `!` on the `#if` or not.

flixel/FlxG.hx
```diff
@@ -541,6 +548,44 @@ class FlxG
 		#end
 	}
 	
+	private static function initRenderMethod():Void
+	{
+		renderMethod = BLIT;
+		
+		#if FLX_RENDER_TILE
```
Gama11: These defines are gone, why still check for them?

flixel/FlxObject.hx
```diff
@@ -22,7 +22,7 @@ class FlxObject extends FlxBasic
 	/**
 	 * Default value for FlxObject's pixelPerfectPosition var.
 	 */
-	public static var defaultPixelPerfectPosition:Bool = #if FLX_RENDER_BLIT true #else false #end;
+	public static var defaultPixelPerfectPosition:Null<Bool> = null;
```
Gama11: This shouldn't be a `Null<Bool>`.

flixel/FlxCamera.hx
```diff
@@ -526,119 +529,128 @@ class FlxCamera extends FlxBasic
...
-		}	
+			_bounds.set(0, 0, width, height);
+			var isColored:Bool = (colors != null && colors.length != 0);
+			var drawItem:FlxDrawTrianglesItem = startTrianglesBatch(graphic, smoothing, isColored, blend);
+			drawItem.addTriangles(vertices, indices, uvtData, colors, position, _bounds);
+		}
 	}
-#end
+
+//else if FlxG.renderBlit
```
Gama11: Did you wanna get rid of these comments? IIRC you said they were temporary?

flixel/FlxG.hx
larsiusprime: Cruft from refactoring, removing them now.

flixel/FlxCamera.hx
larsiusprime: Sure, but do you think I should add some comments or documentation specifying these are only for FlxG.renderBlit?

Gama11: I'm not sure.. These sections are so spread out that I doubt that would be maintained...

Gama11:
Could you [squash the commits](http://eli.thegreenplace.net/2014/02/19/squashing-github-pull-requests-into-a-single-commit)? Not usually one to care too much about commit history, but 22 commits is a little much. :)

larsiusprime:
Okay, there were a lot of merge conflicts, so hope everything's still okay.

flixel/FlxCamera.hx
larsiusprime: I'll just remove the comments then.

larsiusprime:
Travis is failing because of some minor glitches in flixel-addons, here's an isolated PR that patches those, if we merge it in then we should be able to restart the jobs here and see them pass.

https://github.com/HaxeFlixel/flixel-addons/pull/216

larsiusprime:
I went ahead and merged my fix for the flixel-addons demos since it was a very small change and I tested it locally as well (and travis was green).

Now I've restarted the Travis build here -- if I did everything right, this one should become green, too.

larsiusprime:
The travis build is failing because of FlxBackdrop, which needs changes. I've fixed and tested those locally, and they're now in this PR:

https://github.com/HaxeFlixel/flixel-addons/pull/215

PR #1670, opened 2015-12-04
Add "flipX / flipY" property to animations

larsiusprime:
...and also make it interact properly with FlxSprite's own flipX / flipY.

Rationale: for games with big sprite sheets, it can save a lot of memory to re-use frames in animations by vertically flipping them. So if I can re-use a walk cycle and just tell it to flip the frames right in the animation call, that's extremely convenient. Much easier to do than messing around with individual frames in the atlas.

Before this pull, we have had the ability to apply flipX/flipY to individual flxFrames (such as those loaded from an atlas) -- so this means the user can now set flipX/flipY in three places:

1) On the sprite itself
2) On an individual frame (e.g. tagging the metadata as part of atlas loading)
3) On an individual animation

The effect should cascade properly. So you can have a particular frame in an animation that has flipX=true, used in an animation that has flipX=true, and then the sprite itself has flipX=true, which will resolve to true because (true XOR true XOR true) == true.

### DIFF
--- a/flixel/FlxSprite.hx
+++ b/flixel/FlxSprite.hx
@@ -943,7 +943,7 @@ class FlxSprite extends FlxObject
 			}
 			else
 			{
-				framePixels = _frame.paintRotatedAndFlipped(framePixels, _flashPointZero, FlxFrameAngle.ANGLE_0, checkFlipX(), checkFlipY(), false, true);
+				framePixels = _frame.paintRotatedAndFlipped(framePixels, _flashPointZero, FlxFrameAngle.ANGLE_0, doFlipX, doFlipY, false, true);
 			}
 			
 			if (useColorTransform)
@@ -1361,26 +1361,26 @@ class FlxSprite extends FlxObject
 	
 	private inline function checkFlipX():Bool
 	{
-		var doFlipX:Bool = FlxMath.boolXOR(flipX, _frame.flipX);
-		
 		if (animation.curAnim != null)
 		{
-			doFlipX = FlxMath.boolXOR(flipX, animation.curAnim.flipX);
+			return ((flipX != _frame.flipX) != animation.curAnim.flipX);
 		}
-		
-		return doFlipX;
+		else
+		{
+			return (flipX != _frame.flipX);
+		}
 	}
 	
 	private inline function checkFlipY():Bool
 	{
-		var doFlipY:Bool = FlxMath.boolXOR(flipY, _frame.flipY);
-		
 		if (animation.curAnim != null)
 		{
-			doFlipY = FlxMath.boolXOR(flipY, animation.curAnim.flipY);
+			return ((flipY != _frame.flipY) != animation.curAnim.flipY);
 		}
-		
-		return doFlipY;
+		else
+		{
+			return (flipY != _frame.flipY);
+		}
 	}
 }
 
--- a/flixel/FlxSprite.hx
+++ b/flixel/FlxSprite.hx
@@ -650,7 +650,7 @@ class FlxSprite extends FlxObject
 			}
 			else
 			{
-				_frame.prepareMatrix(_matrix, FlxFrameAngle.ANGLE_0, flipX, flipY);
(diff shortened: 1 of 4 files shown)


larsiusprime:
I tried that first, but it's only a numeric operator, not a boolean. There's no boolean xor.

Gama11:
You could also just use `a != b`. ;)

flixel/FlxSprite.hx
```diff
@@ -934,15 +934,16 @@ class FlxSprite extends FlxObject
...
+			var doFlipY:Bool = checkFlipY();
+			
 			if (!doFlipX && !doFlipY && _frame.type == FlxFrameType.REGULAR)
 			{
 				framePixels = _frame.paint(framePixels, _flashPointZero, false, true);
 			}
 			else
 			{
-				framePixels = _frame.paintRotatedAndFlipped(framePixels, _flashPointZero, FlxFrameAngle.ANGLE_0, flipX, flipY, false, true);
+				framePixels = _frame.paintRotatedAndFlipped(framePixels, _flashPointZero, FlxFrameAngle.ANGLE_0, checkFlipX(), checkFlipY(), false, true);
```
Gama11: Why call the check functions again here when the results are locally cached a few lines earlier in `doFlipX` / `Y`?

flixel/FlxSprite.hx
```diff
@@ -1357,6 +1358,30 @@ class FlxSprite extends FlxObject
 	{
 		return antialiasing = value;
 	}
+	
+	private inline function checkFlipX():Bool
+	{
+		var doFlipX:Bool = FlxMath.boolXOR(flipX, _frame.flipX);
```
Gama11: There's not really much of a reason for this line to be always executed - it's just the else case of `if (animation.curAnim != null)`.

flixel/animation/FlxAnimation.hx
```diff
@@ -49,6 +49,16 @@ class FlxAnimation extends FlxBaseAnimation
 	public var reversed(default, null):Bool = false;
 	
 	/**
+	 * Whether or not the frames of this animation are horizontally flipped
+	 */
+	public var flipX(default, null):Bool = false;
```
Gama11: Considering these are checked during each render, is there a reason to make them read-only? Setting them after creation should work.

larsiusprime:
@gama11 how bout now?

larsiusprime:
Ah, you're right. Well in any case the logic as it stands now is what I actually intend.

Gama11:
Ah. Well, in that case, I'd prefer the original approach, since doing it this way introduces some redundancy. :P

larsiusprime:
To be clear, this is what you mean?

```Haxe
	private inline function checkFlipX():Bool
	{
		var doFlipX = (flipX != _frame.flipX);
		if (animation.curAnim != null)
		{
			return (doFlipX != animation.curAnim.flipX);
		}
		return doFlipX;
	}
```

Gama11:
Yes.

larsiusprime:
Cool. Incoming.

MSGhero:
Oh, right. Yeah this looks a bit nicer.

larsiusprime:
Is this good to go then @Gama11 and @MSGhero or does it need more changes?

MSGhero:
Why are flipX and Y parameters added only to a couple of the flxanimcontroller functions?

larsiusprime:
The use case I have in mind is that you set those when you create a new animation, so I put it in the functions that create new FlxAnimations. Are there some others you think it should be added to? If so, which ones?

MSGhero:
:+1: I guess it doesn't make sense adding it to `append()`.

larsiusprime:
Ready for final review @MSGhero @Gama11

MSGhero:
I can't test it right now, but it looks fine to me.

larsiusprime:
Bump for final review?
@Beeblerox @gamedevsam @Gama11

PR #1671, opened 2015-12-05
Add insert function to FlxGroup class.

Corneguidouille:
Hi, I just added a function for inserting an object into a FlxGroup at a specific position.
It's very useful when the order of members are importants. In this case, using the existing add function is not what we need.

Tested on Neko and Mac targets.


MSGhero:
:+1:

PR #1673, opened 2015-12-07
FlxSave: Fixes for latest openfl

larsiusprime:
Latest OpenFL removes the need for the #if flash stuff

### DIFF
--- a/flixel/util/FlxSave.hx
+++ b/flixel/util/FlxSave.hx
@@ -105,19 +105,35 @@ class FlxSave implements IFlxDestroyable
 			return false;
 		}
 		_onComplete = OnComplete;
+		#if (flash && openfl <= "3.4.0")
+		var result:String = null;
+		#else
 		var result:SharedObjectFlushStatus;
+		#end
 		try 
 		{ 
-			result =  _sharedObject.flush(MinFileSize); 
+			#if (!js && openfl <= "3.4.0")
+			result = _sharedObject.flush(MinFileSize); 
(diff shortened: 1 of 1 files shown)


Gama11:
That would break compatibility with older OpenFL versions though, right?

larsiusprime:
Ah, I guess so.

larsiusprime:
This should be compatible with both old and new versions of OpenFL.

Gama11:
Ah, so this is actually required for new openfl versions and not just cleanup?

Shouldn't the defines use `||` rather than `&&`?

Never really sure what to do with breaking openfl changes as we don't define a "minimum required version" with it, unlike Haxe. I guess it's better to stay compatible than not...

larsiusprime:
Yeah, it is actually a requirement for the latest stuff.

And I'm *pretty* sure that the defines should be using ```&&```. If you're using the latest version of OpenFL, everything's nice and unified, including on flash (and js).

It's only using previous versions of OpenFL that the # if flash, #if js, blocks are still required.

Gama11:
How can the flash behavior have changed? Shouldn't the code that flash needed before OpenFL 3.4.0 still be needed?

larsiusprime:
See:

https://github.com/openfl/openfl/commit/7ac3b7dc71a8d9ffa9a314988230ffb6fd3a7732
https://github.com/openfl/openfl/commit/cd9c1eaa010def8e9778e9f7fd638e8114711d82

PR #1675, opened 2015-12-17
Fix button images not loading in html5

larsiusprime:


Gama11:
Hm... this is more of a workaround than a solution - which would be fine, as long as it's still eventually fixed in OpenFL.

include.xml
```diff
@@ -8,6 +8,8 @@
 	<assets path="assets/fonts/nokiafc22.ttf" embed="true" />
 	<assets path="assets/fonts/monsterrat.ttf" embed="true" />
 	
+	<assets path="assets/images/ui" include="*" if="html5" />
```
Gama11: Why include everything in `/ui` if only one file is used?

larsiusprime: Fair enough point. I guess I'll just include the one.

PR #1679, opened 2015-12-22
Added option for repeating a bitmap fill

thegoodideaco:
If this is not a good thing to do, I totally understand :)

### DIFF
--- a/flixel/FlxStrip.hx
+++ b/flixel/FlxStrip.hx
@@ -67,7 +67,7 @@ class FlxStrip extends FlxSprite
 			}
 			
 			getScreenPosition(_point, camera);
-			camera.drawTriangles(graphic, vertices, indices, uvtData, colors, _point, blend, antialiasing);
+			camera.drawTriangles(graphic, vertices, indices, uvtData, colors, _point, blend, antialiasing, true);
 		}
 	}
 }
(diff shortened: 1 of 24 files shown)


Gama11:
What's the use case for this?

`FlxStrip` is the only place that calls `camera.drawTriangles()` to my knowledge, so I'm not sure an option would be needed. Or are you manually calling it and need to be able to specify `repeat = false`?

thegoodideaco:
I made a class that extends FlxStrip, It's being used to wrap the bitmap around a ribbon mesh, so the ground can curve:

https://github.com/thegoodideaco/CurvyTerrain

Having issues with setting uvtData values past 1, where it seems to be stretching the pixels. setting this to true fixes it in -Dnext, however it only works on legacy if I make the flashGfxSprite cacheAsBitmap();

These don't sound like optimal solutions considering I don't know what's really going on underneath, and there might be a better solution that I'm just not aware of

Gama11:
Not really an expert here myself. @Beeblerox might be able to help.

flixel/FlxStrip.hx
```diff
@@ -67,7 +67,7 @@ class FlxStrip extends FlxSprite
 			}
 			
 			getScreenPosition(_point, camera);
-			camera.drawTriangles(graphic, vertices, indices, uvtData, colors, _point, blend, antialiasing);
+			camera.drawTriangles(graphic, vertices, indices, uvtData, colors, _point, blend, true, antialiasing);
```
Beeblerox: it's better to add public `repeating` property to `FlxStrip` and use it here

PR #1680, opened 2016-01-06
compile fix for latest openfl (apparently removes colors & blending .…

larsiusprime:
…from drawtriangles API)

### DIFF
--- a/flixel/graphics/tile/FlxDrawTrianglesItem.hx
+++ b/flixel/graphics/tile/FlxDrawTrianglesItem.hx
@@ -60,11 +60,7 @@ class FlxDrawTrianglesItem extends FlxDrawBaseItem<FlxDrawTrianglesItem>
 		}
 		
 		camera.canvas.graphics.beginBitmapFill(graphics.bitmap, null, true, (camera.antialiasing || antialiasing));
-		#if flash
 		camera.canvas.graphics.drawTriangles(vertices, indices, uvtData, TriangleCulling.NONE);
-		#else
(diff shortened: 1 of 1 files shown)


larsiusprime:
Do we use drawtriangles much or is it still experimental? Anyways, saw that it wouldn't compile without this, figure we can ask OpenFL if there's any questions about it.

Tiago-Ling:
I used it for my game - drawTriangles is necessary to do "free form
deformation" using Spine (inside flixel-addons).

I think we should keep it, afaik vertex manipulation is only possible with
drawTriangles right?

Tiago Ling Alexandre
Tel: +55 41 8819-3191


2016-01-06 12:37 GMT-02:00 Lars Doucet <notifications@github.com>:

larsiusprime:
This doesn't remove drawtriangles, it just removes the color & blend parameters that OpenFL just removed. I vote for keeping it to, this change just lets it compile again. If we need the color & blend properties, we'll need to talk to Joshua about that. Do you use them?

Tiago-Ling:
Oh, ok - i thought it was about to be removed.

I do not use color or blending at the moment (don't know if FlxSpine does
use them internally, but i don't think so).


Tiago Ling Alexandre
Tel: +55 41 8819-3191


2016-01-06 13:49 GMT-02:00 Lars Doucet <notifications@github.com>:

MSGhero:
It matches the flash API now if that's any consolation.

Gama11:
I prefer the version of this fix that lets us keep this functionality for leagcy: #1681

larsiusprime:
works for me

Issue #951, opened 2014-03-21
Unexpected behaviour with certain usage of tweens due to pooling


Gama11:
Thanks @PaulGene for finding this bug and [creating a demo for it](https://github.com/PaulGene/FlxInvadersAtlas).

This is the interesting part:

``` haxe
private function tweenOff():Void
{
    if (_tween != null)
        _tween.cancel();    // We need this because previous tween can still be active, which means you would get 2 tweens scaling to different values!

    _tween = FlxTween.tween(this.scale, { x : 1, y : 1 }, 0.5, { ease : FlxEase.bounceOut});
}
```

The problem is that by the time we call `cancel()`, the old tween has finished and been put back into the pool - since this happens in a couple of objects, that tween has also been retrieved from the pool and started again, which means we're cancelling the tween for another object accidentally.

This issue isn't limited to `cancel()`, it also affects `destroy()` (not a huge issue, that's not really intended to be used by the user anyway) and `start()`.

@gamedevsam Any thoughts on this? I don't think there's a great solution for this... :/

PR #1692, opened 2016-01-18
FlxTween: make start() public per docs

GimmickyApps:
"FlxTween.PERSIST: stops when it finishes. Unlike ONESHOT, this type of tween stays attached to the core container when it finishes. This means you can keep a reference to this tween and call start() whenever you need it."

From this, I believe it was originally intended that start() be public.

- http://haxeflixel.com/documentation/flxtween/


MSGhero:
Newlines get altered in almost every commit lol. Good catch :+1:

Gama11:
Good point.. Looks like I made them private when fixing #951, but that's somewhat irrelevant now that tween pooling has been removed.

PR #1700, opened 2016-01-20
Color offsets for DrawTiles

larsiusprime:


flixel/FlxCamera.hx
```diff
@@ -518,7 +520,8 @@ class FlxCamera extends FlxBasic
 		}
 	}
 	
-	public function drawPixels(?frame:FlxFrame, ?pixels:BitmapData, matrix:FlxMatrix, cr:Float = 1.0,
+	public function drawPixels(?frame:FlxFrame, ?pixels:BitmapData, matrix:FlxMatrix, cr:Float = 1.0, 
+		cg:Float = 1.0, cb:Float = 1.0, ca:Float = 1.0, croff:Int = 0, cgoff:Int = 0, cboff:Int = 0, caoff:Int = 0, ?blend:BlendMode = null, ?smoothing:Bool = false):Void
 		cg:Float = 1.0, cb:Float = 1.0, ca:Float = 1.0, ?blend:BlendMode, smoothing:Bool = false):Void
```
Gama11: Are you sure this compiles?

include.xml
```diff
@@ -4,7 +4,7 @@
 	
 	<assets path="assets/sounds" include="*.ogg" if="html5" unless="unit-test" />
 	<assets path="assets/sounds" include="*.mp3" if="flash" embed="true" />
-	<assets path="assets/sounds" include="*.ogg" unless="flash || html5" embed="true" />
+	<assets path="assets/sounds" include="*.ogg" unless="flash || html5" embed="false" />
```
Gama11: This seems unrelated.

larsiusprime: Yup, I'll exclude it.

flixel/FlxCamera.hx
larsiusprime: merge artifact, arg.

flixel/FlxSprite.hx
```diff
@@ -1207,6 +1225,7 @@ class FlxSprite extends FlxObject
 	
 	private function set_color(Color:FlxColor):Int
 	{
+		Color = 0x00FFFFFF & Color;
```
Gama11: Why do we need this change, color seems unrelated to colorOffset?

flixel/FlxSprite.hx
```diff
@@ -143,6 +143,12 @@ class FlxSprite extends FlxObject
 	 */
 	public var color(default, set):FlxColor = 0xffffff;
 	
+	/**
+	 * Brightens the whole sprite by adding a color (0xRRGGBB format). You can use
+	 * 0xAARRGGBB colors, but the alpha value will simply be ignored. To change the opacity use alpha. 
+	 */
+	public var colorOffset(default, set):FlxColor = 0x000000;
```
Gama11: Didn't we talk about how FlxColor doesn't really make sense as a representation of color offsets? (negative offsets and all that)

flixel/FlxSprite.hx
larsiusprime: Ah, indeed it is. Another change from a local fix I was working on. It's still the correct logic, because set_color shouldn't be alpha-aware based on its logic later on, there's some edge cases where not doing this causes text to sometimes become transparent as it inherits this function. 

Should I separate it as its own PR?

flixel/FlxSprite.hx
larsiusprime: yeah, I remember that now. What is the API you would prefer?

flixel/FlxSprite.hx
Gama11: I'm pretty sure the idea here is that the alpha part is not stripped away so this does not cause unexpected behavior:

```haxe
sprite.color = FlxColor.RED;
trace(sprite.color == FlxColor.RED); // you'd expect this to trace `true`
```

What edge cases are you talking about specifically?

flixel/FlxSprite.hx
Gama11: I think the conclusion was to not add any new properties / functions, since this is technically already exposed via setColorTransform() (and the colorTransform object is public anyway).

flixel/FlxSprite.hx
larsiusprime: It was from a week or so ago, let me remove and see if I notice any issues in my code. If not, I'll remove it.

larsiusprime: Okay, it definitely isn't necessary. The bug I was worried about was later traced to something else now that I'm looking at it again. Reverted.

flixel/FlxSprite.hx
larsiusprime: kay

flixel/graphics/tile/FlxDrawTilesItem.hx
```diff
@@ -33,7 +33,7 @@ class FlxDrawTilesItem extends FlxDrawBaseItem<FlxDrawTilesItem>
 	}
 	
 	override public function addQuad(frame:FlxFrame, matrix:FlxMatrix,
-		red:Float = 1, green:Float = 1, blue:Float = 1, alpha:Float = 1):Void
+		red:Float = 1, green:Float = 1, blue:Float = 1, alpha:Float = 1, redOffset:Float = 0, greenOffset:Float = 0, blueOffset:Float = 0, alphaOffset:Float = 0):Void
```
Gama11: Shouldn't the offsets be integers?

flixel/graphics/tile/FlxDrawTilesItem.hx
```diff
@@ -11,7 +11,7 @@ class FlxDrawTilesItem extends FlxDrawBaseItem<FlxDrawTilesItem>
 {
 	public var drawData:Array<Float> = [];
 	public var position:Int = 0;
-	
+	public var hasColorOffsets:Bool = false;
```
Gama11: This should probably be in `FlxDrawBaseItem` like `colored`.

flixel/FlxSprite.hx
```diff
@@ -636,7 +646,7 @@ class FlxSprite extends FlxObject
 				}
 				
 				_point.copyToFlash(_flashPoint);
-				camera.copyPixels(_frame, framePixels, _flashRect, _flashPoint, cr, cg, cb, alpha, blend, antialiasing);
+				camera.copyPixels(_frame, framePixels, _flashRect, _flashPoint, cr, cg, cb, alpha, croff, cgoff, cboff, caoff, blend, antialiasing);
```
Gama11: Man, 8 parameters for color multipliers / offsets. I feel like passing around a `ColorTransform` instance would be much nicer.

flixel/FlxCamera.hx
```diff
@@ -359,7 +359,7 @@ class FlxCamera extends FlxBasic
 	
 	#if !FLX_RENDER_TRIANGLE
 	@:noCompletion
-	public function startQuadBatch(graphic:FlxGraphic, colored:Bool, ?blend:BlendMode, smooth:Bool = false):FlxDrawTilesItem
+	public function startQuadBatch(graphic:FlxGraphic, colored:Bool, ?blend:BlendMode = null, smooth:Bool = false, hasColorOffsets:Bool = false):FlxDrawTilesItem
```
Gama11: Regarding parameter order, I think having `hasColorOffsets` after `colored` makes more sense.

flixel/graphics/tile/FlxDrawTilesItem.hx
larsiusprime: openfl drawtiles takes an array of floats, I suppose I could make them integers here but they would be converted to floats later

Gama11: If the color params were replaced with a ColorTransform, that wouldn't be relevant anyway.

flixel/FlxSprite.hx
larsiusprime: To be clear, do you think I should make that change now?

flixel/FlxCamera.hx
larsiusprime: done

flixel/graphics/tile/FlxDrawTilesItem.hx
larsiusprime: done

flixel/FlxSprite.hx
larsiusprime: done

flixel/graphics/tile/FlxDrawTilesItem.hx
larsiusprime: Okay, so you do want me to go ahead and replace the params with a ColorTransform?

flixel/FlxSprite.hx
Gama11: Seems like a sensible time to make the change? Could still do it later though if you'd rather be done with the PR.

flixel/graphics/tile/FlxDrawTilesItem.hx
```diff
@@ -73,6 +80,11 @@ class FlxDrawTilesItem extends FlxDrawBaseItem<FlxDrawTilesItem>
 				tempFlags |= Tilesheet.TILE_RGB;
 			}
 			
+			if (hasColorOffsets)
+			{
+				tempFlags |= Tilesheet.TILE_TRANS_COLOR;
```
Gama11: Referencing this constant breaks compatibility with legacy and versions of openfl next that don't have this feature yet.

larsiusprime: I can guard it using ```!openfl_legacy``` and check for openfl version 3.5.4 (the next tag), but Joshua hasn't applied that yet. I just pinged him about it, tho.

flixel/graphics/tile/FlxDrawTilesItem.hx
larsiusprime: done

larsiusprime:
@gama11 -- made all your changes. The colortransform parameter cleans things up a lot but required changes in a lot of places. I tried to make sure that I wasn't making assumptions about how alpha was being used, and I did check that the colortransform is only ever used as a package -- it's never referenced, red/green/blue/alpha offsets & multipliers are just read straight out of it. This also saves us a duplication in at least one place (take a color transform, send raw colors, then make a color transform out of THOSE...), so that's a bonus.

Gama11:
Nice!

The version check isn't good enough it seems, we also shouldn't set the offset draw data in that case, otherwise that causes visual bugs (see Mode for example).

Also, that issue revealed that the `hasColorTransform` check is incorrect (after all, Mode doesn't use offsets anywhere, so that code path shouldn't even trigger). Should be != 0, not 1, I'd guess.

larsiusprime:
@Gama11  fixed and fixed.

Note that if you want to see the new behavior you have to locally edit the version guards to say 3.5.3, at least until Joshua puts out a new release tag on openfl.

larsiusprime:
@gama11 anything else? I'm about to leave for the day.

Gama11:
Travis is failing.

larsiusprime:
Okay, figured it out, that was mostly things in flixel-addons that needed updating.

I'm trying to run the coverage test locally since this PR requires changes in two libraries at once (which I presume Travis can't cover). If the coverage test compiles, does that mean it has passed? When I build in flash I just get an empty window and in neko it just closes immediately. Is that supposed to happen?

Gama11:
Yes, the coverage tests is just an empty projects that attempts to compile everything. As long as the test compiles, it was successful.

larsiusprime:
Okay, my coverage test was successful on flash, neko, and windows then. Just pushed a fix to this branch, the companion PR for flixel-addons is here:

https://github.com/HaxeFlixel/flixel-addons/pull/221

flixel/FlxCamera.hx
```diff
@@ -555,13 +568,23 @@ class FlxCamera extends FlxBasic
 		{
 			_helperMatrix.identity();
 			_helperMatrix.translate(destPoint.x + frame.offset.x, destPoint.y + frame.offset.y);
-			var isColored:Bool = (cr != 1.0) || (cg != 1.0) || (cb != 1.0) || (ca != 1.0);
```
Gama11: It seems a little strange to me that alpha would be included in the `isColored` check, when it ends up being used for the `TILE_RGB` flag, which doesn't include alpha... Instead, alpha is always added to the draw data array. The original check might already have been wrong?

Gama11:
There is an unacceptable amount of redundancy caused by the checks for `isColored` (repeated twice) and `hasColorOffsets` (repeated four times).

I'm thinking some static extension methods for `ColorTransform` in a `flixel.util.FlxColorTransformUtil` would make sense:

- `setMultipliers(red, green, blue, alpha)`
- `hasMulitpliers():Bool`
- `setOffsets(red, green, blue, alpha)`
- `hasOffsets():Bool`

The setters should be used to clean up the changes in `FlxBitmapText`.

flixel/FlxCamera.hx
larsiusprime: Yeah, suppose so. I guess that "isColored" check is in FlxSprite, want me to remove it from there?

Gama11: Which check do you mean? Are you talking about the entire check now or just the alpha part?

larsiusprime: Just the alpha part, anywhere we do an "isColored" check

larsiusprime:
FlxColorTransformUtil added, Redundancy removed. I add RGB and RGBA versions of the multiplier / offset checks so we can use whichever is most appropriate in each case and be very explicit.

flixel/graphics/tile/FlxDrawTilesItem.hx
```diff
@@ -73,6 +82,13 @@ class FlxDrawTilesItem extends FlxDrawBaseItem<FlxDrawTilesItem>
 				tempFlags |= Tilesheet.TILE_RGB;
 			}
 			
+			#if (!openfl_legacy && openfl >= "3.5.4")
```
Gama11: Looks like OpenFL 3.6.0 just got released. These check would technically work, but should probably still be updated to avoid confusion about a non-existant 3.5.4 version.

larsiusprime: Okay, cool.

larsiusprime:
Done & done.

Gama11:
Looks like your forgot one of the version checks?

larsiusprime:
Okay, that's done now. Also caught a logic error at the last minute (was making things transparent that shouldn't be :P)

Anything else? And when it's done do you want me to rebase+squash it? (Finally figured out how to do that. Backing up the branch before messing with it helps a lot!)

Gama11:
I was just testing the actual feature of this PR again, and the test I used before doesn't seem to work anymore:

```haxe
// Mode MenuState.hx:
_playButton.setColorTransform(1, 1, 1, 1, 100); // should lead to a slightly red button
```

And yes, a squash would be nice with the amount of commits. :)

Btw, if you didn't merge anything in from another branch, you could use SourceTree's "Reset to commit" instead of rebase: right click on the commit before the first commit of the PR, Reset current branch to this commit, Mixed, Ok. You then have the changes of all the removed commits in your working copy and can commit them again.

larsiusprime:
Reproducing your bug now, fixing...

Gama11:
Nevermind actually, I can't really expect that to work without `<set name="next" />`... -.-

With that it works fine. :+1:

larsiusprime:
D'oh. Same mistake here.

larsiusprime:
https://github.com/HaxeFlixel/flixel/pull/1705

PR #1710, opened 2016-02-01
FlxSprite: getFlxFrameBitmapData -> updateFramePixels()

Gama11:
I really dislike having the `Flx` prefix in method names. It's also inconsistent with the rest of the codebase.

I feel like `get` wasn't really expressing the intent very well - it primarily _updates_ `framePixels`, and happens to also return it for convenience.


gamedevsam:
:+1:

PR #1711, opened 2016-02-01
toPoint() -> getPosition(), toRect() -> getHitbox()

Gama11:
Not a fan of these names anymore. `sprite.toPoint()` doesn't "convert it" to a point, it just retrieves the position - which is also much more in line with `getScreenPosition()`.

`getHitbox()` I'm a bit less sure about, but it's in principle the same. However, the API only uses the term `hitbox` in docs, not actual method or variable names. Also not sure if we need to be concerned about people being confused that setting the rect's properties doesn't change the actual hitbox, but I suppose it's the same with `getPosition()` and it doesn't really seem like a concern there?

This is technically a breaking change, however these methods weren't part of an official release yet.

### DIFF
--- a/flixel/FlxObject.hx
+++ b/flixel/FlxObject.hx
@@ -826,6 +826,11 @@ class FlxObject extends FlxBasic
 		return point.subtract(Camera.scroll.x * scrollFactor.x, Camera.scroll.y * scrollFactor.y);
 	}
 	
+	public inline function getPosition():FlxPoint
+	{
+		return FlxPoint.get(x, y);
+	}
+	
 	/**
 	 * Retrieve the midpoint of this object in world coordinates.
 	 * 
@@ -839,6 +844,11 @@ class FlxObject extends FlxBasic
 			point = FlxPoint.get();
 		}
 		return point.set(x + width * 0.5, y + height * 0.5);
+	}
+	
+	public inline function getHitbox():FlxRect
+	{
+		return FlxRect.get(x, y, width, height);
 	}
 	
 	/**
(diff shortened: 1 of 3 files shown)


flixel/FlxObject.hx
```diff
@@ -841,6 +846,11 @@ class FlxObject extends FlxBasic
 		return point.set(x + width * 0.5, y + height * 0.5);
 	}
 	
+	public inline function getHitbox():FlxRect
```
gamedevsam: I'd prefer to structure this code similar to ```getMidPoint``` (with optional parameter), That way developers can easily optimize code to avoid allocations or pooling costs.

flixel/FlxObject.hx
```diff
@@ -826,6 +826,11 @@ class FlxObject extends FlxBasic
 		return point.subtract(Camera.scroll.x * scrollFactor.x, Camera.scroll.y * scrollFactor.y);
 	}
 	
+	public inline function getPosition():FlxPoint
```
gamedevsam: I'd prefer to structure this code similar to getMidPoint (with optional parameter), That way developers can easily optimize code to avoid allocations or pooling costs.

Gama11: Hm.. I was never a huge fan of that, but then again, it might make sense purely for consistency..

Gama11:
Updated.

Gama11:
I actually found a really nice use case for having the optional point param: passing a `FlxPoint.weak()`, so you can directly pass the `getPosition()` result into another function.

PR #1712, opened 2016-02-01
Remove FlxPathManager, let FlxObject take care of updating it (#1087)

Gama11:
This takes care of any substate issues with paths still being updated while the object they belong to isn't. This solution works because every object can only follow one path at a time - for timers and tweens a different solution is needed. Similar to the original AS3 Flixel API, FlxObject now has a public `path` property.

Downside: it introduces a circular dependency between FlxObject and FlxPath.

### DIFF
--- a/flixel/FlxObject.hx
+++ b/flixel/FlxObject.hx
@@ -1190,7 +1190,8 @@ class FlxObject extends FlxBasic
 			return path;
 		
 		if (this.path != null)
-			path.object = null;
+			this.path.object = null;
+		
 		if (path != null)
 			path.object = this;
 		return this.path = path;
--- a/flixel/FlxObject.hx
+++ b/flixel/FlxObject.hx
@@ -10,6 +10,7 @@ import flixel.tile.FlxBaseTilemap;
 import flixel.util.FlxAxes;
 import flixel.util.FlxColor;
 import flixel.util.FlxDestroyUtil;
+import flixel.util.FlxPath;
 import flixel.util.FlxSpriteUtil;
 import flixel.util.FlxStringUtil;
 
@@ -543,6 +544,11 @@ class FlxObject extends FlxBasic
 	public var ignoreDrawDebug:Bool = false;
 	#end
 	
+	/**
+	 * The path this object follows.
+	 */
+	public var path(default, set):FlxPath;
+	
 	private var _point:FlxPoint = FlxPoint.get();
 	private var _rect:FlxRect = FlxRect.get();
 	
@@ -620,10 +626,11 @@ class FlxObject extends FlxBasic
 		last.x = x;
 		last.y = y;
 		
+		if (path != null && path.active)
+			path.update(elapsed);
+		
 		if (moves)
-		{
 			updateMotion(elapsed);
-		}
 		
 		wasTouching = touching;
 		touching = NONE;
@@ -994,6 +1001,9 @@ class FlxObject extends FlxBasic
 		for (camera in cameras)
 		{
 			drawDebugOnCamera(camera);
+			
+			if (path != null && !path.ignoreDrawDebug)
+				path.drawDebug();
 		}
 	}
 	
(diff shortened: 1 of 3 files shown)


flixel/util/FlxPath.hx
```diff
@@ -207,12 +191,21 @@ class FlxPath implements IFlxDestroyable
...
-	 * Generally this function is called automatically by preUpdate().
 	 * The first half of the function decides if the object can advance to the next node in the path,
 	 * while the second half handles actually picking a velocity toward the next node.
 	 */
 	public function update(elapsed:Float):Void
 	{
+		if (object == null)
+			return;
+		
+		if (_firstUpdate)
```
Gama11: This is a terrible hack, but I couldn't think of a better solution right off the bat now that `start()` doesn't have an `Object` parameter anymore.

flixel/FlxObject.hx
```diff
@@ -620,10 +626,11 @@ class FlxObject extends FlxBasic
 		last.x = x;
 		last.y = y;
 		
+		if (path != null && path.active)
+			path.update(elapsed);
```
gamedevsam: Doesn't seem like the path nulls itself out when it completes.

Gama11: No, that would be a bit weird, since it got assigned to the `FlxObject` by the user. He might want to `restart()` it.

gamedevsam: Ahh, makes sense.

gamedevsam:
I wonder if the better approach would be to migrate the plugins to exist on FlxState level (when new Timer / Path / other plugin object was created, it would get its manager from the active state). Everything has to be inside of a state anyways, so why not make each state contain its own set of IUpdatable plugins?

Gama11:
I've thought about that. It's not really a solution to the underlying issue, it just prevents conflicts with substates. You might be manipulating the `elapsed` value of the `super.update()` call, and you'd expect the timers / tweens belonging to an object to slow down / speed up appropriately (the issue being that timers and tweens don't really belong to an object right now).

I think @JoeCreates' solution was essentially to add a lazily initialized manager to `FlxBasic` (which we could probably do without breaking changes if still keeping the old API available). Anyway, for `FlxPath`, a manager is overkill, since every object can only have one at a time.

flixel/util/FlxPath.hx
gamedevsam: A better approach would be to have two a private vars: ```_immovable``` and ```_wasImmovable```, and have a public property ```public var immovable(get, set);``` that users would use to manually modify these values.

flixel/FlxObject.hx
```diff
@@ -1173,4 +1183,16 @@ class FlxObject extends FlxBasic
...
 		return allowCollisions = Value;
 	}
+	
+	private function set_path(path:FlxPath):FlxPath
+	{
+		if (this.path == path)
+			return path;
+		
+		if (this.path != null)
+			path.object = null;
```
gamedevsam: We should probably call ```path.destroy();``` here instead.

Gama11: To me, `destroy()` implies setting `path` to null and letting it be garbage-collected. Like I mentioned earlier, the user might want to `restart()` the path.

gamedevsam: There's a bug here:
```if (this.path != null)
			path.object = null;
```
Should be:
```if (this.path != null)
			this.path.object = null;
```

With regards to destroy, I suggest we destroy the path that is being replaced (the one being nulled out).

Gama11: Fixed.

flixel/util/FlxPath.hx
Gama11: Hm, I'm not sure that approach is better. It leads to an even tigther coupling between `FlxPath` and `FlxObject` (accessing `path._desiredImmovable `), I already don't like it as is..

gamedevsam: Both approaches are hacky, it's a bit late to avoid tight coupling between these two objects. Your implementation introduces a bug where the user's immovable value could get clobbered by the Path if they're not aware of the inner implementation.

The way to avoid the tight coupling would be to take a completely different approach to this problem. Avoiding tighter coupling is not a good excuse for leaving this bug in.

gamedevsam:
I approve, :+1:

PR #1713, opened 2016-02-01
FlxCamera#follow(): remove the Offset parameter, closes #1056

### LINKED ISSUE #1056: FlxCamera offset is weird. No way to offset target.
The offset point you can provide in FlxCamera seems rather weird. I can't at all think why it exists, as it seems only to affect the size of the deadzone.

In my ludum dare game, I needed to be able to offset the actual target position of a sprite, so for example, the target position would be somewhere above the target sprite. I couldn't find any way of doing this, so I added a targetOffset point to the class which is added to the target x and y when updating the following.

I would make this change a PR, however I wanted to check what the existing purpose of the offset point is. Is this a badly implemented attempt at offsetting the target, or does it have a legitimate purpose?

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/2025371-flxcamera-offset-is-weird-no-way-to-offset-target?utm_campaign=plugin&utm_content=tracker%2F254189&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F254189&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Gama11:
Its usefulness is questionable as discussed in #1056.

### DIFF
--- a/flixel/FlxCamera.hx
+++ b/flixel/FlxCamera.hx
@@ -1028,13 +1028,12 @@ class FlxCamera extends FlxBasic
 	/**
 	 * Tells this camera object what FlxObject to track.
 	 * 
-	 * @param	Target	The object you want the camera to track.  Set to null to not follow anything.
+	 * @param	Target	The object you want the camera to track. Set to null to not follow anything.
 	 * @param	Style	Leverage one of the existing "deadzone" presets. Default is LOCKON. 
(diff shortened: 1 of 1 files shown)


gamedevsam:
:+1:

PR #1714, opened 2016-02-02
Add native PSVita gamepad controls (WiiU, PS4, XBOX1 to follow soon, …

larsiusprime:
…as well as general cleanup now that I know what I'm doing)


MSGhero:
Do you think it would be better in general to make the ids cross-library? I guess a separate lib that flixel can include by default. 

In the future, not near 4.0 release or anything.

larsiusprime:
You mean just store these id mappings in a big ol' list of constants and then make it something flixel just makes use of?

I'm not 100% sure of the usefulness of that -- these id's are all totally specific to OpenFL anyway, and so it would only be relevant for libraries built on top of Lime/OpenFL, so it wouldn't help out Kha or Luxe or Flambe at all. I guess it might save a step for HaxePunk tho...

PR #1715, opened 2016-02-05
Gamepad

larsiusprime:
- Unifies XInputID.hx and XBox360ID.hx, eliminating the latter
- Fixes XInputID.hx with latest tests from Windows, Mac, and Linux, on legacy and next.
- Fixes PS4ID.hx with latest tests from Windows, Mac, and Linux, on legacy and next.
- Added notes about which drivers are assumed (sensible defaults to recommend to users -- on mac linux, without them, controllers often won't work *at all* and they're the top google hits for those platforms' controller support).

Couple of things: I recently upgraded to El Capitan on my macbook and something annoying with apple's audio API is keeping openfl from compiling flixel applications under -Dnext. So although I have full test results on Linux and Windows confirmed, on Mac I could only test legacy. However, I was able to run the openfl-samples GamepadInput test (in next), which gave me the outputs I've put in this PR. From what I've seen, the GamepadInput test results are extremely reliable and extremely consistent cross platform. Usually the only discrepancy between platforms is whether or not the GUIDE button works or is reserved by the system, or whether an analog stick axis needs to be inverted by default. The only other divergence I saw was the result of using an old and unsupported driver (xboxdrv on Linux, which was terrible and made me switch to Valve's version of xpad), and it vanished when I installed a better driver.

PS4 controller is recognized as a PS4 controller on both Mac and Windows, but on Linux xpad makes it indistinguishable from an XInput device.

PS3 controller is an extremely difficult little bugger and has never consistently worked on any platform anywhere without a crazy 3rd party tool completely disguising its identity as a 360 controller. So I feel good about dropping all support for PS3 controllers, and focusing on something more worthwhile, like Logitech.


flixel/input/gamepad/id/XInputID.hx
```diff
@@ -4,52 +4,210 @@ import flixel.input.gamepad.FlxGamepad;
...
+		// Axis indices
+		public static var LEFT_ANALOG_STICK(default, null) = new FlxGamepadAnalogStick(0, 1);
+		public static var RIGHT_ANALOG_STICK(default, null) = new FlxGamepadAnalogStick(2, 3);
+	
+		public static inline var LEFT_TRIGGER:Int = 4;
+		public static inline var RIGHT_TRIGGER:Int = 5;
+	
+		
+	#elseif FLX_JOYSTICK_API
+		#if !web
```
Gama11: Why `#if !web`? That means there are no IDs here for `FLX_JOYSTICK_API && web`?

larsiusprime: The way we currently have it set up, FLX_JOYSTICK_API basically just means "legacy," and there is no legacy target for "web" -- flash has no next/legacy distinction (or joystick API for that matter, it only has access to the gamepad API), and html5 *only* works with next (and I haven't checked but it's a solid bet that it only has access to the gamepad API as well).

The !web may very well be redundant, but it doesn't represent a hole -- there is no "legacy / joystick" + "flash / html5" mode. Flash target will take the #if flash path and HTML5 target will take the FLX_GAMEINPUT_API path.

Gama11: I see.

>The !web may very well be redundant

Should it just be removed then? Seems a bit confusing.

Gama11:
Would be nice if you could squash the commits when you're done.

Please also look into why travis is failing.

larsiusprime:
Sure, I'll try to get this done by today.

larsiusprime:
I'll check back to see if travis passes; I'll wait to squash until the whole series of commits has the thumbs up

larsiusprime:
@gama11 Okay travis is green, any last comments before I squashify?

Gama11:
Gonna checkout your branch locally and double-check if everything works...

Gama11:
IIRC, Flash used to recongize the Guide button? Not happening on latest dev or this branch though.

larsiusprime:
The Guide button is horribly inconsistent -- it's almost always reserved, or plain doesn't work. It may be best practice to assume it's always reserved but I like to expose whatever the system does. If it used to work, perhaps the recent openfl gamepad API changes were the result of that. On Windows there's even OS system level interrupts that happen when you push the xbox button on a 360/XB1 pad.

I have had better success with the GUIDE button for other models of controller, XInput is where it runs into trouble.

Gama11:
After testing all the relevant targets: it seems to only work on native legacy. Which is strange, since you'd expect it to work on Next as well then

larsiusprime:
I bet what it is, is that the gamepad API mostly reserves the GUIDE button, but in legacy we're using the joystick API, which is just raw hardware input. This might be a good reason to take a second look at the joystick API (works on both next & legacy) and consider making that the basis for everything.

Later, though.

flixel/input/gamepad/id/XInputID.hx
```diff
@@ -4,52 +4,209 @@ import flixel.input.gamepad.FlxGamepad;
...
+		public static var LEFT_ANALOG_STICK(default, null) = new FlxGamepadAnalogStick(0, 1);
+		public static var RIGHT_ANALOG_STICK(default, null) = new FlxGamepadAnalogStick(2, 3);
+	
+		public static inline var LEFT_TRIGGER:Int = 4;
+		public static inline var RIGHT_TRIGGER:Int = 5;
+	
+		
+	#elseif FLX_JOYSTICK_API
+		
+		#if (windows || linux)
```
Gama11: Seems like this could mostly be merged with the `#if mac` block, there's a lot of redundancy.

larsiusprime: If I remove the redundancy it looks like this:

```Haxe
#if (windows || linux || mac)
		
			// Button IDs
			public static inline var A:Int = 0;
			public static inline var B:Int = 1;
			public static inline var X:Int = 2;
			public static inline var Y:Int = 3;
			
			public static inline var LB:Int = 4;
			public static inline var RB:Int = 5;
			
			#if (windows || linux)
				
				public static inline var BACK:Int = 6;
				public static inline var START:Int = 7;
				
			#elseif mac
				
				public static inline var BACK:Int = 9;
				public static inline var START:Int = 8;
				
			#end
			
			#if windows
			
				public static inline var LEFT_STICK_CLICK:Int = 8;
				public static inline var RIGHT_STICK_CLICK:Int = 9;
			
			#elseif linux
			
				public static inline var LEFT_STICK_CLICK:Int = 9;
				public static inline var RIGHT_STICK_CLICK:Int = 10;
			
			#elseif mac
			
				public static inline var LEFT_STICK_CLICK:Int = 6;
				public static inline var RIGHT_STICK_CLICK:Int = 7;
				
			#end
			
			#if (windows || mac)
				
				public static inline var GUIDE:Int = 10;
				
			#elseif linux
			
				public static inline var GUIDE:Int = 8;
			
			#end
				
			//"fake" IDs, we manually watch for hat axis changes and then send events using these otherwise unused joystick button codes
			public static inline var DPAD_UP:Int = 11;
			public static inline var DPAD_DOWN:Int = 12;
			public static inline var DPAD_LEFT:Int = 13;
			public static inline var DPAD_RIGHT:Int = 14;
			
			// Axis indices
			public static inline var LEFT_TRIGGER:Int = 2;
			public static inline var RIGHT_TRIGGER:Int = 5;
			
			public static var LEFT_ANALOG_STICK(default, null) = new FlxGamepadAnalogStick(0, 1);
			public static var RIGHT_ANALOG_STICK(default, null) = new FlxGamepadAnalogStick(3, 4);
			
		#end
```

Less redundant, but takes a hit to readability, because there's no less than six different logical groups:

1. windows only
2. mac only
3. linux only
4. windows OR linux OR mac
5. windows OR linux (but not mac)
6. windows OR mac (but not linux)

I'm not sure a little redundancy for the sake of clarity is all that bad here as these are platform-specific lists of mappings anyway. Anyways, up to you.

gamedevsam:
I approve :+1:

flixel/input/gamepad/id/XInputID.hx
Gama11: Good point, didn't look like there were that many differences at first. I'm fine with leaving it like that.

larsiusprime: Cool, I also think it's good to leave lists kind of "as-is" when large swaths are not obviously identical, because it makes it easier for a contributor to update an incorrect ID for a single platform without having to know what's going on with the other platforms.

larsiusprime:
I have two factor authentication on my Github account which makes command line pushing a pain. I just squashed and made a new PR, it's over here:
https://github.com/HaxeFlixel/flixel/pull/1717

Gama11:
Hm, ok. I wish SourceTree would add a GUI option for force pushes.. probably too dangerous though.

PR #1716, opened 2016-02-05
FlxMath: remove getDistance()

Gama11:
It's pretty redundant since `FlxPoint` has a `distanceTo()`.

Technically it would be nicer if all distance-related functions in `FlxMath` were using that + `getPosition()`, but that might be bad for performance? Either way not an issue here, since you already have a `FlxPoint` object when calling `FlxMath.getDistance()`.

### DIFF
--- a/flixel/ui/FlxButton.hx
+++ b/flixel/ui/FlxButton.hx
@@ -435,7 +435,7 @@ class FlxTypedButton<T:FlxSprite> extends FlxSprite implements IFlxInput
 	private function checkInput(pointer:FlxPointer, input:IFlxInput, justPressedPosition:FlxPoint, camera:FlxCamera):Bool
 	{
 		if (maxInputMovement != Math.POSITIVE_INFINITY &&
-			FlxMath.getDistance(justPressedPosition, pointer.getScreenPosition()) > maxInputMovement &&
+			justPressedPosition.distanceTo(pointer.getScreenPosition(FlxPoint.weak())) > maxInputMovement &&
 			input == currentInput)
 		{
 			currentInput == null;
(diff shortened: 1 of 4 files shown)


flixel/ui/FlxButton.hx
```diff
@@ -435,7 +435,7 @@ class FlxTypedButton<T:FlxSprite> extends FlxSprite implements IFlxInput
 	private function checkInput(pointer:FlxPointer, input:IFlxInput, justPressedPosition:FlxPoint, camera:FlxCamera):Bool
 	{
 		if (maxInputMovement != Math.POSITIVE_INFINITY &&
-			FlxMath.getDistance(justPressedPosition, pointer.getScreenPosition()) > maxInputMovement &&
+			justPressedPosition.distanceTo(pointer.getScreenPosition(FlxPoint.weak())) > maxInputMovement &&
```
gamedevsam: Does this compile? First parameter to ```getScreenPosition``` is an FlxCamera. I suggest passing no parameters to ```getScreenPosition```

Gama11: Sure, Haxe allows skipping parameters if they have different types.

Passing `FlxPoint.weak()` means the weakly allocated point will be `put()` back autmoatcally in `distanceTo()`.

gamedevsam: Ohh, I didn't know that, weird. :+1:

gamedevsam:
Good here :+1:

flixel/ui/FlxButton.hx
Gama11: I just puts `null` as the first argument.

PR #1726, opened 2016-02-18
VCRFrontEnd#stopRecording(): add an OpenSaveDialog argument, closes #1725

### LINKED ISSUE #1725: Allow foregoing a save dialog on VCRFrontEnd.stopRecording() with flash target
- **Flixel version:** dev, updated just now
- **OpenFL version:** 3.6.0
- **Lime version:** 2.9.0
- **Affected targets:** flash (d/r)

```haxe
		var save:String = FlxG.vcr.stopRecording();
```

**Observed behavior:** A native Save dialog box opens, allowing the user to save the recording.
**Expected behavior:** A way to avoid this, so that I can use the VCR for integration testing.  (Or, suppose you want to handle the data internally, like for an instant replay a la Worms, without saving the file somewhere.)

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/30972785-allow-foregoing-a-save-dialog-on-vcrfrontend-stoprecording-with-flash-target?utm_campaign=plugin&utm_content=tracker%2F254189&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F254189&utm_medium=issues&utm_source=github).
</bountysource-plugin>

GimmickyApps:
Default true so as to not break existing code.

### DIFF
--- a/flixel/system/debug/VCR.hx
+++ b/flixel/system/debug/VCR.hx
@@ -89,7 +89,7 @@ class VCR
 	public inline function recording():Void
 	{
 		recordBtn.changeIcon(new GraphicRecordOn(0, 0));
-		recordBtn.upHandler = FlxG.vcr.stopRecording;
+		recordBtn.upHandler = FlxG.vcr.stopRecording.bind(true);
 	}
 
 	/**
--- a/flixel/system/frontEnds/VCRFrontEnd.hx
+++ b/flixel/system/frontEnds/VCRFrontEnd.hx
@@ -219,9 +219,11 @@ class VCRFrontEnd
 	/**
 	 * Stop recording the current replay and return the replay data.
 	 * 
+	 * @param	OpenSaveDialog	If true, and targeting flash, open an OS-native save dialog for the user to choose where to save the data, and save it there.
+	 * 
 	 * @return	The replay data in simple ASCII format (see FlxReplay.save()).
 	 */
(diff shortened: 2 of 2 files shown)


Gama11:
The change looks good to me, not merging for now though - this is still a new feature and as such belongs in 4.1.0. We might want to do a patch release (4.0.1) first to sort out some bugs.

GimmickyApps:
Cool, thanks for explaining :)

GimmickyApps:
It's interesting the Travis build completed.  I was getting a compiler error when trying to use it in a project.

Gama11:
Yeah, that's interesting. The reason is that the failing code only compiles in debug mode and the demos are compiled in release mode.

It's also slipped through the coverage tests (one of them defines `FLX_RECORD`), since debug mode is only activated in the _other_ case...

We might want to think about changing that to debug mode.. We shouldn't really "miss out" on anything with that, it should just compile more code..

GimmickyApps:
For me, the code was failing to compile in debug mode, too.

Gama11:
Actually, I've decided it's probably best to `dev` for the next minor release (`4.1.0`) and a new `patch` branch for the next patch release (4.0.1, if there is one).

Issue #1165, opened 2014-06-09
Multiple outlines, shadows and gradients for FlxText


JoeCreates:
Presently, it is very difficult to make a `FlxText` look distinct. I wish to improve `FlxText` to make it possible to have multiple outlines, multiple shadows, gradients, and perhaps other effects. What I would like would be a breaking change, so before I go ahead and try it, I'd like some feedback.

I would like to change the following:-
- Remove `borderStyle` and its related properties. (Since when did a shadow count as a border?)
- Make a general `FlxTextEffect` to represent the various filters.
- Make a `FlxTextOutline`, `FlxTextGradient`, and `FlxTextShadow`.
- Add an array of `FlxTextEffect`s to `FlxText`.
- Add shorthand methods, `addOutline()`, `addShadow()`, and `addGradient()`?

Note that effects would be applied cumulatively. This would make text significantly more customizable. Instead of just having a single border or outline, you could have outlined text with a shadow, or shadowed text with an outline around the shadow etc..
##

adrianulima:
Seeing this issue now. What if this new class `FlxEffectText` extends the newly released `FlxEffectSprite` and use the `IFlxEffect`s? The missing effects could added and would work for every `FlxEffectSprite` as well, and I'm sure you guys could also help to improve existing effects.

Gama11:
Using a separate `FlxText` class for this doesn't sound good.

adrianulima:
Right, not a new class so. Passing current `FlxText` as target to `FlxEffectSPrite` would work?

PR #1728, opened 2016-02-18
FlxText: fix blurry lines on Flash with center alignment

adrianulima:
This is a better fix for the old blurry lines bug on flash.

original issue: https://github.com/HaxeFlixel/flixel/issues/1422
last tricky to fix it: https://github.com/adrianulima/flixel/commit/cbbd73090b9ff6b2291e66e67d190f06cbdcdc70
remove commit: https://github.com/HaxeFlixel/flixel/commit/1040f3242caba8989fa3ba754c4c5f985d3b405a

### DIFF
--- a/flixel/text/FlxText.hx
+++ b/flixel/text/FlxText.hx
@@ -161,10 +161,12 @@ class FlxText extends FlxSprite
 	
 	private var _hasBorderAlpha = false;
 	
+	#if flash
 	/**
 	 * Helper to draw line by line used at drawTextFieldTo().
 	 */
 	private var _textFieldRect:Rectangle = new Rectangle();
+	#end
 	
 	/**
 	 * Creates a new FlxText object at the specified position.
@@ -821,10 +823,13 @@ class FlxText extends FlxSprite
 		resetFrame();
 	}
 	
+	/**
+	 * Internal function to draw textField to a bitmapData, if flash it calculates every line x to avoid blurry lines.
+	 */
 	private function drawTextFieldTo(graphic:BitmapData):Void
 	{
 		#if flash
-		if (alignment == FlxTextAlign.CENTER)
+		if (alignment == FlxTextAlign.CENTER && isTextBlurry())
 		{
 			var h:Int = 0;
 			var tx:Float = _matrix.tx;
@@ -852,6 +857,24 @@ class FlxText extends FlxSprite
 		
 		graphic.draw(textField, _matrix);
 	}
+	
+	#if flash
+	/**
+	 * Helper function for drawTextFieldTo(), this checks if thw workaround is needed to prevent blurry lines.
+	 */
+	private function isTextBlurry():Bool
+	{
+		for (i in 0...textField.numLines) 
+		{
+			var lineMetricsX = textField.getLineMetrics(i).x;
+			if (lineMetricsX - Std.int(lineMetricsX) != 0)
+			{
+				return true;
+			}
+		}
+		return false;
+	}
+	#end
 	
 	override public function draw():Void 
 	{
--- a/flixel/text/FlxText.hx
+++ b/flixel/text/FlxText.hx
@@ -162,6 +162,11 @@ class FlxText extends FlxSprite
 	private var _hasBorderAlpha = false;
 	
 	/**
+	 * Helper to draw line by line used at drawTextFieldTo().
+	 */
+	private var _textFieldRect:Rectangle = new Rectangle();
+	
+	/**
 	 * Creates a new FlxText object at the specified position.
 	 * 
 	 * @param   X              The X position of the text.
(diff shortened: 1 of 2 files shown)


flixel/text/FlxText.hx
```diff
@@ -804,33 +805,47 @@ class FlxText extends FlxSprite
...
 	
-	private function avoidSingleLineBlur():Void
+	private function drawTextFieldTo(graphic:BitmapData):Void
 	{
 		#if flash
-		// If it's a single, centered line of text, we center it ourselves so it doesn't blur to hell
-		if (textField.numLines > 1 || alignment != FlxTextAlign.CENTER)
+		if (alignment == FlxTextAlign.CENTER)
+		{
+			// Prevents blurry lines.
```
Gama11: This comment could be a bit more specific, maybe mention issue number and under which circumstances the issue occurs.

flixel/text/FlxText.hx
```diff
@@ -804,33 +805,47 @@ class FlxText extends FlxSprite
...
+			var h:Int = 0;
+			var old_tx:Float = _matrix.tx;
+			var rect:Rectangle = new Rectangle();
```
Gama11: This is not a good idea, that's an allocation every time... `FlxText` inherits inherits a `_flashRect`, can we use that here?

flixel/text/FlxText.hx
```diff
@@ -804,33 +805,47 @@ class FlxText extends FlxSprite
...
+	private function drawTextFieldTo(graphic:BitmapData):Void
 	{
 		#if flash
-		// If it's a single, centered line of text, we center it ourselves so it doesn't blur to hell
-		if (textField.numLines > 1 || alignment != FlxTextAlign.CENTER)
+		if (alignment == FlxTextAlign.CENTER)
+		{
+			// Prevents blurry lines.
+			var h:Int = 0;
+			var old_tx:Float = _matrix.tx;
```
Gama11: `old_tx` isn't camelCase.

flixel/text/FlxText.hx
adrianulima: It has `_flashRect` and `_flashRect2` I can use any safely? or it is better to create a new one?

Gama11: Hm, it looks like messing with either could cause problems..

adrianulima:
Fixed that, any other thing here?

Gama11:
Under the right circumstances, this fix can lead to a _ridicolous_ amount of draw-calls:

- lots of lines (one draw call per line)
- border style `OUTLINE` (the number of lines times 8 (!))
- frequent text updates (has to be redrawn each frame)



```haxe
package;

import flixel.FlxG;
import flixel.FlxState;
import flixel.text.FlxText;
using StringTools;

{	
	var textTemplate = 
		"Text [ticks]\n" +
		"Text\n" +
		"Text\n" +
		"Text\n" +
		"Text\n" +
		"Text\n";
	
	private var text:FlxText; 

	{
		text = new FlxText(0, 0, 0, "Test");
		text.alignment = FlxTextAlign.CENTER;
		text.borderStyle = OUTLINE;
		add(text);
	}
	
	{
		text.text = textTemplate.replace("[ticks]", Std.string(FlxG.game.ticks));
		super.update(elapsed);
	}
}
```

It seems like we should be able to just render the non-blurry text once, and then copy that to different positions, without multiplying the draw-calls like that.. In general I'm wondering if we couldn't use `copyPixels()` for rendering the borders..

larsiusprime:
I've been thinking that we should perhaps think of making borders a generic thing you can do with *any* FlxSprite -- basically all you're doing is taking the "basic" pixel data, tint it white, then stamp it all around using the desired color, plus offsets for thickness, etc, as defined in FlxBorderStyle.

Gama11:
@larsiusprime -> https://github.com/HaxeFlixel/flixel/issues/1165

There already is a `FlxOutlineEffect` in flixel-addons which should work on any sprite.

larsiusprime:
Wow, new stuff discovered every day!

MSGhero:
Wow where did that class come from? Addons has like 100 random things no one knows about.

flixel/text/FlxText.hx
```diff
@@ -942,7 +962,7 @@ class FlxText extends FlxSprite
 	{
 		var graphic:BitmapData = _hasBorderAlpha ? _borderPixels : graphic.bitmap;
 		_matrix.translate(x, y);
-		graphic.draw(textField, _matrix);
+		drawTextFieldTo(graphic);
```
adrianulima: @Gama11 We can draw the borders without calling `drawTextFieldTo()`, it will draw ~0.5 pixels displaced but the result is not too bad. This would reduce `drawCalls`, but I don't like the idea.

We also could remove the border property and use the `IFlxEffect` here as we need to use with any `FlxSprite`. But the current `FlxOutlineEffect` is more expensive than `FlxTextBorderStyle.OUTLINE` and `FlxTextBorderStyle.OUTLINE_FAST`. I openned an issue to improve `IFlxEffect`s, including outile and also a new one called `FlxShadowEffect`
https://github.com/HaxeFlixel/flixel-addons/issues/228

Gama11: We won't remove the border styles any time soon, that would be a big breaking change.

Gama11:
Considering how expensive this workaround can be, we should probably add a property with which it can be turned off, like `centerManually`.

Gama11:
In single-line cases especially, this is a lot more expensive than the previous workaround with border style `OUTLINE`, which doesn't seem ideal.

flixel/text/FlxText.hx
```diff
@@ -161,6 +162,11 @@ class FlxText extends FlxSprite
 	private var _hasBorderAlpha = false;
 	
 	/**
+	 * Helper to draw line by line used at drawTextFieldTo().
+	 */
+	private var _textFieldRect:Rectangle = new Rectangle();
```
Gama11: This only needs to exist `#if flash`.

Gama11:
Thanks!

PR #1733, opened 2016-02-19
VCR-FlxButton MOUSE_UP fix (closes #1729), VCR format improvement

### LINKED ISSUE #1729: VCR doesn't record mouseups unless cursor moved
- **Flixel version:** dev
- **OpenFL version:** 3.6.0
- **Lime version:** 2.9.0
- **Affected targets:** flash, probably all?

```haxe
//in update()
if (FlxG.mouse.justPressed) {
	trace( 'mouse down at ', FlxG.mouse.getWorldPosition() );
}
if (FlxG.mouse.justReleased) {
	trace( 'mouse up at ', FlxG.mouse.getWorldPosition() );
}
```
When you record, make sure to click without moving the cursor position at all.
**Observed behavior:**  Your .fgr will contain lines like this:
```
84km478,95,2,0
85km478,95,1,0
86km478,95,1,0
87km
88km
89km
90km
```
Thus, on playback, the above code snippet does not trace any mouseups, even though during recording it does.

**Expected behavior:**  The .fgr should contain one more update to trigger a mouseup, I think:
```
84km478,95,2,0
85km478,95,1,0
86km478,95,1,0
87km478,95,0,0
88km
89km
90km
```

IBwWG:
This is working in my project correctly now.  Hopefully in yours too :)


GimmickyApps:
Sorry, still getting used to git workflow, I forgot to rebase this after fixing my own dev branch.

Gama11:
What do you mean by "VCR format improvement"? Just some code cleanup, or is the output different now? We can't really change the output, we need to remain compatible with old recordings.

GimmickyApps:
The output is different (smaller) but the loader remains able to load .fgr files that were made in the old format.

GimmickyApps:
Huh.  Is the build failing because String isn't nullable on neko or something?  (I'm not using neko myself, so I'm not super familiar with it.)

flixel/system/replay/FlxReplay.hx
```diff
@@ -125,20 +122,22 @@ class FlxReplay
...
+		
+		var output:String = seed + "\n";
+		var curLine:String;
 		var i:Int = 0;
 		while (i < frameCount)
 		{
-			output += _frames[i++].save() + "\n";
+			curLine = _frames[i++].save();
+			if (curLine != null)
+				output += curLine + "\n";
```
gamedevsam: What happens during playback if we skip save frames where nothing happens? Does the VCR recording properly wait for input during frames where nothing happens?

flixel/system/replay/FrameRecord.hx
```diff
@@ -60,6 +60,9 @@ class FrameRecord
 	public function save():String
 	{
 		var output:String = frame + "k";
+	
+		if ((keys == null) && (mouse == null))
```
gamedevsam: I think this is a problem, if we don't save the frame, how does the playback know to wait during frames where no input occurred?

flixel/system/replay/FlxReplay.hx
GimmickyApps: I don't understand what you mean--you can't record during playback.  Or were your two questions unrelated to each other?  In that case your second question still doesn't make sense to me--what input are you waiting for on a frame where nothing happens?  (Doesn't "nothing happens" mean no input occurs on that frame?)

flixel/system/replay/FrameRecord.hx
GimmickyApps: It was apparently originally (before this PR) written with this purpose in mind, to [wait during frames not mentioned in the save file](https://github.com/HaxeFlixel/flixel/blob/dev/flixel/system/replay/FlxReplay.hx#L198) as I mentioned here about [the old loader being able to load the new format](https://github.com/HaxeFlixel/flixel/pull/1733#issuecomment-186262691).

IBwWG:
Conflicts here too?  Maybe it was a bad idea to try separating these two VCR-related PRs.

IBwWG:
Ah, OK, it was just the same conflicts as before.  Hopefully this is good now?

Gama11:
Seems that way.

Separating things makes it easier to review a pull request. This already does two separate things, changes the recording format _and_ fixes a bug, adding a third wouldn't help.

IBwWG:
Is there a way to trigger a new bulid, or do I just do a dummy checkin?

IBwWG:
OK, this should be good to go now, too.  I've also tested it recently with my project, so it works "in the wild" as well as in the unit test.  :)

IBwWG:
There, in case it helped to update the base, this is caught up to dev...not sure what's with that import line, but anyway it's there again.

IBwWG:
Ups, guess I tempted fate there.  I don't understand what's broken though--no test errors, and build completed successfully, yet exit code 1?  Also only on a few targets, not all?  Reminds me of a popup I once got in Win98..."Error: the operation was a success."  ;)

IBwWG:
Maybe it's this?

```
PASSED

Tests: 271  Passed: 271  Failed: 0 Errors: 0 Ignored: 2 Time: 0.31342

X Error of failed request:  BadRequest (invalid request code or no such operation)

  Major opcode of failed request:  147 (GLX)

  Minor opcode of failed request:  168 ()

  Serial number of failed request:  173

  Current serial number in output stream:  173

Building coverage tests...
```

IBwWG:
How can this change cause graphics driver trouble?  Is this another one of those situations where the build system is broken and I just happened to push a change at the wrong moment?

Gama11:
I don't know what's causing the Travis failures right now, but it's unrelated to your changes.

I'm having a hard time reviewing this, because it mixes a bugfix with formatting changes (which might not intend to break backwards compatibility, but definitely has the potential to, so it has to be tested).

Could you open two separate pull requests for these things?

PR #1736, opened 2016-02-20
FlxSound: add loopTime, closes #1662

### LINKED ISSUE #1662: Request: Loop music from specific point
So, after working with a couple of different music artists on projects, something that I keep getting asked is: "Can you make this track loop from a specific point, after the first play?"

Apparently, this is something that is pretty common in a lot of games? A track will have a few seconds of 'lead' for the first time it's played, but when it loops, it's meant to skip that section...

I'm not sure the best way to implement, or how we have it work, but it would be a nice feature to add.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/28458687-request-loop-music-from-specific-point?utm_campaign=plugin&utm_content=tracker%2F254189&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F254189&utm_medium=issues&utm_source=github).
</bountysource-plugin>

winterismute:
Also, tiny cleanup of some FlxSound things.
Related for example to request https://github.com/HaxeFlixel/flixel/issues/1662

### DIFF
--- a/flixel/system/FlxSound.hx
+++ b/flixel/system/FlxSound.hx
@@ -99,6 +99,10 @@ class FlxSound extends FlxBasic
 	@:isVar
 	public var looped(get, set):Bool;
 	/**
+	 * In case of looping, the point (in milliseconds) from where to restart the sound when it loops back
+	 */
+	public var loopTime(default, set):Float;
+	/**
 	 * Internal tracker for a Flash sound object.
 	 */
 	private var _sound:Sound;
@@ -144,10 +148,6 @@ class FlxSound extends FlxBasic
 	 * Helper var to prevent the sound from playing after focus was regained when it was already paused.
 	 */
 	private var _alreadyPaused:Bool = false;
-	/**
-	 * In case of looping, from where to restart the sound when it loops back
-	 */
-	private var loopPoint(default, set):Float;
 	
 	/**
 	 * The FlxSound constructor gets all the variables initialized, but NOT ready to play a sound yet.
@@ -173,7 +173,7 @@ class FlxSound extends FlxBasic
 		_volume = 1.0;
 		_volumeAdjust = 1.0;
 		looped = false;
-		loopPoint = 0.0;
+		loopTime = 0.0;
 		_target = null;
 		_radius = 0;
 		_proximityPan = false;
@@ -387,8 +387,9 @@ class FlxSound extends FlxBasic
 	 * Call this function to play the sound - also works on paused sounds.
 	 * 
 	 * @param	ForceRestart	Whether to start the sound over or not.  Default value is false, meaning if the sound is already playing or was paused when you call play(), it will continue playing from its current position, NOT start again from the beginning.
+	 * @param	Time			At which point to start plaing the sound, in milliseconds
 	 */
-	public function play(ForceRestart:Bool = false, Position:Float = 0.0):FlxSound
+	public function play(ForceRestart:Bool = false, Time:Float = 0.0):FlxSound
 	{
 		if (!exists)
 		{
@@ -410,7 +411,7 @@ class FlxSound extends FlxBasic
 		}
 		else
 		{
-			startSound(Position);
+			startSound(Time);
 		}
 		return this;
 	}
@@ -451,34 +452,19 @@ class FlxSound extends FlxBasic
 		return this;
 	}
 	
-	/**
-	 * Sets the sound point at which to restart playing after a loop
-	 * @param	newloopPoint The new position in the sound, in milliseconds
-	 * @return	The updated value in milliseconds
-	 */
-	public function set_loopPoint(newloopPoint:Float):Float
+	private function set_loopTime(newloopTime:Float):Float
 	{
-		if (looped && playing && loopPoint == 0.0)
+		// If the sound is looping, already playing and the looptime WAS 0
+		//	we need to restart the sound and NOT call play on the _sound property
+		// 	with max int as number of loops, but with 0 instead, handling the loop with the event
+		if (looped && playing && loopTime == 0.0)
 		{
-			loopPoint = newloopPoint;
-			var pos:Float = _channel.position;
-			cleanup(false, false);
-			startSound(pos);
-			return loopPoint;
+			loopTime = newloopTime;			
+			restartSound();
+			return loopTime;
 		}
-		loopPoint = newloopPoint;
-		return loopPoint;
-	}
-	
-	/**
-	 * Sets the sound point at which to restart playing after a loop using the more natural 
-	 * 	notation mm:ss.
-	 * @param	min	The minute part of the new position eg. 2 in case of 02:32
-	 * @param	sec	The seconds part of the new position eg. 32 in case of 02:32
-	 */
-	public inline function setloopPointMinSec(min:Int, sec:Int)
-	{
-		loopPoint = (min * 60 + sec) * 1000;
+		loopTime = newloopTime;
+		return loopTime;
 	}
 	
 	/**
@@ -558,16 +544,16 @@ class FlxSound extends FlxBasic
 	/**
 	 * An internal helper function used to attempt to start playing the sound and populate the _channel variable.
 	 */
-	private function startSound(Position:Float):Void
+	private function startSound(Time:Float):Void
 	{
 		if (_sound == null)
 		{
 			return;
 		}
 		
-		var numLoops:Int = looped && loopPoint == 0.0 ? FlxMath.MAX_VALUE_INT : 0;
+		var numLoops:Int = looped && loopTime == 0.0 ? FlxMath.MAX_VALUE_INT : 0;
 		
-		time = Position;
+		time = Time;
 		_paused = false;
 		_channel = _sound.play(time, numLoops, _transform);
 		if (_channel != null)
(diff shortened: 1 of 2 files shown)


flixel/system/FlxSound.hx
```diff
@@ -144,6 +144,10 @@ class FlxSound extends FlxBasic
 	 * Helper var to prevent the sound from playing after focus was regained when it was already paused.
 	 */
 	private var _alreadyPaused:Bool = false;
+	/**
+	 * In case of looping, from where to restart the sound when it loops back
+	 */
+	private var loopPoint(default, set):Float;
```
Gama11: This should be `public` (otherwise what's the point of making it a property with a setter?) and declared after `looping`.

Gama11: I'm also not sure I like the name... How about `loopTime` (there already is a `time`) or `loopPosition`?

flixel/system/FlxSound.hx
```diff
@@ -442,11 +447,41 @@ class FlxSound extends FlxBasic
...
+		cleanup(autoDestroy, true);
 		return this;
 	}
 	
 	/**
+	 * Sets the sound point at which to restart playing after a loop
+	 * @param	newloopPoint The new position in the sound, in milliseconds
+	 * @return	The updated value in milliseconds
+	 */
+	public function set_loopPoint(newloopPoint:Float):Float
```
Gama11: Setters should always be private.

flixel/system/FlxSound.hx
```diff
@@ -442,11 +447,41 @@ class FlxSound extends FlxBasic
...
-		cleanup(autoDestroy, true, true);
+		cleanup(autoDestroy, true);
 		return this;
 	}
 	
 	/**
+	 * Sets the sound point at which to restart playing after a loop
+	 * @param	newloopPoint The new position in the sound, in milliseconds
+	 * @return	The updated value in milliseconds
+	 */
```
Gama11: There's no point to having documentation for setters, the docs should go to the property.

flixel/system/FlxSound.hx
```diff
@@ -704,22 +738,19 @@ class FlxSound extends FlxBasic
...
 	private inline function set_looped(loop:Bool):Bool
 	{
+		looped = loop;
+		
 		// If we're going from looping to not looping while playing,
 		// the channel needs to be updated so it won't loop next time.
-		
 		if (!loop && looped && playing)
 		{
 			looped = loop;
```
Gama11: This line is now redundant.

flixel/system/frontEnds/SoundFrontEnd.hx
```diff
@@ -95,6 +96,7 @@ class SoundFrontEnd
 		music.volume = Volume;
 		music.persist = true;
 		music.group = (Group == null) ? defaultMusicGroup : Group;
+		music.set_loopPoint(LoopPoint);
```
Gama11: Should never directly call a property setter, only access it via the property.

flixel/system/FlxSound.hx
```diff
@@ -442,11 +447,41 @@ class FlxSound extends FlxBasic
...
+		return loopPoint;
+	}
+	
+	/**
+	 * Sets the sound point at which to restart playing after a loop using the more natural 
+	 * 	notation mm:ss.
+	 * @param	min	The minute part of the new position eg. 2 in case of 02:32
+	 * @param	sec	The seconds part of the new position eg. 32 in case of 02:32
+	 */
+	public inline function setloopPointMinSec(min:Int, sec:Int)
```
Gama11: I think we can do without this for now.

flixel/system/FlxSound.hx
```diff
@@ -383,7 +388,7 @@ class FlxSound extends FlxBasic
 	 * 
 	 * @param	ForceRestart	Whether to start the sound over or not.  Default value is false, meaning if the sound is already playing or was paused when you call play(), it will continue playing from its current position, NOT start again from the beginning.
 	 */
-	public function play(ForceRestart:Bool = false):FlxSound
+	public function play(ForceRestart:Bool = false, Position:Float = 0.0):FlxSound
```
Gama11: Adding `Position` here creates the awkward situation that the argument is not settable from the `SoundFrontEnd` function. We really should add any more arguments to those though if possible though, they already have too many.

It would be much better to implement a setter for `time`. That should be done separately though.

flixel/system/FlxSound.hx
Gama11: `loopTime` probably makes the most sense?

flixel/system/FlxSound.hx
```diff
@@ -442,11 +447,41 @@ class FlxSound extends FlxBasic
...
+	{
+		if (looped && playing && loopPoint == 0.0)
+		{
+			loopPoint = newloopPoint;
+			var pos:Float = _channel.position;
+			cleanup(false, false);
+			startSound(pos);
+			return loopPoint;
+		}
+		loopPoint = newloopPoint;
```
Gama11: This could be moved before the `if` and save a few lines - it's executed in either case.

flixel/system/FlxSound.hx
Gama11: Hm.. at least call it `Time` then for consistency with the public property.

flixel/system/FlxSound.hx
Gama11: Right... I'm not sure I really understand that if-logic anyway. Why is it necessary?

winterismute: I think: you need to spot the case in which the sound is playing, is looping, and has no loopPoint (starts from the beginning). In that case, it meant it was played (startSound()) with MAX_INT as number of loops, which means that it will never stop. In this case, you have to update the channel to set the num of loops to be 0, so that after each completed play "stopped" is called and you can call then "play" with the time parameter. It' similar to what "set_looped" does.

Gama11: Hm, so it's a matter of having to manually manage the looping now... I wonder why we don't always do that then, seems less complicated. Either way, you should:

- create a helper method that handles this to avoid the redundancy with set_looped()
- add a comment to the call here explaning why it's necessary since it doesn't seem super-intuitive

flixel/system/FlxSound.hx
```diff
@@ -382,8 +387,9 @@ class FlxSound extends FlxBasic
 	 * Call this function to play the sound - also works on paused sounds.
 	 * 
 	 * @param	ForceRestart	Whether to start the sound over or not.  Default value is false, meaning if the sound is already playing or was paused when you call play(), it will continue playing from its current position, NOT start again from the beginning.
+	 * @param	Time			At which point to start plaing the sound, in milliseconds
 	 */
-	public function play(ForceRestart:Bool = false):FlxSound
+	public function play(ForceRestart:Bool = false, Time:Float = 0.0):FlxSound
```
gamedevsam: This ```time``` variable should be named ```delay``` to better reflect its behavior.

flixel/system/frontEnds/SoundFrontEnd.hx
```diff
@@ -79,8 +79,9 @@ class SoundFrontEnd
 	 * @param	Volume		How loud the sound should be, from 0 to 1.
 	 * @param	Looped		Whether to loop this music.
 	 * @param	Group		The group to add this sound to.
+	 * @param	LoopPoint	Point at which the sound will restart when looping back (in case of looping)
```
gamedevsam: Should be ```LoopTime```

flixel/system/FlxSound.hx
```diff
@@ -442,10 +448,25 @@ class FlxSound extends FlxBasic
...
+		cleanup(autoDestroy, true);
 		return this;
 	}
 	
+	private function set_loopTime(newloopTime:Float):Float
+	{
+		// If the sound is looping, already playing and the looptime WAS 0
+		//	we need to restart the sound and NOT call play on the _sound property
+		// 	with max int as number of loops, but with 0 instead, handling the loop with the event
+		if (looped && playing && loopTime == 0.0)
```
gamedevsam: This function could be condensed for brevity:
```
var requiresRestart:Bool = looped && playing && loopTime == 0.0;
loopTime = newloopTime;
if(requiresRestart)
{
	restartSound();
}
return loopTime;
```

flixel/system/FlxSound.hx
```diff
@@ -523,16 +544,16 @@ class FlxSound extends FlxBasic
...
 		if (_sound == null)
 		{
 			return;
 		}
 		
-		var numLoops:Int = looped && Position == 0 ? FlxMath.MAX_VALUE_INT : 0;
+		var numLoops:Int = looped && loopTime == 0.0 ? FlxMath.MAX_VALUE_INT : 0;
 		
-		time = Position;
+		time = Time;
```
gamedevsam: This looks weird, perhaps the variable could be renamed ```startTime```.

flixel/system/FlxSound.hx
gamedevsam: This variable should be renamed ```StartTime``` for clarity.

winterismute:
I have fixed the possibility of looping using the second argument of Sound.play(), now we always use the event. In native builds has 0 effect on performance, not sure about flash, although it leads to nicer code.
Fixed the naming as pointed out by @gamedevsam too.

Gama11:
Thanks!

PR #1740, opened 2016-02-23
Fix VCR to record simultaneous keys+mouse

IBwWG:
I'm not 100% sure why github also records my pull request #1 here, which was just catching my fork up to the main dev branch of flixel.  I suppose it shouldn't create any conflicts though.  :)

Nor why it says "GimmickyApps added some commits 3 hours ago" when I *just* submitted it now.


GimmickyApps:
I suppose if you pull this is will make part of the logic that I added in the other PR unnecessary.  But I'll have to check.

BTW @Gama11 now that I see how [the recorder changed back in 2013](https://github.com/HaxeFlixel/flixel/commit/b5adf6891c26e177c7902aa76872d2854bbb3085#commitcomment-16263605), I believe that the "new format" I talk about is actually the *even older* format that got broken at that point, i.e. output starting becoming needlessly verbose.  :)

Gama11:
Seems like a good candidate for a unit test.

GimmickyApps:
True.  Apparently there aren't even any for VCR yet.

Gama11:
I don't think anyone has even touched the vcr code since we have unit tests. :D

GimmickyApps:
There are a few now.  With the windows target, this new test passes with this PR and fails without it.  Hopefully Mr. Travis agrees.

Gama11:
Wow, _several_ even. :D

Please stick to the flixel code style - `{` in a new line and no extra spaces after `(` and before `)` in function calls.

GimmickyApps:
;)  Oh no...I thought I had seen a shift away from those styles in Flixel code in particular.  I've been consciously changing everywhere to look like this.  I guess I should have just gone by the official style guide...

GimmickyApps:
Although, I wasn't even doing it that way consistently, either, it seems.  Well, hopefully now it's correct.

GimmickyApps:
Yay!  Working tests.  Even on neko.

gamedevsam:
This looks good to me, @Gama11 your call.

IBwWG:
What conflicts need resolving?  I had just merged changes that were in HF's dev so I could keep up-to-date *without* there arising conflicts...

IBwWG:
Oops, wait.  I need to undo that.  I keep forgetting I have the two different VCR PRs waiting.

IBwWG:
I think the conflict is very minor.  I tried to fix it myself somehow but I got a different result than I thought I had seen in the merge editor.

```
import flash.display.Bitmap;
import flash.display.BitmapData;
import flash.display.Sprite;
import flash.display.Stage;
import flash.events.Event;
import flash.events.MouseEvent;
import flash.geom.Matrix;
import flash.geom.Point;
import flash.Lib;
import flash.ui.Mouse;
import flash.Vector;
import flixel.FlxG;
import flixel.input.FlxInput.FlxInputState;
import flixel.input.IFlxInputManager;
import flixel.input.mouse.FlxMouseButton;
import flixel.system.FlxAssets;
import flixel.system.replay.MouseRecord;
import flixel.util.FlxDestroyUtil;
```

If we have these, that should cover it (above the compiler directive.)

IBwWG:
Um, gee.  Details = "no analysis present"?  Am I missing something with codeclimate here?

IBwWG:
Yay, fixed codeclimate anyway.  Didn't fix my branch problem, but oh well.  Hopefully this is good to merge in now.

Gama11:
Sorry for the delay here.

The logic change in `recordFrame()` seems sound to me - in fact, that's how it worked back in AS3: 
https://github.com/AdamAtomic/flixel/blob/master/org/flixel/system/FlxReplay.as#L153

I suspect the addition of mouse / keyboard conditonals required a few changes and the logic got mixed up in the process.

There are still conflicts here that prevent me from merging this (probably caused by some code style fixes I did to get down to 0 checkstyle warnings). Two things you could do:
- create a new branch based on current dev and reapply these changes there and create a new pull request (probably the easier solution, considering there aren't a lot of changes here)
- rebase against current dev and force-push

I don't really like the use of `Std.random()` in a unit test... That makes the tests non-deterministic, meaning they could fail during one execution and succeed in another, without any changes, making it difficult to reproduce an issue. If you do need randomness, you could at least use a `FlxRandom` instance and initialize it with a constant seed to get determinsitic results.

IBwWG:
I was trying to avoid magic numbers.  Apparently there is some debate.

I guess in this case, seeding for determinism wouldn't do much more than hardcoding a prerandomized magic number, right?  So if you're against random, I might as well just put something like 300 frames and be done with it.  Then the test shows exactly one case, as it would with a deterministic random.

Gama11:
That would be the other way of doing it, yes. Either way, I think unit tests should be deterministic.

PR #1746, opened 2016-02-26
Make analog sticks emit "fake" digital actions for left/right/up/down…

larsiusprime:
… movement, and add ID's that make it simple to use them. Makes it easy to use an Analog stick as if it were a DPAD, a very popular input pattern.

This behavior was already implemented on legacy, but the API for it was buried. I added top level FlxGamepadInputID's for the digital events, and refactored things so that Next can emit these events as well.
Also, I added a "value" property to the FlxGamepadButton for storing last frame's analog input value, which should remove the requirement for FlxActions to call check() every frame in order to get accurate JUST_MOVED / JUST_STOPPED behavior.

(This pull request helps the FlxAction PR, but can be considered in isolation and does not depend on it)


flixel/input/gamepad/FlxGamepad.hx
```diff
@@ -261,10 +269,10 @@ class FlxGamepad implements IFlxDestroyable
 		manager = null;
 		
 		#if FLX_JOYSTICK_API
-		hat = FlxDestroyUtil.put(hat);
+		hat  = FlxDestroyUtil.put(hat);
```
Gama11: Fyi, I'm not a fan of this formatting for vertical alignment. That's just a lot of extra maintenance effort with little benefit.

flixel/input/gamepad/FlxGamepadAnalogStick.hx
```diff
@@ -32,7 +32,7 @@ class FlxGamepadAnalogStick
 	/**
 	 * when analog inputs are received, how to process them digitally
 	 */
-	public var mode(default, null):FlxAnalogToDigitalMode = ONLY_ANALOG;
+	public var mode(default, null):FlxAnalogToDigitalMode = BOTH;
```
Gama11: Changing the default here seems like a breaking change.

flixel/input/gamepad/FlxGamepadAnalogStick.hx
```diff
@@ -42,12 +42,12 @@ class FlxGamepadAnalogStick
...
 			return;
 
-		mode = (settings.mode != null) ? settings.mode : ONLY_ANALOG;
+		mode = (settings.mode != null) ? settings.mode : BOTH;
 		rawUp = (settings.up != null) ? settings.up : -1;
 		rawDown = (settings.down != null) ? settings.down : -1;
 		rawLeft = (settings.left != null) ? settings.left : -1;
 		rawRight = (settings.right != null) ? settings.right : -1;
-		digitalThreshold = (settings.threshold != null) ? settings.threshold : -1;
+		digitalThreshold = (settings.threshold != null) ? settings.threshold : 0.5;
```
Gama11: Changing the default here seems like a breaking change

flixel/input/gamepad/FlxGamepadButton.hx
```diff
@@ -31,4 +31,9 @@ class FlxGamepadButton extends FlxInput<Int>
 		super.press();
 	}
 	#end
+	
+	/**
+	 * Optional analog value, so we can check when the value has changed from the last frame
+	 */
+	public var value:Float = 0;
```
Gama11: Variables are declared at the top of the class.

flixel/input/gamepad/FlxGamepadInputID.hx
```diff
@@ -81,14 +81,24 @@ abstract FlxGamepadInputID(Int) from Int to Int
...
+	var LEFT_STICK_DIGITAL_LEFT        = 33;
+	
+	var RIGHT_STICK_DIGITAL_UP         = 34; 
+	var RIGHT_STICK_DIGITAL_RIGHT      = 35;
+	var RIGHT_STICK_DIGITAL_DOWN       = 36;
+	var RIGHT_STICK_DIGITAL_LEFT       = 37;
+	
 	/**an extra digital button that doesn't fit cleanly into the universal template**/
-	var EXTRA_0              = 30;
+	var EXTRA_0              = 46;
```
Gama11: Changing these IDs is a breaking change because this abstract allows `from Int` and `to Int`.

flixel/input/gamepad/FlxGamepadManager.hx
```diff
@@ -453,58 +453,18 @@ class FlxGamepadManager implements IFlxInputManager
...
 			var isForMotion = gamepad.mapping.isAxisForMotion(i);
+			
 			if (!isForStick && !isForMotion)
 			{
 				// in legacy this returns a (-1, 1) range, but in flash/next it
 				// returns (0,1) so we normalize to (0, 1) for legacy target only
 				newAxis[i] = (newAxis[i] + 1) / 2;
 			}
-			else if (isForStick)
+			else if(isForStick)
```
Gama11: Flixel's code style always has a space after `if`, `switch`, `while` and `for`.

flixel/input/gamepad/FlxGamepadManager.hx
```diff
@@ -453,58 +453,18 @@ class FlxGamepadManager implements IFlxInputManager
...
-						else if (i == stick.y) digitalButton = stick.rawDown;
-						handleButtonUp(device, digitalButton);
-					}
-					
-					if (stick.mode == ONLY_DIGITAL)
-					{
-						//still haven't figured out how to suppress the analog inputs properly. Oh well.
-					}
-				}
+				var newVal = newAxis[i];
```
Gama11: These temp variables seem a little unnecessary?

flixel/input/gamepad/id/LogitechID.hx
```diff
@@ -55,7 +52,9 @@ class LogitechID
 	//TODO: Someone needs to look this up and define it! (NOTE: not all logitech controllers have this)
 	public static inline var LOGITECH:Int = -5;
 	
-	public static var LEFT_ANALOG_STICK(default, null) = new FlxGamepadAnalogStick(0, 1);
-	public static var RIGHT_ANALOG_STICK(default, null) = new FlxGamepadAnalogStick(2, 3);
 #end
+
+public static var LEFT_ANALOG_STICK(default, null)  = new FlxGamepadAnalogStick(0, 1, {up:24, down:25, left:26, right:27});
```
Gama11: Why is there no indent here?

flixel/input/gamepad/FlxGamepad.hx
```diff
@@ -606,6 +614,61 @@ class FlxGamepad implements IFlxDestroyable
 		return getAnalogYAxisValue(Stick);
 	}
 
+	@:allow(flixel.input.gamepad.FlxGamepadManager)
+	private function handleAxisMove(axis:Int, newVal:Float, oldVal:Float)
```
Gama11: Abbreviating `value` to `val` seems a bit silly, barely saves any chars and definitely doesn't help readability.

flixel/input/gamepad/FlxGamepad.hx
```diff
@@ -606,6 +614,61 @@ class FlxGamepad implements IFlxDestroyable
...
+	{
+		newVal = applyAxisFlip(newVal, axis);
+		oldVal = applyAxisFlip(oldVal, axis);
+		
+		//check to see if we should send digital inputs as well as analog
+		var stick:FlxGamepadAnalogStick = this.getAnalogStickByAxis(axis);
```
Gama11: Unnecessary `this` here.

flixel/input/gamepad/FlxGamepad.hx
```diff
@@ -606,6 +614,61 @@ class FlxGamepad implements IFlxDestroyable
...
+		if (stick.mode == ONLY_DIGITAL || stick.mode == BOTH)
+		{
+			var neg = stick.digitalThreshold * -1;
+			var pos = stick.digitalThreshold;
+			var digitalButton = -1;
+			
+			trace("axis(" + axis + ") val = " + newVal);
```
Gama11: Leftover trace.

flixel/input/gamepad/FlxGamepad.hx
```diff
@@ -606,6 +614,61 @@ class FlxGamepad implements IFlxDestroyable
 		return getAnalogYAxisValue(Stick);
 	}
 
+	@:allow(flixel.input.gamepad.FlxGamepadManager)
```
Gama11: This is unnecessary, the class already has `@:allow(flixel.input.gamepad)`.

flixel/input/gamepad/FlxGamepad.hx
```diff
@@ -606,6 +614,61 @@ class FlxGamepad implements IFlxDestroyable
...
+		if (stick.mode == ONLY_DIGITAL || stick.mode == BOTH)
+		{
+			var neg = stick.digitalThreshold * -1;
+			var pos = stick.digitalThreshold;
+			var digitalButton = -1;
+			
+			trace("axis(" + axis + ") val = " + newVal);
+			
+			//pressed/released for digital LEFT/UP
+			if (newVal < neg && oldVal >= neg)
```
Gama11: The left/up and right/down blocks seem a little redundant and could probably be abstracted into a helper function?

flixel/input/gamepad/id/MFiID.hx
```diff
@@ -25,8 +25,8 @@ class MFiID
 	public static inline var DPAD_LEFT:Int = 19;
 	public static inline var DPAD_RIGHT:Int = 20;
 	
-	public static var LEFT_ANALOG_STICK(default, null) = new FlxGamepadAnalogStick(0, 1);
-	public static var RIGHT_ANALOG_STICK(default, null) = new FlxGamepadAnalogStick(2, 3);
+	public static var LEFT_ANALOG_STICK (default, null) = new FlxGamepadAnalogStick(0, 1, {up:21, down:22, left:23, right:24});
```
Gama11: Where do these up / down / left / right IDs come from? Just arbitrary?

flixel/input/gamepad/FlxGamepadAnalogStick.hx
larsiusprime: It's not strictly speaking breaking; previously it was reporting only analog actions, this would make it report both analog and digital. As far as I can tell, the analog-to-digital actions weren't even mapped up correctly and thus weren't being reported. 

The only difference users should see is new digital actions being reported, and will only affect things if they choose to poll for them.

flixel/input/gamepad/FlxGamepadAnalogStick.hx
larsiusprime: See above note.

flixel/input/gamepad/FlxGamepadInputID.hx
larsiusprime: Okay, so just move the new ids to the end of the value list?

flixel/input/gamepad/id/MFiID.hx
larsiusprime: They are "arbitrary" in that they don't match any physical inputs, but are specifically chosen to not conflict with any existing button ID's. They are basically new virtual digital button inputs that represent the analog stick being pushed past the digital threshold in each direction.

flixel/input/gamepad/FlxGamepadInputID.hx
Gama11: Seems sensible.

flixel/input/gamepad/FlxGamepadManager.hx
larsiusprime: cruft from a refactor, removing...

flixel/input/gamepad/id/MFiID.hx
Gama11: In that case, wouldn't it be better if they weren't so close to the actual IDs? Maybe have them all start at 100-something an an indicator that they aren't real IDs?

flixel/input/gamepad/FlxGamepad.hx
larsiusprime: result of a refactor, those were the old variable names, updating now.

flixel/input/gamepad/id/MFiID.hx
larsiusprime: Sure, that's totally fine, as long as you're okay with these other two considerations:

- The "FAKE" id's for legacy are adjacent, we'd be departing from that convention
- These ID's ultimately map to array indeces, so it'd implicitly create a lot of empty space in the underyling array structures

flixel/input/gamepad/FlxGamepadAnalogStick.hx
larsiusprime: To clarify in this situation: since the default setting everywhere was to only report Analog, I'm pretty sure nobody was actually accessing this behavior. Accessing analog-to-digital virtual stick directions was buried in the API, essentially undocumented, and only worked on legacy. This PR is basically the first time the feature's actually been exposed to daylight in any practical sense.

flixel/input/gamepad/id/MFiID.hx
Gama11: The array indices thing is a good argument against my suggestion. I think we might even have had the exact same conversation about those other fake IDs a while ago.. :D

larsiusprime:
@Gama11: I think I got everything you mentioned. Anything else I should hit before squashing?

Gama11:
The mapping classes still use `if(` instead of `if (`. :P

adrianulima:
@Gama11 Do you have any FlashDevelop or other text editor macro that check this code styles?
Maybe we could have one to execute before make a PR, this would avoid some common errors like:
`if(`-`if (` and `if () {`- `if ()\n{`, same for `while`, `for`, `else` ...

Gama11:
@adrianulima No.. Somebody suggested using [checkstyle](https://github.com/adireddy/haxe-checkstyle) a while ago to check the code style on Travis.. However I never looked into how well that tool works, and I also dislike idea of having failed Travis builds because of code stlye issues... What's really needed is an auto-formatter for Haxe.

@larsiusprime Squash? :)

larsiusprime:
@Gama11 here:
https://github.com/HaxeFlixel/flixel/pull/1752

Tried to force push but github two factor auth makes that hard for me to do.

Gama11:
Actually, checkstyle seems like a great tool, I should've looked into that way earlier...

PR #1755, opened 2016-03-01
Memory usage optimizations and OCD style cleanup.

gamedevsam:


gamedevsam:
This is causing an error in Flash, will look into it.

gamedevsam:
It's hard to say, the changes I made are focused on reducing runtime garbage generation, not necessarily to improve performance per se (just trying to reduce GC hit).

flixel/FlxCamera.hx
```diff
@@ -103,19 +103,19 @@ class FlxCamera extends FlxBasic
 	/**
 	 * Lower bound of the cameras scroll on the x axis
 	 */
-	public var minScrollX:Null<Float>;
+	public var minScrollX:Float = FlxMath.MIN_VALUE_FLOAT;
```
Gama11: I don't think we can change these types for now, that would be a fairly big breaking change.

flixel/math/FlxMath.hx
```diff
@@ -44,6 +44,11 @@ class FlxMath
 	 */
 	public static inline var EPSILON:Float = 0.0000001;
 	
+	private static var _cachedX:Float = 0;
```
Gama11: Caching floats? Ouch.. :/

flixel/FlxGame.hx
```diff
@@ -758,8 +759,6 @@ class FlxGame extends Sprite
 		}
 		FlxArrayUtil.clearArray(FlxG.swipes);
 		#end
-		
-		filters = filtersEnabled ? _filters : null;
```
Gama11: I think there was a reason why this happens every frame.. I think the contents of the array can change without a `setFilters()` call, and that wouldn't trigger OpenFL's setter.

flixel/math/FlxPoint.hx
```diff
@@ -45,8 +45,10 @@ class FlxPoint implements IFlxPooled
 		return point;
 	}
 	
-	public var x(default, set):Float = 0;
-	public var y(default, set):Float = 0;
+	public var _x:Float = 0; // raw access for maximum speed
```
Gama11: Surely these are meant to be private?

flixel/math/FlxPoint.hx
```diff
@@ -96,6 +98,20 @@ class FlxPoint implements IFlxPooled
 	}
 	
 	/**
+	 * Set the coordinates of this point (inline).
+	 * 
+	 * @param	X	The X-coordinate of the point in space.
+	 * @param	Y	The Y-coordinate of the point in space.
+	 * @return	This point.
+	 */
+	public inline function setInline(X:Float, Y:Float):FlxPoint
```
Gama11: Not a fan of this.. How big of a speed boost is this really?

flixel/FlxGame.hx
gamedevsam: Seems like we should have a function called updateFilters or something then, because this is silly.

flixel/math/FlxMath.hx
gamedevsam: It saves a boatload of allocs.

flixel/FlxGame.hx
Gama11: You mean a public one the user has to call?

flixel/math/FlxPoint.hx
gamedevsam: No, they are meant to be public, for inline set.

flixel/math/FlxPoint.hx
gamedevsam: It's definitely noticeable in my profiles, although we could omit this in favor of direct inline variable access (via ```_x``` and ```_y```).

flixel/math/FlxPoint.hx
Gama11: I don't support this at all then, it creates a really nasty / confusing API.

flixel/FlxCamera.hx
gamedevsam: Agreed.

flixel/math/FlxPoint.hx
Gama11: It seems like the real issue here is the variables having setters for `FlxCallbackPoint`. This is just a really nasty workaround for that...

flixel/FlxGame.hx
gamedevsam: Actually I don't think we need a new function, we should just expect that users call ```setFilters``` if they change something in that array.

Gama11: Maybe, but again: that's a breaking change with the potential for stuff to silently stop working. Really worth it here? Doubt it...

flixel/math/FlxPoint.hx
gamedevsam: I made them private to avoid changing the public API.

flixel/FlxGame.hx
gamedevsam: Yes... it is arguable whether it's worth making this change. Maybe someone else can weigh in on it?

flixel/FlxCamera.hx
```diff
@@ -1198,6 +1200,7 @@ class FlxCamera extends FlxBasic
 	public function setFilters(filters:Array<BitmapFilter>):Void
 	{
 		_filters = filters;
+		flashSprite.filters = filtersEnabled ? _filters : null;
```
gamedevsam: Same issue as https://github.com/HaxeFlixel/flixel/pull/1755/files#r54653375

flixel/FlxCamera.hx
```diff
@@ -799,9 +801,8 @@ class FlxCamera extends FlxBasic
 	 */
 	public function updateScroll():Void
 	{
-		//Make sure we didn't go outside the camera's bounds
-		scroll.x = FlxMath.bound(scroll.x, minScrollX, (maxScrollX != null) ? maxScrollX - width : null);
```
Gama11: I think this needs to be reverted as well?

gamedevsam:
I'm closing this PR in favor of creating separate, more granular (safer) PRs.

Gama11:
Good idea!

Tiago-Ling:
Any chance to take a look at FlxQuadTree allocations too?

Tiago Ling Alexandre
Tel: +55 41 8819-3191


2016-03-02 15:00 GMT-03:00 Gama11 <notifications@github.com>:

PR #1760, opened 2016-03-02
Safe alloc optimizations + usage of flixel math objects in PixelPerfectCollision function

gamedevsam:


flixel/input/FlxPointer.hx
```diff
@@ -15,6 +15,7 @@ class FlxPointer
 	
 	private var _globalScreenX:Int = 0;
 	private var _globalScreenY:Int = 0;
+	private static var _cachedPoint:FlxPoint = new FlxPoint();
```
Gama11: Should maybe just be `_point`? That's how it's called in most other classes.

flixel/util/FlxCollision.hx
```diff
@@ -21,15 +24,17 @@ import flixel.tile.FlxTileblock;
 class FlxCollision 
 {
 	// Optimization: Local static vars to reduce allocations
-	private static var pointA:Point = new Point();
```
Gama11: This is just code cleanup, or does it have any performance gains too?

gamedevsam: I noticed a considerable improvement in PixelPerfectCollision demo (45fps vs 55fps, mostly gains from using faster Flx classes).

There is also some work to reduce unnecessary allocations in FlxMath functions.

flixel/input/FlxPointer.hx
gamedevsam: I'm trying to make it obvious that these are static cached values (to prevent confusion or accidental deletion), but I am open to changing this to better fit existing conventions if you'd prefer.

flixel/util/FlxCollision.hx
Gama11: That's interesting, I didn't know Flixel point / rects were faster than their OpenFL counterparts.

gamedevsam: They are in some cases due to aggressive inlining.

flixel/input/FlxPointer.hx
Gama11: I was specifically talking about this `FlxPoint` var here. I agree that the floats would be <strike>quite</strike> even more confusing without the prefix.

gamedevsam: I'd rather stick with the ```_cached``` convention, if there are a ```_point``` variables that are static for caching purposes, we should rename it to ```_cachedPoint```.

Gama11:
I partially merged this (mostly the parts that didn't have merge conflicts: `FlxCollision` and `FlxPointer`). I still really dislike the readability implications of caching basic types and didn't merge that part...

PR #1761, opened 2016-03-02
Removing color hack + preventing filter changes every frame

gamedevsam:
We need to ensure this doesn't cause massive breakages, but I support this change for a major patch.

### DIFF
--- a/flixel/FlxCamera.hx
+++ b/flixel/FlxCamera.hx
@@ -789,8 +789,6 @@ class FlxCamera extends FlxBasic
 		updateFade(elapsed);
 		updateShake(elapsed);
 		
-		flashSprite.filters = filtersEnabled ? _filters : null;
-		
 		updateFlashSpritePosition();
 	}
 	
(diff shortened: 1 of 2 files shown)


flixel/FlxCamera.hx
```diff
@@ -1256,10 +1255,8 @@ class FlxCamera extends FlxBasic
 			{
 				return;
 			}
-			// This is temporal fix for camera's color
```
Gama11: What was this even for?

Gama11:
What's a "major patch"? SemVer is `Major.Minor.Patch`, "major patch" is an oxymoron. :D

What's the benefit of this change? Does the setter cause allocations you can see in hxscout?

flixel/FlxCamera.hx
gamedevsam: I don't know, seemed very hacky to me, so I removed it. As far as I can tell everything is running fine.

gamedevsam:
Yes, the set causes a lot of Array allocations on the OpenFL DisplayObject.

http://screencast.com/t/mdy5AXue

flixel/FlxCamera.hx
Gama11: Wow, this code was _old_! https://github.com/HaxeFlixel/flixel/commit/227a4647b8b1f38b11eacc98bf1be57203c1381e#diff-9ef1aac582dab97c9bde1a8fb693a1b7R845

PR #1762, opened 2016-03-02
Vector optimizations

gamedevsam:
I propose optimizing ```FlxVector``` to avoid overhead of ```FlxPoint``` due to requirements of ```FlxCallbackPoint```. This gives us an easy way to optimize for performance (by changing ```FlxPoint``` to ```FlxVector``` in performance critical portions of engine), while avoiding any API changes to end users.

### DIFF
--- a/flixel/math/FlxVector.hx
+++ b/flixel/math/FlxVector.hx
@@ -10,7 +10,13 @@ class FlxVector extends FlxPoint
 	public static inline var EPSILON:Float = 0.0000001;
 	public static inline var EPSILON_SQUARED:Float = EPSILON * EPSILON;
 	
-	private static var _pool = new FlxPool<FlxVector>(FlxVector);
+	/**
+	 * Pooling mechanism to minimimze runtime allocations.
+	 * 
+	 * TODO: We need more evidence to decide whether liberal use of pooled objects
+	 *       is worthwhile, or whether static vars are a better alternative. - gamedevsam
+	 */
+	public static var pool(default, never):FlxPool<FlxVector> = new FlxPool<FlxVector>(FlxVector);
 	
 	private static var _vector1:FlxVector = new FlxVector();
 	private static var _vector2:FlxVector = new FlxVector();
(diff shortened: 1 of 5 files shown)


gamedevsam:
I also propose a new code convention:

For non-static variables that should be pooled, and recycled upon destruction of container object, use:
```
var pooledPoint = FlxPoint.get();
```

Don't forget to release it inside __destroy__:
```
function destroy()
{
	super.destroy();
	pooledPoint = FlxDestroyUtil.put(pooledPoint);
}
```

For static variables that are used for caching purposes, I propose we allocate them with __new__ to future highlight the differences (cached static variables should never get destroyed):

```
static var _cachedPoint = new FlxPoint();
```

Beeblerox:
i'd rather merged FlxVector in FlxPoint, since FlxPoint is much more familiar concept to flixel users. And added FlxVector typedef (which will just point to FlxPoint class). What do you think about it?

Tiago-Ling:
+1 to that. There shouldn't be the need to use a specialized class for
vector math. I'd rather use the FlxVector name instead of FlxPoint, but
since it's a breaking change, merging it inside FlxPoint makes sense.

Off-topic but would facilitate a lot for vector operations on sprites and
objects: having a 'position' FlxPoint variable instead of separate x and y.
We could keep these just as getters and setters for backward compatibility
(users could keep using x and y normally)

Gama11:
Let's move this discussion to a new issue. This is completely off-topic.

Tiago-Ling:
I created #1856 to discuss these things :)

flixel/math/FlxVector.hx
```diff
@@ -10,6 +10,13 @@ class FlxVector extends FlxPoint
 	public static inline var EPSILON:Float = 0.0000001;
 	public static inline var EPSILON_SQUARED:Float = EPSILON * EPSILON;
 	
+	/**
+	 * Pooling mechanism to minimimze runtime allocations.
+	 * 
+	 * TODO: We need more evidence to decide whether liberal use of pooled objects
```
Gama11: Putting notes like this in public docs doesn't seem like a good idea to me, doesn't exactly induce confidence in API useres.

Gama11:
I'm still not sure how I feel about this in principle, but either way, there are merge conflicts by now.

PR #1763, opened 2016-03-03
Correct Misspelling

buckle2000:


Gama11:
Pull Requests should generally target the dev branch. Master is always up to date with the latest haxelib release, which it wouldn't be anymore if we merged PRs to it.

In this case however, you might actually want to target the `patch` branch (going to be used for patch releases like `4.0.1`), since this only affects docs.

PR #1764, opened 2016-03-05
Fixed Bug

buckle2000:
Fixed Null Pointer Exception when used with FlxNapeTileMap.hx:103;
coords may be null before this fix, result in program halt.
This may cause problem in other places if they expect null.

### DIFF
--- a/flixel/tile/FlxTilemap.hx
+++ b/flixel/tile/FlxTilemap.hx
@@ -42,10 +42,10 @@ class GraphicAutoAlt extends BitmapData {}
 class FlxTilemap extends FlxBaseTilemap<FlxTile>
 {
 	/** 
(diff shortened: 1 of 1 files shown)


Gama11:
This must be based on an outdated version of `FlxTilemap`? There are "1,260 additions and 1,265 deletions." in the diff.

Also, from your description it sounds like this issue was already fixed on flixel-addons: https://github.com/HaxeFlixel/flixel-addons/issues/233

MSGhero:
Prolly just converted all the tabs into spaces or CRLFs got messed up.

Gama11:
Hm yeah, here it is with `?w=1`: https://github.com/HaxeFlixel/flixel/pull/1764/files?w=1

PR #1765, opened 2016-03-07
Implementing FlxTIleFilter logic

SeiferTim:
I tried to implement this before, but, there was something wrong with my repo and I never figured out how to fix it - and then promptly forgot about it.

So, here it is again, hopefully better than before: FlxTileFilter.

Instead of setting a single callback for a tile/range of tiles in a FlxTilemap, you can now set multiple filters to be triggered by different Classes.

You can set it up so that your Enemy class can pass through certain tiles, while your Player cannot, and much more.

Usage is `myTilemap.addTileFilter(5, MyPlayerObject, FlxObject.UP, collisionCallback, processCallback, 2);`

This will set tiles 5 and 6 to have the `MyPlayerObject` collide with the top of the tile only, call `processCallback` to do some pre-collision checks, and then `collisionCallback` if the `processCallback` returns `true`.

You can use multiple `addTileFilter`s on the same Tilemap.

Use `removeTileFilter` to remove a specific filter.

### DIFF
--- a/flixel/tile/FlxBaseTilemap.hx
+++ b/flixel/tile/FlxBaseTilemap.hx
@@ -5,6 +5,7 @@ import flixel.group.FlxGroup.FlxTypedGroup;
 import flixel.math.FlxPoint;
 import flixel.math.FlxRect;
 import flixel.system.FlxAssets.FlxTilemapGraphicAsset;
+import flixel.tile.FlxTile.FlxTileFilter;
 import flixel.util.FlxArrayUtil;
 import openfl.Assets;
 using StringTools;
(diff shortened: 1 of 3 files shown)


flixel/tile/FlxBaseTilemap.hx
```diff
@@ -609,38 +610,79 @@ class FlxBaseTilemap<Tile:FlxObject> extends FlxObject
...
-	 * @param	Range				If you want this callback to work for a bunch of different tiles, input the range here. Default value is 1.
+	 * @param	Tile		The tile id that this filter is for, or the first tile id in a range of tiles.
+	 * @param	Filter		Specify a class for this filter to work with, defaults to `FlxObject`
+	 * @param	Collisions	Specify Collision flags to be used for this tile. Defaults to ANY
+	 * @param	Callback	The function to trigger when overlap is detected. e.g. `tileCallback(Tile:FlxObject, Object:FlxObject)`
+	 * @param	ProcessCallback	This function, if set, will be called when overlap is detected, and will return true
+	 * @param	Range		If you want this filter to be applied to a range of different tiles, input the range here. Default is 1.
 	 */
-	public function setTileProperties(Tile:Int, AllowCollisions:Int = FlxObject.ANY, ?Callback:FlxObject->FlxObject->Void, ?CallbackFilter:Class<FlxObject>, Range:Int = 1):Void
+	public function addTileFilter(Tile:Int, ?Filter:Class<FlxObject>, ?Collisions:Int = FlxObject.ANY, ?Callback:FlxTile-> FlxObject->Void, ?ProcessCallback:FlxTile-> FlxObject->Bool, ?Range:Int = 1):Void
```
Gama11: So this renames `setTileProperties()` and changes the function signature? That's a pretty big breaking change and can't be merged at this point.

SeiferTim: Correct. I couldn't find a good way/reason to keep the old behavior in tact.
Not too worried that it can't be merged now - just want to make sure to keep it. I use this logic in several of my current projects and accidentally lost it.

larsiusprime: Could we just add `addTileFilter` as a new function, and mark `setTileProperties` as deprecated, then remove it in the next point release?

Gama11: What do you mean by "point release"? 5.0.0?

MSGhero: Haxe itself has breaking (and broken) changes between minor versions, but are we sticking to just major versions?

This change would bring flixel's physics closer to something I could use, though the syntax of this function is gross.

larsiusprime: @Gama11 : Whatever semantic versioning says is the next bump for a breaking change?

Gama11: Just because Haxe does it doesn't mean we have to.

@larsiusprime Semantic versioning only allows breaking changes in major versions.

MSGhero: I guess my point was how often will/should we up the major version? Haxe takes forever, and I personally love breaking changes.

Gama11:
Like mentioned earlier, in the current state, this has too many breaking changes to be merged.

PR #1767, opened 2016-03-08
update FlxObject.angle doc comment

medavox:
Specified that FlxObject.angle uses degrees for its unit, not radians.


Gama11:
We don't merge pull requests to `master`. Normally they should target `dev`, but since this is just a doc change, you can target `patch`.

Gama11:
I added the unit to the docs in cb9931b0da781670b34ecacbdefcacd8a1bb2355.

medavox:
OK, thanks for fixing my contribution :)

PR #1770, opened 2016-03-09
Add FlxMath.fastFloor()

MSGhero:
It kinda looks like the function call to `Math.floor` takes longer than actually flooring on js. `fastFloor` was about 20x faster in my little test suite with analyzer off.

Not sure if there's a `fastCeil`. I was also thinking about using `fastSin` and `fastCos` throughout the library, but that's technically a breaking change?

### DIFF
--- a/flixel/math/FlxMath.hx
+++ b/flixel/math/FlxMath.hx
@@ -201,7 +201,7 @@ class FlxMath
 		
 		if (useWorldCoords)
 		{
-			return pointInFlxRect(Math.floor(FlxG.mouse.x), Math.floor(FlxG.mouse.y), rect);
+			return pointInFlxRect(fastFloor(FlxG.mouse.x), fastFloor(FlxG.mouse.y), rect);
 		}
 		else
 		{
--- a/flixel/math/FlxMath.hx
+++ b/flixel/math/FlxMath.hx
@@ -568,4 +568,12 @@ class FlxMath
 		return (a > 0) ? a : -a;
 	}
 	
+	/**
+	 * Floors the argument. Same as Math.floor, but faster.
+	 */
+	public static inline function fastFloor(f:Float):Int
+	{
+		return f >= 0 ? Std.int(f) : Std.int(f - 1);
+	}
+	
 }
--- a/flixel/math/FlxMath.hx
+++ b/flixel/math/FlxMath.hx
@@ -570,6 +570,7 @@ class FlxMath
 	
 	/**
 	 * Floors the argument. Same as Math.floor, but faster.
+	 * WARNING: This will not work for NaN or numbers that are outside the bounds of a 32-bit integer.
 	 */
 	public static inline function fastFloor(f:Float):Int
 	{
--- a/flixel/animation/FlxPrerotatedAnimation.hx
+++ b/flixel/animation/FlxPrerotatedAnimation.hx
@@ -1,6 +1,7 @@
 package flixel.animation;
 
 import flixel.animation.FlxBaseAnimation;
+import flixel.math.FlxMath;
 
 /**
  * ...
@@ -26,14 +27,14 @@ class FlxPrerotatedAnimation extends FlxBaseAnimation
 	private function set_angle(Value:Float):Float
 	{
 		var oldIndex:Int = curIndex;
-		var angleHelper:Int = Math.floor(Value % 360);
+		var angleHelper:Int = FlxMath.fastFloor(Value % 360);
 		
 		while (angleHelper < 0)
 		{
 			angleHelper += 360;
 		}
 		
-		var newIndex:Int = Math.floor(angleHelper / baked + 0.5);
+		var newIndex:Int = FlxMath.fastFloor(angleHelper / baked + 0.5);
 		newIndex = Std.int(newIndex % rotations);
 		if (oldIndex != newIndex)
 		{
--- a/flixel/math/FlxPoint.hx
+++ b/flixel/math/FlxPoint.hx
@@ -281,12 +281,12 @@ class FlxPoint implements IFlxPooled
 	}
 	
 	/**
-	 * Rounds x and y using Math.floor()
+	 * Rounds x and y using FlxMath.fastFloor()
 	 */
 	public inline function floor():FlxPoint
 	{
-		x = Math.floor(x);
-		y = Math.floor(y);
+		x = FlxMath.fastFloor(x);
+		y = FlxMath.fastFloor(y);
 		return this;
 	}
 	
--- a/flixel/math/FlxRandom.hx
+++ b/flixel/math/FlxRandom.hx
@@ -82,7 +82,7 @@ class FlxRandom
 			
 			if (Excludes == null)
 			{
-				return Math.floor(Min + generate() / MODULUS * (Max - Min + 1));
+				return FlxMath.fastFloor(Min + generate() / MODULUS * (Max - Min + 1));
 			}
 			else
 			{
@@ -90,7 +90,7 @@ class FlxRandom
 				
 				do
 				{
-					result = Math.floor(Min + generate() / MODULUS * (Max - Min + 1));
+					result = FlxMath.fastFloor(Min + generate() / MODULUS * (Max - Min + 1));
 				}
 				while (Excludes.indexOf(result) >= 0);
 				
--- a/flixel/math/FlxRect.hx
+++ b/flixel/math/FlxRect.hx
@@ -258,14 +258,14 @@ class FlxRect implements IFlxPooled
 	}
 	
 	/**
-	 * Rounds x, y, width and height using Math.floor()
+	 * Rounds x, y, width and height using FlxMath.fastFloor()
 	 */
 	public inline function floor():FlxRect
 	{
-		x = Math.floor(x);
-		y = Math.floor(y);
-		width = Math.floor(width);
-		height = Math.floor(height);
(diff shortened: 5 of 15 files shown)


Gama11:
Is this specifically an issue on js, and ok on other platforms?

How about fixing it in the standard library instead? ;)

MSGhero:
JS is the only one I checked (I did it in try.haxe.org), but this should be faster on every platform.

larsiusprime:
I can't find a reference at the moment, but I think there are some non-intuitive reasons why standard library floor/ceil might be implemented in a "slow" way.

Regardless of which way you go here, my first step would be to identify which target it's slow on, put in an #if for just that target, add the "fast" implementation, and fall back to the default implementation everywhere else until you've tested that it's actually better on all those other targets. Makes the PR much more likely to be accepted if there's some edge case you didn't think of for the others.

Gama11:
I'm not sure how the implementations compare, but I specifically remember removing a few `Math` functions that are also present in the standard lib, because the performance differences weren't significant enough (if even present at all): https://github.com/HaxeFlixel/flixel/pull/412

MSGhero:
I guess I can test the other platforms, but on js using 1 million iterations:
`Math.floor` (native js): `0.02600s`
noninlined fastFloor: `0.00700s`
inlined fastFloor: `0.0009999s`

The function call alone takes 7 times longer than just flooring. Perhaps the native version is doing some kind of NaN checking or whatever.

MSGhero:
As far as #412, custom implementations of any of those functions are only faster if you restrict yourself to outputting ints, where you can do arcane bitwise magic, or inlining their raw implementations, where the function call is the slowest part of the function.

Gama11:
Let's see what the Haxe unit tests say about this implementation: https://travis-ci.org/Gama11/haxe/builds/114854752

Not sure if there are a lot of tests for `Math.floor()` though.

Gama11:
Looks like it failed already, on:

```haxe
eq( Math.floor( -10000000000.7)*1.0, -10000000001. );
eq( Math.ffloor( -10000000000.7), -10000000001. );

Math.ffloor(Math.POSITIVE_INFINITY) == Math.POSITIVE_INFINITY;
Math.ffloor(Math.NEGATIVE_INFINITY) == Math.NEGATIVE_INFINITY;
Math.isNaN(Math.ffloor(Math.NaN)) == true;
```

MSGhero:
Haha that's actually amusing. Idk what's up with the first two, but since + and - infinity are special IEEE implementations, `Std.int` is just chopping off all the bits that define infinity or not.

MSGhero:
So, this won't be a haxe-default implementation, but nothing in flixel is NaN or >2 billion except for potentially things inside `FlxRandom`. But that should be upper bounded by 0x7fffffff anyway.

Gama11:
Should definitely mention the limitations in the docs (`NaN` and Infinity are not considered).

larsiusprime:
I assume putting in checks for NaN and Infinity will erase the speed gains?

MSGhero:
Anything bigger than 2 billion or less than negative 2 billion can't use `Std.int`, so I think handling those edge cases isn't worth it. Then we'd have to use `Std.int64` if that exists. I can try it, but it's no longer a 20x speedup.

gamedevsam:
I recommend calling the functions just ```FlxMath.floor``` and ```FlxMath.ceil```. We can put in the docs of the function calls details about their limitations and performance advantages.

MSGhero:
So just rename the function? I'm on my phone, but I think I updated the docs already.

Any other changes?

Gama11:
I guess the idea with the naming was consistency with `fastSin()` / `fastCos()`?

MSGhero:
Up to you three.

Gama11:
I did a bit of benchmarking (debug build). According to these results, `fastFloor()` is __slower on all targets__ (considerably even on Neko and CPP). It gets better with inline, but `fastFloor()` only beats `Math.floor()` on Flash, by a factor of ~2.

|       | Math.floor() | FlxMath.fastFloor() | inline FlxMath.fastFloor() |
|-------|--------------|---------------------|----------------------------|
| **Flash** | 0.127s       | 0.183s              | 0.066s                     |
| **HTML5** | 0.001s       | 0.002s              | 0.002s                     |
| **Neko**  | 0.0267s      | 0.0819s             | 0.0573s                    |
| **CPP**  | 0.0202s      | 0.368s              | 0.194s                     |

```haxe
package;

import flixel.FlxState;
import haxe.Timer;

{
	{
		Timer.measure(testMathFloor);
		Timer.measure(testFastFloor);
		Timer.measure(testFastFloorInline);
	}

	function testMathFloor()
	{
		for (i in -500000...500000)
			Math.floor(i);
	}

	function testFastFloor()
	{
		for (i in -500000...500000)
			fastFloor(i);
	}

	function testFastFloorInline()
	{
		for (i in -500000...500000)
			inlineFastFloor(i);
	}
	
	function fastFloor(f:Float):Int
	{
		return f >= 0 ? Std.int(f) : Std.int(f - 1);
	}

	inline function inlineFastFloor(f:Float):Int
	{
		return f >= 0 ? Std.int(f) : Std.int(f - 1);
	}
}
```

@MSGhero I'm curious why my results are so different from yours - what does you code look like?

MSGhero:
@Gama11 I just copy pasted your code into try.haxe.org, js no analyzer.

Minus flixel, static functions:
image

Minus flixel, non-static functions:
image

**???** I'll run it in a test project as well.

larsiusprime:
Maybe compare debug vs release?

MSGhero:
I don't feel like setting it up for flash release (no debug traces), but here are mine:

 |Math.floor|fastFloor|inlineFastFloor
-----|-------|---------|----------|-------------
Flash debug|0.0845|0.131|**0.0523**
Neko debug/release|**0.0207**|0.0660|0.0463
Windows debug|**0.0142**|0.184|0.0976
Windows release|0.00878|0.00860|**0.00588**
HTML5 debug/release|0.197|0.182|**0.0580**

MSGhero:
@Gama11 your cpp result looks like you forgot a 0 compared to mine, and my js results are far more drastic **???**

MSGhero:
Debug Windows' and d/r neko's `Math.floor` are faster than just `Std.int` for whatever reason. Un-analyzed js is 20x faster for me (analyzed just gets rid of all function calls since they don't do anything).

I'll `#if` for neko and debug cpp then? Anything else?

gamedevsam:
Doing performance profiling on debug builds is misleading. Always profile with release builds.

Gama11:
Either way, with these benchmark results I'm not sure this is worth it. What do you think @gamedevsam? A million executions is kind of an unrealistic scenario anyway, and it's not even consistenly faster there.

MSGhero:
@gama11 so the whole reason I even added it is because the simplex noise addition to `FlxRandom` I'll add later had it as a private function. There, you're more realistically doing a million ops: BMD width x height x octaves of noise.

gamedevsam:
I believe we should include this functionality, but only if we can ensure there is no performance degradation in ANY platform. So we would need a unit test that would compare the speed of flooring numbers between ```Math.floor``` and ```FlxMath.floor```, and fail the test if ```FlxMath.fastFloor``` is slower (within some margin of error, say 5%).

I wouldn't be comfortable merging this change without some kind of failsafe in the event that ```Math.floor``` is further optimized in platforms where we override its behavior.

flixel/math/FlxMath.hx
```diff
@@ -568,4 +568,13 @@ class FlxMath
 		return (a > 0) ? a : -a;
 	}
 	
+	/**
+	 * Floors the argument. Same as Math.floor, but faster.
+	 * WARNING: This will not work for NaN or numbers that are outside the bounds of a 32-bit integer.
+	 */
+	public static inline function fastFloor(f:Float):Int
+	{
+		return f >= 0 ? Std.int(f) : Std.int(f - 1);
```
gamedevsam: We need to carefully profile this a very simple perf test on all platforms, to ensure it's actually faster than ```Math.floor``` alternative.

gamedevsam:
This is a big change and I understand Gama's skepticism. We need to be sure we don't cause performance regressions due to making assumptions based on one or two platform's behavior.

I think we should develop a Flixel project that can be easily run on any platform that would run performance tests and output clear, unbiased results regarding our optimization assumptions.

MSGhero:
@gamedevsam that's fair, but where would this perf test go? Demos, addons? Or some template file? And say it's faster on half of the targets and slower on the others, should we `#if` or avoid changing it?

Not sure what the conflicts are now, I guess something changed under me?

Gama11:
I don't think either really fits. Maybe we need a performance subfolder for /tests.

Don't worry about the merge conflicts.

MSGhero:
Ok, if we make a subfolder for performance, what exactly are we testing? Are we `assert`ing that one thing is faster than another on each platform? What about if something becomes slower but adds greater functionality?

Gama11:
This is interesting: https://github.com/HaxeFoundation/haxe/commit/6c80e4c4d242f83fc0d8f23aaba21740ac470b68

Gama11:
There are merge conflicts here by now.

I'm not opposed to this in principal, but it definitely needs to be considered and teseted carefully. For instance, the tests I added to `FlxStringUtil.formatMoney()` after this PR was created would fail if that method was changed to use `fastFloor()` like you did here.

Issue #1771, opened 2016-03-13
Transparent FlxTextFormat doesn't work.


buckle2000:
- **Flixel version:** 3.3.12
- **OpenFL version:** 3.5.3
- **Lime version:** 2.8.3
- **Affected targets:** tested on flash, neko, but should be a problem on other platform

________________________________

**Code snippet reproducing the issue**:

```haxe
		line1 = new FlxText(0, 0, 0, "sometext"); 
		line2 = new FlxText(0, 0, 0, "othertext");
		line1.setFormat("Cambria", FontSize, FlxColor.BLACK, null, NONE);
		line2.setFormat("Cambria", FontSize, FlxColor.BLACK, null, NONE);
		var ff = new FlxTextFormat(FlxColor.TRANSPARENT); // When I use WHITE, it works, but white text whill overlay other elements in ui. I do not wish that.
		line1.addFormat(ff,1,10); // no error here, nothing to do with the number 10;
		line2.addFormat(ff,1,10);
```
It might be a problem of openfl, but I am not sure.
________________________________

**Observed behavior:** compilation success, but texts are still BLACK, not transparent.
**Expected behavior:** texts should be visually disappeared

buckle2000:
I've found the reason.
Sorry I didn't see this: 
`@param	Color			The color of the text in traditional flash 0xRRGGBB format.`
FlxText.setFormet  accept FlxColor, which is very confusing.
FlxColor constants are ints like this: 0x12345678,
but the color of textformat can only receive 0x345678, leave the alpha value behind.

PR #1772, opened 2016-03-13
Update FlxText.hx

buckle2000:
Increase readability.

### DIFF
--- a/flixel/FlxGame.hx
+++ b/flixel/FlxGame.hx
@@ -561,7 +561,7 @@ class FlxGame extends Sprite
 		
 		if (_skipSplash || FlxSplash.nextState != null) // already played
 		{
-			_requestedState = cast Type.createInstance(_initialState, []);
+			_requestedState = cast (Type.createInstance(_initialState, []));
 			if (FlxSplash.nextState == null)
 			{
 				_gameJustStarted = true;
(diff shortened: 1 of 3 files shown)


buckle2000:
#1771

Gama11:
Hm, ARBG border colors should be supported..

Gama11:
Oh, right, you're using 3.3.12. Alpha support for border colors has been added in 4.0.0.

buckle2000:
@Gama11 
Be more precise, I was using `flixel: 3.3.12 [4.0.0]`.
Which is the actual version number?
default

Gama11:
You have both 3.3.12 and 4.0.0 installed, but 4.0.0 is active (since it's within square brackets).

Gama11:
Hm, transparency does seem to work, but it also appears to create a gray background for the text...

```haxe
package;

import flixel.FlxState;
import flixel.text.FlxText;
import flixel.util.FlxColor;

{
	{
		var text = new FlxText(0, 0, 0, "Test");
		var transparentGreen = FlxColor.GREEN;
		transparentGreen.alphaFloat = 0.3;
		text.borderColor = transparentGreen;
		text.borderStyle = FlxTextBorderStyle.OUTLINE;
		add(text);
	}
}
```

buckle2000:
But just borders can be transparent, not the fill color (main color) of text.
Due to poor flash API design. 
Also, please check this if you have not: #1771

buckle2000:
Fixed a stupid error.

PR #1784, opened 2016-03-19
Clearer lerp() sig and doc

GimmickyApps:
The formula actually works perfectly well for cases where "Max" < "Min", so they shouldn't be called that.

### DIFF
--- a/flixel/math/FlxMath.hx
+++ b/flixel/math/FlxMath.hx
@@ -88,7 +88,7 @@ class FlxMath
 	 *					Another example, 2 will result in a number twice as far from NumberA as NumberB is.
 	 *					Finally, negatives are allowed too: -1 will result in a number the same distance as NumberB is from NumberA,
 	 *                  but on the other side of NumberA.
-	 * @return 			Interpolated number.
+	 * @return 			Interpolated/extrapolated number.
 	 */
 	public static inline function lerp(NumberA:Float, NumberB:Float, Ratio:Float):Float
 	{
--- a/flixel/math/FlxMath.hx
+++ b/flixel/math/FlxMath.hx
@@ -79,11 +79,15 @@ class FlxMath
 	}
 	
 	/**
-	 * Returns linear interpolated value between two numbers
+	 * Returns the linear interpolated/extrapolated value of two numbers
 	 *
 	 * @param NumberA 		Any number.
 	 * @param NumberB	 	Any number.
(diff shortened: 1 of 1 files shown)


Gama11:
Why change the parameter names?

Gama11:
I was gonna comment about the parameter names, but `FlxMath` seems to be totally inconsistent about that anyway. `num1` and `num2`, `f1` and `f2`, `a` and `b`..

GimmickyApps:
Regardless of the other inconsistencies, "Max" implies that it must be greater than "Min", when actually the two numbers can be anything and the function still interpolates between them correctly, so these names do not make sense here, only something more generic does.

flixel/math/FlxMath.hx
```diff
@@ -79,16 +79,16 @@ class FlxMath
...
 	/**
-	 * Returns linear interpolated value between Max and Min numbers
+	 * Returns linear interpolated value between two numbers
 	 *
-	 * @param Min 		Lower bound.
-	 * @param Max	 	Higher bound.
-	 * @param Ratio 	Defines which number is closer to desired value.
+	 * @param NumberA 		Any number.
+	 * @param NumberB	 	Any number.
+	 * @param Ratio 	How far to interpolate between the two, on a scale of 0 (NumberA) to 1 (NumberB).  For example, 0.5 will result in the midpoint between the two numbers.
```
Gama11: Not sure about this doc, the forumla still works with numbers outside the `0...1` range. Of course then it's not really interpolation anymore.

GimmickyApps: Touchee.

Gama11:
Not sure how I feel about explicitly mentioning extrapolation in a "lerp" function (which stands for linear _interpolation_, so that might lead to some confusion).

Looking through some API docs, it seems like the ratio is often clamped to values between 0 and 1- Unity does that for example. That would be a breaking change though...

GimmickyApps:
It doesn't only stand for interpolation.  E.g. http://www.mathworks.com/matlabcentral/fileexchange/28432-lerp--fast-n-dimensional-linear-interpolation---extrapolation

Gama11:
Not so sure about that, https://en.wikipedia.org/wiki/Lerp_(computing) redirects to "Linear Interpolation".

GimmickyApps:
I just meant there's precedent.  Why bother capping it?  It's useful.  Advertise its usefulness.

Gama11:
Yeah, I guess as long as we're not changing the implementation, might as well document it..

PR #1787, opened 2016-03-22
Correct FlxKeyManager docs

ashes999:
Mention key pressed is ID, not name, and returns -1 if not found.
It should probably also link you to some example or API where you can consume this integer in a human-readable way (eg. `FlxG.keys.justPressed.SPACE`).

### DIFF
--- a/flixel/input/FlxKeyManager.hx
+++ b/flixel/input/FlxKeyManager.hx
@@ -73,9 +73,9 @@ class FlxKeyManager<Key:Int, KeyList:FlxBaseKeyList> implements IFlxInputManager
 	}
 	
 	/**
-	 * Get the first key which is currently pressed.
+	 * Get the ID of the first key which is currently pressed.
 	 * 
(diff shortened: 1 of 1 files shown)


Gama11:
Thanks!

ashes999:
Thanks for the lightning-fast merge!

Any idea which class exposes the key constants? I couldn't find it (surprisingly ...)

Gama11:
`FlxKey` for `FlxKeyboard` or `FlxAndroidKey` for `FlxAndroidKeys`. Which makes documenting that here in the base class a bit awkward... However, these should probably at least return the `Key` type parameter instead of plain Integers.

ashes999:
That's a pretty sensible change. Using strings almost made more sense (easy to read/identify, but troubling to compare).

Maybe you can open an issue to address returning the proper `Key` type. I'd love to do it, but I have too many games on the go right now :)

PR #1788, opened 2016-03-23
Anim prefix fix, closes #1786

### LINKED ISSUE #1786: crash using animation.addByPrefix()
- **Flixel version:** dev
- **OpenFL version:** 3.6.0
- **Lime version:** 2.9.0
- **Affected targets:** flash, probably all

```haxe
package;

import flixel.FlxState;

{
	var leftFighter:FlxSprite;
	{
		super.create();
		leftFighter = cast add(new FlxSprite().loadGraphic(AssetPaths.katatedoriShihonage_000__png, true, 452, 414, true));
		leftFighter.animation.addByPrefix("Katatedori Shihonage", "assets/images/katatedoriShihonage_", 21, true);
		leftFighter.animation.play("Katatedori Shihonage");
	}
}
```

**Observed behavior:** 

```
TypeError: Error #1009: Cannot access a property or method of a null object reference.
	at StringTools$/startsWith()[C:\HaxeToolkit\haxe\std\StringTools.hx:142]
	at flixel.animation::FlxAnimationController/findByPrefix()[C:\HaxeToolkit\haxe\lib\flixel\git\flixel\animation\FlxAnimationController.hx:719]
	at flixel.animation::FlxAnimationController/addByPrefix()[C:\HaxeToolkit\haxe\lib\flixel\git\flixel\animation\FlxAnimationController.hx:452]
at (my playstate)
```
Or, if I leave off the `loadGraphic()` call, the flixel debugger says `[WARNING] No animation called "Katatedori Shihonage"` and I get a placeholder image, but otherwise no clues are given from `addByPrefix()` as to what I might be doing wrong.

**Expected behavior:**  Animation loaded and played...

I had been loading frames like this, with one big sprite sheet:

```haxe
	{
		super.create();
		leftFighter = cast add(new FlxSprite().loadGraphic(AssetPaths.katatedoriShihonage__png, true, 452, 414, true));
		var frames:Array<Int> = [];
		for (i in 0...45)
			frames.push(i + 1);
		leftFighter.animation.add("Katatedori Shihonage", frames, 21, true);
		leftFighter.animation.play("Katatedori Shihonage");
```

But the problem is, when I have as many frames as I plan to in the end, they will not fit on a single sprite sheet without hitting the common 4096x4096 texture size limit, even if I use multiple rows and columns rather than a single strip.  So I wanted to load them individually from files suffixed with _000, _001, _002, and so on.  From [the documentation](https://github.com/HaxeFlixel/flixel/blob/fc4abe05e19e13992f133b299d6cf36dfd013eed/flixel/animation/FlxAnimationController.hx#L437) for `addByPrefix()` it sounded like that's exactly what that's for, individual PNG files for each frame, named this way:

```
	/**
	 * Adds a new animation to the sprite.
	 * 
	 * @param	Name		What this animation should be called (e.g. `"run"`).
	 * @param	Prefix		Common beginning of image names in atlas (e.g. `"tiles-"`).
	 * @param	FrameRate	The speed in frames per second that the animation should play at (e.g. `40` fps).
	 * @param	Looped		Whether or not the animation is looped or just plays once.
	 * @param	FlipX		Whether the frames should be flipped horizontally.
	 * @param	FlipY		Whether the frames should be flipped vertically.
	 */
```

I'm likely doing it wrong, but I hope this functionality is actually supported...in this case maybe we can touch up the documentation a tiny bit for future newcomers.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/32148287-crash-using-animation-addbyprefix?utm_campaign=plugin&utm_content=tracker%2F254189&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F254189&utm_medium=issues&utm_source=github).
</bountysource-plugin>

MSGhero:
Is there a way to update my local flixel without a pull request from github? I don't like how I have 2 commits for this.


MSGhero:
***Without using haxelib

Gama11:
What does this have to do with haxelib?

You just need to pull from origin/dev (if origin is the main repo, not your fork).

MSGhero:
Origin is my fork. Could I do all of this without having my own fork? Or do I need it when it comes time for github pulls?

Or could main repo be my origin, and I push changes to my fork...? Sourcetree.

Gama11:
Yeah, you need a fork to create a branch you can push changes to.

Your origin could be the main repo, and then you'd have to set up your fork as a remote. That way you can push to it separately (and also pull from origin). That's actually how I tend to do it. But it doesn't really matter which one is origin and which one an additional remote.

tests/unit/src/flixel/animation/FlxAnimationControllerTest.hx
```diff
@@ -88,6 +88,15 @@ class FlxAnimationControllerTest extends FlxTest
 		Assert.areEqual(1, timesCalled);
 	}
 	
+	@Test // #1786
+	function testNullFrameName():Void
+	{
+		loadSpriteSheet();
+		sprite.animation.addByPrefix("Test", "test");
+		
+		Assert.isNull(sprite.animation.getByName("Test"));
```
MSGhero: `Assert.previousLineDidntCauseAnError()`

Gama11: You could just try-catch it and call `Assert.fail()` in the `catch()`.

MSGhero: Hmm, this is actually pretty good still. Makes sure if `findByPrefix` doesn't find anything, no animation gets made. I don't know how necessary of a check that is, though.

Gama11: Actually, munit detects uncaught exceptions anyway.

PR #1789, opened 2016-03-24
Include .mp3 sounds in HTML5 for Internet Explorer support

Nallebeorn:
Currently, the html5 export won't run in Internet Explorer. This is because it expects mp3 versions of all audio files, and when those aren't found, it gets stuck on the loading screen.
Therefore, I simply modified include.xml so that mp3s are included in HTML5 as well as Flash. Ogg sounds are still included alongside; some browsers like Firefox depend on the OS for mp3 codecs, so ogg is preferred in those cases.

### DIFF
--- a/include.xml
+++ b/include.xml
@@ -3,7 +3,7 @@
 	<icon path="assets/images/logo/HaxeFlixel.svg" />
 	
 	<assets path="assets/sounds" rename="flixel/sounds" include="*.ogg" if="html5" unless="unit-test" />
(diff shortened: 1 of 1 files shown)


sruloart:
Oh :) funny, you should really take a look at my old related issues...short answer is it's not legal, long answer is there's no reliable audio solution for HTML5.

Nallebeorn:
Ah, I see. It's a bummer that IE won't just not play any sounds, but refuse to run the game at all...

PR #1811, opened 2016-04-11
Auto detect window size and etc.

buckle2000:
Tested. works. [see here](https://github.com/HaxeFlixel/flixel/pull/1811/commits/feae5239bfa1eec2c8824155fe137fe99c8cc8c2)


buckle2000:
Also, please change [the template of Main.hx](https://github.com/HaxeFlixel/flixel-templates/blob/dev/default/source/Main.hx.tpl) (Main.hx.tpl) into a static file (Main.hx).
${WIDTH} and ${WIDTH} can be simply replaced with 0.

buckle2000:
Added function `multiply` to FlxPoint.
It might be useful.
e.g.
> sprite.scale.multiply(2.0)

Also, try operators overloading?

flixel/math/FlxPoint.hx
```diff
@@ -152,6 +152,19 @@ class FlxPoint implements IFlxPooled
 	}
 	
 	/**
+	 * Multiply this point by a scalar.
+	 * 
+	 * @param	scalar	How much times this point will be as it before.
+	 * @return	This point.
+	 */
+	public function multiply(scalar:Float):FlxPoint
```
Gama11: `FlxVector#scale()` is the same thing, so this seems redundant.

flixel/FlxGame.hx
```diff
@@ -561,7 +569,7 @@ class FlxGame extends Sprite
 		
 		if (_skipSplash || FlxSplash.nextState != null) // already played
 		{
-			_requestedState = cast Type.createInstance(_initialState, []);
+			_requestedState = cast (Type.createInstance(_initialState, []));
```
Gama11: There shouldn't be any parens here, `cast` is not a function.

flixel/math/FlxPoint.hx
```diff
@@ -358,7 +371,7 @@ class FlxPoint implements IFlxPooled
 			{
 				angle = c2 - c1 * ((x + ay) / (ay - x));
 			}
-			angle = ((y < 0) ? -angle : angle) * FlxAngle.TO_DEG;
+			angle = ((y < 0) ? - angle : angle) * FlxAngle.TO_DEG;
```
Gama11: Accidentally reformatted?

flixel/FlxGame.hx
```diff
@@ -243,6 +243,14 @@ class FlxGame extends Sprite
 		// Super high priority init stuff
 		_inputContainer = new Sprite();
 		
+		if (GameSizeX == 0)
+		{
+			GameSizeX = Std.int(Lib.current.stage.window.width);
```
Gama11: Would be shorter / nicer to use `FlxG.stage`.

flixel/FlxGame.hx
```diff
@@ -243,6 +243,14 @@ class FlxGame extends Sprite
 		// Super high priority init stuff
 		_inputContainer = new Sprite();
 		
+		if (GameSizeX == 0)
```
Gama11: The special meaning of `0` for `GameSizeX ` and `GameSizeY` should be documented in the `@param` docs.

flixel/FlxGame.hx
buckle2000: No way. FlxG is not initailized at that time.

Gama11: I don't think that matters? `get_stage()` is just a shortcut for `Lib.current.stage`.

flixel/math/FlxPoint.hx
buckle2000: It is not redundant.
Some of HaxeFlixel's code use FlxPoint instead of FlxVector.
So you have to write code like this:
```
point.x*=scalar;
point.y*=scalar;
```

Gama11: I'm aware, that doesn't make it any less redundant - both functions have the same implementation. Since `FlxVector` extends `FlxPoint`, `FlxVector` would have a `scale()` **and** a `multiply()`.

If anything, `scale()` should be moved to `FlxPoint()`.

buckle2000: Oh I did not know that.

buckle2000:
Done.

Gama11:
>Also, try operators overloading?

Operator overloading only works for abstracts. I've experimented with it a bit once... You would need a `FlxPoint` abstract and a `FlxPointImpl` class or something of the sort. I think I had some issues with methods that return `FlxPoint` instances / `this`, but I don't remember why exactly.

Operator overloading might be supported via static extensions in Haxe 4 (https://github.com/HaxeFoundation/haxe/issues/2803), which would be quite convenient. Of course, that's not gonna be available anytime soon.

flixel/math/FlxPoint.hx
```diff
@@ -152,6 +152,19 @@ class FlxPoint implements IFlxPooled
 	}
 	
 	/**
+	 * Scale this point.
+	 * 
+	 * @param	k - scale coefficient
+	 * @return	scaled vector
+	 */
+	public inline function scale(k:Float):FlxVector
```
Gama11: This still returns a `FlxVector` and says `@return	scaled vector`.

Gama11: Hm, this is going to result in `sprite.scale.scale()`, which looks quite weird... Oh well.

flixel/math/FlxVector.hx
```diff
@@ -96,19 +96,18 @@ class FlxVector extends FlxPoint
...
-		return this;
-	}
-	
+    /**
+     * Scale this vector.
+     * 
+     * @param   k - scale coefficient
+     * @return  scaled vector
+     */
+    public inline override function scale(k:Float):FlxVector
```
Gama11: An inlined function can't be overriden. Are you actually compiling this code locally? ;)

flixel/FlxGame.hx
```diff
@@ -223,16 +223,16 @@ class FlxGame extends Sprite
 	/**
 	 * Instantiate a new game object.
 	 * 
-	 * @param	GameSizeX		The width of your game in game pixels, not necessarily final display pixels (see Zoom).
-	 * @param	GameSizeY		The height of your game in game pixels, not necessarily final display pixels (see Zoom).
+	 * @param	GameSizeX		The width of your game in game pixels, not necessarily final display pixels (see Zoom). If it is less than 0, HaxeFlixel will auto detect the width according to the current window.
```
Gama11: These docs are not quite accurate, they only account for the `< 0` case, not `== 0`.

buckle2000:
[Also, template change](https://github.com/HaxeFlixel/flixel-templates/pull/25).

flixel/math/FlxVector.hx
buckle2000: Oh no. both with override or without override failed.
I tried this:

```
class A{
    public static function main(){
        new B().a();
    }
}

class B extends C {
    public **override** inline function a():B {
        return this;
    }
}

class C {
    public inline function a():C {
        return this;
    }
}
```
with `override`: Field a is inlined and cannot be overridden
without: Field a should be declared with 'override' since it is inherited from superclass C

Gama11: The problem is that `FlxPoint#scale()` is `inline` - you can't override a function that is inline, otherwise the compiles wouldn't be able to inline it anymore.

You definitely can't omit hthe override keyword.

flixel/FlxGame.hx
```diff
@@ -243,6 +243,14 @@ class FlxGame extends Sprite
 		// Super high priority init stuff
 		_inputContainer = new Sprite();
 		
+		if (GameSizeX <= 0)
```
Gama11: Maybe we should stick to only `== 0`. Something like `new FlxGame(-10, -50, PlayState)` would result in windowSize x windowHeight, which seems a little weird.

buckle2000: Sorry, I don't understand what do you mean by `windowSize x windowHeight`.
Another reason I can think of is to **reserve** the negative values for possible more implements in the future.

buckle2000:
just keep both `multiply` and `scale`.

flixel/FlxGame.hx
Gama11: I meant windowWidth x windowHeight.

What I meant is that it seems confusing to the reader that random negative values would turn into a `FlxGame` instance which uses the window dimensions, wheareas with `0`, you can at least guess that it probably has some special meaning.

Gama11:
Why? I don't agree with that. Having two functions in `FlxVector` that do the same thing is bad API design.

flixel/FlxGame.hx
```diff
@@ -243,6 +243,14 @@ class FlxGame extends Sprite
...
+		{
+			GameSizeX = Std.int(Lib.current.stage.window.width);
```
Gama11: We can't use `.window` here, needs to be `stageWidth` and `stageHeight` (Travis is currently failing because `Stage#window` doesn't exist in OpenFL legacy).

buckle2000: I have tried using `Lib.current.stage.width` but not `Lib.current.stage.*window*.width`.
BUT, `Lib.current.stage.width` **IS 0!** I don't know why.
I figured that out when targeting flash.

Gama11: `stage.stageWidth`, not `stage.width`.

buckle2000:
Well, if not, how to solve the "inline thing"?
remove `inline`?
use some dark magic of haxe to make `multiiply` un-accessible in `FlxVector`?

flixel/FlxGame.hx
buckle2000: Sorry, I thought they are the same.
stage.stageWidth does work.

flixel/FlxGame.hx
buckle2000: I changed the comments, so it is easier to be understood.
Just think if somehow a negative value is passed in, `FlxGame` can deal with it.

Gama11:
Yeah, just remove the `inline`. It's only a minor optimization.

buckle2000:
OK.
Another question:
as you mentioned:
> Having two functions in FlxVector that do the same thing is bad API design.
What about `FlxVector.clone` and `FlxPoint.copyTo`?

Gama11:
Yeah, not a fan of that.

PR #1814, opened 2016-04-11
redundant cast

buckle2000:


Gama11:
Ah, good point. :) Thanks!

PR #1818, opened 2016-04-20
FlxObject: fix overlapsPoint() if overlap is at x or y = 0

IBwWG:
The top and left edges of checked objects were not registering as overlapping, when they should be.  (Consider the example of a single pixel at 0,0: the former could would never register an overlap with the mouse, always looking for your position to be > 0 and < 1.  Even though it's a float, it still gets updated to integer values when you move the mouse.)

May seem trivial, but when you have a button at the edge of a full-screen game, it becomes important.

### DIFF
--- a/flixel/FlxObject.hx
+++ b/flixel/FlxObject.hx
@@ -781,7 +781,7 @@ class FlxObject extends FlxBasic
 	{
 		if (!InScreenSpace)
 		{
-			return (point.x > x) && (point.x < x + width) && (point.y > y) && (point.y < y + height);
+			return (point.x >= x) && (point.x < x + width) && (point.y >= y) && (point.y < y + height);
 		}
 		
 		if (Camera == null)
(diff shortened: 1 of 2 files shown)


Gama11:
This seems to be causing a failure in `FlxButtonTest` on neko and cpp:

```
FAIL: massive.munit.AssertionException: Value [1] was not equal to expected value [0] at flixel.ui.FlxButtonTest#testTriggerAnimationOnce (77)
```

IBwWG:
Thanks, I'll have a look.

IBwWG:
That's probably expected behaviour, if the mouse cursor is at 0,0 during the test, and the button is at 0,0.  :)

Gama11:
I just checked this with a screen magnifier and a `FlxButton` - wow, how did nobody notice this before?  This code has been around since the AS3 days. :)

Before mering though - this seems like the sort of thing that we should really have a unit test for, and that should also be pretty easy to test.

IBwWG:
I had noticed, because I had a bunch of buttons at the top of the screen that testers were complaining did not function well.  I had thought I was somehow not understanding hitboxes correctly or something.  When I finally dug into it I was thinking the same thing ;)

Hopefully this test is sufficient.

IBwWG:
Super, it passed on Travis too, not just locally.

PR #1822, opened 2016-04-21
Fix draw debug rect off by one

### LINKED ISSUE #1821: drawDebug hitbox dimensions are off
- **Flixel version:** dev

Per https://github.com/HaxeFlixel/flixel/commit/647795c29b3d14cc34296686d0248787fa6b05ca#commitcomment-17198787

Actually, [that particular line](https://github.com/IBwWG/flixel/blob/5bfea4406e9868dc99c6bedfc6065b89a27cb3eb/flixel/FlxObject.hx#L122) is OK.  "x + width" if you sub in the simplest example of 0 and 1, makes the equation true.  Likewise for other x+width and y+height spots.

Where it might not work is `x < FlxG.worldBounds.right`, but I have to look into how `right` is calculated.

BTW now that I think of it, when I first was looking into this issue, I noticed that the bounding boxes were drawn off by 1px by the debugger:

hitbox-vis
hitbox-invis

Since my fix worked I didn't really revisit that.  (Sorry for the white border, you'll have to view those in an editor to see, but you probably would anyway to zoom in.)  I wonder if that also needs fixing somewhere.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/33196699-investigate-off-by-one-boundaries-and-overlap-checks?utm_campaign=plugin&utm_content=tracker%2F254189&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F254189&utm_medium=issues&utm_source=github).
</bountysource-plugin>

IBwWG:
Fixes #1821.


IBwWG:
Fixes #1821.

Gama11:
Looks good on Flash, but unfortunately not so much on neko:



And here's Neko + OpenFL Next:



Less blurred, but still off. Would be nice if we could get this to behave consistently on all targets. IIRC, this logic used `lineTo()` instead of `drawRect()` a while back, maybe that's more accurate no native targets or  something?

This seems like a pretty big inconsistency in OpenFL's implementation that should probably be isolated and reported (if not already done by somebody else). 

Didn't bother to check HTML5 yet.
__________________

Test code:

```haxe
package;

import flixel.FlxG;
import flixel.FlxSprite;
import flixel.FlxState;

{
	{
		FlxG.debugger.drawDebug = true;
		
		var sprite = new FlxSprite(10, 10);
		sprite.makeGraphic(10, 10);
		sprite.screenCenter();
		add(sprite);
	}
}
```

Gama11:
Fyi, here's the original AS3 Flixel implementation: https://github.com/AdamAtomic/flixel/blob/master/org/flixel/FlxObject.as#L429

Already using `lineTo()`, so better cross-platformness wasn't the reason for earlier HaxeFlixel versions using that. But who knows, might still be more accurate.

IBwWG:
Hmm...wouldn't it also have looked wrong before, then?  I imagine neko+next would already have the correct-looking right + bottom edges but left+top would be too far out by 1px, no?

Gama11:
Indeed. Just less noticably wrong.

IBwWG:
So...in that case, this PR shouldn't change, but we should hold off on merging it until OpenFL fixes their end?  Or we should make a wordaround in flixel?

Gama11:
I'm not too optimistic this is going to get fixed in OpenFL any time soon...

IBwWG:
So, workaround?

IBwWG:
I couldn't find an issue on the openfl tracker.  Maybe I'll open one, then.

Gama11:
If you can find one that's not too much of a hack, sure.

IBwWG:
It would be a bunch of branches, pretty hacky.

Gama11:
Let me guess, add / subtract pixels depending on the target? :)

IBwWG:
Yeah, exactly.  My only other idea was rounding to int first, but that didn't work; after checking, turned out they were already rounded OK anyway, even if visually they seem to be in between pixels.

Gama11:
Did you check if using `lineTo()` makes a difference?

IBwWG:
Yep, see the openfl issue for screenshots.

IBwWG:
@Gama11 maybe you can check out the openfl thread again.  I had trouble reproducing it in pure openfl (although there were still rendering discrepancies, but they weren't the between-pixels kind), would you know of any general rendering settings of openfl's that flixel would change from default?

PR #1825, opened 2016-04-22
Fix VCR to record simultaneous keys+mouse

IBwWG:
C.f. https://github.com/HaxeFlixel/flixel/pull/1740, this is a fresh branch redo.  Tests pass locally on cpp.


tests/unit/src/flixel/system/replay/FlxReplayTest.hx
```diff
@@ -0,0 +1,69 @@
...
+	@Before
+	function before()
+	{
+		seed = 123456789;
+		frameCount = 300;
+		fgr = seed + "\n";
+		var key:Int;
+		var x:Int;
+		var y:Int;
+		var possibleKeys = [8, 9, 13, 16, 17, 18, 20, 27, 32, 33, 34, 35, 36, 37, 38, 39, 40, 45, 46, 48, 49, 50, 51, 52, 53, 54, 55, 56, 57, 65, 66, 67, 68, 69, 70, 71, 72, 73, 74, 75, 76, 77, 78, 79, 80, 81, 82, 83, 84, 85, 86, 87, 88, 89, 90, 96, 97, 98, 99, 100, 101, 102, 103, 104, 105, 106, 107, 109, 110, 112, 113, 114, 115, 116, 117, 118, 119, 120, 121, 122, 123, 186, 187, 188, 189, 190, 191, 219, 221, 222, 301];
```
Gama11: This is a bit silly... You could probably use `FlxKey.toStringMap.keys()` (as long as `-1` and `-2` are excluded).

tests/unit/src/flixel/ui/FlxButtonTest.hx
```diff
@@ -14,6 +17,7 @@ class FlxButtonTest extends FlxTest
 	{
 		button = new FlxButton();
 		destroyable = button;
+		FlxG.mouse.playback(new MouseRecord(0, 0, FlxInputState.RELEASED, 0)); // put the mouse in the upper-left corner to prepare for highlight testing
```
Gama11: Why is this suddenly necessary?

tests/unit/src/flixel/system/replay/FlxReplayTest.hx
IBwWG: Also backslash and graveaccent, so you don't bring up the debugger and throw off other tests?  This test ended up throwing off the flxbutton test because it moved the mouse, so I guess they're not set up to be independent, exactly.

Gama11: Really? That's weird.. There should be a `FlxG.resetGame()` call after each test class at least.

IBwWG: And I went to do it as you said, but that returns an iterator.  I then learned a new trick, thanks!  :)

IBwWG: For others reading, c.f. https://github.com/HaxeFlixel/flixel/pull/1825/files/49c53e39e3bfe81b97a61224f3229a851f02d6d8#r60766546

tests/unit/src/flixel/ui/FlxButtonTest.hx
IBwWG: I would not have thought it so, except that after starting from the current dev branch, this test was failing.  After resetting the cursor, it was succeeding.

IBwWG: I guess resetGame() either doesn't run between the tests, or it does, but doesn't touch the cursor...?

tests/unit/src/flixel/system/replay/FlxReplayTest.hx
Gama11: An iterator can be converted to an array fairly easily.

IBwWG: I know, see my latest commit.

tests/unit/src/flixel/ui/FlxButtonTest.hx
Gama11: Hm, yeah.. I thought resetGame would completely recreate all instances in FlxG, guess it's actually far from that. All this global stuff is pretty annoying for unit testing... :/

We should at least handle the mouse reset in `FlxTest#resetGame()`, so it's not just a random hack, but happens consistently. Perhaps we can even manually call `FlxG.init()`, so things _really_ get reset (new `FlxMouse` instance).

IBwWG: That's interesting.  FlxTest#resetGame() calls FlxG.resetGame(), which eventually should mean hitting FlxG.reset(), which calls input.reset(), which ought to, ya know, reset the inputs, I guess.  But actually FlxMouse's reset() just resets the mouse button states.  I guess that makes sense, otherwise you might have a weird cursor jump for the user, but I'm not sure how resetGame() gets used in the wild anyway.

Gama11: Good point..  and I'm not sure if that makes sense actually. I'd expect the mosue to be reset after a mouse reset.

IBwWG: As a programmer I would, but as a user, having the cursor jump to 0,0 even when I am resetting a game might be unexpected.

IBwWG: But anyway it's moot if we use FlxG.init, which I'm trying now.

Gama11: Fyi, `reset()` gets called on all inputs on state switches too, so it's probably safer not to mess with that for now.

IBwWG: What do you mean exactly by "mess with that"?  Call FlxG.init(), or change FlxMouse.reset() to make the cursor go to 0,0?

Gama11: Change FlxMouse.

IBwWG: Ah, OK.  Meanwhile this approach in FlxTest is causing trouble:

```haxe
	@:access(flixel.FlxG.init)
	function resetGame()
	{
		FlxG.init(FlxG.game, FlxG.width, FlxG.height, FlxG.initialZoom);
		step();
	}
```

Gama11: Could you be a bit more specific than "causing trouble"?

IBwWG: Sorry, see edit.  It was the trouble I'd already described (crashing, other tests failing) but there's more detail now.

IBwWG: I just tried a new approach:

```haxe
		Lib.current.stage.removeChild(FlxG.game);
		Lib.current.stage.addChild(new FlxGame(640, 480, FlxState, 1, 60, 60, true));
```
Which works better: the FlxButtonTest and the others that broke now pass.  However, it still makes the FlxReplayTest fail with accumulated keys.  I find this strange, but we're getting closer...

Gama11: That seems to have a pretty big impact on the runtime of the tests. Perhaps it's best to keep it as simple as we need it for now - we could just create a new FlxMouse instance manually. There's even already a setter that handles all the details of that since we needed that in flixel-ui at some point.

IBwWG: Or, I could just make the final playback frame in the flxrecord test put things back where they were.

IBwWG:
Tests pass locally.

Gama11:
Thanks for your patience!

PR #1826, opened 2016-04-22
Added support for shader based scaling on Next

Seanw265:


Gama11:
Does this still work with OpenFL legacy?

I guess the old pull request can be closed?

flixel/input/FlxPointer.hx
```diff
@@ -120,8 +120,8 @@ class FlxPointer
 	 */
 	public inline function setGlobalScreenPositionUnsafe(newX:Float, newY:Float):Void 
 	{
-		_globalScreenX = Std.int(newX / FlxG.scaleMode.scale.x);
-		_globalScreenY = Std.int(newY / FlxG.scaleMode.scale.y);
+		_globalScreenX = Std.int(FlxG.scaleMode.mouseMultiplier.x * newX / FlxG.scaleMode.scale.x);
```
Gama11: Since this is in `FlxPointer`, it affects touches as well. Because of that, `pointerMultiplier` is probably a better name than `mouseMultiplier`.

flixel/system/scaleModes/shaders/ScaleShaderFilter.hx
```diff
@@ -0,0 +1,42 @@
...
+
+	public function new(shader:Shader) {
+		super(shader);
+	}
+	
+	override public function clone():BitmapFilter {
+		return super.clone();
+	}
+	
+	override function __growBounds (rect:Rectangle) {
```
Gama11: Are these safe to override? Methods prefixed by `__` are meant to be internal in OpenFL afaik.

Seanw265:
It doesn't actually scale on legacy but it doesn't crash either. It just runs as BaseScaleMode. 

I suppose its possible to reimplement it the old way with a compiler conditional. I'll look into that right now.

And yes the other one can be closed.

larsiusprime:
Looks cool :) What's the use code, same as the other one?

flixel/system/scaleModes/shaders/ScaleShaderFilter.hx
Seanw265: This was the original function in the super class. As far as I know it is just called once before the shader is applied unless the `__passes` variable is increased. It maintains functionality.
```haxe
override function __preparePass(pass:Int):Shader {
		return shader;
}
```
Perhaps it should return `super.__preparePass(pass);`?

Seanw265:
@larsiusprime Yup same as the other one!

Gama11:
You don't have to update each file individually using GitHub's web UI, that seems very cumbersome. You can use a git client like SourceTree to push to your fork's branch.

Gama11:
Btw, we have automated code style checks, yours doesn't match Flixel's in a few places:

https://codeclimate.com/github/HaxeFlixel/flixel/pull/1826

Feel free to ignore the `Invalid method name signature` ones as false positives.

flixel/system/scaleModes/shaders/ScaleShaderFilter.hx
Gama11: I'm a bit confused, I commented on `__growBounds`, not `__preparePass`. And it was meant as more of a general question - should methods prefixed by `__` be overriden? My understanding was that these are "internal" in OpenFL.

flixel/system/scaleModes/shaders/ScaleShaderFilter.hx
```diff
@@ -0,0 +1,42 @@
...
+	public var scaleX:Float;
+	public var scaleY:Float;
+	public var resolution:Array<Float>;
+
+	public function new(shader:Shader) {
+		super(shader);
+	}
+	
+	override public function clone():BitmapFilter {
+		return super.clone();
```
Gama11: This method can probably be removed, considering it just calls super?

flixel/system/scaleModes/ShaderScaleMode.hx
```diff
@@ -0,0 +1,102 @@
...
+
+	#if !openfl_legacy
+	private var scaleX:Float;
+	private var scaleY:Float;
+	private var strength:Float;
+
+	public var filter:shaders.ScaleShaderFilter;
+	public var filters:Array<BitmapFilter>;
+
+	public function new(?shader:OneOfTwo<ShaderScaleEnum, openfl.display.Shader> = ShaderScaleEnum.NEAREST, ?scaleX:Float = 1, ?scaleY:Float = 1){
```
Gama11: There's no reason for the `?` here, the default values already make the params optional. Additionally, `?` turns basic types into nullable ones, creating unnecessary runtime overhead.

http://stackoverflow.com/questions/31914614/is-question-mark-operator-in-haxe-really-necessary/31916466#31916466

flixel/system/scaleModes/ShaderScaleMode.hx
```diff
@@ -0,0 +1,102 @@
...
+		FlxG.game.filtersEnabled = true;
+
+		FlxG.signals.postDraw.add(postDraw);
+	}
+
+	public function postDraw(){
+		this.filter.resolution = [FlxG.stage.width, FlxG.stage.height];
+	}
+
+	public function setScale(?scaleX:Float = 1, ?scaleY:Float = 1){
```
Gama11: Again, no need for `?` params.

flixel/system/scaleModes/ShaderScaleMode.hx
```diff
@@ -0,0 +1,102 @@
...
+		this.scaleX = scaleX;
+		this.scaleY = scaleY;
+		this.filter.scaleX = scaleX;
+		this.filter.scaleY = scaleY;
+		this.mouseMultiplier = new flixel.math.FlxPoint(1/scaleX, 1/scaleY);
+	}
+
+	public function setStrength(?strength:Float = 1){
```
Gama11: Again, no need for `?` params.

MSGhero:
SourceTree is confusing af to start with, but it's super useful once you do. Alternatively, github also has a desktop client

Seanw265:
So I have a bunch of changes that improve on the classes I made that I'll push as soon as I'm back at the computer. It adds support for legacy and cleans up a lot of the styling issues. 
I took care of the override of `__preparePass` 
There might be a way to take care of overriding `__growBounds` but changing that back to the original breaks the shaders' math. That math is very convoluted and complicated and I'm tempted to say we should leave it as is for now.

Seanw265:
Ok I just pushed another commit that gets rid of all the internal openfl method overrides. It also gets rid of really weird funky math that was necessary in each shader in order to position the screen correctly.

Gama11:
So, is this on hold because of https://github.com/openfl/openfl/issues/1128?

MSGhero:
That issue was so recent and this PR has been sitting a while

Gama11:
I don't see how that answers my question? :P

Seanw265:
Hang on I may have found a small bug. I'll run some tests tonight and get back to you guys soon.

MSGhero:
Oh I thought you were asking "Has this been on hold this whole time bc of this issue?"

Is it the same result if openfl implements it, given that this is acting on a single DisplayObject (FlxGame) and openfl's would act on the visible stage/screen buffer?

Gama11:
Ah, I should have said "is this _now_ on hold".

Well, there usually isn't too much going on in a flixel game that isn't part of `FlxGame`.

larsiusprime:
No, I don't think this should be on hold; it's not clear whether OpenFL will agree it should be part of the core API (although I think it should) and this is working splendidly in DQ.

Just noticed CI / codeclimate are failing, guess we can fix those up.

Tiago-Ling:
Seems like this could solve the weird, inconsistent FlxGame scaling in Android devices when using Next. I'd argue that currently there's no way to use HaxeFlixel + Android + Next because of the game scaling problem.

Tiago-Ling:
Any news on this?

JoeCreates:
@tiago-ling What is the android+next game scaling issue you mentioned? Is there another issue open for that?

Tiago-Ling:
@JoeCreates Actually no issues opened, but it may be the situation for one - not directly related to the scaling though. 

The scaling problem is this:
* Scaling is done with nearest neighbor algorithm, which can really mess up pixel art when using non-integer factors. This is expected and can be countered to an extent through scaling thresholds and letter boxing.
* The real problem is that a lot of devices use soft buttons for their home, back and menu system buttons. Since there's no way (i think) to really hide them this leads to some really unconventional stage sizes which will result in a sub-100% zoom and will visibly affect non-antialiased graphics (especially text).

The idea would be to use this scaling with shaders to alleviate this problem. One way to try this is to scale the game first using nearest neighbor to the desired integer (2x, 3x, etc) then apply the decimal part using a bilinear filter or something better than nn. This is what Shovel Knight does, according to https://github.com/openfl/openfl/issues/1128

On the other hand, the problem which might need an issue is that i couldn't get `ScaleMode` to work correctly in Android (probably iOS too) when using `next`. I had to manually scale the game here to achieve what works perfectly in `legacy`.

JoeCreates:
Wasnt the first scaling issue always a problem even before next? I actually
prefer nearest neighbour for pixel art games, but I certainly see that its
a problem for higher res stuff.

Tiago-Ling:
In legacy the game scaling works fine here - in next i've experienced (i'll create an issue later if there isn't one already):
* Game centered on the top-left part of the screen, with only the bottom-right part inside the screen
* Game scaled to 1/4 of the screen

Nearest neighbor causes problems for non-integer scales - try scaling a pixel-art game to 1.15, 1.25 to see what i mean: some hard edges will be of different sizes and things will be uneven. This is a bigger problem for pixel-perfect fonts which will get terribly jagged and lose a lot of readability.

JoeCreates:
I know, but if you scale more, say 3 or more times, those extra lines
become negligable. You dont tend to scale a pixel art game by 1.5 times. I
would prefer the extra pixels to be a color in the existing palette rather
than have bilinear create new colors.

What I meant regarding lagacy was that legacy always used nearest
neighbour, too, so that issue isn't new.

flixel/system/scaleModes/ShaderScaleMode.hx
```diff
@@ -0,0 +1,296 @@
...
+	vec4 t0 = mix(tmp1, tmp2, fv);
+    
+    gl_FragColor = t0;//texture2D(uImage0, vec2(tc.x/scaleX,tc.y/scaleY));
+}";
+
+	public function new(shader:OneOfTwo<ShaderScaleEnum, Shader> = ShaderScaleEnum.NEAREST, scaleX:Float = 1, scaleY:Float = 1)
+	{
+		super();
+
+		if (Std.is(shader, Shader))
```
gamedevsam: Isn't this unnecessary since we're leveraging Haxe's type system to ensure people send valid Shader types?

flixel/system/scaleModes/ShaderScaleMode.hx
```diff
@@ -0,0 +1,296 @@
...
+		this.setStrength(1);
+	}
+
+	public function setScale(scaleX:Float = 1, scaleY:Float = 1):Void
+	{
+		postProcess.setUniform("scaleX", scaleX);
+		postProcess.setUniform("scaleY", scaleY);
+		this.scaleX = scaleX;
+		this.scaleY = scaleY;
+		this.pointerMultiplier = new flixel.math.FlxPoint(1 / scaleX, 1 / scaleY);
```
gamedevsam: Unnecessary memory allocation, just do: ```this.pointerMultiplier.set(1 / scaleX, 1 / scaleY);```

Gama11:
This is a bit outdated by now and would have to be updated to support OpenFL 8.

Issue #1828, opened 2016-04-23
FlxReplayTest fails on Travis


Gama11:
cc @IBwWG 

```
Class: flixel.system.replay.FlxReplayTest ....ALSA lib confmisc.c:768:(parse_card) cannot find card '0'
ALSA lib conf.c:4241:(_snd_config_evaluate) function snd_func_card_driver returned error: No such file or directory
ALSA lib confmisc.c:392:(snd_func_concat) error evaluating strings
ALSA lib conf.c:4241:(_snd_config_evaluate) function snd_func_concat returned error: No such file or directory
ALSA lib confmisc.c:1251:(snd_func_refer) error evaluating name
ALSA lib conf.c:4241:(_snd_config_evaluate) function snd_func_refer returned error: No such file or directory
ALSA lib conf.c:4720:(snd_config_expand) Evaluate error: No such file or directory
ALSA lib pcm.c:2217:(snd_pcm_open_noupdate) Unknown PCM default
AL lib: (EE) alsa_open_playback: Could not open playback device 'default': No such file or directory
x
    ERROR: massive.munit.UnhandledException: Not Sound at flixel.system.replay.FlxReplayTest#testReplayRecordSimultaneousKeydownAndMouseDown at openfl/_legacy/media/Sound.hx (141)
	at openfl/_legacy/Assets.hx (2288)
	at DefaultAssetLibrary.hx (490)
	at DefaultAssetLibrary.hx (234)
	at openfl/_legacy/Assets.hx (484)
	at flixel/system/ui/FlxSoundTray.hx (133)
	at flixel/system/frontEnds/SoundFrontEnd.hx (306)
	at flixel/system/frontEnds/SoundFrontEnd.hx (295)
	at flixel/system/frontEnds/SoundFrontEnd.hx (333)
	at flixel/FlxGame.hx (748)
	at flixel/FlxGame.hx (663)
	at FlxTest.hx (37)
	at flixel/system/replay/FlxReplayTest.hx (69)
	at massive/munit/TestRunner.hx (330)
```

https://travis-ci.org/HaxeFlixel/flixel/jobs/125192905#L433-L456

I don't think this is a problem with the test, as it's fine locally, probably just something with the Travis setup.

I tried setting `AUDIODEV=null` (dd983cc1e9b1cf926405cd1801cda5e1de33330d) according to http://stackoverflow.com/questions/15859196/travisci-no-audio-device (and OpenFL's Travis also has this), but looks like that didn't help.

[...]

PR #1829, opened 2016-04-23
Travis flxreplaytest: no sound

IBwWG:
Stopgap for #1828.


Gama11:
Thanks, I think it's better to avoid this issue alltogether though. Now the default sounds are simply excluded from the unit tests.

PR #1831, opened 2016-04-23
Fix bug in accelerateFromAngle

16th-earl:
When setting maxVelocity in accelerateFromAngle, it's necessary to set it to the absolute value, so it's never negative. Also, added some documentation for the accelerateFromAngle function.


Gama11:
Changes are merged to the `dev` branch - could you create a new pull request targetting that?

In what scenario does this cause problems?

PR #1833, opened 2016-04-25
Fix bug in accelerateFromAngle, and add comment

16th-earl:
(Sorry I sent the previous version of this suggestion to the wrong branch, hopefully this is ok…)

The accelerateFromAngle function has a problem in that if sine or cosine of the angle is negative, then the maximum velocity will end up being negative, which it's never supposed to be.

For example, If your current velocity was 10, and you wanted a limit of 100; depending on the angle of rotation you can suddenly end up with a maximum velocity of -100. This makes your velocity of 10 look to be 110 over the maximum. Then you get an extreme change of velocity. In the app I was working on I ended up with a constant oscillation between -100 and +100 from one frame to the next. Adding the Math.abs functions inside accelerateFromAngle, fixes this.

### DIFF
--- a/flixel/math/FlxVelocity.hx
+++ b/flixel/math/FlxVelocity.hx
@@ -265,16 +265,26 @@ class FlxVelocity
 		}
 		return Velocity;
 	}
-	
+
+	/**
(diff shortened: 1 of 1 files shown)


Gama11:
Thanks!

PR #1840, opened 2016-04-30
fixed  -DhscriptPos support in ConsoleUtil.hx

x2f:
With:
  haxeflixel version 4.0.1
  hscript version >= 2.0.5
...compiling on Neko at least with -DhscriptPos gives compilation error:

flixel/4,0,1/flixel/system/debug/console/ConsoleUtil.hx:156: characters 9-42 : Class<hscript.Error> has no field EInvalidAccess

The proposed fix uses the error() method now defined in hscript's Interp class (following usage in hscript's Interp.hx)


Gama11:
Thanks!

Issue #1801, opened 2016-04-02
FlxTilemap buffers are not updated on camera resize or replacement


Gama11:
Nicely demonstrated by a sample project from @jscott1989: https://github.com/jscott1989/camera-test

buckle2000:
true. [strange behavior](https://github.com/buckle2000/camera-test/blob/master/source/MenuState.hx)

buckle2000:
I think there is not a easy solution but call `tilemap.updateBuffers()` on changing cameras.

Because `FlxBasic` is not notified when its main camera changed:
```
private function get_camera():FlxCamera
	{
		return (_cameras == null || _cameras.length == 0) ? FlxCamera.defaultCameras[0] : _cameras[0];
	}
```

Gama11:
I mostly solved this locally, it involves adding a few new `FlxSignal`s for certain "camera events". I struggled with adding proper unit tests for this though..

Gama11:
Eh, I didn't mean to close this just yet.

PR #1841, opened 2016-05-01
fix for issue #1801

Beeblerox:
This commit adds camera resize signal to the core and adds some logic into FlxTilemap. Plus i've added _camera var to FlxTilemapBuffer, so various methods don't need camera argument now (so i deleted them)

### DIFF
--- a/flixel/FlxCamera.hx
+++ b/flixel/FlxCamera.hx
@@ -1439,6 +1439,7 @@ class FlxCamera extends FlxBasic
 			updateFlashOffset();
 			updateScrollRect();
 			updateInternalSpritePositions();
+			FlxG.signals.cameraResized.dispatch(this);
 		}
 		return Value;
 	}
@@ -1452,6 +1453,7 @@ class FlxCamera extends FlxBasic
 			updateFlashOffset();
 			updateScrollRect();
 			updateInternalSpritePositions();
+			FlxG.signals.cameraResized.dispatch(this);
 		}
 		return Value;
 	}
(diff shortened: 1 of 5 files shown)


Gama11:
Didn't you see that I already have a fix for this and assigned myself to the issue? :/

I don't think this fixes the issue for newly added cameras, only for resized cameras. When I tried fixing this, I added a `cameraAdded` (and `cameraRemoved` just for consistency) signal too.

I added the signals to `CameraFrontEnd`, that seemed a bit more fitting.

Also, I think it's better to have some sort of "dirty" flag and wait until the next render loop to actually resize the buffers. Otherwise multiple changes in one frame are unnecessarily expensive (and changing both widht and height of a camera already results in two changes).

I'm not sure we should really be messing with the `FlxTilemapBuffer` API for minor releases. It's mostly internal, but could still be used for things like that isometric tilemap we added `FlxBaseTilemap` for.. (haven't actually checked though).

Beeblerox:
@Gama11 sorry. i hadn't noticed it.
My version fixes both cases (added camera and resized camera).
added camera handled by this line https://github.com/HaxeFlixel/flixel/pull/1841/files#diff-77f0a1439c662b73a06aa33b6dfbd41fR404
And you're right about dirty flag for tilemap buffer. i'd call it `needToResize` or something like that

Gama11:
I still prefer my solution, I don't think we should put _all_ signals in `SignalFrontEnd`. These signals fit much better in `CameraFrontEnd`.

I'm going to work on finishing my fix and integrate your changes.

Issue #1749, opened 2016-02-27
Refactor and fix FlxPath


IBwWG:
- **Flixel version:** 4.0.0

`start()` is undocumented, but it'd be great to know:

1. Why does it take a nonoptional list of nodes?  From reading the docs it at first looks like you would add them via `add()`, `addAt()`, etc.  If those are only meant for adding nodes after `start()`ing maybe they should say so.  (But are they?)
1. Why doesn't it set `active` to true?  Is it meant that someone `start()`s a path and then sets `active` to true themselves?
1. It could maybe be made clear at the beginning that you attach a path to a FlxObject by setting the FlxObject's path property.  (I didn't know at first that they had such a property, so I was looking around the FlxPath API wondering how you're supposed to link the two.)
1. In what order should the `.path = myFlxPath`, `myFlxPath.start()`, and `myFlxPath.active=true` happen?
1. What units are `start()`'s `speed` parameter in?

GimmickyApps:
OK, just from trying it out, for 4. I found that .path, then .active, then .start() works...

winterismute:
Honestly, more than better documentation, I would say "refactor and fix flxPath". I mean, it has an empty constructor which does not even initialize a list of points, and then it's full of methods (head(), tail()) which do not check for null list (the base condition of the object), and other oddities. I actually wanted to do it myself one day.

GimmickyApps:
Sounds good.

winterismute:
I actually made something for this, it's here: https://github.com/HaxeFlixel/flixel/pull/1875
Honestly, I have realized I have messed up a bit, since I did start from the changes of https://github.com/HaxeFlixel/flixel/compare/issue-1749
but I did not integrate them directly, because I did not notice they were from the haxeFlixel tree, I thought they came from the fork of Beeblerox...
But it is the same rationale, in general making sure the _nodes list is ever empty (and therefore methods do not crash), the unit tests added (all failing before) should be enough to understand the changes.

PR #1842, opened 2016-05-02
some changes for FlxPath. see issue #1749

Beeblerox:


Gama11:
I think we should start using backticks for doc comments to have them show up as `inline code blocks` like on GitHub in our API docs page. see here for example, in some cases is even smart enough to turn them into links:

http://api.haxeflixel.com/flixel/util/FlxColor.html?#colorLookup

flixel/util/FlxPath.hx
```diff
@@ -135,16 +166,59 @@ class FlxPath implements IFlxDestroyable
...
+	public function start(?Nodes:Array<FlxPoint>, Speed:Float = 100, Mode:Int = FlxPath.FORWARD, AutoRotate:Bool = false):FlxPath
+	{
+		nodes = (Nodes != null) ? Nodes : nodes;
+		setPathProperties(Speed, Mode, AutoRotate);
+		restart();
+		return this;
+	}
+	
+	/**
+	 * Just sets nodes for this path, plus returns this path object. Could be useful for chaining.
```
Gama11: >Could be useful for chaining.

With that logic, we would need a setter method for every property... I don't really see a good reason to have this function.

flixel/util/FlxPath.hx
```diff
@@ -175,8 +249,9 @@ class FlxPath implements IFlxDestroyable
 	 * Change the path node this object is currently at.
 	 * 
 	 * @param  NodeIndex    The index of the new node out of path.nodes.
+	 * @return	This path object.
```
Gama11: :+1:

flixel/util/FlxPath.hx
```diff
@@ -115,16 +115,47 @@ class FlxPath implements IFlxDestroyable
...
+	 * You can do this in one line:
+	 * object.path = new FlxPath().start([new FlxPoint(0, 0), new FlxPoint(100, 0)], 50, FlxPath.FORWARD);
+	 * 
+	 * or using some more chaining:
+	 * object.path = new FlxPath().add(0, 0).add(100, 0).start(50, FlxPath.FORWARD);
+	 * 
+	 * @param	Nodes	An optional array of points which will be used for path.
 	 */
-	public function new() {}
+	public function new(?Nodes:Array<FlxPoint>)
```
Gama11: What's the benefit of having a `Nodes` param here? You're gonna need to call `start()` anyway if you want to use the path.

flixel/util/FlxPath.hx
```diff
@@ -135,16 +166,59 @@ class FlxPath implements IFlxDestroyable
...
+	/**
+	 * Sets path following properties: speed, mode and auto rotation.
+	 * 
+	 * @param	Speed			The speed at which the object is moving on the path.
+	 * @param	Mode			Path following behavior (like looping, horizontal only, etc).
+	 * @param	AutoRotate		Whether the object's angle should be adjusted to the path angle during path follow behavior.
+	 * 
+	 * @return	This path object.
+	 */
+	public function setPathProperties(Speed:Float = 100, Mode:Int = FlxPath.FORWARD, AutoRotate:Bool = false):FlxPath
```
Gama11: What's the use case for this function? It's really just `start()` with one less argument? Or is it mostly about the `restart()` call?

flixel/util/FlxPath.hx
```diff
@@ -698,4 +778,17 @@ class FlxPath implements IFlxDestroyable
 		}
 	}
 	#end
+	
+	private function get_nodes():Array<FlxPoint>
+	{
+		if (_nodes == null)
+			_nodes = new Array<FlxPoint>();
```
Gama11: It seems the only reason this is a property is to avoid `_nodes` being null. In that case, we could just initialize `_nodes` to `[]`. It's not like `start()` allows `null` to be assigned anyway.

flixel/FlxObject.hx
```diff
@@ -545,6 +545,10 @@ class FlxObject extends FlxBasic
 	
 	/**
 	 * The path this object follows.
+	 * Example:
```
Gama11: :+1:

Gama11:
@winterismute's PR (#1875) seems to be a more updated version of this with unit tests, so I'm closing this.

Issue #1812, opened 2016-04-11
FlxTypedButton behave strange after scaling.


buckle2000:
- **Flixel version:** 4.0.1
- **Affected targets: All**

---

**Code snippet reproducing the issue**:

``` haxe
package;

import flixel.FlxState;
import flixel.ui.FlxButton;

class PlayState extends FlxState
{
    override public function create():Void
    {
        var a = new FlxButton(0, 0, "asdfghjkl");
        a.scale.x*=2;
        a.scale.y*=2;
        add(a);
        // insert your code here
    }
}
```

---

**Observed behavior:** ![a picture](https://cloud.githubusercontent.com/assets/10850402/14416541/d99be9f4-ffdd-11e5-9951-ee3995f5d449.PNG)
The bounding box did not change, and text wasn't scaled.
##

buckle2000:
`FlxSpriteButton` is the same.

Gama11:
`scale` of a sprite doesn't automatically affect the hitbox. You can call `updateHitbox()` manually.

Automatically scaling the text would be a bit tricky regarding positioning..

Changing either of these behaviors could be a breaking change.

buckle2000:
I think the latter one is not a breaking change. It is what FlxTypedButton supposed to be.

Gama11:
Merged into #2047.

PR #1843, opened 2016-05-02
fix for the issue #1812

Beeblerox:
Render scaled button's label at correct position

### DIFF
--- a/flixel/ui/FlxButton.hx
+++ b/flixel/ui/FlxButton.hx
@@ -312,6 +312,8 @@ class FlxTypedButton<T:FlxSprite> extends FlxSprite implements IFlxInput
 		
 		if (_spriteLabel != null && _spriteLabel.visible)
 		{
+			_spriteLabel.scale.set(scale.x, scale.y);
+			updateLabelPosition();
 			_spriteLabel.cameras = cameras;
 			_spriteLabel.draw();
 		}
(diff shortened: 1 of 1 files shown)


flixel/ui/FlxButton.hx
```diff
@@ -312,6 +312,8 @@ class FlxTypedButton<T:FlxSprite> extends FlxSprite implements IFlxInput
 		
 		if (_spriteLabel != null && _spriteLabel.visible)
 		{
+			_spriteLabel.scale.set(scale.x, scale.y);
```
Gama11: I'm not sure this is a good idea... it breaks all code that sets the `scale` of the label directly, since those values will just always be overriden by the button's `scale` now.

Gama11:
This breaks quite a bit of stuff in RPGInterface / flixel-ui:







cc @larsiusprime

Beeblerox:
i guess we should change button's `scale` from `FlxPoint` to `FlxCallbackPoint`, so it would change label scale in callback methods.
plus we'll need some changes in flixel-ui code related to buttons.
will look into this next week

Gama11:
I wonder if it's really worth it... We not only break flixel-ui, which we can fix, but potentially also user code which relies on this behavior.

Gama11:
Closing this since there hasn't been any activity in a while and the changes breaks flixel-ui in their current state.

Issue #679, opened 2013-10-16
FlxTimer array manipulation bug


sinnus:
The issue is like #665 but now for timers. FlxTimer has complete callback https://github.com/HaxeFlixel/flixel/blob/dev/flixel/util/FlxTimer.hx#L167 where we can start new timer (add new timer into timers array), for example. Callback is invoked in TimerManager#update function https://github.com/HaxeFlixel/flixel/blob/dev/flixel/plugin/TimerManager.hx#L53. As you can see we can change array inside array iteration.

Beeblerox:
i think that my old commit should fix this issue

Gama11:
That pull request caused a unit test to hang if I remember correctly.

Beeblerox:
i'll try one more time then

PR #1845, opened 2016-05-03
fix for issue #679

Beeblerox:


flixel/util/FlxTimer.hx
```diff
@@ -156,16 +156,23 @@ class FlxTimer implements IFlxDestroyable
...
-				cancel();
-			}
+		}
+	}
+	
+	/**
+	 * Called by timer manager when timer finishes loop.
+	 */
+	@:allow(flixel.util.FlxTimerManager)
+	private inline function onLoop():Void
```
Gama11: I don't think this should be inline?

Gama11:
A unit test for this would be nice. I think we might already have one for tweens where we had the same issue?

Beeblerox:
@Gama11 cant' find test for that issue with tweens, and have no idea how to implement it

flixel/util/FlxTimer.hx
```diff
@@ -156,16 +156,23 @@ class FlxTimer implements IFlxDestroyable
...
-				cancel();
-			}
+		}
+	}
+	
+	/**
+	 * Called by timer manager when timer finishes loop.
+	 */
+	@:allow(flixel.util.FlxTimerManager)
+	private function onLoop():Void
```
Gama11: Should be `onLoopFinished()`, which I guess also makes the doc comment obsolete. :)

flixel/util/FlxTimer.hx
```diff
@@ -232,11 +241,31 @@ class FlxTimerManager extends FlxBasic
 	 */
 	override public function update(elapsed:Float):Void
 	{
+		var timerLoops:Int;
```
Gama11: I prefer to declare local vars as close to where they're actually needed as possible.

Gama11:
Closing; this fix causes `FlxTimerTest` to go into an infinite loop (which is why Travis was failing here).

PR #1846, opened 2016-05-09
Fixes to make the preloader work on HTML5

### LINKED ISSUE #1804: `FlxPreloader` doesn't work in HTML5.
As mentioned in #1803, `FlxPreloader` should (but doesn't) work in HTML5.

I'm not sure what specific issues need to be fixed (probably at least blend modes).

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/32526747-flxpreloader-doesn-t-work-in-html5?utm_campaign=plugin&utm_content=tracker%2F254189&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F254189&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Nallebeorn:
This modifies FlxPreloader and FlxBasePreloader to work on the HTML5 target. Because blend modes don't work on html5, the default preloader doesn't look quite as good as in Flash... but it runs. More importantly, FlxBasePreloader works fine, so custom preloaders can be made (which don't use blend modes :))
There are some quirks to be aware of with regard to embedded BitmapData and TextFields with custom fonts (and probably other things I haven't encountered yet) – this is documented to some extent in the comments in FlxPreloader.hx. Unfortunately, it inevitably leads to somewhat messier code, but I've tried to keep it as clean as possible.

Fixes #1804.

### DIFF
--- a/flixel/system/FlxBasePreloader.hx
+++ b/flixel/system/FlxBasePreloader.hx
@@ -163,20 +163,17 @@ class FlxBasePreloader extends NMEPreloader
 	 * @param	onLoad				Executed once the bitmap data is finished loading in HTML5, and immediately in Flash. The new Bitmap instance is passed as an argument.
 	 * @return  The Bitmap instance that was created.
 	 */
-	private function createBitmap(bitmapDataClass:Class<BitmapData>, ?onLoad:Bitmap->Void):Bitmap
+	private function createBitmap(bitmapDataClass:Class<BitmapData>, onLoad:Bitmap->Void):Bitmap
 	{
 		#if html5
-		if (onLoad != null && Type.getClassFields(bitmapDataClass).indexOf("preload") != -1)
-		{
-			var bmp = new Bitmap();
-			bmp.bitmapData = Type.createInstance(bitmapDataClass, [0, 0, true, 0xFFFFFFFF, function(_) onLoad(bmp)]);
-			return bmp;
-		}
-		#end
-		var bmp = new Bitmap(Type.createInstance(bitmapDataClass, [0, 0]));
-		if (onLoad != null)
-			onLoad(bmp);
+		var bmp = new Bitmap();
+		bmp.bitmapData = Type.createInstance(bitmapDataClass, [0, 0, true, 0xFFFFFFFF, function(_) onLoad(bmp)]);
 		return bmp;
+		#else
+		var bmp = new Bitmap(Type.createInstance(bitmapDataClass, [0, 0]));
+		onLoad(bmp);
+		return bmp;
+		#end
 	}
 	
 	/**
@@ -193,15 +190,12 @@ class FlxBasePreloader extends NMEPreloader
 	private function loadBitmapData(bitmapDataClass:Class<BitmapData>, onLoad:BitmapData->Void):BitmapData
 	{
 		#if html5
-		if (onLoad != null && Type.getClassFields(bitmapDataClass).indexOf("preload") != -1)
-		{
-			return Type.createInstance(bitmapDataClass, [0, 0, true, 0xFFFFFFFF, onLoad]);
-		}
-		#end
+		return Type.createInstance(bitmapDataClass, [0, 0, true, 0xFFFFFFFF, onLoad]);
+		#else
 		var bmpData = Type.createInstance(bitmapDataClass, [0, 0]);
-		if (onLoad != null)
-			onLoad(bmpData);
+		onLoad(bmpData);
 		return bmpData;
+		#end
 	}
 	
 	/**
--- a/flixel/system/FlxBasePreloader.hx
+++ b/flixel/system/FlxBasePreloader.hx
@@ -52,7 +52,7 @@ class FlxBasePreloader extends NMEPreloader
 	 * @param	MinDisplayTime	Minimum time (in seconds) the preloader should be shown. (Default = 0)
 	 * @param	AllowedURLs		Allowed URLs used for Site-locking. If the game is run anywhere else, a message will be displayed on the screen (Default = [])
 	 */
-	public function new(?MinDisplayTime:Float = 0, ?AllowedURLs:Array<String>)
+	public function new(MinDisplayTime:Float = 0, ?AllowedURLs:Array<String>)
 	{
 		super();
 		
--- a/flixel/system/FlxBasePreloader.hx
+++ b/flixel/system/FlxBasePreloader.hx
@@ -154,6 +154,57 @@ class FlxBasePreloader extends NMEPreloader
 	}
 	
 	/**
+	 * This should be used whenever you want to create a Bitmap that uses BitmapData embedded with the
+	 * @:bitmap metadata, if you want to support both Flash and HTML5. Because the embedded data is loaded
+	 * asynchronously in HTML5, any code that depends on the pixel data or size of the bitmap should be
+	 * in the onLoad function; any such code executed before it is called will fail on the HTML5 target.
+	 * 
+	 * @param	bitmapDataClass		A reference to the BitmapData child class that contains the embedded data which is to be used.
+	 * @param	onLoad				Executed once the bitmap data is finished loading in HTML5, and immediately in Flash. The new Bitmap instance is passed as an argument.
+	 * @return  The Bitmap instance that was created.
+	 */
+	private function createBitmap(bitmapDataClass:Class<BitmapData>, ?onLoad:Bitmap->Void):Bitmap
+	{
+		#if html5
+		if (onLoad != null && Type.getClassFields(bitmapDataClass).indexOf("preload") != -1)
+		{
+			var bmp = new Bitmap();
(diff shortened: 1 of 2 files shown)


flixel/system/FlxBasePreloader.hx
```diff
@@ -152,11 +154,62 @@ class FlxBasePreloader extends NMEPreloader
...
+	 * the @:bitmap metadata. Often, you'll want to use the BitmapData in a Bitmap object; in this case, createBitmap()
+	 * can should be used instead. Because the embedded data is loaded asynchronously in HTML5, any code that depends on
+	 * the pixel data or size of the bitmap should be in the onLoad function; any such code executed before it is called
+	 * will fail on the HTML5 target.
+	 * 
+	 * @param	bitmapDataClass		A reference to the BitmapData child class that contains the embedded data which is to be used.
+	 * @param	onLoad				Executed once the bitmap data is finished loading in HTML5, and immediately in Flash. The new BitmapData instance is passed as an argument.
+	 * @return  The BitmapData instance that was created.
+	 */
+	private function loadBitmapData(bitmapDataClass:Class<BitmapData>, onLoad:BitmapData->Void):BitmapData
```
Gama11: This function doesn't even seem to be used?

flixel/system/FlxBasePreloader.hx
```diff
@@ -152,11 +154,62 @@ class FlxBasePreloader extends NMEPreloader
...
+	 * in the onLoad function; any such code executed before it is called will fail on the HTML5 target.
+	 * 
+	 * @param	bitmapDataClass		A reference to the BitmapData child class that contains the embedded data which is to be used.
+	 * @param	onLoad				Executed once the bitmap data is finished loading in HTML5, and immediately in Flash. The new Bitmap instance is passed as an argument.
+	 * @return  The Bitmap instance that was created.
+	 */
+	private function createBitmap(bitmapDataClass:Class<BitmapData>, ?onLoad:Bitmap->Void):Bitmap
+	{
+		#if html5
+		if (onLoad != null && Type.getClassFields(bitmapDataClass).indexOf("preload") != -1)
```
Gama11: Can you explain why you are checking for a `preload` field? This seems quite hacky.

flixel/system/FlxBasePreloader.hx
```diff
@@ -109,8 +111,8 @@ class FlxBasePreloader extends NMEPreloader
 	 */
 	private function onEnterFrame(E:Event):Void
 	{
-		var time:Int = Lib.getTimer();
-		var min:Int = Std.int(minDisplayTime * 1000);
+		var time = Date.now().getTime() - _startTime;
```
Gama11: Why use `Date.now()` instead of  `Lib.getTimer()` - does the latter not work in HTML5?

Gama11:
Do you know what's responsible for the textfield being resized like this after a few frames?

flixel/system/FlxBasePreloader.hx
Nallebeorn: It isn't used in the default preloader, but I included it for the benefit of those writing custom preloaders, in case anyone wants to load an embedded bitmapdata without immediately creating a Bitmap object using it.

Gama11: Either way, the implementation is completely redundant. This method should be used in the preloader, wrapping the `BitmapData` in a `Bitmap` can easily be done there.

Also not sure this is really the right place for a utility method like this...

flixel/system/FlxBasePreloader.hx
Nallebeorn: It _is_ quite very hacky... When @:bitmap is used, OpenFl will (on html5) add the `onload` argument to the constructor, as well as the static field `preload`. So the constructor only has 5 argument on html5, and only in classes created with @:bitmap. That check is there to ensure we're not passing too many constructor arguments if for whatever reason a "normal" BitmapData class is passed.

flixel/system/FlxBasePreloader.hx
Nallebeorn: On Flash, Lib.getTimer() returns the time since the app was started, but on html5, it returns the current unix timestamp.

flixel/system/FlxBasePreloader.hx
Gama11: Can't we just assume that that argument is there? The docs you added even state that it's intended only for `@:bitmap` assets.

Nallebeorn:
No, I'm not sure what is causing that. Well, it's obviously because there's a delay until it starts using the correct font, but I don't know why that is. My guess is it's similar to the bitmap loading issues, i.e. it takes a while until the specified font has actually been loaded... But that's purely a guess.

Gama11:
Are you seeing the same effect / have you looked into possible fixes?

flixel/system/FlxBasePreloader.hx
Gama11: `FlxAssets` would probably be a better place for it (as `getEmbeddedBitmapData()` maybe)?

Gama11: Although "embedded" is a bit ambigious, there's also the option of embedding assets included via the project.xml...

Nallebeorn: Actually no... if you try to do something like
```haxe
var bitmap = new Bitmap(loadBitmapData(GraphicClass, function (bmp) bmp.width = 0));
```
it won't work, because the onLoad function will be passed the BitmapData, not the Bitmap. Of course, you can't reference the local bitmap var in the function. And if you make an instance variable to hold it, that will work on html5, but on Flash, the helper function won't be able to call the onLoad method automatically – because the variable won't have a value assigned to it yet.

Nallebeorn: Hmm... I wasn't quite sure where it would be best to put it either, I admit. But I thought FlxBasePreloader.hx made the most sense, because it's only needed in preloaders. FlxAssets might be a good idea, though...

Gama11: Hm, what a mess. Perhaps something like I did for the debugger buttons (see https://github.com/HaxeFlixel/flixel/blob/dev/flixel/system/debug/DebuggerUtil.hx#L41) would be better solution, that way you can just directly pass width and height. It uses Reflection, but it's hacky either way... And you could get rid of the callbacks I think?

It would be really nice to have a proper solution for openfl/openfl#682....

Nallebeorn:
I've seen that effect, yes... But for me, it has always resized so quickly I don't really notice it unless looking for it, and I've never seen it as that big a problem.
Actually, it's probably most noticeably in games that load really quickly, because the preloader often freezes a short while before starting the game – and that will be _before_ the TextField had changed.
I haven't really looked much into it, no. But I'll do some research.
freezes a short while before starting the game – and that will be _before_ the TextField had changed.

flixel/system/FlxBasePreloader.hx
Nallebeorn: I suppose... If you are fine with that, then sure :) I just thought it was best not to assume anything.

Gama11: If it's specified accordingly in the docs, it's definitely ok IMO.. especially if it avoids hacks like this.

Nallebeorn: Fair enough.

flixel/system/FlxBasePreloader.hx
Nallebeorn: Ah... But if I get this right, the bitmapdata's dimensions should be set to the actual size of the image that is loaded, right? (not the the size you want it to stretch to when displayed). This means we'd have to manually make sure to pass the correct values. I'm not sure that would really be cleaner or easier... :/

Gama11: That's correct. We already do that in the debugger, so at least it's a consistent workaround... And getting rid of the callback awkwardness counts for something too I guess?

Nallebeorn: Yeah, the callback system is pretty messy... But remember that it's not only about the default FlxPreloader, many people write custom preloaders. And having to manually supply all images' size (and keep it up to date whenever I switch images!) feels, at least to me, like it'd been even more awkward than the callback system.
But I dunno... I guess there isn't _any_ good solution.

Gama11: The good solution would be a fix at OpenFL level.

Nallebeorn: True. Although it's not obvious what a good solution on that level would be, either.

Gama11: Lars' solution seemed quite sensible to me:

> On export, have the tools
inspect all the bitmapdata objects for size and write that data out, so that we can
have an additional Assets.getSize() command that gives you the width/height
of a bitmapdata without actually loading it.

Gama11:
Thanks for your contribution! :) I merged the changes manually in 00c0e20 since there were merge conflicts by now.

PR #1847, opened 2016-05-10
FlxObject: allow configuring debug rect colors

starry-abyss:
So I want to cut the decorative (tile.allowCollisions == FlxObject.NONE) tiles' rects from debug draw: http://forum.haxeflixel.com/topic/8/flxtilemap-debug-draw/2
Desirable by default :-)


flixel/tile/FlxTilemap.hx
```diff
@@ -115,6 +115,8 @@ class FlxTilemap extends FlxBaseTilemap<FlxTile>
 	private var _scaledTileHeight:Float = 0;
 	
 	#if FLX_DEBUG
+	public var debugHideDecorativeTiles:Bool = true;
```
Gama11: I don't like the idea of introducing new terminology for this (what is considered a "decorative" tile? seems a bit ambigious). In general, I would prefer a more flexible solution (setting the color per debug tile type like you originally proposed sounds good, that way you'd have full control, similar to `debugBoundingBoxColor` in `FlxObject`. Which makes me wonder what that does for `FlxTilemap`, if anything...).

starry-abyss:
Ok, I'll work further on it. Can you provide old terminology maybe? This is from docs:

> The behavior of tiles in FlxTilemap is slightly different:

    Blue for allowCollisions == FlxObject.NONE
    Green for allowCollisions == FlxObject.ANY
    Pink for other values of allowCollisions

Should I maybe make a function var like Int -> FlxColor ? E.g. collision flags to color converter, and refactor the original code as the default function.
@Gama11 what do you think about hiding blue ones by default?

starry-abyss:
As a variant I could also draw lines instead of rects to indicate the sides of enabled collisions :-) That would simplify three-color scenario to two-color. One could be changed by current debugBoundingBoxColor, other - by adding second color parameter.

Gama11:
I had something like `Map<FlxDebugTileType, FlxColor>` in mind, but that requires setting up an entire enum. Using `allowCollisions` is a lot more elegant. Not sure whether a map or a function is better. Are there any particular advantages to using a function?

I'm not opposed to hiding them by default. I'll have to think about it a bit.

Gama11:
Hm, that sounds interesting as well. I'd like to see how that looks! :)

starry-abyss:
Can't say about particular function advantages...

I tried different coloring approaches, this is the least confusing so far (with green lines, but also with pink rects):

http://i.imgur.com/bsJGVoC.png

But I still have user-coloring not implemented yet :D

starry-abyss:
But it draws collider lines inside while they seem to have more meaning outside of the tile (in the opposite direction) :/

Gama11:
Drawing the "collider lines" on the outer rectangl doesn't work well?

starry-abyss:
Did the outer version, it feels natural to me:

http://i.imgur.com/rSMy8xX.png

Gama11:
I don't like how that makes the debug draw bigger than the tile bounds. What I meant was drawing the collision lines _on top of_ the bounding rectangle.

starry-abyss:
But you don't see the direction that way - to which of neighbouring tiles the edge belongs среда, 11 мая 2016г., 01:12 +0300 от Gama11 < notifications@github.com> :

starry-abyss:
So that's the benefit of using customizable function: you can provide your the renderer for one tile without copy-pasting whole drawDebugOnCamera, and this function can be theoretically reused for standalone FlxObjects. среда, 11 мая 2016г., 01:12 +0300 от Gama11 < notifications@github.com> :

starry-abyss:
@Gama11 so I post the version you requested, that I shot during the experimenting: http://i.imgur.com/NjGNuZZ.png
And now I have a common drawDebugBoundingBox subfunction between FlxObject and FlxTilemap. (except renderBlit mode, which I didn't touch yet)
The only difference is in FlxObject typedef-variable, which is set in constructor:
// FlxObject
```
debugColorScheme = { solid: FlxColor.RED, highlighted: FlxColor.GREEN, notSolid: FlxColor.BLUE };
```
// FlxTilemap
```
debugColorScheme = { solid: FlxColor.PINK, highlighted: FlxColor.GREEN, notSolid: null };
```

But I mixed the labels :D Solid and highlighted are inverted for FlxTileMap!
Also I can override now just the small drawDebugBoundingBox to provide fancy collision edges. Not sure if I need the callback function I proposed here, postponing it for now :-)

starry-abyss:
Now you can change colors for tiles in Flash with limitations:
1. Apparently it's slow
2. Colors are only updated when you change the whole debugColorScheme, and not only a field of it

Also I think about customization of alpha values too.

starry-abyss:
I hacked Mode demo a bit here to test (compile in Debug mode): https://github.com/starry-abyss/flixel-demos/tree/tilemap_debug_color_test/Platformers/Mode
Press E to randomly change color scheme for tilemap debug draw.
FPS drops a bit in neko, not sure if it's my fault, any recommended way of finding leaks, profiling in Haxe?

Gama11:
Could you enable this new checkbox? https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/#enabling-repository-maintainer-permissions-on-existing-pull-requests

starry-abyss:
@Gama11, sure, did it work?

Gama11:
Looks like it did, thanks. :)

Gama11:
I checked your hacked Mode demo - it seems like after pressing E, you still need to move enough for the tilemap to be redrawn for color changes to show? Didn't you implement the setter for this?

starry-abyss:
Oh, cool, back then I thought updating color was deferred! Looks like you have to move for it to update. I'm newbie in Flash (and original Flixel) development, but what setter are you referring to and do you want me to implement it?

Gama11:
I meant the `set_debugColorScheme` setter you already added - I thought that was the reason you added it, in fact. :)

starry-abyss:
I think I override it in some derived class. Apparently, I need to call something else in the setter for Flash to work properly?

starry-abyss:
I changed the key for switching sprites' color scheme in the demo to E too. It's applied immediately, but tilemap's is not (on neko both are OK).

Beeblerox:
So, what should be done to make decision about this PR? What's missing?

Gama11:
There are some things I don't like about the API:

- I don't like the name `highlighted` in `DebugColorScheme`. I can't really think of something better right now though
- I dislike that you can change the colors, but they won't be updated if you don't trigger the setter (reassign the structure)

Also, with blitting, the colors don't update right away even if you do trigger the setter. See https://github.com/HaxeFlixel/flixel/pull/1847#issuecomment-246445620.

starry-abyss:
For some reason colors are updated on Flash only when you move camera

flixel/FlxObject.hx
```diff
@@ -540,6 +549,17 @@ class FlxObject extends FlxBasic
 	 * Overriding this will force a specific color to be used for debug rect.
 	 */
 	public var debugBoundingBoxColor:Null<Int> = null;
+	
+	public var debugColorScheme(default, null):DebugColorScheme;
+	
+	/**
+	 * Can be overriden for RenderBlit mode to re-render stuff with new colors
+	 */
+	public function setDebugColorScheme(debugColorScheme:DebugColorScheme)
```
Beeblerox: why not change this method to debugColorScheme setter?
```
public var debugColorScheme(default, set):DebugColorScheme;

public function set_debugColorScheme(value:DebugColorScheme):DebugColorScheme
{
 	return debugColorScheme = value;
}
```

flixel/FlxObject.hx
```diff
@@ -1034,26 +1057,33 @@ class FlxObject extends FlxBasic
 		}
 		
 		var rect = getBoundingBox(camera);
-		
+		var gfx:Graphics = beginDrawDebug(camera);
+		drawDebugBoundingBox(gfx, rect, allowCollisions, immovable);
+		endDrawDebug(camera);
+	}
+	
+	function drawDebugBoundingBox(gfx:Graphics, rect:FlxRect, allowCollisions:Int, highlight:Bool)
```
Beeblerox: maybe rename `highlight` parameter to `partial` as well?

flixel/tile/FlxTilemap.hx
```diff
@@ -128,9 +128,18 @@ class FlxTilemap extends FlxBaseTilemap<FlxTile>
 	private var _scaledTileHeight:Float = 0;
 	
 	#if FLX_DEBUG
-	private var _debugTileNotSolid:BitmapData;
-	private var _debugTilePartial:BitmapData;
-	private var _debugTileSolid:BitmapData;
+	private var updateDebugTileBoundingBoxesInit = false;
```
Beeblerox: i'd rather remove this var and check if `_tileWidth` and `_tileHeight` are initialized (have values more than 0).

so we could have something like:
```
override private function set_debugColorScheme(value:DebugColorScheme):DebugColorScheme
	{
		this.debugColorScheme = value;
		
		if (_tileWidth > 0 && _tileHeight > 0)
		{	
			updateDebugTileBoundingBoxes();
		}
		
		return value;
	}
```

flixel/tile/FlxTilemap.hx
```diff
@@ -290,14 +301,42 @@ class FlxTilemap extends FlxBaseTilemap<FlxTile>
...
+	
+	#if FLX_DEBUG
+	private function updateDebugTileBoundingBoxes():Void 
+	{
 		if (FlxG.renderBlit)
 		{
-			_debugTileNotSolid = makeDebugTile(FlxColor.BLUE);
-			_debugTilePartial = makeDebugTile(FlxColor.PINK);
-			_debugTileSolid = makeDebugTile(FlxColor.GREEN);
+			if (_debugTileSolid == null)
```
Beeblerox: too much repetitions.

Maybe it would be better have this?
```
private function updateDebugTileBoundingBoxes():Void 
	{
		if (FlxG.renderBlit)
		{
			_debugTileSolid = updateDebugTile(_debugTileSolid, debugColorScheme.solid);
			_debugTilePartial = updateDebugTile(_debugTilePartial, debugColorScheme.partial);
			_debugTileNotSolid = updateDebugTile(_debugTileNotSolid, debugColorScheme.notSolid);
		}
	}
	
	private function updateDebugTile(tileBitmap:BitmapData, color:FlxColor):BitmapData
	{
		if (tileBitmap != null && (tileBitmap.width != _tileWidth || tileBitmap.height != _tileHeight))
			tileBitmap = FlxDestroyUtil.dispose(tileBitmap);
		
		if (tileBitmap == null)
			tileBitmap = makeDebugTile(color);
		else
		{
			tileBitmap.fillRect(tileBitmap.rect, FlxColor.TRANSPARENT);
			drawDebugTile(tileBitmap, color);
		}
		
		return tileBitmap;
	}
```

flixel/tile/FlxTilemap.hx
```diff
@@ -128,9 +128,18 @@ class FlxTilemap extends FlxBaseTilemap<FlxTile>
...
+	private var updateDebugTileBoundingBoxesInit = false;
+	
+	override public function setDebugColorScheme(debugColorScheme:DebugColorScheme)
+	{
+		this.debugColorScheme = debugColorScheme;
+		if (updateDebugTileBoundingBoxesInit)
+			updateDebugTileBoundingBoxes();
+	}
+	
+	private var _debugTileNotSolid:BitmapData = null;
```
Beeblerox: and don't forget to dispose these bitmaps in `destroy()` method. Change lines
```
_debugTileNotSolid = null;
_debugTilePartial = null;
_debugTileSolid = null;
```
to 
```
_debugTileNotSolid = FlxDestroyUtil.dispose(_debugTileNotSolid);
_debugTilePartial = FlxDestroyUtil.dispose(_debugTilePartial);
_debugTileSolid = FlxDestroyUtil.dispose(_debugTileSolid);
```

(this don't have to be the part of this PR, but it will make code a tiny-tiny bit better :)

flixel/FlxObject.hx
starry-abyss: Because as Gama11 said it triggers on assigning the structure, but doesn't on field value change

Beeblerox: i don't quite understand this statement. Could you explain?

I've made this change in my fork and it works like i expect it to work. Maybe there is some misunderstanding?

starry-abyss: What you propose is the original way I've done it.
On native targets it works fine because it just gets the colors every frame and uses them.
On Flash (+ FlxTilemap) there is some update stuff to be done on each change.

starry-abyss: Apparently `_tileMap.debugColorScheme.solid = Std.random(0xFFFFFF);` doesn't trigger the update

Beeblerox: of course it doesn't trigger it, but this is expected. Various classes have such behavior, for example you need to reset filters for display objects every time you change properties of one of the filters.
So this isn't bug in my opinion. We just need to document this behavior.

You can have the same "issue" with this code:
```
var colorScheme = { solid: FlxColor.RED, partial: FlxColor.GREEN, notSolid: FlxColor.BLUE };
object.setDebugColorScheme(colorScheme);
colorScheme.solid = Std.random(0xFFFFFF);
```
this example won't trigger update as well. So that's why i'm for original way (with `debugColorScheme` as a property with setter)

starry-abyss: I agree with @Gama11 that it's a bit confusing this way :-/ Maybe he has better idea?

Gama11: Ony way to work around this is to simply have a separate class field for each color.

Beeblerox: yes, it would be a simple solution.

Beeblerox: @starry-abyss so do you agree to add 3 debug color properties instead of `sebugColorScheme`?

starry-abyss: Yeah, no problem

starry-abyss:
@Beeblerox So... any clue how to force tile update on Flash without waiting for camera movement?

flixel/FlxObject.hx
```diff
@@ -548,16 +539,26 @@ class FlxObject extends FlxBasic
 	/**
 	 * Overriding this will force a specific color to be used for debug rect.
 	 */
-	public var debugBoundingBoxColor:Null<Int> = null;
-	
-	public var debugColorScheme(default, null):DebugColorScheme;
+	public var debugBoundingBoxColor(default, set):Null<Int> = null;
+	public var debugBoundingBoxColorNotSolid(default, set):Null<Int> = null;
+	public var debugBoundingBoxColorPartial(default, set):Null<Int> = null;
```
Gama11: I think these should be `FlxColor` instead of `Null<Int>`? And also be initialized with their default values right away?

starry-abyss: I'm not sure. I want to use `debugBoundingBoxColor` which is the original field prior to my PR. Won't changing the type cause any breaking? Or should I not mess with the original field at all and should create a new one?

Gama11: I'm not sure that one should be touched at all. Isn't the use-case for that "use the color I assigned to it no matter what value for `solid` etc is"?

starry-abyss: OK, makes sense

PR #1848, opened 2016-05-11
GLSL shaders support for individual objects (sprites, tilemaps, etc.)

Beeblerox:


flixel/graphics/tile/FlxDrawTilesItem.hx
```diff
@@ -91,7 +95,11 @@ class FlxDrawTilesItem extends FlxDrawBaseItem<FlxDrawTilesItem>
 			#end
 			
 			tempFlags |= blending;
+			#if (!openfl_legacy && openfl >= "3.3.9")
+			camera.canvas.graphics.drawTiles(graphics.tilesheet, drawData, (camera.antialiasing || antialiasing), tempFlags, shader, position);
```
Gama11: Why do we draw to `camera.canvas.graphics` in this case and `graphics.tilesheet` in the other?

Beeblerox: ok, i'll replace both lines, so they use `camera.canvas.graphics.drawTiles`

Gama11: Ideally there's only one line, with only the shaders argument conditionalized, to avoid redundancy.

PR #1850, opened 2016-05-12
add antialiasing to FlxTilemap

larsiusprime:


flixel/tile/FlxTilemap.hx
```diff
@@ -1013,6 +1018,11 @@ class FlxTilemap extends FlxBaseTilemap<FlxTile>
 	}
 	#end
 	
+	private function set_antialiasing(value:Bool):Bool
```
Gama11: Why do we need a setter here?

Gama11:
Does this work with both `FlxG.renderTile` and `FlxG.renderBlit`?

flixel/tile/FlxTilemap.hx
larsiusprime: I'm just following the convention from FlxSprite

larsiusprime:
@Gama11: It works with rendertile, not sure about renderblit. Let me see.

larsiusprime:
Okay, looks like it doesn't -- in BLIT mode, the draw call happens on the FlxTilemapBuffer, which passes a null (and thus false) value for "smoothing" to draw().

Serendipitously, this creates a non-speculative use for the antialiasing setter on the tilemap -- we need to add an antialiasing flag to the FlxTilemapBuffers, and that's a natural place to update it.

Gama11:
I was about to tell you that there's a reason why `FlxSprite` has an empty setter, namely that it's overriden by `FlxSpriteGroup`. Actually looking into that though, that doesn't seem to be the case...

Gama11:
Hm, looks like it was needed by the old version of the bitmap text class, but apparently not anymore in the refactored `FlxBitmapText`: https://github.com/HaxeFlixel/flixel/commit/f4829a8f8f68d1ba93f8d866748cc3a106d76963

larsiusprime:
Okay, I have the change for Blit, tested and works. one second.

flixel/tile/FlxTilemap.hx
```diff
@@ -970,6 +975,10 @@ class FlxTilemap extends FlxBaseTilemap<FlxTile>
 	{
 		var buffer = new FlxTilemapBuffer(_tileWidth, _tileHeight, widthInTiles, heightInTiles, camera, scale.x, scale.y);
 		buffer.pixelPerfectRender = pixelPerfectRender;
+		if (FlxG.renderBlit)
```
Beeblerox: you can skip this check for render mode and just set `antialiasing` directly

flixel/tile/FlxTilemap.hx
```diff
@@ -1013,6 +1022,19 @@ class FlxTilemap extends FlxBaseTilemap<FlxTile>
 	}
 	#end
 	
+	private function set_antialiasing(value:Bool):Bool
+	{
+		antialiasing = value;
+		if (FlxG.renderBlit)
```
Gama11: Same with this check I guess.

Issue #1836, opened 2016-04-26
Pre/Post Update and Draw signals should include elapsed


JoeCreates:
Pre/Post Update and Draw signals should include elapsed. This time will always be calculated by the time the calls are required, and this information is useful in many cases, such as doing late updates to physics.

It seems FlxG.elapsed could be used in most cases, but not in the case of preUpdate because elapsed isn't calculated until after preUpdate is dispatched.

If the value is passed instead, it would also make the listeners more flexible.
##

Gama11:
Hm, this is unfortunately a breaking change...

I think we had those signals before there was an `elapsed` argument in `update()`.

JoeCreates:
As an intermiary non-breaking improvement, the dispatch of the preupdate could be moved to after calculation of FlxG.elapsed, so that at least preUpdate listeners could use that.

Gama11:
That seems reasonable.

MSGhero:
@Gama11 can you think of any reason these lines can't be moved up to the beginning of the function, after the early exit? https://github.com/HaxeFlixel/flixel/blob/dev/flixel/FlxGame.hx#L725-L736

It's breaking if someone needed to use last frame's `elapsed` in this frame's `preUpdate`, but no one has probably ever thought about that before. I don't see coming before `switchState` being a problem.

Gama11:
Why do you need `elapsed` during `switchState()` / what's wrong with Joe's suggestion?

MSGhero:
If you can't come up with a reason why, then Joe's suggestion is good and I can make a quick PR. I wanted to make sure I wasn't overlooking anything.

Gama11:
Go ahead.

Geokureli:
Clearing out 5.0.0 tasks, was this completed? I see preUpdate is called after FlxG.elapsed is set, we don't pass elapsed in draw calls, but do we want to pass it on pre/post update signals? I could go either way, honestly

Geokureli:
Gonna close this out, for pre/post update they can just use FlxG.elapsed, and draw calls shouldn't use elapsed to begin with

PR #1853, opened 2016-05-13
Calculate elapsed third

MSGhero:
#1836 

So that `preUpdate` signals can have the current `elapsed`.

### DIFF
--- a/flixel/FlxGame.hx
+++ b/flixel/FlxGame.hx
@@ -708,6 +708,19 @@ class FlxGame extends Sprite
 			return;
 		}
 		
+		if (FlxG.fixedTimestep)
+		{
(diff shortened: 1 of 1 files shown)


Gama11:
Hm.. while we're at it, this code should probably be moved into a separate function. ;)

Also, I thought you were gonna implemnet Joe's suggestion (i.e. just swap with the preUpdate signal dispatch), not yours?

MSGhero:
Separate function within FlxGame? And I honestly didn't think about where I placed it when I dragged the code up. Would you rather it go below `ticks` and above `preUpdate`?

Gama11:
Yeah, less chance to break things if the change is smaller I guess.

No, just a private method (`updateElapsed()` I guess?).

Gama11:
Thanks! :+1:

PR #1860, opened 2016-05-27
Change shuffleArray to use unbiased algorithm

16th-earl:
For proper shuffling when making card games or anything similar, it's necessary to have an unbiased shuffle algorithm so that every permutation of the array is equally likely. The current shuffleArray function doesn't provide an unbiased result, and depending on the HowManyTimes parameter may do a shuffle of varying quality. This modification replaces the shuffleArray routine with the Fisher-Yates algorithm. It maintains compatibility with current user code by still accepting the HowManyTimes parameter, even though it's not needed anymore.

### DIFF
--- a/flixel/math/FlxRandom.hx
+++ b/flixel/math/FlxRandom.hx
@@ -300,34 +300,37 @@ class FlxRandom
 		return selected;
 	}
 	
-	/**
-	 * Shuffles the entries in an array into a new pseudorandom order.
-	 * 
-	 * @param   Objects        An array to shuffle.
-	 * @param   HowManyTimes   How many swaps to perform during the shuffle operation. 
-	 *                         A good rule of thumb is 2-4 times the number of objects in the list.
+    /**
+	 * Shuffles the entries in an array into a new pseudorandom order, using an 'unbiased' shuffle.
+	 * This means that every permutation of the array is equally likely.
+	 *
+	 * @param   Objects        An array to shuffle in-place.
+	 * @param   HowManyTimes   This parameter is no longer relevant and should not be used.
+	 *                         However if a value of 0 (or less) is specified then no shuffle will be done.
 	 * @return  The newly shuffled array.
 	 */
-	@:generic
-	public function shuffleArray<T>(Objects:Array<T>, HowManyTimes:Int):Array<T>
-	{
-		HowManyTimes = Std.int(Math.max(HowManyTimes, 0));
-		
-		var tempObject:Null<T> = null;
-		
-		for (i in 0...HowManyTimes)
-		{
-			var pick1:Int = int(0, Objects.length - 1);
-			var pick2:Int = int(0, Objects.length - 1);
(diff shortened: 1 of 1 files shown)


Gama11:
Hmm... I think this is still technically a breaking changes since users might rely on getting the same results from this function with replay files.

MSGhero:
What would this new function be called? `shuffleArraySuperRandomly<T>`? @Gama11

16th-earl:
I wonder if it could just be named `shuffle<T>`? Python and PHP both have a shuffle function (though the Python one shuffles any sequence, not just arrays); and the Java Collections class has a shuffle method.

sruloart:
You could add an (optional) ```?unbiased``` bool to ```shuffleArray<T>(...)``` that leads to this commit algorithm.

16th-earl:
Another possibility… how about if the HowManyTimes parameter was changed to a nullable `?HowManyTimes:Int`? If an integer is passed then the old shuffling behaviour occurs, but if the value is omitted (null) then the Fisher-Yates shuffle is used. Would this still be considered a breaking change?

MSGhero:
@16th-earl people seem to avoid nullable ints for some reason, but I would prefer `shuffle()` to `shuffle(-1)` if `1` has to be the default parameter, and `-1` or `0` was how you told it to do the normal algo. So yeah, nullable int would not be breaking.

Gama11:
I don't like the idea of changing the algorithm used based on the value of some parameter, it's way **too implicit**. Feel free to open a new PR if you still feel this is an important addition. `shuffleFisherYates()` might be an appropriate function name if that's the official name of the algorithm I suppose?

Also, could you explain what makes the current algorithm biased and the new one unbiased (or link to an explanation)?

JoeCreates:
Wow. I'm actually surprised that the old algorithm was ever included and has been here so long. It's a common newbie mistake when implementing shuffling and I'd personally prefer to see a breaking change that removes it from flixel.

Fisher-Yates is pretty standard and guarantees a result where every permutation is equally likely by doing exactly one swap for every slot in the array. It means that every slot will have been swapped to be a random element in only n iterations where n is the array size.

The existing algorithm is a very naive approach that does far more work for a statistically less random result. This is the case because if you loop only n times then many slots will likely not be swapped at all. To alleviate this it suggests you  perform 4n swaps even though this doesn't solve the problem entirely. Really the existing algorithm makes no sense at all and I wouldn't recommend anyone use it.

I also feel that adding in additional methods would make the api more messy and less intuitive. The shuffle <T> option with the deprecated tag would best best otherwise, provided we can eventually remove the deprecated version as its inclusion is a little embarrassing.

Gama11:
I see. I think we'll have to deprecate it initially, a breaking change isn't really acceptable here since it has the potential to break backwards compatibility with all replay files that rely on shuffling. With `@deprecated`, every usage should cause a compiler warning.

PR #1862, opened 2016-06-03
Debugger: add "FlxObjects-are-draggable"-mode

Dovyski:
This pull request contains a port of Flixel community's [interactive debug](https://github.com/FlixelCommunity/flixel/pull/217) feature. The new "interactive debug" mode can be activated by clicking the pencil icon at the top of the debugger.

Available tools (functionalities) during interaction:
- Pointer: select this tool on the toolbar on the right, then click any object on the screen. Multiple objects can be selected by holding CTRL during the clicks;
- Mover: move selected object(s) by holding SHIFT while clicking and dragging the objects.
- Eraser: delete the selected objects. It's also possible to delete selected objects by pressing DELETE; by using SHIFT + DELETE the object will be deleted and removed from memory.

This implementation has the idea of a toolbar (appears at the left of the screen when the interactive debug is activated). I am waiting for feedback on this. If you guys like the idea, I can fix any bugs and improve the toolbar (make it look more like a toolbar), otherwise I can remove it all together.

I implemented the toolbar because I envision more tools in the future, like a tile manipulator, rotation, etc.

***NOTICE***: a lot of the methods I created are not documented. I will add documentation after the review process has settled all details/changes.

Here is a gif of the whole thing working:

interactive_debug_move_delete


Gama11:
On Neko, Mode seems to crash immediately:

```
Invalid operation (-)
Called from flixel.system.debug.interaction.Interaction::keyJustPressed line 315
Called from flixel.system.debug.interaction.tools.Eraser::update line 36
Called from flixel.system.debug.interaction.Interaction::preUpdate line 167
Called from flixel.util._FlxSignal.FlxSignal0::dispatch0 line 292
Called from flixel.FlxGame::update line 723
Called from flixel.FlxGame::step line 663
Called from flixel.FlxGame::onEnterFrame line 530
Called from openfl._legacy.events.EventDispatcher::dispatchEvent line 98
Called from openfl._legacy.display.DisplayObject::__dispatchEvent line 182
Called from openfl._legacy.display.DisplayObject::__broadcast line 161
Called from openfl._legacy.display.DisplayObjectContainer::__broadcast line 286
Called from openfl._legacy.display.Stage::__render line 1103
Called from openfl._legacy.display.Stage::__checkRender line 351
Called from openfl._legacy.display.Stage::__pollTimers line 1084
Called from openfl._legacy.display.Stage::__doProcessStageEvent line 430
```

Gama11:
The UI is a bit confusing to me. 100% alpha seems to mean that the tool is not active, while a lowered alpha means that the tool _is_ active? I'd have expected this to behave the other way around, similar to the debugger window toggle buttons.

I'm also not sure how it makes sense that multiple tools can be selected at once. I haven't really been able to make the eraser work. And the cursor only ever seems to change to the first tool?

It also seems like whenever the cursor changes to the first tool, the system cursor is also shown, which is distracting because of the mouse lag.

Gama11:
> a lot of the methods I created are not documented. I will add documentation after the review process has settled all details/changes.

I wouldn't worry about documentation too much. These are just internal debugger methods (not API), there isn't a whole lot to be documented about them as long as they are aptly named.

Dovyski:
> On Neko, Mode seems to crash immediately:

I forgot to check Neko (and all other platforms but Flash). I'm sorry, single platform habit. I will fix the problem.

> The UI is a bit confusing to me. 100% alpha seems to mean that the tool is not active, while a lowered alpha means that the tool is active? I'd have expected this to behave the other way around, similar to the debugger window toggle buttons.

Yeah, the UI is not final. I want to replace the current alpha buttons with "real" buttons (the ones with background, border, etc). I think it will make everything easier to understand/use. I will only go for that if you approve the idea of a toolbar.

> I'm also not sure how it makes sense that multiple tools can be selected at once. I haven't really been able to make the eraser work. And the cursor only ever seems to change to the first tool?

Those are all bugs that I will fix after we decide on the use of a toolbar or not.

> It also seems like whenever the cursor changes to the first tool, the system cursor is also shown, which is distracting because of the mouse lag.

Hum, I need to add some custom mouse handling code to the debugger to fix that (as we previously discussed).

Gama11:
I was wondering whether it makes sense to have a dropdown with the available tools open upon clicking the pencil button instead of having a dedicated window for it - with the assumption that only one tool can be active at a time, I could see that working.

Dovyski:
I think that approach is a bit confusing. If I want to use the pointer tool, for instance, I click the pencil then click the pointer. Will the pointer icon be displayed instead of the pencil to indicate the currently active tool? What if I am in the middle of clicking and moving things around and I decide to stop the editing process altogether, should I click the pencil tool again?

As a side note, it will require two clicks to activate a tool (which could be fixed by showing the dropdown when the cursor hovers the icon, but it still doesn't fix the problem I described above).

I think the current window approach allows more freedom for the user to organize things, especially if more tools are added.

Gama11:
Fair enough.

Dovyski:
I think I am finally done with it :tada: I've fixed all bugs I could find and polished the interaction mode a bit (UI, usability). I will wait for feedback now.

Gama11:
- the tools shouldn't be active by default when it's finished (I assumed you kept that for easier debugging until now). I'd also prefer if the pencil icon in the top bar was positioned to the left of the bitmap log.
- ocassionally, there's a subtle flickering of the cursor (flickering between the selection and Mode's cursor I think)



- I'm not sure I really like the new graphic for the selection cursor, any particular reason you replaced the old one?
- it can be surprsingly hard to select another tool with the "move" cursor, I think this comes from the cursor still using the top left as the cursor position while visually, it seems like the center should be what you "click with"
- the tools window has a lot of empty space caused by the window title having a certain width - this could be fixed by aligning the tools horizontally instead of vertically
 


- debugger windows have a certain minimum distance to the window edges, 2 px or something. the initial tools window positioning doesn't respect that (x=0)
- the pencil icon could use a bit of antialiasing
- the general workflow seems a bit cumbersome - "select the selection tool, select an object, select the move tool, then move it". Couldn't we just skip the first two steps by moving whatever object is under the cursor when the move tool is selected? wouldn't even need the first tool that way.
- if I understand correctly, the eraser tool works unlike the other tools in that it doesn't change the cursor, it just acts on the currently selected object. This seems counter-intuitive, as it's presented the same way in the UI as the other tools. It's also greyed out (meaning that it can never be not greyed out / "selected"?)
- the delete and move tools should probably work on tiles as well
- it's quite awkward to move objects that have a velocity - perhaps any motion updates could be disabled while an object is moved? perhaps even "pause" the object entirely by setting it's active to false? This could have unintended side effects though.
- it seems like objects can't be selected while the game is paused via the vcr button in the middle (this is quite an important use case IMO - pause the game, move a few objects, resume is probably a more efficient workflow than trying to do that while the game is in progress in a lot of cases).
- selecting a `FlxButton` still counts as a regular click (effectively making it impossible to select the "play" button in Mode)

Dovyski:
Thanks for all the feedback!

> the tools shouldn't be active by default when it's finished (I assumed you kept that for easier debugging until now). I'd also prefer if the pencil icon in the top bar was positioned to the left of the bitmap log.

Ok, that's easy to fix.

> ocassionally, there's a subtle flickering of the cursor (flickering between the selection and Mode's cursor I think)

That's caused by the way the debugger handles the mouse cursor (show it when over a window, hide it when moving anywhere else). At the moment, I override such cursor changes  by updating the mouse cursor every frame.

In order to fix that, I need to re-work the mouse code in the debugger (add a flag/property to control how the mouse should show/hide when interacting with window, etc). I can do that if you think the current flickering of the cursor is a big problem.

> I'm not sure I really like the new graphic for the selection cursor, any particular reason you replaced the old one?

I am using native cursors for Flash, so I am not able to position the cursor icon at `(-width / 2, -height / 2)`, only `(0, 0)` (top left-corner). I think it can be worked by doing some math with the cursor position before doing any colision/selection calculations.

> it can be surprsingly hard to select another tool with the "move" cursor, I think this comes from the cursor still using the top left as the cursor position while visually, it seems like the center should be what you "click with"

Yeah, the same problem I described above. I will try to work this out and keep the "click point" at the center of the cursor icon.

> the tools window has a lot of empty space caused by the window title having a certain width - this could be fixed by aligning the tools horizontally instead of vertically

I agree, but I was following the design principals of other tools such as Photoshop, Gimp, etc. If we add more tools in the future, the tools window will grow too wide with an horizontal alignment approach. Maybe by placing two tools per row and let it grow vertiaclly will solve the empty spaces and still give some room for future grow? 

> debugger windows have a certain minimum distance to the window edges, 2 px or something. the initial tools window positioning doesn't respect that (x=0)

Ok.

> the pencil icon could use a bit of antialiasing

Ok.

> the general workflow seems a bit cumbersome - "select the selection tool, select an object, select the move tool, then move it". Couldn't we just skip the first two steps by moving whatever object is under the cursor when the move tool is selected? wouldn't even need the first tool that way.

Yeah, you can do that by using keyboard shortcuts. You can move  selected objects at any time by holding SHIFT then clicking and dragging the object. The problem is that if you click an empty space, even holding SHIFT, it will clear any previously existent selection.

The move tool prevents that: after selecting object(s), you can use the move tool no matter where you click on the screen. The move tool is more of a "fine tuning" tool.

> if I understand correctly, the eraser tool works unlike the other tools in that it doesn't change the cursor, it just acts on the currently selected object. This seems counter-intuitive, as it's presented the same way in the UI as the other tools. It's also greyed out (meaning that it can never be not greyed out / "selected"?)

I agree. Maybe I can turn the eraser tool into a button in a different place, or just hide it and allow the user to delete things only by pressing delete.

> the delete and move tools should probably work on tiles as well

Maybe that's a question of class hierarquy that is preventing the tiles from being selected. I will check that, but depending on the structure of the classes, anything tile-related might need its dedicated "tile tool" in the future.

> it's quite awkward to move objects that have a velocity - perhaps any motion updates could be disabled while an object is moved? perhaps even "pause" the object entirely by setting it's active to false? This could have unintended side effects though.

That's a tricky thing to do. The debugger will completely interfere with the game, so we will need to keep track of the changes in order to restore them after the interaction is over. What if a class overrided the `active` setter and we keep on calling it during the debug interaction? 

> it seems like objects can't be selected while the game is paused via the vcr button in the middle (this is quite an important use case IMO - pause the game, move a few objects, resume is probably a more efficient workflow than trying to do that while the game is in progress in a lot of cases).

I agree and I've tried to implement that, but I couldn't do it easily. I need more time to investigate that approach better.

> selecting a FlxButton still counts as a regular click (effectively making it impossible to select the "play" button in Mode)

That's probably another tricky thing to fix, because I need to somehow intercept any mouse event performed in the engine when the debugger/interaction is active. Right now I just added a new event listener for mouse events, but I probably need something more sophisticated.

Dovyski:
I've made the changes you asked for, @Gama11  

- No more active tools by default when use is finished;
- Fix all mouse issues (flickering, hard to hit clicks, etc);
- Tools window is now horizontal (no empty spaces);
- Add 2px screen padding;
- Pencil icon has more antialiasing and is placed to the left of bitmap log;
- Interactive debug works with vcr (e.g. when game is paused or a replay is being played);
- Removed the eraser tool from tools bar (it's still possible to delete things by pressing DEL on keyboard though);

Some thoughts:

- Buttons can be selected/moved using the interactive tools when the game is paused. This is the same behavior when any other interaction happens when the game is running (e.g. moving the main character in Mode is affected by gravity).

- Selecting tiles/tilemaps will require a dedicated interaction tool because there are too many corners cases to cover in the current pointer tool.

- I want to create another window to house actions to be performed when something is selected. For instance, if you select a sprite, the actions window will show buttons to delete/show info, etc. It's pretty much what happens when you use a tool in Photoshop/Gimp.

I would rather like to gradually add the actions window and the tile tool in the future using dedicated PRs than continue to bloat this PR here. What do you think?

Gama11:
>I would rather like to gradually add the actions window and the tile tool in the future using dedicated PRs than continue to bloat this PR here. What do you think?

Makes sense, I prefer small PRs anyway.

Dovyski:
Great! I will wait for any issues with this PR then. I will work on the new tools after it gets merged.

Gama11:
I feel like at this point, this is polished enough to be merged - thanks for the amazing contribution! :+1: 

Without knowing the keyboard shortcut(s), it's a bit cumbersome right now, so it'd be nice to have tooltips or somethings as hints. Of course, that'd require implementing tooltips for the debugger first.. :D The need to counteract the velocities of moving objects also bothers me a bit when using this in Mode.

Minor polish issue: there's a fraction of a second at startup, at least on Flash, where the "move" cursor is visible instead of Mode's default cursor.

It would be nice if the new functionality could be documented here: http://haxeflixel.com/documentation/debugger/ (sources are located in the flixel-docs repo, you can also click "Edit" in the upper right).

Note that this should target the `dev` branch of flixel-docs, not `master`.

Gama11:
I did a bit of code cleanup. I hope I didn't break anything in the process.

Your code is quite consistent and clean (I also like the architecture), but a bit AS3-y. ;)

Dovyski:
>  I feel like at this point, this is polished enough to be merged - thanks for the amazing contribution! :+1:

Thank you for all the guidance! It's a pleasure to contribute to HaxeFlixel :smile: :tada:  

>  Without knowing the keyboard shortcut(s), it's a bit cumbersome right now, so it'd be nice to have tooltips or somethings as hints. Of course, that'd require implementing tooltips for the debugger first.. :D 

Agreed! I am planning to give UX a bit of love in the upcoming interactions (e.g. animated, dashed line for selected items, better handling of clicks to prevent unselecting things by accident, probably a Ctrl+Z functionality). I will add tooltips to that list too. 

>The need to counteract the velocities of moving objects also bothers me a bit when using this in Mode.

You can avoid that by using the "pause" button of VCR. I will add that info to the docs. 

>  Minor polish issue: there's a fraction of a second at startup, at least on Flash, where the "move" cursor is visible instead of Mode's default cursor.

I will check on that.

>  It would be nice if the new functionality could be documented here: http://haxeflixel.com/documentation/debugger/ (sources are located in the flixel-docs repo, you can also click "Edit" in the upper right). Note that this should target the dev branch of flixel-docs, not master.

Sure! I think I could write the docs as a mini tutorial, with small gifs explaining each tool and how they can be used. How about that?

Additionally since more things will be added in the future (e.g. the tile tool), I suggest we create a separate page for interactive debug, otherwise http://haxeflixel.com/documentation/debugger/ will get a lot of text and images.
  
Just let me know how you prefer it.

> I did a bit of code cleanup. I hope I didn't break anything in the process.

I've made some tests and everything looks ok. 

> Your code is quite consistent and clean (I also like the architecture), but a bit AS3-y. ;)

Thanks! I've carefully inspected your changes (sorry about giving you that much work!) and will adherence to that in the future. About the AS3-y, I'm still migrating to Haxe :smile:

Gama11:
>Agreed! I am planning to give UX a bit of love in the upcoming interactions

Looking forward to it.

>You can avoid that by using the "pause" button of VCR. I will add that info to the docs.

Yeah, I noticed that that's working now, great work on that. :) Definitely something that should be noted in the docs.

>[...]  I suggest we create a separate page for interactive debug

Sounds very sensible, that page is already fairly long.

>I've made some tests and everything looks ok.

Awesome!

>sorry about giving you that much work!

Please don't apologize for contributing! Besides, it's my problem if I'm this obsessed about code style. ;)

Dovyski:
Ok, awesome! I will start working on the docs soon. 

> Besides, it's my problem if I'm this obsessed about code style. ;)

You have to be this obsessed, otherwise we will end up with a mess at the end of the day :D

Gama11:
Something you should probably know, since it's bitten me a few times: OpenFL has a bit of code in its externs that leads to what would otherwise have been a compiler error being a runtime error (for field access of `DisplayObjcet` instances specifically). See https://github.com/openfl/openfl/issues/1065 for more info.

To fix this, go to `C:\HaxeToolkit\haxe\lib\openfl\3,6,1\extern\flash\display\DisplayObject.hx` and remove this code:

```haxe
implements #if flash Dynamic #else Dynamic<DisplayObject> #end
```

Dovyski:
Ok, thanks for the heads up!

MSGhero:
When my game first starts up (flash target, haven't tried others), the cursor is the drag one instead of the regular one. It seems like it resolves itself by frame 2, but it's apparent for me because my frame 1 takes a while with random generation.

Dovyski:
I will check on that @MSGhero, thanks for the heads up.

Gama11:
I actually just fixed that: 80cd87f

Issue #972, opened 2014-04-06
Debugger: add "FlxObjects-are-draggable"-mode


Gama11:
Apparently you can do this in HaxePunk.

This would be extremely handy for UI-building and other things of that nature. `MouseEventManager` could be used for this - would need to iterate over `FlxState` and add a onClick callback to everything that is a `FlxObject` (MouseEventManager supporting FlxObject is useful after all, @impaler, @gamedevsam  :P).

Gama11:
FlixelCommunity/flixel#217 might be a good reference for this.

Dovyski:
I can work on this. I know HaxeFlixel has a way more advanced debugger than FlixelCommnity, but the plugin architecture is quite similar. I remember that everything of this feature was implemented as a plugin in Flixel Community, so maybe I could do a direct port and iterate from that.

How about it?

Gama11:
That seems reasonable as a starting point.

Quickly glancing over that old flixel-community, I noticed you're using FlxG.mouse and keys in the plugin. Does this mean interaction doesn't work while the vcr is paused? Debugger components usually register their own Flash event listeners because of this.

Dovyski:
I implemented everything on the engine side avoiding whenever possible anything that was connected to Flash directly, which should help us here.

The only way to stop the interaction is to prevent the `update()` method of the plugin from being invoked. It should nener happen because it was using the `preUpdate` signal to update itself, as well as the `postDraw` signal to render things. I never tested things together with VCR though.

As a side note, VCR was removed from the core and made a plugin in flixel-community.

Dovyski:
I've done the initial port and at least one tool (the pointer used to select things on the screen) is rudimentary working. I have a problem with mouse visibility though.

I've tried setting `FlxG.mouse.visible = true` when the interactive debug is activated, but the mouse does not become visible at all. I am using Mode for development and in the `PlayState` the mouse visibility is set to false. If I comment that line out, the mouse is always visible.

Am I doing something wrong to make the mouse visible or is it a bug?

Gama11:
Sorry for the delay!

`FlxDebugger` has some special code to make the mouse appear while the user hovers over a debugger window and then make it return to the previous state after it leaves the window again. This is a bit tricky because there are at least 3 cases to consider:

- custom sprite cursor
- system cursor
- native system cursor

[...]

PR #1864, opened 2016-06-04
Add two new signals: debuggerShowed and debuggerHid

Dovyski:
Those signals are dispatched when the debugger is showed/hid (usually by pressing the key below ESC). The change was implemented by overriding the setter of the `visible` property in the `FlxDebugger` class. As a consequence, ***there is no API changes/breaks***.

The new signals are useful for classes interested in monitoring any changes in the debugger visibility; they can now use the (reliable) signals instead of making checks in the update method, for instance.

Those signals are essential to implement custom cursors in #972 without introducing a lot of hacky code all over the place.

### DIFF
--- a/flixel/system/debug/FlxDebugger.hx
+++ b/flixel/system/debug/FlxDebugger.hx
@@ -444,6 +444,29 @@ class FlxDebugger extends Sprite
 		return result;
 	}
 	
+	#if flash @:setter(visible) #else override #end
+	public function set_visible(Value:Bool): #if flash Void #else Bool #end
+	{
+		#if flash
+		super.visible = Value;
+		#else
+		super.set_visible(Value);
+		#end
+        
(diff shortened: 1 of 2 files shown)


Gama11:
Pretty much the same as `FlxG.debugger.drawDebugChanged`?

Gama11:
I think this should be more similar to `FlxG.debugger.drawDebugChanged` (part of `DebuggerFrontEnd`, just a single signal, perhaps `visibilityChanged`). It would also be nice to use the setter in `DebuggerFrontEnd` instead to avoid the hacky flash conditionals for `@:setter` etc (in theory, a visibility change could be missed that way, but all visibility changes go through the front end anyway I think).

Dovyski:
Ok, thanks for the feedback. I will issue a new pull request with your suggested approach.

PR #1865, opened 2016-06-05
Add visibilityChanged signal to debugger

Dovyski:
Add the `visibilityChanged` signal to the `DebuggerFrontEnd` following the same convention/idea of the already existing `drawDebugChanged` signal.

### DIFF
--- a/flixel/system/frontEnds/DebuggerFrontEnd.hx
+++ b/flixel/system/frontEnds/DebuggerFrontEnd.hx
@@ -159,9 +159,9 @@ class DebuggerFrontEnd
 			FlxG.stage.stageFocusRect = false; // don't show yellow focus rect on flash
 			FlxG.stage.focus = FlxG.game;
 		}
-		#end
 		
 		visibilityChanged.dispatch(Value);
+		#end
 		
 		return visible = Value;
 	}
--- a/flixel/system/frontEnds/DebuggerFrontEnd.hx
+++ b/flixel/system/frontEnds/DebuggerFrontEnd.hx
@@ -159,10 +159,16 @@ class DebuggerFrontEnd
 			FlxG.stage.stageFocusRect = false; // don't show yellow focus rect on flash
 			FlxG.stage.focus = FlxG.game;
 		}
-		
-		visible = Value;
-		visibilityChanged.dispatch();
 		#end
+		
+		if (Value != visible)
+		{
+			visible = Value;
+			
+			#if FLX_DEBUG
+			visibilityChanged.dispatch();
+			#end
+		}
 		
 		return visible;
 	}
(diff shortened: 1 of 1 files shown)


flixel/system/frontEnds/DebuggerFrontEnd.hx
```diff
@@ -35,6 +35,10 @@ class DebuggerFrontEnd
 	 * Dispatched when drawDebug is changed.
 	 */
 	public var drawDebugChanged(default, null):FlxSignal = new FlxSignal();
+	/**
+	 * Dispatched when visibility is changed.
+	 */
+	public var visibilityChanged(default, null):FlxTypedSignal<Bool->Void> = new FlxTypedSignal<Bool->Void>();
```
Gama11: Is it necessary to include the new value here (considering we don't do that for `drawDebugChanged` either)?

Dovyski: No, it's not necessary, I was just trying to make things easier for the signal's user. I can remove the new value for the sake of consistency.

flixel/system/frontEnds/DebuggerFrontEnd.hx
```diff
@@ -155,8 +159,11 @@ class DebuggerFrontEnd
 			FlxG.stage.stageFocusRect = false; // don't show yellow focus rect on flash
 			FlxG.stage.focus = FlxG.game;
 		}
+		
+		visible = Value;
+		visibilityChanged.dispatch();
 		#end
 		
-		return visible = Value;
```
Gama11: This now leads to `set_visible()` not doing anything with `FLX_NO_DEBUG`. I think we should be consistent with the behavior of `drawDebug` here again.

flixel/system/frontEnds/DebuggerFrontEnd.hx
```diff
@@ -155,8 +159,11 @@ class DebuggerFrontEnd
 			FlxG.stage.stageFocusRect = false; // don't show yellow focus rect on flash
 			FlxG.stage.focus = FlxG.game;
 		}
+		
+		visible = Value;
+		visibilityChanged.dispatch();
```
Gama11: We shouldn't always dispatch the signal, only when the value actually _changed_.

flixel/system/frontEnds/DebuggerFrontEnd.hx
Dovyski: Yep, I see it now. I will fix it as soon as I have some free time.

flixel/system/frontEnds/DebuggerFrontEnd.hx
Dovyski: Ok, I will change that.

PR #1882, opened 2016-07-04
Taking scaling into account in overlapsPoint() for sprites.

starry-abyss:
FlxObject's code but with multiplication by scale. Not sure if it's a proper way to go.


starry-abyss:
Seems I just forgot to call updateHitbox(). False alarm...

PR #1887, opened 2016-07-08
Decrease FlxGroup.length in FlxGroup.remove()

Nallebeorn:
Note that `length` is always decremented, even when `Splice` is false. However, I also made it so the `add()` method increments `length` also when replacing a null entry—in other words, `FlxGroup.length` represents the number of actual flixel objects in the group, not the physical size of `FlxGroup.members`.

### DIFF
--- a/flixel/group/FlxGroup.hx
+++ b/flixel/group/FlxGroup.hx
@@ -189,10 +189,7 @@ class FlxTypedGroup<T:FlxBasic> extends FlxBasic
 		{
 			members[index] = Object;
 			
-			if (index >= length)
-			{
-				length = index + 1;
-			}
+			length++;
 			
 			return Object;
 		}
@@ -363,6 +360,8 @@ class FlxTypedGroup<T:FlxBasic> extends FlxBasic
 		else
 			members[index] = null;
 		
+		length--;
+			
 		return Object;
 	}
 	


Gama11:
Why is this necessary? I have a feeling a lot of FlxGroup internals depend on the current behavior.

There also seems to be a failing unit test.

Nallebeorn:
The reason I did this in the first place, was that in a project I checked for `if (group.length == 0)`, and it took me quite a while to figure out why that didn't work... It's very counter-intuitive to have a public length property which isn't changed when removing elements, it should *at least* be decreased when the actual array is spliced. I admit I haven't tested this properly, though, it does indeed seem like this would currently break some functionality in FlxGroup... The getFirstxxx() methods only loop through the first `length` items, for example, so if there are a lot of null entries at the beginning of the array they would cancel before having looped through everything.

The unit test a bit odd, seeing as only Neko and C++ fail. Hmm...

Nallebeorn:
On the other hand, as it is now the getFirstxxx() methods will loop past the end of the members array if anything has been removed with Splice as true, because length will be too large... Which is bad for performance, obviously.

Gama11:
I don't think that's too much of a performance concern... Either way, this PR in its current state causes some issues, so I'm closing here.

Any changes to this logic need to be _really_ well thought-through. Also, I'm wary of changing it because someone might very well rely on the current behavior, which would make this a breaking change.

PR #1889, opened 2016-07-20
Adjust camera scroll bounds to account for zoom

JoeCreates:
Currently there is an issue with FlxCamera's scroll bounds if you zoom. If you are zoomed then move to the edge of the scroll bounds, you will not be able to see a part of the side of the map, because the scroll bounds are only correct at a zoom of 1.

For example, if your camera is targeting a player, and your player is right beside the scroll bounds, then you attempt to zoom on the player, the player will go off-screen in spite of it being your camera's target.

This pull request adds the required adjustments to correctly apply the scroll bounds irrespective of the current zoom. When you are zoomed, your camera's size is effectively shrunk, and it is this shrunk camera that is prevented from passing the scroll bounds, rather than the camera's size at a zoom of 1.

### DIFF
--- a/flixel/FlxCamera.hx
+++ b/flixel/FlxCamera.hx
@@ -872,9 +872,15 @@ class FlxCamera extends FlxBasic
 	 */
 	public function updateScroll():Void
 	{
-		//Make sure we didn't go outside the camera's bounds
-		scroll.x = FlxMath.bound(scroll.x, minScrollX, (maxScrollX != null) ? maxScrollX - width : null);
-		scroll.y = FlxMath.bound(scroll.y, minScrollY, (maxScrollY != null) ? maxScrollY - height : null);
+		// Adjust bounds to account for zoom
+		var minX:Null<Float> = minScrollX == null ? null : minScrollX - (zoom - 1) * width / (2 * zoom);
+		var maxX:Null<Float> = maxScrollX == null ? null : maxScrollX + (zoom - 1) * width / (2 * zoom);
+		var minY:Null<Float> = minScrollY == null ? null : minScrollY - (zoom - 1) * height / (2 * zoom);
+		var maxY:Null<Float> = maxScrollY == null ? null : maxScrollY + (zoom - 1) * height / (2 * zoom);
+		
+		// Make sure we didn't go outside the camera's bounds
+		scroll.x = FlxMath.bound(scroll.x, minX, (maxX != null) ? maxX - width : null);
+		scroll.y = FlxMath.bound(scroll.y, minY, (maxY != null) ? maxY - height : null);
 	}
 	
 	/**


Gama11:
What's the best way to test this change? Any of the demos?

JoeCreates:
Arrows to move, space to zoom. Note what happens when the sprite is near the bounds when you zoom.

```
	var spr = new FlxSprite();
	var tween:FlxTween;
	
		super.create();
		FlxG.camera.setScrollBoundsRect(0, 0, 700, 700, true);
		FlxG.camera.follow(spr);
		var bg = new FlxSprite(20, 20);
		bg.makeGraphic(660, 660, FlxColor.fromHSB(0, 0.3, 0.3));
		for (x in 0...5) {
			for (y in 0...10) {
				FlxSpriteUtil.drawRect(bg, x * 132 + y % 2 * 66, y * 66, 66, 66, FlxColor.fromHSB(0, 0.3, 0.5));
			}
		}
		spr.scale.set(3, 3);
		add(bg);
		add(spr);

	}
	
		super.update(dt);
		spr.velocity.set();
		if (FlxG.keys.pressed.LEFT) spr.velocity.x = -200;
		if (FlxG.keys.pressed.RIGHT) spr.velocity.x = 200;
		if (FlxG.keys.pressed.UP) spr.velocity.y = -200;
		if (FlxG.keys.pressed.DOWN) spr.velocity.y = 200;
		if (FlxG.keys.justPressed.SPACE) {
			if (tween != null) tween.cancel();
			tween = FlxTween.tween(FlxG.camera, {zoom: 2}, 0.5, {ease: FlxEase.quadInOut});
		} else if (FlxG.keys.justReleased.SPACE) {
			if (tween != null) tween.cancel();
			tween = FlxTween.tween(FlxG.camera, {zoom: 1}, 0.5, {ease: FlxEase.quadInOut});
		}
		
		spr.x = FlxMath.bound(spr.x, FlxG.worldBounds.left, FlxG.worldBounds.right - spr.width);
		spr.y = FlxMath.bound(spr.y, FlxG.worldBounds.top, FlxG.worldBounds.bottom - spr.height);
	}
```

JoeCreates:
@Gama11 I re-added some changes which I had somehow lost when making this branch, but it should be good to pull again, now.

JoeCreates:
Note that the FlxCamera demo doesn't exhibit the issue as it effectively performs this calculation itself when you zoom in (https://github.com/HaxeFlixel/flixel-demos/blob/master/Features/FlxCamera/source/PlayState.hx#L164). Doing it that way means that you have to separately store your own values for "bounds at 1x zoom" and use these to recalculate the scroll bounds every time you change the zoom.

This change is therefore breaking, although I think it should be included in flixel. The solution in the FlxCamera is more of a workaround.

Gama11:
Thanks for the fix and the demo code, showcased the issue very nicely. I think the breaking aspect here is ok, as it's essentially a bugfix.

That zoom-in-effect in the demo code is quite nice visually, could you add that to the FlxCamera demo in some capacity? :) Perhaps it could replace the current zooming controls, which seem a bit awkward...

PR #1897, opened 2016-08-03
HTML5FrontEnd: Add platform and mobile detection

DleanJeans:
I added `platform` and `isMobile` variables to `FlxG.html5`.
Tested on Windows and Android. Not sure about the others. Here's a [test](https://dleanjeans.github.io/flixel-html5-platform-test/) (https://github.com/DleanJeans/flixel-html5-platform-test).

### DIFF
--- a/flixel/system/frontEnds/HTML5FrontEnd.hx
+++ b/flixel/system/frontEnds/HTML5FrontEnd.hx
@@ -1,6 +1,7 @@
 package flixel.system.frontEnds;
 
 import flixel.math.FlxPoint;
+import flixel.system.frontEnds.HTML5FrontEnd.IOSPlatform;
 import flixel.util.FlxStringUtil;
 
 #if js
@@ -9,16 +10,18 @@ import js.Browser;
 class HTML5FrontEnd
 {
 	public var browser(default, null):FlxBrowser;
+	public var ios(default, null):IOSPlatform;
+	public var platform(default, null):FlxPlatform;
+	public var isMobile(default, null):Bool;
 	public var browserWidth(get, null):Int;
 	public var browserHeight(get, null):Int;
 	public var browserPosition(get, null):FlxPoint;
-	public var platform(default, null):FlxPlatform;
-	public var isMobile(default, null):Bool;
 	
 	@:allow(flixel.FlxG)
 	private function new()
 	{
 		browser = getBrowser();
+		ios = getiOS();
 		platform = getPlatform();
 		isMobile = getIsMobile();
 	}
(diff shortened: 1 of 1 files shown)


flixel/system/frontEnds/HTML5FrontEnd.hx
```diff
@@ -59,6 +61,49 @@ class HTML5FrontEnd
 	{
 		return Browser.window.innerHeight;
 	}
+	
+	private inline function get_platform():FlxPlatform
```
Gama11: This function is way to long to be a good candidate for `inline`.

flixel/system/frontEnds/HTML5FrontEnd.hx
```diff
@@ -59,6 +61,49 @@ class HTML5FrontEnd
...
+	{
+		
+		if (Browser.navigator.userAgent.indexOf("Win") > -1)
```
Gama11: Should probably wrap `Browser.navigator.userAgent.indexOf()` in a helper function, it's used quite often here.

flixel/system/frontEnds/HTML5FrontEnd.hx
Gama11: We could probably just run this entire thing once in `new()` and cache the result? It's not like it's gonna change at runtime, right? :)

flixel/system/frontEnds/HTML5FrontEnd.hx
Gama11: Oh, and instead of `indexOf()` you can use `.contains()` via `using flixel.util.FlxStringUtil`.

flixel/system/frontEnds/HTML5FrontEnd.hx
```diff
@@ -1,27 +1,37 @@
...
 	public var browserPosition(get, null):FlxPoint;
+	public var platform(default, null):FlxPlatform;
+	public var isMobile(default, null):Bool;
 	
 	@:allow(flixel.FlxG)
-	private function new() {}
+	private function new()
+	{
+		browser = getBrowser();
+		browserWidth = getBrowserWidth();
```
Gama11: Browser width and height could change at runtime, so I don't think caching those is a good idea.

flixel/system/frontEnds/HTML5FrontEnd.hx
```diff
@@ -70,4 +128,17 @@ enum FlxBrowser
 	OPERA;
 	UNKNOWN;
 }
+
+enum FlxPlatform
+{
+	WINDOWS;
+	LINUX;
+	UNIX;
+	MAC;
```
Gama11: Hm... `MAC` isn't used anywhere?

flixel/system/frontEnds/HTML5FrontEnd.hx
```diff
@@ -70,4 +128,17 @@ enum FlxBrowser
 	OPERA;
 	UNKNOWN;
 }
+
+enum FlxPlatform
+{
+	WINDOWS;
+	LINUX;
+	UNIX;
```
Gama11: What exactly does `UNIX` mean in this context? As I understand, both Linux and Mac are technically UNIX systems. That wouldn't make much sense in an enum though.

flixel/system/frontEnds/HTML5FrontEnd.hx
```diff
@@ -70,4 +128,17 @@ enum FlxBrowser
...
+	MAC;
+	ANDROID;
+	BLACKBERRY;
+	WINDOWS_PHONE;
+	IOS;
```
Gama11: It seems `"iPhone"`, `"iPad"` and `"iPod"` all end up as `IOS`. Would it be useful to the able to distuingish them?

DleanJeans:
I think we really need more people to test this

DleanJeans:
Sorry for my ignorance. Here's my [reference](http://stackoverflow.com/questions/9514179/how-to-find-the-operating-system-version-using-javascript). So you're saying it should be removed?

Gama11:
I'm not sure, I was mostly asking out of ignorance as well. ;)

DleanJeans:
Maybe it's for some kind of Unix-based OS which is probably rarely used by our users. So I guess it should be removed anyway.

Gama11:
It feels a bit strange to have IPOD / IPAD / IPHONE in the enum directly, since everything else in there is an OS... It might make sense to continue having `IOS` as an enum value, but with an enum constructor argument that is another enum (which is then either IPOD, IPAD or IPHONE)?

Gama11:
Thanks! :)

PR #1901, opened 2016-08-09
HTML5FrontEnd: Fix iPad check and iOS isMobile

DleanJeans:
- On an iPad, user agent contains "Mac"
 - For some reason, `[IOS(IPHONE), IOS(IPAD), IOS(IPOD)].contains(IOS(IPAD)) == false` so that's my workaround

### DIFF
--- a/flixel/system/frontEnds/HTML5FrontEnd.hx
+++ b/flixel/system/frontEnds/HTML5FrontEnd.hx
@@ -55,7 +55,7 @@ class HTML5FrontEnd
 		{
 			return WINDOWS;
 		}
-		else if (userAgentContains("Mac"))
+		else if (userAgentContains("Mac") && !userAgentContains("iPad"))
 		{
 			return MAC;
 		}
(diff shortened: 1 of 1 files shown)


flixel/system/frontEnds/HTML5FrontEnd.hx
```diff
@@ -92,8 +92,7 @@ class HTML5FrontEnd
 	
 	private function getIsMobile():Bool 
 	{
-		return [ANDROID, BLACKBERRY, WINDOWS_PHONE,
-			IOS(IPHONE), IOS(IPAD), IOS(IPOD)].contains(platform);
+		return Type.enumConstructor(platform) == "IOS" || [ANDROID, BLACKBERRY, WINDOWS_PHONE].contains(platform);
```
Gama11: This seems a bit hacky.. How about using a switch-case:

```haxe
return switch (platform) {
    case ANDROID:
    case BLACKBERRY:
    ...
    case IOS(_):
        true; 
    default:
        false;
}
```

DleanJeans: Nah, switch-case doesn't work. It still returns `false` no matter what.

DleanJeans: I'll take that back. It works now. My bad :P

Gama11:
Looks good, thanks!

PR #1903, opened 2016-08-11
Ignorable emitter color

IBwWG:
The current implementation forces a sort of colour-tweening for the lifespan of an emitted particle.  This change allows you to force-add your own particles with whatever static colours you want, without FlxEmitter overwriting them--all you have to do is set your emitter's `.color = null`.

Docs updated, no defaults changed.  Should be simple and non-breaking.


IBwWG:
Huh?

Are there general build issues right now or what?

Gama11:
Hm.. Apparently? Not sure what's causing those errors.

Gama11:
I think the intended way to do this is to set `FlxParticle#colorRange.active` to `false`.

IBwWG:
@Gama11  Quite odd, I never got a notification that you commented here.  I had actually discovered that and tried it out.  The problem is the line in FlxEmitter, `particle.color = particle.colorRange.start;` which then overrides whatever color you had set for a particular particle, as it emits it.

Gama11:
I see. I don't think that code should be run of `colorRange.active` was set to false. This seems like a bug, and this is more of a workaround than a fix.

Gama11:
Urgh, it seems like `active` of range / bounds objects is currently false by default...

Gama11:
Okay, as of the latest two commits this should now work as expected:

```haxe
package;

import flixel.FlxG;
import flixel.FlxState;
import flixel.effects.particles.FlxEmitter;
import flixel.effects.particles.FlxParticle;
import flixel.util.FlxColor;
using flixel.util.FlxSpriteUtil;

{
    {
		var emitter = new FlxEmitter();
		emitter.particleClass = Particle;
		emitter.makeParticles();
		add(emitter);
		emitter.start();
		emitter.x = FlxG.width / 2;
		emitter.y = FlxG.height / 2;
		
		emitter.color.active = false;
    }
}

class Particle extends FlxParticle
{
	public function new()
	{
		super();
		color = FlxColor.GREEN;
	}
}
```

I don't think the changes should break anything, but some testing would be good... I checked against the Particles and FlxPexParser demo and didn't notice any changes.

PR #1904, opened 2016-08-13
FlxG: Add FlxG.onMobile

DleanJeans:
Also I renamed `isMobile` in HTML5FrontEnd to `onMobile` since it sounds better.

### DIFF
--- a/flixel/FlxG.hx
+++ b/flixel/FlxG.hx
@@ -117,7 +117,8 @@ class FlxG
 	 */
 	public static var drawFramerate(default, set):Int;
 	
-	public static var onMobile(get, null):Bool;
+	@:isVar
+	public static var onMobile(get, never):Bool;
 	
 	public static var renderMethod(default, null):FlxRenderMethod;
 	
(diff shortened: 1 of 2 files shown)


flixel/FlxG.hx
```diff
@@ -707,6 +709,22 @@ class FlxG
 	{
 		return game._state;
 	}
+	
+	private static inline function get_onMobile():Bool
+	{
+		if (onMobile == null)
```
MSGhero: What's the point of this check? Why not make it `@:isVar public static var onMobile(get, never):Bool` and
```haxe
private static inline function get_onMobile():Bool {
   return #if js blah #elseif mobile true #else false #end;
}
```

DleanJeans:
Not sure why thus

DleanJeans:
Not sure why Travis didn't build. It's been working OK in my project so fast.

MSGhero:
Travis fails every project for the same reason, something weird on cpp and neko.

DleanJeans:
All the builds were successful but still exited with 1?

MSGhero:
Yeah, if you look closely,
```
X Error of failed request:  BadRequest (invalid request code or no such operation)
  Major opcode of failed request:  147 (GLX)
  Minor opcode of failed request:  168 ()
  Serial number of failed request:  173
  Current serial number in output stream:  173
```
on Haxe 3.2.0 neko and Haxe dev neko/cpp for every recent pull request. This pull is fine since all the demos and coverages passed.

Gama11:
Thanks!

I'm not a huge fan of putting even more stuff into `FlxG`, but I don't really have a better alternative right now and this does seem useful.

The Travis failures are indeed unrelated, I'm noto sure what's causing them yet.

I don't like the formatting of `get_onMobile()`'s body, cramping everything into one line doesn't make it very readable.

PR #1908, opened 2016-08-21
Fix FlxText applyMarkup for markers with length > 1.

bendmorris:
I tried using applyMarkup with some colored text tags, using multi-character markers which all start with the same character, like [red], [blue], [green]. The current implementation only looks at the first character of the marker, so each left brace matches every marker, resulting in the positions of the format ranges being off. Doing a full comparison against the marker fixes this issue.

Before:

screen shot 2016-08-20 at 11 08 31 pm

After (correct):

screen shot 2016-08-20 at 11 07 43 pm

### DIFF
--- a/flixel/text/FlxText.hx
+++ b/flixel/text/FlxText.hx
@@ -299,12 +299,13 @@ class FlxText extends FlxSprite
 			if (rule.marker != null && rule.format != null)
 			{
 				var start:Bool = false;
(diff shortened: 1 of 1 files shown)


Gama11:
Thanks! Great-looking game btw. :)

Gama11:
I added an example of this to the demo: https://github.com/HaxeFlixel/flixel-demos/commit/e218192f7d3616c388814bd5290cad303726d5a8

PR #1910, opened 2016-08-21
Optional step after console command, closes #1884

### LINKED ISSUE #1884: Optional step() after console command?
https://github.com/HaxeFlixel/flixel/blob/dev/flixel/system/debug/console/Console.hx#L247-L251

```haxe
// Step forward one frame to see the results of the command
#if (flash && FLX_DEBUG)
if (FlxG.vcr.paused)
	FlxG.game.debugger.vcr.onStep();
#end
```

I have to say that I really hate this line of code when debugging. Most of the time, I want my game paused on a single frame where I input a series of commands. Sometimes, I just want to see the current value of 2 things, but advancing a frame changes the second value. Sometimes, I use it as a calculator when deciding what to do during that frame. And then sometimes, I write several lines of actual haxe code that I want to run.

If the current behavior should be kept, the only thing I can think of is setting a debugger/console boolean or adding a special character before the string that gets parsed out. Thoughts?

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/35743025-optional-step-after-console-command?utm_campaign=plugin&utm_content=tracker%2F254189&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F254189&utm_medium=issues&utm_source=github).
</bountysource-plugin>

MSGhero:
Adds a boolean to see if the user wants to step after inputting a single console command (default behavior) or input multiple commands in a single frame. Adds the `step()` console command so you don't have to mouse over to the little stepper arrow each time you do want to advance. The boolean is regularly accessible from the console if it needs to be changed at runtime.

Cleaned up some defines that didn't seem to be relevant any longer. The entire `ConsoleCommands` class is wrapped in `#if FLX_DEBUG`, so there's no need to repeat those internally. Neko seemed fine with respect to taking out the `#if flash` defines.

image
Ugh, I'll get this right one day.

### DIFF
--- a/flixel/system/debug/console/Console.hx
+++ b/flixel/system/debug/console/Console.hx
@@ -149,11 +149,9 @@ class Console extends Window
 		#end
 		
 		#if FLX_DEBUG
-		#if flash 
 		// Pause game
 		if (FlxG.console.autoPause)
 			FlxG.vcr.pause();
-		#end
 		
 		// Block keyboard input
 		#if FLX_KEYBOARD
@@ -174,11 +172,10 @@ class Console extends Window
 		#end
 		
 		#if FLX_DEBUG
-		#if flash
 		// Unpause game
 		if (FlxG.console.autoPause)
 			FlxG.vcr.resume();
-		#end
+		
 		// Unblock keyboard input
 		#if FLX_KEYBOARD
 		FlxG.keys.enabled = true;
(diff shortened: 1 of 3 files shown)


Gama11:
Wow, the VCR step actually seems to work as expected on Neko now. I think there were a lot of issues with it back in the day related to text field focus.

>Ugh, I'll get this right one day.

Doesn't matter commits get squased anyway. ;)

PR #1921, opened 2016-08-31
FlxAnimation: clean up update() + improve some var names

IBwWG:
Edit: I did two things here (ignore the branch name please):

1. renamed NumFramesMinusOne to something that tells its purpose
2. refactored the update logic for clarity (it didn't even need NumFramesMinusOne after the refactor)


Gama11:
Hm.. I do like the code cleanup, but I'd generally prefer to keep cleanup and adding new features separate  (the diff becomes a bit useless otherwise). Would it be too much of a hassle to remove the new feature here and add a separate PR for it?

IBwWG:
Isn't this how it's done all the time?  Just not by outsiders I guess :)

I don't think in this case it's hard to see the diff.  The pingpong has its own if clause and its own bool, and that's it--two neat blocks.  I can separate it if you want but I'm going to do it later if so.

MSGhero:
At least two commits, but two PRs is clearer.

Gama11:
Yeah, that'd help too. But the commit history here isn't that clean. ;) (which doesn't matter btw, commits are squashed on merge anyway)

IBwWG:
OK, pingpong removed.  I didn't like how it functioned anyway, and will do a separate PR using a different method.

flixel/animation/FlxAnimation.hx
```diff
@@ -204,24 +204,19 @@ class FlxAnimation extends FlxBaseAnimation
 		while (_frameTimer > delay && !finished)
 		{
 			_frameTimer -= delay;
-			
-			if (looped)
+			if (reversed)
```
Gama11: This actually makes the code _a lot_ more readable. :+1: 

Btw, previously it was hard to tell if the changes in this section were from cleanup or related to the new feature, which was what caused my confusion for the most part.

PR #1923, opened 2016-09-01
FlxAtlasFrames#fromLibGdx(): fix offset for whitespace-stripped atlases

IBwWG:


Gama11:
I'm going to wait on this one until https://github.com/libgdx/libgdx/issues/4288 gets a reply.. I'm not familiar with the format at all.

IBwWG:
Sure, if you want.  But with flixel how it is now, it produces funny output like this:

mg-libgdx-offset-bug

This PR fixes that.

Gama11:
The TexturePackerAtlas simply doesn't have any assets that use this feature? I didn't notice any issues there.

IBwWG:
You mean the demo?  It has one that uses it (although maybe not with whitespace stripping used?  not sure), but you'll only notice on sprites that aren't vertically centred, if you also know exactly where you expect the sprite to be (debug mode boxes help here.)

IBwWG:
Correction, I think the demo does not use a whitespace-stripped atlas: https://github.com/HaxeFlixel/flixel-demos/blob/master/Editors/TexturePackerAtlas/assets/libgdx/test-me.pack#L11 and similar lines all have no offset.  The offset tells where the smaller image is, within the extra transparency of the original image.

Gama11:
I think it would be good if we had an example with this issue in the demo.

IBwWG:
Here's the source for the demo, for reference.  It includes yesterday's nightly of libgdx standalone texture packer.  Made for Windows.
[libgdx-demo.zip](https://github.com/HaxeFlixel/flixel/files/449486/libgdx-demo.zip)

IBwWG:
Windows source for the demo is here if anyone needs to test libgdx output.  Includes yesterday's nightly of the standalone packer: [libgdx-demo.zip](https://github.com/HaxeFlixel/flixel/files/449541/libgdx-demo.zip)

PR for the demo is at https://github.com/HaxeFlixel/flixel-demos/pull/245

Gama11:
With the demo updated to showcase the bug (and working correctly with your changes), I actually feel fairly confident merging this. Thanks!

Issue #1888, opened 2016-07-09
No way to modify animation frames (without instantiating new animation)


JoeCreates:
```FlxAnimation._frames``` is private, meaning we cannot edit the frames of an animation after it has already been instantiated. ```FlxAnimationController``` provides methods append frames to an animation, but there is no way to remove or change the frames.

The reason I wish to do this is that I have pooled objects that may require slightly different animations once recycled. Simply using add, however, will be constantly instantiating new ```FlxAnimation``` objects, so I would prefer to avoid this.

Should ```FlxAnimation._frames``` be public, or perhaps we could add a new method to ```FlxAnimationController``` to set the frames of an existing animation?

IBwWG:
I'd also find this handy for read access.

PR #1925, opened 2016-09-01
Add pingpong helper function

IBwWG:


IBwWG:
I did it this way instead of like in #1921 originally, because this way you aren't forced to loop.

flixel/animation/FlxAnimation.hx
```diff
@@ -277,4 +277,20 @@ class FlxAnimation extends FlxBaseAnimation
 	{
 		return _frames.length;
 	}
+	
+	/**
+	 * Appends the reverse of the animation to its frames, without duplicating the original last frame.
+	 * 
+	 * If looped is true, the original first frame is also not duplicated.
+	 * 
+	 * E.g. [0,1,2,3] becomes [0,1,2,3,2,1]   (looped)     -- Frame 0 is not played twice in a row when the loop returns to the beginning.
```
Gama11: This seems like a very strange special case to me, which might be different depending on the animation you have...

flixel/animation/FlxAnimation.hx
```diff
@@ -277,4 +277,20 @@ class FlxAnimation extends FlxBaseAnimation
...
+	 * but  [0,1,2,3] becomes [0,1,2,3,2,1,0] (non-looped) -- Animation ends looking how it started.
+	 */
+	public function pingPongifyFrames()
```
Gama11: Hm... not a huge fan of the name... thinking about it more, do we really need this if https://github.com/HaxeFlixel/flixel/issues/1888 is implemented? With that it would be as simple as:

```haxe
anim.frames = anim.frames.concat(anim.frames.reverse());
```

flixel/animation/FlxAnimation.hx
IBwWG: Really?  Why would you normally want to duplicate a frame?  Then you would have a weird pause in the middle of your animation.

flixel/animation/FlxAnimation.hx
IBwWG: I'm open to name suggestions.  :)

Didn't see #1888.  That would help, then my function could just live in my own code.  But I thought it would be a nice helper function to have, since I can picture many animations that one might want to play this way.  E.g. a bouncing ball, a person jumping, etc.  It is a cheap technique, but that's the point: you only need half the number of rendered frames this way, therefore half the memory usage.  I figured that would be a common enough scenario/technique that it was worth including in the core.

The method you indicate here would duplicate both the starting and ending frame, adding two extra pauses in the animation if you're looping it, and one extra pause in the middle if not.  Hence my function taking away 2 or 1 from the number of frames depending on `looped`, in line 292.

Gama11:
I don't feel good about adding this as its implemented right now (or maybe in general). It somehow feels a little bit too specific and doesn't seem to fit that well with the remaining `FlxAnimation` API.

Might be a case of trying to do too much for the user - there are probably quite a few people needing something like this, but perhaps in a slightly different way.

IBwWG:
If you say so.  I mean, there's FlxTween.PINGPONG, that's where I got the name.  This just does the same thing to your animation frames.  I'm still not seeing a different way one would want to do it.  But if #1888 allows our own solutions, I guess this is here if anyone goes looking.

PR #1928, opened 2016-09-05
FlxCollision: fix an animation-related crash with renderBlit

IBwWG:
This needs a unit test, but in case anyone else was having flash crash on them during a collision check, before I get to that...


Gama11:
Definitely need a way of reproducing the issue.

Gama11:
Any update on this one? In what situation does this fix a crash?

IBwWG:
As the commit says, on flash target, "if you play an animation that has a larger frame size than the previously playing one" and then do a collision check.  I'll do up a test case now.

Gama11:
Oh, right. I missed that commit comment.

I'm a bit worried about this change because this code is fairly performance-critical, and doing two `drawFrame()` calls each time seems like it'd impact things negatively. Might want to check if this has an impact on FPS in the `PixelPerfectCollision` demo.

IBwWG:
I suppose we could first check whether the frame size changed, and only run it then.

IBwWG:
Hmm, having trouble reproducing it.  Very strange, I definitely needed this fix for my project.

IBwWG:
Ah, OK it's even a bit more specific, it seems.  Got it breaking now.

IBwWG:
Here's a repro, I'm just turning it into a test case:

```haxe
package;

import flash.display.BitmapData;
import flash.geom.Matrix;
import flash.geom.Point;
import flash.geom.Rectangle;
import flixel.FlxSprite;
import flixel.FlxState;
import flixel.graphics.frames.FlxAtlasFrames;
import flixel.util.FlxCollision;
import flixel.util.FlxColor;

class PlayState extends FlxState{
        var sprite:FlxSprite = cast add(new FlxSprite(30, 0));
		sprite.makeGraphic(100, 100, FlxColor.WHITE, true);
        var sourceSprite = new FlxSprite();
		sourceSprite.makeGraphic(100, 100, FlxColor.RED, true);
		var animatedSprite:FlxSprite = cast add(new FlxSprite());
		// load animations with varying frame sizes
		animatedSprite.frames = FlxAtlasFrames.fromLibGdx(sourceSprite.pixels, '
test.png
size: 100,100
format: RGBA8888
filter: Nearest,Nearest
repeat: none
a/0
  rotate: false
  xy: 0, 0
  size: 10, 10
  orig: 10, 10
  offset: 0, 0
  index: -1
b/0
  rotate: false
  xy: 90, 90
  size: 20, 20
  orig: 50, 50
  offset: 40, 0
  index: -1
'); // reuse sprite's graphic, it doesn't really matter what's there
		// add animations
		for (name in ["a","b"])
			animatedSprite.animation.addByPrefix(name, name);
		// play smaller-framed anim
		animatedSprite.animation.play("a");
		trace(FlxCollision.pixelPerfectCheck(sprite, animatedSprite));
		// play larger-framed anim
		animatedSprite.animation.play("b");
		// run PPC
		trace(FlxCollision.pixelPerfectCheck(sprite, animatedSprite));
    }
	
	

}
```

IBwWG:
Gee, it'd help if the test cases ran before my tests.  How'd this get committed?

```
C:\HaxeToolkit\haxe\lib\flixel\git\tests\unit>lime test flash
src/TestSuite.hx:32: characters 7-50 : Type not found : flixel.system.frontEnds.
DebuggerFontEndTest
src/TestMain.hx:15: lines 15-64 : Defined in this class
```

`font` should be `front` in both places...

But even with that for some reason it's not working:

```
C:\HaxeToolkit\haxe\lib\flixel\git\tests\unit>lime test flash
src/TestSuite.hx:96: characters 6-12 : Unknown identifier : flixel

C:\HaxeToolkit\haxe\lib\flixel\git\tests\unit>lime test flash
massive.munit.MUnitException: ExternalInterface not available at massive.munit.c
lient.ExternalPrintClientJS#new (320)
        at massive.munit.client::ExternalPrintClientJS()
        at massive.munit.client::RichPrintClient/init()
        at massive.munit.client::AbstractTestResultClient()
        at massive.munit.client::PrintClientBase()
        at massive.munit.client::RichPrintClient()
        at TestMain()
        at TestMain$/main()
        at Function/http://adobe.com/AS3/2006/builtin::apply()
        at Function/<anonymous>()
        at ApplicationMain$/start()
        at ApplicationMain$/init()
        at lime.app::Event_Void_Void/dispatch()
        at lime.app::Preloader/start()
        at openfl.display::Preloader/display_onComplete()
        at flash.events::EventDispatcher/dispatchEventFunction()
        at flash.events::EventDispatcher/dispatchEvent()
        at NMEPreloader/onLoaded()
        at openfl.display::Preloader/start()
        at lime.app::Preloader/current_onEnter()

```

Am I running the wrong command?

IBwWG:
Er, OK, I try `lime test windows` and instead it complains about `front` and likes `font` instead.  I'm not sure how the test suite gets out of sync with the rest of the codebase when I pull updates?

Gama11:
Try using these files: https://github.com/HaxeFlixel/flixel/tree/dev/tests/unit/targets

They run `munit gen`, which regenerates the `TestSuite.hx` file.

IBwWG:
Sorry, what command should I use?

I got it completed with `lime test windows` so at least the test as-is works.  Travis should show it passing.  The test is then, remove my changes to FlxCollision and then the test should fail.

Gama11:
https://github.com/HaxeFlixel/flixel/blob/dev/tests/unit/README.md

IBwWG:
Missed that layer.  Thanks!

I guess also I was wrong earlier about confirming the passing, if it didn't regen the suite, it won't include my new test.

IBwWG:
Yeah, that'll surely fail.  Forgot to finish the before section entirely.

tests/unit/src/flixel/util/FlxCollisionTest.hx
```diff
@@ -0,0 +1,73 @@
...
+import massive.munit.Assert;
+
+class FlxCollisionTest extends FlxTest
+{
+        var sprite:FlxSprite = cast add(new FlxSprite(30, 0));
+		var sourceSprite = new FlxSprite();
+		var animatedSprite:FlxSprite = cast add(new FlxSprite());
+		// play smaller-framed anim
+		animatedSprite.animation.play("a");
+		trace(FlxCollision.pixelPerfectCheck(sprite, animatedSprite));
```
Gama11: This doesn't look like it should even compile...

IBwWG: Hence 'that'll surely fail' :)

flixel/util/FlxCollision.hx
```diff
@@ -18,6 +18,7 @@ import flixel.tile.FlxTileblock;
  * @link http://www.photonstorm.com
  * @author Richard Davey / Photon Storm
 */
+@:access(flixel.FlxSprite)
```
Gama11: Hm, not a fan of this... Have you checked whether it makes a noticable performance difference to call `drawFrame()`? Thinking about it some more, it might actually not matter much, since it's a non-op if the sprite isn't `dirty`.

IBwWG: I haven't done a performance test.  Oh, I see...that's interesting, I had misread drawFrame()'s if statement.  I guess in this case it's properly marked dirty when it needs to be, otherwise this wouldn't be fixing anything.

Gama11:
Hm, how is it that your test case looks different on Flash than on Neko? (even with the fix)

IBwWG:
Feel free to file a new issue :)  I almost never use neko, it seems to just require even more adjustments to all my function calls than flash does, and not provide much benefit to me over flash.

Gama11:
Well, it's not exactly Neko-specific, looks the same on cpp...

tests/unit/src/flixel/util/FlxCollisionTest.hx
```diff
@@ -0,0 +1,71 @@
...
+');
+		// add animations
+		for (name in ["a", "b"])
+			animatedSprite.animation.addByPrefix(name, name);
+		
+		destroyable = sprite;
+	}
+	
+	@Test
+	function pPCOnAnim()
```
Gama11: `pPCO`? What's that supposed to mean?

tests/unit/src/flixel/util/FlxCollisionTest.hx
```diff
@@ -0,0 +1,71 @@
...
+		// play larger-framed anim
+		animatedSprite.animation.play("b");
+		// run PPC
+		Assert.isTrue(FlxCollision.pixelPerfectCheck(sprite, animatedSprite));
+	}
+
+	@After
+	public function testTeardown():Void
+	{
+		animatedSprite.destroy();
```
Gama11: This isn't really necessary, we don't have any teardown methods in other tests either.

IBwWG:
Ah, to be fair, it's bad input, I just noticed.  I should probably change that... (the offset, size, and orig don't make sense together in b/0.)

tests/unit/src/flixel/util/FlxCollisionTest.hx
IBwWG: Out of curiosity then, what's the point in the `destroyable = ` in e.g. flxsprite's test?

IBwWG:
Although, even with that fixed, it still looks wrong.  Try setting the offset to 30,0 and you'll see the resulting size is still 10x10 instead of 20x20.

tests/unit/src/flixel/util/FlxCollisionTest.hx
Gama11: It's to test that it's safe calling `FlxSprite#destroy()` multiple times (happens in `FlxTest`), which might happen in some scenarios (the engine already destroys add()ed sprites for instance, and if the user does it too, it shouldn't blow up).

Gama11:
Different bug? :)

tests/unit/src/flixel/util/FlxCollisionTest.hx
IBwWG: Ah, good to know.  So I can skip that line in this new test too.

IBwWG:
I think it's definitely a different bug.  Maybe related to the cause of #1936, not sure.

tests/unit/src/flixel/util/FlxCollisionTest.hx
```diff
@@ -36,7 +37,7 @@ b/0
   xy: 90, 90
   size: 20, 20
   orig: 50, 50
-  offset: 40, 0
+  offset: 30, 0
```
IBwWG: Forgot to mention this in commit msg.  Adjust so this is legal input.

IBwWG:
It's quite odd I didn't notice that bug with my own project, you'd think that would've broken something...

Anyway, hopefully Travis passes and this is all good now.

Gama11:
I'm gonna double-check the performance implications of always running `drawFrame()`.

The current workaround is too hacky for my taste... (target-specific conditionals, but especially access to private fields...). If it's indeed necessary not to always run `drawFrame()`, there might be a better / less hacky workaround of doing that.

IBwWG:
Maybe it's not target-specific but renderBlit-specific, like you mentioned in the other issue.  I haven't tried this on html5.  I suppose you could have it always run drawFrame() at this point in the code, and then make sure that animation.play() or whatever changes the current animation checks the frame size and marks the sprite dirty if the size changed.

Gama11:
@Beeblerox Do you have an idea how to fix this in a less hacky way?

flixel/util/FlxCollision.hx
```diff
@@ -98,12 +99,17 @@ class FlxCollision
 		matrixB.identity();
 		matrixB.translate(-(intersect.x - boundsB.x), -(intersect.y - boundsB.y));
 		
+	#if flash
+		if (Contact._frame.sourceSize.x != Contact.framePixels.width || Contact._frame.sourceSize.y != Contact.framePixels.height
+		|| Target._frame.sourceSize.x != Target.framePixels.width || Target._frame.sourceSize.y != Target.framePixels.height)
+	#else
 		if (FlxG.renderTile)
+	#end
```
Beeblerox: we don't need both checks:
```
#if flash
		if (Contact._frame.sourceSize.x != Contact.framePixels.width || Contact._frame.sourceSize.y != Contact.framePixels.height || Target._frame.sourceSize.x != Target.framePixels.width || Target._frame.sourceSize.y != Target.framePixels.height)
 #else
  		if (FlxG.renderTile)
 #end
```

so we just call `Contact.drawFrame();` and `Target.drawFrame();`.

`drawFrame()` (and following methods which are called by it) have checks for value of `dirty` flag, so if sprite doesn't need to update its frame graphic then it will skip this step.
The only perfomance problem is with animated sprites when animation could have series of the same frames: when animation changes the frame (even if it has the same index) then it automatically resets sprite's frame, whih set dirty to true

PR #1932, opened 2016-09-07
Flxanim cleanup

IBwWG:


IBwWG:
Huh, it says it has conflicts, but doesn't say what exactly...maybe it's time to rebase all my "catch up with upstream" PRs on my fork?

IBwWG:
Oh, it conflicts with itself.  I didn't realize that was possible.  So much for trying to keep things separate.

Gama11:
What do you mean by "itself"?

You should probably create a fresh branch based on origin/dev anyway, look at all the merge commits etc you have in your history. :)

IBwWG:
Well, have a look at the conflict.  I only saw what it was when I went to merge it into my own fork: the two commits of this PR seemed to be what conflicted.

I wish I could create a fresh branch, but I have my monkey patches that I need to keep in order to build.  I looked into it once before and the accepted wisdom seemed to be to keep pulling a reverse PR from upstream, but it sure looks ugly (and creates false info: I made these changes today, not in April.)

MSGhero:
What is the point of these changes?

Gama11:
I think the point is code cleanup. There's some explanations in the 2 most recent commits.

Anyway, can you solve the merge conflict @IBwWG? Discussing the content of the PR is a bit pointless otherwise, since I wouldn't be able to merge anyway. :)
I didn't really understand your monkey-patch argument. That branch is separate from the one for this PR anyway, right?

MSGhero:
Oh I just ignored all of the commits. You can have your monkey patches be one branch, and then make a branch that is initially based on dev when you want to change something here.

IBwWG:
It doesn't let me from here, it says I need write access.  Just take the line with `delay ==`...I'm not even sure how this is considered a conflict, just because I changed the same line on two different commits.

IBwWG:
@MSGhero specifically the point is not having extra checks for nothing, and also not having logic checks that imply a possibility that isn't there, which makes it harder to understand what's happening in the code and what's expected to happen.  These are lines that caused me to ask, "OK, why does that happen?"  Upon finding out there was no reason, I realized I had wasted time.

IBwWG:
As for the branches, I guess my setup is not very good: my "origin" locally is my fork.  Can I force a new branch to be based somewhere else?  Ah...let me try.

IBwWG:
OK try #1935 .

PR #1933, opened 2016-09-07
FlxTimerManager: add completeAll()

IBwWG:


tests/unit/src/flixel/util/FlxTimerTest.hx
```diff
@@ -5,6 +5,7 @@ import massive.munit.Assert;
 class FlxTimerTest extends FlxTest
 {
 	var timer:FlxTimer;
+	var loopsCompleted:Int;
```
Gama11: This can be a local variable.

PR #1935, opened 2016-09-08
Anim cleanup

IBwWG:
Hopefully this doesn't have a conflict.  I think I saw what was happening, I hadn't pulled from upstream the last changes (renaming to lastFrameIndex) before making the branch.  So it would have been OK, but this is still a lot less merges to see before seeing the actual changes.


Gama11:
This doesn't compile, see Travis.

IBwWG:
Oh, the add-ons don't have unit tests...I guess I should test compile with a demo or somesuch.

IBwWG:
Funny, compiles locally...

IBwWG:
Oh wait, I had confused threads here.

PR #1939, opened 2016-09-08
FlxSprite: fix position discrepancy between simple and complex render

### LINKED ISSUE #1917: simpleRender and complexRender position discrepancy
I initially discovered this issue when tweening the y scale of a FlxText. At a scale of 1 the text appeared correct, but at other scales the whole text would appear to shift left by a pixel. It seems that this is due to a discrepancy between simpleRender and complexRender, the latter of which seems to incorrectly handle pixelPerfectRender.

**Flixel version:** dev

```haxe
package;

import flixel.FlxG;
import flixel.FlxSprite;
import flixel.FlxState;
import flixel.text.FlxText;
import flixel.util.FlxColor;
import openfl.display.BlendMode;

@:access(flixel.FlxSprite)
class ComplexRenderBugState extends FlxState {
	
	var text:FlxText;
	var spr:FlxSprite;
	
	var textDesc:FlxText;
	var sprDesc:FlxText;
	
		super.create();
		
		text = new FlxText(100, 20, 0, "This is a test", 40);
		spr = new FlxSprite(100, 120);
		spr.makeGraphic(101, 40, FlxColor.RED);
		spr.updateHitbox();
		
		add(text);
		add(spr);
		
		textDesc = new FlxText(100, 85, 0, 16);
		sprDesc = new FlxText(100, 165, 0, 16);
		
		add(textDesc);
		add(sprDesc);
	}
	
		super.update(dt);
		
		if (FlxG.keys.justPressed.SPACE) {
			if (spr.blend == null) {
				spr.blend = BlendMode.NORMAL;
				text.blend = BlendMode.NORMAL;
			} else {
				spr.blend = null;
				text.blend = null;
			}
		}
		
		textDesc.text = desc(text);
		sprDesc.text = desc(spr);
	}
	
	private function desc(spr:FlxSprite):String {
		return "simpleRender = " + spr.isSimpleRenderBlit(FlxG.camera) +
			", x = " + spr.x + ", width = " + spr.width + ", _matrix.tx = " + spr._matrix.tx;
	}
}
```

**Observed behavior:** There is a discrepancy in the position of FlxSprites depending on whether simpleRender or complexRender is used. It is only an issue when sprites have an on width (or height, though that is not demonstrated in this test). Pressing space causes the render mode to change by adding a blend mode, after which you can see the rendered x position of the sprites change.
**Expected behavior:** The render mode should not cause the sprites to render at different positions.

I have already looked into this issue and I think I have a solution for which I will post a pull request imminently. It would be good if I could get a sanity check on this though.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/37250836-simplerender-and-complexrender-position-discrepancy?utm_campaign=plugin&utm_content=tracker%2F254189&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F254189&utm_medium=issues&utm_source=github).
</bountysource-plugin>

JoeCreates:
See #1917

Before for sprites with an odd width of height, the initial translation (using the offset) would be at a .5.

I'm not completely sure that this is correct, so it would be good to get a second opinion from @beeblerox.

### DIFF
--- a/flixel/FlxSprite.hx
+++ b/flixel/FlxSprite.hx
@@ -666,11 +666,15 @@ class FlxSprite extends FlxObject
 				_matrix.rotateWithTrig(_cosAngle, _sinAngle);
 		}
 		
-		_point.add(origin.x, origin.y);
-		if (isPixelPerfectRender(camera))
-			_point.floor();
-		
+		_point.add(origin.x, origin.y);		
 		_matrix.translate(_point.x, _point.y);
+		
+		if (isPixelPerfectRender(camera))
+		{
+			_matrix.tx = Math.floor(_matrix.tx);
+			_matrix.ty = Math.floor(_matrix.ty);
+		}
+		
 		camera.drawPixels(_frame, framePixels, _matrix, colorTransform, blend, antialiasing, shader);
 	}
 	


Gama11:
Looks ok to me, @Beeblerox gave his Ok as well.

Thanks for providing that test case in the issue, that really helped!

PR #1940, opened 2016-09-09
OpenFL 4 compatibility

Beeblerox:
This pull request adds openfl 4 compatibilty to HaxeFlixel and it still will work on openfl 3 also.
Openfl 3 backend still uses Tilesheet class, but openfl 4 uses custom DisplayObject. I had to make my own class, because Tilemap class is too restrictive and doesn't allow us to have multiple cameras support.
I had to reorganize the `FlxCamera` class: almost all rendering related logic has been moved to `FlxCameraView` class and its subsclasses, which are kind of backends for different targets. And every camera has `view` property which is an instance of `FlxCameraView` subclass. All drawing commands sent to the camera are forwarded to its view.
There are:
- `FlxBlitView` which works when game uses blit render mode
- `FlxHardwareView` which works on native targets.
  Plus there are plenty other helper classes added for batching handling.

One more good thing is that shaders are kinda work, but they aren't completely compatible since shader format in openfl had been changed (you'll need little work to convert them to work on openfl 4, nothing special).

The bad news is that this version contains some breaking changes:
1. Couple of demos (FlxBlur and FlxBloom) won't compile without small changes since drawing logic had been moved out of FlxCamera
2. I've mad some changes to `FlxStrip` class for the sake of optimization - you'll need to set `dirty = true;` every time you change its `vertices`, `indices` or `uvtData`, or it won't update it's visuals.

I'll start updating demos this sunday and hope to make another pull request to demos repo next week.


Gama11:
Wow, that's a lot of commits / changes! :+1: 

I think this will have to wait until 4.3.0 / after 4.2.0 is released, for several reasons:
- this has breaking changes, so 4.2.0 will be another release that's pretty stable
- this needs to be tested thoroughly

Right now there are also merge conflicts that need to be resolved.

larsiusprime:
Excellent work!

Beeblerox:
one more problem: it doesn't work with html5 yet. I've found couple blocking issues for me:
1. ids in `XInputID` aren't defined on html5. And i don't know what values to select for this platform
2. i can't see anything to render on screen (except for one `FlxText` object), even if i set XInput ids to random values for html5 and force renderMethod to `BLITTING`.
So i need to finish html5 support for openfl4 also. Hope to get help with XInputIDs values.

Gama11:
I get this error when trying to compile this branch right now:

```
C:/HaxeToolkit/haxe/lib/flixel/git/flixel/system/render/blit/FlxBlitView.hx:85: characters 16-52 : Array<Unknown<0>> should be openfl.Vector<Float>
```

I guess you're not getting that? What Haxe / OpenFL / Lime versions are you testing with?

Beeblerox:
i'm on Haxe 3.2.1, OpenFl 4.0.3 and Lime 3.0.3
I guess i need to change lines:
```
private static var rectVertices:Vector<Float> = [];
private static var rectUVs:Vector<Float> = [];
private static var rectIndices:Vector<Int> = [];
```
in `FlxBlitView`

How it's compiling now?

Gama11:
Compiles to Flash, but Neko still has issues:

```
C:/HaxeToolkit/haxe/lib/flixel/git/flixel/system/render/blit/FlxBlitView.hx:158: characters 41-49 : flixel.system.render.common.DrawData<Float> should be openfl.Vector<Float>
C:/HaxeToolkit/haxe/lib/flixel/git/flixel/system/render/blit/FlxBlitView.hx:158: characters 41-49 : Array<Float> should be openfl.Vector<Float>
C:/HaxeToolkit/haxe/lib/flixel/git/flixel/system/render/blit/FlxBlitView.hx:158: characters 41-49 : For function argument 'vertices'
C:/HaxeToolkit/haxe/lib/flixel/git/flixel/system/render/blit/FlxBlitView.hx:162: characters 21-29 : flixel.system.render.common.DrawData<Float> should be openfl.Vector<Float>
C:/HaxeToolkit/haxe/lib/flixel/git/flixel/system/render/blit/FlxBlitView.hx:162: characters 21-29 : Array<Float> should be openfl.Vector<Float>
C:/HaxeToolkit/haxe/lib/flixel/git/flixel/system/render/blit/FlxBlitView.hx:162: characters 21-29 : For function argument 'rectVertices'
```

Is this related to me using newer OpenFL versions?

Beeblerox:
Oh. i forgot to check latest changes if they work with openfl 3 before making pull request. I'll try to fix them tomorrow

MSGhero:
OpenFL changed Vector and might have broken toArray implicit conversions. I only skimmed though https://github.com/openfl/openfl/issues/1235

larsiusprime:
@Beeblerox : to get the ID's for a gamepad model on a new platform, there are some samples included with OpenFL and LIme -- one for joystick input and one for gamepad. I have a large collection of various controllers I can test against and report what ID's I get using OpenFL 4 for the given model.

Beeblerox:
@Gama11 i hope that i've fixed compilation issues. Btw i'm also using Openfl 3.6.1 and Lime 2.9.1 for testing compatibility with Openfl 3

Beeblerox:
@larsiusprime well, do you have any XInput device for testing values? because i have only XBox 360 controller.

Gama11:
`#if ((openfl >= "4.0.0") && !flash)` appears 9 times. Maybe it'd be a good idea to introduce a new helper define in `FlxDefines` for it?

Beeblerox:
@Gama11 good idea!

larsiusprime:
@Beeblerox Yes, I have several. I have a XBox 360 controller, an XBox one controller, PS3, PS4, Logitech, pretty much everything.

Beeblerox:
i understood why my code doesn't work on js target with openfl 4 - for some reason `gl.drawElements()` isn't working, while `gl.drawArrays()` works fine. 
This means i need to rewrite my code to work with this method (and this method requires a bit different input data), hope to make it before weekend. Plus i'll compare perfomance for both render methods on cpp target

MSGhero:
@Beeblerox do different browsers handle that differently?

Beeblerox:
@MSGhero i've tested this only in Chrome, but i've found that all openfl classes use `gl.drawArrays()` in OpenGL mode, even shader example in openfl-samples had been rewritten. So i guess, it's better to stick to this method for browsers

Beeblerox:
ok, made it to work on js target in webgl mode. need to clean it up now.
And about perfomance comparison: i know that it can be optimized (make less buffer uploads), but current gl.drawArrays() rendering is much slower than gl.drawElements (at least in Chrome with BunnyMark demo). So i'll stick with gl.drawElements() on cpp and fallback to gl.drawArrays on js target (because gl.drawElements() just doesn't work there for some reason).

sruloart:
@Beeblerox I don't know whether it's relevant or not (usually isn't), but [this](https://github.com/openfl/openfl/issues/1220) guy has apparently used `drawElements` successfully on html5 quite recently. Maybe there's something in his code that can help you with that?

Beeblerox:
@sruloart thanks for the link, i'll try to find differencies between my code and his

Beeblerox:
ok, now i need to make PostProcess and Filters work with openfl4, so we'll have per camera and game post processing effects.
Then i'll try to play with render textures for individual objects, so they could have more complex effects and fix some visual glitches (like in case of bitmap text with shadow).

flixel/tile/FlxTilemap.hx
```diff
@@ -48,13 +51,6 @@ class FlxTilemap extends FlxBaseTilemap<FlxTile>
 	 */
 	private static var _helperBuffer:FlxTilemapBuffer = Type.createEmptyInstance(FlxTilemapBuffer);
 	
-	// TODO: remove this hack and add docs about how to avoid tearing problem by preparing assets and some code...
-	/**
-	 * Try to eliminate 1 px gap between tiles in tile render mode by increasing tile scale, 
-	 * so the tile will look one pixel wider than it is.
-	 */
-	public var useScaleHack:Bool = true;
```
bendmorris: I agree with removing this, but it doesn't look like you've solved the problem with tile seams yet. The "TODO" above also glosses over the problem - if I'm using non-integer scaling, there's no way to prepare my assets to avoid tearing. Alpha bleed and tile gaps caused by the renderer are two different issues.

I had a pair of HaxePunk commits to fix the gap problem a few years ago. The idea was to always increment the draw position by an integer when moving to the next row or column, which would sometimes result in 1px difference in tile width/height but no gaps: https://github.com/HaxePunk/HaxePunk/commit/6622764cf0bce9fb568abffd0e0837f055ac7baa 
https://github.com/HaxePunk/HaxePunk/commit/2b8c81d1f42b5860a033415ed3ed6795942af261

Beeblerox: @bendmorris thank you for comment! 
With new gl renderer it should be possible to avoid this problem with shader based game scaling. But i'll try to reimplement your work for flixel as well.

Beeblerox: ok, i'll wait for PR then :)

flixel/util/FlxBitmapData.hx
```diff
@@ -1,10 +1,25 @@
...
 import openfl.display.BitmapData;
+import openfl.display.IBitmapDrawable;
 import openfl.display3D.textures.Texture;
+import openfl.filters.BitmapFilter;
+import openfl.geom.ColorTransform;
+import openfl.geom.Point;
+import openfl.geom.Rectangle;
 
 // TODO: implement it...
+// TODO: hack in openfl and lime asset management system to create FlxBitmapData objects (not just BitmapData)...
```
Gama11: Hm.. is that really a good idea?

flixel/util/FlxBitmapData.hx
```diff
@@ -1,10 +1,25 @@
...
 
 /**
  * I need this class for flash target with stage3D support
```
Gama11: Are you adding Stage3D support in this PR as well? I thought this was just about OpenFL 4 support.

flixel/util/FlxBitmapData.hx
Beeblerox: the only reason i'm adding this class is for easier texture management support on flash target (maybe it would be better to have on openfl level, but i'm not sure).
i think i could force users to use this class. This will be another breaking change, so i think i'd better to include this class here as well.

Gama11: I'm not a fan of that, that seems like a pretty big breaking change.

flixel/util/FlxBitmapData.hx
Beeblerox: this class will be part of stage3d support, and i can add the rest in the next PR.

Gama11: Yeah, I think it makes sense to keep that separate. There's enough changes to review and test here as is.

flixel/util/FlxBitmapData.hx
Beeblerox: yeah, i understand you concern. i see 2 options: 
1) implement it this way (and users will have to change all occurences of `openfl.display.BitmapData` with `flixel.util.FlxBitmapData`)
2) make PR to openfl with texture generation functionality and hope that it will be merged in the core.

Beeblerox: Well, i'll try to make pull request. 
But i don't think that it would be huge hack

Beeblerox: i've opened issue about it on openfl repo: https://github.com/openfl/openfl/issues/1270

Beeblerox:
Today i found that my renderer isn't good for features like render textures. While renderer works like old tilesheet renderer (have almost the same fuctionality), it doens't allow to have new features for move advanced effects. So here is the plan: we could merge this version and then i'll update it, or i could remake it while this PR isn't merged.

sruloart:
Considering @Gama11 plans to make the new renderer available only on the 4.3.0 release, it might be more efficient to continue working on the (upgraded) renderer. Such a major breaking change should not be continuously broken by even more changes.

Beeblerox:
@sruloart i'm planning to continue work on renderer in any way, but next iteration hardly will have breaking changes

Gama11:
>it doesn't allow to have new features for move advanced effects

What exactly do you mean by that?

larsiusprime:
> Today i found that my renderer isn't good for features like render textures. While renderer works like old tilesheet renderer (have almost the same fuctionality), it doens't allow to have new features for move advanced effects. So here is the plan: we could merge this version and then i'll update it, or i could remake it while this PR isn't merged.

I think this comes down to one big question:
Does the stated deficiency ("doesn't allow to have new features for more advanced effects") also apply to the old version? IE, is the new version *no worse* than the old version in this regards?

If so, then merge it as soon as it's stable, and we'll deal with the refactor to allow more flexibility for new features as the next step. If not, then add the new functionality before merging.

Beeblerox:
new version doesn't support support color offsets (but it can be fixed with current code).
And by `it doesn't allow to have new features for move advanced effects` i mean that you can, for example, draw sprite on bitmapdata and use this bitmapdata as a texture source (things work this way now), but you can's draw sprite on native render texture and use this texture later (plus it should be faster)

Gama11:
I think ideally, we should not drop any features with these changes - for instance, we support color offsets right now (for OpenFL < 4 + Next), and if we suddenly don't in 4.3.0, people using it may want to stay on outdated Flixel versions (which we should definitely avoid).

Beeblerox:
ok, i'll bring color offsets back

flixel/util/FlxBitmapData.hx
```diff
@@ -0,0 +1,231 @@
...
+import openfl.geom.Point;
+import openfl.geom.Rectangle;
+
+// TODO: hack in openfl and lime asset management system to create FlxBitmapData objects (not just BitmapData)...
+
+/**
+ * I need this class for flash target with stage3D support
+ * @author Zaphod
+ */
+class FlxBitmapData extends BitmapData
```
Gama11: I don't think this should be included in this PR if it's not needed?

flixel/system/render/hardware/gl/HardwareRenderer.hx
```diff
@@ -0,0 +1,322 @@
...
+					gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MAG_FILTER, gl.LINEAR);
+					gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MIN_FILTER, gl.LINEAR);	
+				} 
+				else 
+				{
+					gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MAG_FILTER, gl.NEAREST);
+					gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MIN_FILTER, gl.NEAREST);
+				}
+				
+				#if !js
```
Gama11: Does this mean something is broken on JS? Or are these calls just not needed there?

flixel/system/render/hardware/gl/GLRenderHelper.hx
```diff
@@ -0,0 +1,352 @@
...
+		var filterIndex:Int = 0;
+		var i:Int = 0;
+		
+		while (i < passes && filterIndex < numFilters)
+		{
+			shader = null;
+			
+			if (Std.is(filters[filterIndex], ShaderFilter))
+			{
+				shader = cast(filters[filterIndex], ShaderFilter).shader;
```
Gama11: Probably shouldn't use a safe cast here - they have unnecessary runtime overhead (unnecessary you're already checking with `Std.is()` anyway - safe casts are only necessary when you want to try-catch it).

https://haxe.org/manual/expression-cast.html

flixel/system/render/common/FlxCameraView.hx
```diff
@@ -0,0 +1,302 @@
...
+	/**
+	 * Tracks total number of drawTiles() calls made each frame.
+	 */
+	public static var _DRAWCALLS:Int = 0;
+	
+	/**
+	 * Display object which is used as a container for all the camera's graphic.
+	 * This object is added to display tree.
+	 */
+	public var display(get, null):DisplayObject;
```
Gama11: Shouldn't this be `(get, never)`?

flixel/math/FlxRect.hx
```diff
@@ -389,6 +389,33 @@ class FlxRect implements IFlxPooled
 		return FlxRect.get(x0, y0, x1 - x0, y1 - y0);
 	}
 	
+	public inline function inflate(x:Float, y:Float):FlxRect
```
Gama11: This seems too long to `inline`.

Gama11:
Have you done any performance benchmarks yet? Comparing the current dev branch + -Dnext with your changes in FlxBunnyMark or something like that.

flixel/system/FlxAssets.hx
```diff
@@ -156,7 +161,7 @@ class FlxAssets
 		{
 			return cast Graphic;
 		}
-		else if (Std.is(Graphic, Class))
+		else if (Std.is(Graphic, Class)) // TODO: maybe remove ability to get BitmapData from class???
```
Gama11: Why?

flixel/system/render/hardware/gl/HardwareRenderer.hx
```diff
@@ -0,0 +1,322 @@
...
+import openfl.geom.Rectangle;
+
+// TODO: try to add general vertex and index arrays to minimize data upload operations (gl.bufferData() calls). Like it's done in GL implementation of Tilemap renderer...
+// TODO: multitexture batching...
+// TODO: sprite materials with multiple textures...
+// TODO: support for colorOffsets???
+
+/**
+ * ...
+ * @author Yanrishatum
```
Gama11: Who's this? :)

flixel/util/FlxBitmapData.hx
Beeblerox: ok, i'll remove it from this PR

flixel/system/render/hardware/gl/HardwareRenderer.hx
Beeblerox: it's Pavel Alexandrov aka Yanrishatum https://github.com/Yanrishatum 
i took big part from his fork of HaxePunk and he didn't mind that i used his work as a basis
Will add links in the next commit

Beeblerox:
@Gama11  yes, i've made some tests and my code is much slower than Tilesheet version, but i'm working on making it faster

flixel/math/FlxRect.hx
Beeblerox: agree, will remove `inline` in the next commit

flixel/system/render/common/FlxCameraView.hx
Beeblerox: will do :)

flixel/system/render/hardware/gl/HardwareRenderer.hx
Beeblerox: yes, it causes errors on js target. something like incorrect parameter value.
i just want to be sure that texture repeating is on.

bendmorris:
Hi @Beeblerox, I have a similar HaxePunk pull request in that you might be interested in: https://github.com/HaxePunk/HaxePunk/pull/406

For one thing, I was able to avoid extending DisplayObject by instead grabbing the RenderSession from the Stage and drawing to an existing DisplayObject. This made it possible to create a new drawTiles API which seems to work exactly as before, and other graphic classes didn't need to change at all.

I don't know a lot about Flixel rendering, just posting this in case it's helpful. Thanks for all your great work.

Beeblerox:
@bendmorris i've seen your code. it's great that you've managed to avoid extending DisplayObject. The code is clean and easy to read.
But unfortunately flixel uses more than just Tilesheet class. It uses drawTrianlges operations as well and should be able to "mix" them. And while drawTiles can be seen as subset of drawTriangles operation, it could be optimized (like precalculation of values in index array), so i need to 2 sets for each of operations.
Plus multiple cameras support requires to have ability to draw colored (non-textured) quads for background fill. So i'd rather stick to my way and will try to optimize it. I'm already have an idea how to do it and will show results soon (hopefully).

starry-abyss:
Is this far from completion yet? I remember to have some problems with compilation, which I by mistake posted in addons repo

Beeblerox:
@starry-abyss unfortunately, yes. i still need to do (and rewrite) a lot of work. but i think it should be completed by the end of the year :(

larsiusprime:
@Beeblerox -- anything I can do to help this one along?

Beeblerox:
@larsiusprime i'm working on it, you can track progress here: https://github.com/Beeblerox/flixel/tree/pixi-batcher 
i've changed the first version of batcher in order to make less data uploads. Now i'm working on compatibility with older versions of openfl (3.6.1) and lime (2.9.1)

Beeblerox:
i'm closing this PR, because of multiple conflicts.
New PR could be found here: https://github.com/HaxeFlixel/flixel/pull/2032

Issue #1929, opened 2016-09-06
Screenshake behavior during pause


MSGhero:
- **Flixel version:** latest
- Flash target, maybe others

________________________________

When you screenshake, pause, and step the game forward one frame, the display starts moving (and doesn't stop) in the random direction that was chosen by `updateShake`. This is because of the 3rd linked snippet below, where `draw()` gets called from FlxGame's enterframe despite things being paused. Shake is updated during the draw phase in the 1st linked snippet, which moves the display.

I'm not sure what other FX there are, but they are probably affected too. This is just an issue while paused, as the display will snap back to normal on the next frame.

https://github.com/HaxeFlixel/flixel/blob/dev/flixel/FlxCamera.hx#L1409-L1413
https://github.com/HaxeFlixel/flixel/blob/dev/flixel/FlxGame.hx#L899
https://github.com/HaxeFlixel/flixel/blob/dev/flixel/FlxGame.hx#L540

Beeblerox:
i think that other effects (fade and flash) shouldn't be affected.
the easiest way to fix this issue is to change this line https://github.com/HaxeFlixel/flixel/blob/dev/flixel/FlxCamera.hx#L1409 to something like
```
if (((_fxShakeOffset.x != 0) || (_fxShakeOffset.y != 0)) && FlxG.state.active)
```
But i don't like it.

bendmorris:
HaxePunk used to have this issue for particles, the fix is simply to not do state modification in the draw method, but move it into update methods. In this case you could modify flashSprite.x/y in updateShake instead of in draw. Easy fix, PR incoming.

Beeblerox:
or move this block of code https://github.com/HaxeFlixel/flixel/blob/dev/flixel/FlxCamera.hx#L1409-L1413 into `updateShake()` method, which is not completely right too, since this method can be called few times per frame (update rate can be higher than draw rate)

bendmorris:
>update rate can be higher than draw rate

If that's the case it's not really a problem, right? If flashSprite gets recentered and moved to a random offset multiple times before drawing, the end result is the same, it's offset by some random amount.

Beeblerox:
yes, right!

MSGhero:
Thanks @bendmorris

PR #1941, opened 2016-09-10
Move FlxCamera screen-shake movement to updateShake.

bendmorris:
Fixes https://github.com/HaxeFlixel/flixel/issues/1929

### DIFF
--- a/flixel/FlxCamera.hx
+++ b/flixel/FlxCamera.hx
@@ -859,11 +859,11 @@ class FlxCamera extends FlxBasic
 		updateScroll();	
 		updateFlash(elapsed);
 		updateFade(elapsed);
-		updateShake(elapsed);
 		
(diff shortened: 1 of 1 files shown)


bendmorris:
With this change I don't think _fxShakeOffset is even needed anymore. It's used in `shake` to determine whether another shake is already happening, but that seems wrong, could use _fxShakeDuration for that.

Gama11:
Nice fix, less bugs _and_ less code. ;)

MSGhero:
Travis? https://travis-ci.org/HaxeFlixel/flixel/jobs/159004450#L1484-L1486

Only on dev cpp tho

Gama11:
Yeah, not sure what that's about. That happens every now and then, apparently randomnly. I guess I'll open a separate issue about that because it's getting pretty annoying...

Nothing to do with these changes though!

Gama11:
See #1942.

PR #1943, opened 2016-09-10
FlxSound: add endTime, closes #1906

### LINKED ISSUE #1906: Feature request: allow setting sound looping end point as well
FlxSound.loopTime presumes you want to loop from point X until the end of the track.  What if you want to loop from the start of the track to point X?  Or from point X to point Y?  If it's possible, I think it'd be much more flexible to have a loopEndTime as well.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/36997943-feature-request-allow-setting-sound-looping-end-point-as-well?utm_campaign=plugin&utm_content=tracker%2F254189&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F254189&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Beeblerox:
See issue #1906

### DIFF
--- a/flixel/system/FlxSound.hx
+++ b/flixel/system/FlxSound.hx
@@ -740,21 +740,18 @@ class FlxSound extends FlxBasic
 	
 	private function set_loopTime(value:Float):Float
 	{
-		loopTime = value;
 		loopTime = FlxMath.bound(value, 0, length);
 		return value;
 	}
 	
 	private function set_loopEndTime(value:Float):Float
 	{
-		loopEndTime = value;
 		loopEndTime = (value <= loopTime) ? length : value;
 		return value;
 	}
 	
 	private function set_endTime(value:Float):Float
 	{
-		endTime = value;
 		endTime = (value <= 0) ? length : value;
 		return value;
 	}
--- a/flixel/system/FlxSound.hx
+++ b/flixel/system/FlxSound.hx
@@ -104,7 +104,15 @@ class FlxSound extends FlxBasic
 	/**
 	 * In case of looping, the point (in milliseconds) from where to restart the sound when it loops back
 	 */
-	public var loopTime:Float;
+	public var loopTime(default, set):Float;
+	/**
+	 * In case of looping, the point (in milliseconds) from where to restart the sound when it reaches this time.
+	 */
+	public var loopEndTime(default, set):Float;
+	/**
+	 * In case of non-looping, at which point to stop playing the sound, in milliseconds
+	 */
+	public var endTime(default, set):Float;
 	/**
 	 * The tween used to fade this sound's volume in and out (set via `fadeIn()` and `fadeOut()`)
 	 */
@@ -189,6 +197,8 @@ class FlxSound extends FlxBasic
 		_volumeAdjust = 1.0;
 		looped = false;
 		loopTime = 0.0;
+		loopEndTime = 0.0;
+		endTime = 0.0;
 		_target = null;
 		_radius = 0;
 		_proximityPan = false;
@@ -272,6 +282,15 @@ class FlxSound extends FlxBasic
 			amplitudeRight = 0;
 			amplitude = 0;			
 		}
+		
+		if (looped && _time >= loopEndTime)
+		{
+			stopped(null);
+		}
+		else if (!looped && _time >= endTime)
+		{
+			stopped(null);
+		}
 	}
 	
 	override public function kill():Void
@@ -314,7 +333,8 @@ class FlxSound extends FlxBasic
 		}
 		
 		// NOTE: can't pull ID3 info from embedded sound currently
-		looped = Looped; 
+		looped = Looped;
+		loopEndTime = _sound.length;
 		autoDestroy = AutoDestroy;
 		updateTransform();
 		exists = true;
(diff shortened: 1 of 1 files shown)


flixel/system/FlxSound.hx
```diff
@@ -713,6 +737,36 @@ class FlxSound extends FlxBasic
 		}
 		return _time = time;
 	}
+	
+	private function set_loopTime(value:Float):Float
+	{
+		loopTime = value;
+		if (_sound != null)
```
Gama11: Couldn't we use the `length` property to avoid the null checks?

Beeblerox: yes, of course. i've missed this new property)

Gama11: You forgot to remove the null checks. :)

Beeblerox: ok. removed them

flixel/system/FlxSound.hx
```diff
@@ -713,6 +731,21 @@ class FlxSound extends FlxBasic
...
 	}
+	
+	private function set_loopTime(value:Float):Float
+	{
+		return loopTime = FlxMath.bound(value, 0, length);
+	}
+	
+	private function set_loopEndTime(value:Float):Float
+	{
+		return loopEndTime = (value <= loopTime) ? length : value;
```
Gama11: So, you prevent invalid `loopEndTime` values here... However, what these values are depends on `loopTime`. What happens if `loopTime` is set _after_ `loopEndTime`?

flixel/system/FlxSound.hx
```diff
@@ -104,7 +104,15 @@ class FlxSound extends FlxBasic
 	/**
 	 * In case of looping, the point (in milliseconds) from where to restart the sound when it loops back
 	 */
-	public var loopTime:Float;
+	public var loopTime(default, set):Float;
+	/**
+	 * In case of looping, the point (in milliseconds) from where to restart the sound when it reaches this time.
+	 */
+	public var loopEndTime(default, set):Float;
```
Gama11: Thinking about this some more, why do we need two separate properties? They don't really seem that different from each other. Wouldn't a single `endTime` be enough, which then handles either the loop point if `looped == true`, or the end point if `looped == false`?

When having two separate properties, users could get into confusing situations, where they use the wrong one (for example `endTime` for a looping sound) and nothing happens.

flixel/system/FlxSound.hx
```diff
@@ -713,6 +726,18 @@ class FlxSound extends FlxBasic
...
+	private function set_loopTime(value:Float):Float
+	{
+		loopTime = FlxMath.bound(value, 0, length);
+		set_endTime(endTime);
+		return loopTime;
+	}
+	
+	private function set_endTime(value:Float):Float
+	{
+		return endTime = (value <= loopTime) ? length : value;
```
Gama11: I don't think this check is still correct - `value <= loopTime` only matters if the sound is `looped`. This probably means that we'd need a `set_looped()` as well.

flixel/system/FlxSound.hx
```diff
@@ -713,6 +726,18 @@ class FlxSound extends FlxBasic
 		}
 		return _time = time;
 	}
+	
+	private function set_loopTime(value:Float):Float
+	{
+		loopTime = FlxMath.bound(value, 0, length);
```
Gama11: Wouldn't we need the inverse check of `set_endTime` here? It doesn't make sense for `loopTime` to be bigger than `endTime`...

Gama11:
I'm not sure how much is gained by having setters for `loopTime` and `endTime`. As my last two comments show, it's getting quite complex...

Maybe it's better to just trust the user to supply correct values? That way there are no bugs caused by Flixel / bounding the values at least.

Beeblerox:
yeah, i think it would be better. 
sorry for such messed pull request for such simple feature

Gama11:
It's really not as simple as it looks, there's a lot of things to take into account. :)

flixel/system/FlxSound.hx
```diff
@@ -404,8 +415,9 @@ class FlxSound extends FlxBasic
 	 *                         paused when you call play(), it will continue playing from its current
 	 *                         position, NOT start again from the beginning.
 	 * @param   StartTime      At which point to start plaing the sound, in milliseconds
+	 * @param	EndTime			At which point to stop playing the sound, in milliseconds
 	 */
-	public function play(ForceRestart:Bool = false, StartTime:Float = 0.0):FlxSound
+	public function play(ForceRestart:Bool = false, StartTime:Float = 0.0, EndTime:Float = 0.0):FlxSound
```
Gama11: `EndTime == 0` here used to mean "set it to `length`", but now the behavior of `set_endTime()` is different.

Maybe it should be `?EndTime:Float`, and `if (EndTime == null)`, we set it to `length`?

Beeblerox: done

Gama11:
Looks good now, thanks! :+1:

Issue #911, opened 2014-03-02
Serious floating-point artifacts on CPP FlxTilemap & FlxSprite


cwkx:
![pic](https://f.cloud.github.com/assets/5144374/2305507/4590e95e-a25a-11e3-9b98-1af90c01ffee.png)

I made all transparent tiles in my tileMap bright green to highlight the issue. The boundaries of some tiles are showing pixel data from their adjacent neighbours.

(1) This only occurs on some resolutions. Try lots of different resolutions.
(2) This occurs infrequently; best way to "find" it is by moving slowly in x and y axis. Most likely some floating point issue.

Using latest version from github.
##

JoeCreates:
Just wanted to check what the state of this is these days, seeing as it's been a while. Any news or leads?

RawArkanis:
As far I know you need to use `FlxG.camera.pixelPerfectRender = false;` in your FlxState.

Tiago-Ling:
Hey, i solved my problem with artifacts in cpp targets (windows & android) by using the following combination:
* FlxG.camera.pixelPerfectRender = false;
* FlxG.camera.antialiasing = true;
* FlxSprite.pixelPerfectPosition = false; (applied to every sprite)

At least for FlxSprites this fixed the artifact problem in my project (using release versions of Lime, OpenFL and Hxcpp and dev version of HaxeFlixel).

However i'm still suffering from the 'blur' when moving the camera - it is much stronger than in any other framework or game i've seen so far. When the camera is moving fast the objects appear to be 'shaking'. Solving this would make the visual part of any scrolling game made with the framework a hundred times better.

gamedevsam:
Could you try disabling antialiasing and see if the camera blur seems
reduced? I hadn't heard about the camera blur before, hope you figure out
how to fix it...

Tiago-Ling:
I'll try that and to record a video showing the problem (it is hard because of video encoding, frame skipping and other things that hide the problem).

Beeblerox:
i can confirm "blur" issue, but it's not blur, it's more like camera is slightly jumping.

Beeblerox:
and it looks like different objects on the camera scrolls a little bit differently (just by one pixel)

Beeblerox:
i think it might be something with camera's scroll behavior of with sprite onscreen position calculations

Tiago-Ling:
If it has something to do with the camera, it seems to be down to two methods that might be part of the problem:
* updateFollow
[...]

Issue #1951, opened 2016-09-16
Major jittering when scaling and rotating sprites


JoeCreates:
- **Flixel version:** github ev

________________________________

**Code snippet reproducing the issue**:

```package;

import flixel.FlxG;
import flixel.FlxSprite;
import flixel.FlxState;
import flixel.text.FlxText;
import flixel.tweens.FlxTween;
import flixel.ui.FlxButton;
import flixel.math.FlxMath;

class MenuState extends FlxState
{
	override public function create():Void
	{
		FlxG.camera.pixelPerfectRender = true;
		super.create();
		var spr:FlxSprite = new FlxSprite();
		spr.x += 0.5;
		add(spr);
		spr.screenCenter();
		FlxTween.tween(spr.scale, {x: 3, y: 3}, 2, {type: FlxTween.PINGPONG});
	}

	override public function update(elapsed:Float):Void
	{
		super.update(elapsed);
	}
}
```
This seems to have been introduced in a very recent commit. I'm looking into it now. Sprites jitter around when scaling on rotating.

JoeCreates:
Looks like it started after my change in https://github.com/HaxeFlixel/flixel/commit/462d76d7ab0c2de1f355de2231b29f6dcbdba983

Gama11:
This should probably be looked into before 4.2.0 is released, since it's sort of a regression.

JoeCreates:
Mentioned this in chat, but its not really a bug. More just that fact that for complex rendered sprites pixelperfectposititon never worked. Now that it does, gradually scaling things with pixelperfectrender has unexpected results. The solution is to simply set pixelperfectrender to false on things if you want their top left to not be at an exact pixel when it is scaled or rotated, and this will result in the old behaviour of rendering things at float positions.

Gama11:
Right, that's why I said "sort of a regression" - can't really think of a better label. :)

besserwisser:
I do have major jittering as well when rotating or moving sprites. It does not matter if I have pixelperfectposition set to false or true. Also undoing the changes in the FlxSprite Class had no impact on it. Even if i just put this in the update loop:
`_player.angle += 0.1;`
it jitters a lot. 

Do you have found another solution I can try?

JoeCreates:
@besserwisser Its pixelperfectrender you want to set to false, not pixelperfectposition.

besserwisser:
[...]

Issue #1861, opened 2016-05-29
FlxSprite+FlxAtlas distortions on native targets


starry-abyss:
- **Flixel version:** dev, c561696
- **OpenFL version:** 3.6.1
- **Lime version:** 2.9.1
- **Affected targets:** neko, windows

Hello, I've come across an issue with sprite rendering using atlases. It doesn't occur on Flash, and it doesn't occur when not using FlxAtlas nodes for sprite graphics. It _seems_ to occur more frequently with 3x zoom, and subpixel sprite positions.

Here is a little demo that shows the issue:
https://github.com/starry-abyss/render_test_cpp

Here are some grabs from main project:
http://imgur.com/a/cqvbM

Take attention to bear's eyes, ears and honey cut into halves with diagonal corner-to-corner line (apparently, triangles in quad). Also on GIFs you can see that the frame with bear's eyes shut is rendered differently from other frames.

Beeblerox:
i'm sorry. i don't quite understand what is the problem. Could you describe it in another words?

starry-abyss:
Sure.
The problem is that pixels are drawn imprecisely in certain scenarios (atlas is 100% relevant): 
![distortion](http://i.imgur.com/SAIsa7d.png)

Also, the demo above is a bit interactive (press Space to switch between modes).
Note that you **won't see** the distortion **on Flash** target

Beeblerox:
I think that this is a general problem, which could be seen not only flixel but in any other opengl-based renderer. So possible solution could be round rendering position of the sprite.
Try to add `FlxG.camera.pixelPerfectRender = true;` somewhere in your state's `create()` method. This will force rounding position of all objects, being rendered on default camera.

starry-abyss:
Thanks for the answer.
With ```FlxG.camera.pixelPerfectRender=true``` everything looks jittery.
And I don't think this is general, because exactly the same input, but without FlxAtlas, gives pixel perfect results (please check demo when you can). For me it looks like atlas is adding to imprecision.

Beeblerox:
Will try to do it tonight

Beeblerox:
i'm afraid it's something with opengl renderer (not on flixel level, but on openfl level). 
i will create simple demo without flixel dependency to proof it. 
@starry-abyss will you help me with checking demo when it will be ready?

starry-abyss:
@Beeblerox Thanks for your effort, yes, you can reach me!

larsiusprime:
[...]

PR #1953, opened 2016-09-17
Fix for FlxTilemap with pixelPerfectRender.

bendmorris:
The goal is to make useScaleHack on FlxTilemap unnecessary, by adjusting tile widths/heights as needed to fill exactly the right amount of space. This totally eliminates gaps between tiles, and even works with tilesets that use alpha (useScaleHack causes tiles to overlap, which with alpha != 1 creates visible seams.) This also solves some issues when the camera \* screen are scaled at a non-integer value, and flooring coordinates doesn't actually result in integer positions after applying scale.

~~Would appreciate some more eyes on this. There's still one big issue, which is that a FlxSprite on top of a FlxTilemap using pixelPerfectRender appears to "jitter" as the camera moves, due to slightly different scrolling behavior. I'm guessing this is a difference in the order of flooring coordinates, and may be related to https://github.com/HaxeFlixel/flixel/issues/1951?~~ The camera jitter issue is fixed.

### DIFF
--- a/flixel/FlxCamera.hx
+++ b/flixel/FlxCamera.hx
@@ -1538,12 +1538,12 @@ class FlxCamera extends FlxBasic
 	
 	public inline function floorX(x:Float):Float
 	{
-		return Math.floor(x * totalScaleX) / totalScaleX;
+		return FlxG.renderBlit ? Math.floor(x) : (Math.floor(x * totalScaleX) / totalScaleX);
 	}
 	
 	public inline function floorY(y:Float):Float
 	{
-		return Math.floor(y * totalScaleY) / totalScaleY;
+		return FlxG.renderBlit ? Math.floor(y) : (Math.floor(y * totalScaleY) / totalScaleY);
 	}
 	
 	private function set_followLerp(Value:Float):Float
--- a/flixel/FlxCamera.hx
+++ b/flixel/FlxCamera.hx
@@ -1525,6 +1525,27 @@ class FlxCamera extends FlxBasic
 		setScale(scaleX, scaleY);
 	}
 	
+	/**
+	 * Helper function to make sure a point will use integer coordinates after
+	 * applying scale.
+	 */
+	public inline function floorPoint(point:FlxPoint):FlxPoint
+	{
+		point.x = floorX(point.x);
+		point.y = floorY(point.y);
+		return point;
+	}
+	
+	public inline function floorX(x:Float):Float
+	{
+		return Math.floor(x * totalScaleX) / totalScaleX;
+	}
+	
+	public inline function floorY(y:Float):Float
+	{
+		return Math.floor(y * totalScaleY) / totalScaleY;
+	}
+	
 	private function set_followLerp(Value:Float):Float
 	{
 		return followLerp = FlxMath.bound(Value, 0, 60 / FlxG.updateFramerate);
--- a/flixel/FlxObject.hx
+++ b/flixel/FlxObject.hx
@@ -810,28 +810,33 @@ class FlxObject extends FlxBasic
 	/**
 	 * Call this function to figure out the on-screen position of the object.
 	 * 
-	 * @param	Point		Takes a FlxPoint object and assigns the post-scrolled X and Y values of this object to it.
-	 * @param	Camera		Specify which game camera you want.  If null getScreenPosition() will just grab the first global camera.
+	 * @param	point		Takes a FlxPoint object and assigns the post-scrolled X and Y values of this object to it.
+	 * @param	camera		Specify which game camera you want.  If null getScreenPosition() will just grab the first global camera.
 	 * @return	The Point you passed in, or a new Point if you didn't pass one, containing the screen X and Y position of this object.
 	 */
-	public function getScreenPosition(?point:FlxPoint, ?Camera:FlxCamera):FlxPoint
+	public function getScreenPosition(?point:FlxPoint, ?camera:FlxCamera):FlxPoint
 	{
 		if (point == null)
 		{
 			point = FlxPoint.get();
 		}
-		if (Camera == null)
+		if (camera == null)
 		{
-			Camera = FlxG.camera;
+			camera = FlxG.camera;
 		}
 		
 		point.set(x, y);
 		if (pixelPerfectPosition)
 		{
-			point.floor();
+			camera.floorPoint(point);
+		}
+		point.subtract(camera.scroll.x * scrollFactor.x, camera.scroll.y * scrollFactor.y);
+		if (pixelPerfectPosition)
+		{
+			camera.floorPoint(point);
 		}
 		
-		return point.subtract(Camera.scroll.x * scrollFactor.x, Camera.scroll.y * scrollFactor.y);
+		return point;
 	}
 	
 	public function getPosition(?point:FlxPoint):FlxPoint
--- a/flixel/FlxSprite.hx
+++ b/flixel/FlxSprite.hx
@@ -644,6 +644,8 @@ class FlxSprite extends FlxObject
 	
 	private function drawSimple(camera:FlxCamera):Void
 	{
+		if (isPixelPerfectRender(camera))
+			camera.floorPoint(_point);
 		_point.copyToFlash(_flashPoint);
 		camera.copyPixels(_frame, framePixels, _flashRect,
 			_flashPoint, colorTransform, blend, antialiasing);
--- a/flixel/FlxSprite.hx
+++ b/flixel/FlxSprite.hx
@@ -644,9 +644,6 @@ class FlxSprite extends FlxObject
 	
 	private function drawSimple(camera:FlxCamera):Void
 	{
-		if (isPixelPerfectRender(camera))
-			_point.floor();
-		
 		_point.copyToFlash(_flashPoint);
 		camera.copyPixels(_frame, framePixels, _flashRect,
 			_flashPoint, colorTransform, blend, antialiasing);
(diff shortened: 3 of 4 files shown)


MSGhero:
When you say make `useScaleHack` unnecessary, are you planning on removing it from the API? It's still being used after your changes.

bendmorris:
Since the OpenFL 4 PR is still a work in progress, I would suggest merging it into dev first, then @Beeblerox can add this behavior to his PR as well.

I also don't see a need to remove useScaleHack right away (it's good enough for some people, and there could be undiscovered issues with mine) but @Beeblerox will remove it when his PR lands.

Gama11:
This is supposed to close #911?

bendmorris:
Hadn't seen that one. Maybe :)

flixel/FlxCamera.hx
```diff
@@ -1525,6 +1525,27 @@ class FlxCamera extends FlxBasic
 		setScale(scaleX, scaleY);
 	}
 	
+	/**
+	 * Helper function to make sure a point will use integer coordinates after
+	 * applying scale.
+	 */
+	public inline function floorPoint(point:FlxPoint):FlxPoint
```
Gama11: Strikes me as a little weird to have these functions as public API, considering they're only used for `FlxTilemap`.. Do you anticipate they'll be useful elsewhere? If not, we could probably just make them `private` and use `@:access`, or move them to `FlxTilemap` alltogether.

bendmorris: I do think they might be useful elsewhere. Seems like in a lot of places we're flooring before camera scale is applied, resulting in final positions that aren't actually integers. Not sure how many places this might be needed, but I'm guessing more than just FlxTilemap. I originally used these in FlxObject and FlxSprite too, but removed them until I can figure out the jittering issue.

bendmorris:
Added another commit which adds similar flooring in FlxObject and FlxSprite. I'm not sure if this is 100% correct, hopefully someone else can test this out, but with `pixelPerfectPosition = pixelPerfectRender = true` and `useScaleHack = false` I'm seeing zero gaps in a FlxTilemap, and zero jitter on overlaid FlxTilemap/FlxSprite/FlxText objects.

Gama11:
`isPixelPerfectRender()` isn't being checked at all anymore now in `FlxSprite`? That doesn't seem right.

I'd like @JoeCreates to check these changes.

bendmorris:
It is checking pixelPerfectPosition (in getScreenPosition.) From the docs, I'm not really clear what the difference between the two is supposed to be.

bendmorris:
It would be great to have tests for this. Are there any examples of automated tests that assert on the output of the tile renderer that I could look at?

Gama11:
No, I don't think we really have many tests for rendering right now. It's pretty tricky to even get an accurate representation of what's on screen at any particular moment on native targets... see https://github.com/HaxeFlixel/flixel-addons/issues/82

JoeCreates:
I thought the seams issue was pretty much solved by using the padding method on the tileset. This is seams specifically on native we are talking about, right? 

Just to clarify the purpose of pixelPerfectPosition, this is quite differnt to pixelPerfectRender in that it floors the position of sprites hitbox before rendering, rather than the final rendered position of yhe graphic. The reason to do this is that if your camera is slowly moving and you have pixelPerfectRender set to true, the sprites will all jitter is their offsets dont all have the same fractional part. The render position of the graphic is calculated using each sprites offset, which will often all have different fractional parts, and when this is summed with the cameras position they will all be just over or just under some integer, and as the camera moves they will jutter between just over and just under. pixelPerfectPosition should not be treated or used instead of pixelPerfectRender. The former floors the sprite position /before/ calculating the screen position, the latter floors the /graphic's/ position /after/ calculating the screen position.

bendmorris:
Currently there are still cases that show gaps unless you use useScaleHack, which is removed in @Beeblerox's OpenFL 4 pull request (and IMO, it's a hack, removing it is the right call.)

There are two problems that can cause gaps or seams in tiles: (1) texture bleeding (results in colored seams that sample adjacent tiles) and (2) when the screen scale is not a whole number, the renderer actually drawing them in the wrong place, resulting in gaps the same color as the background where nothing was drawn. (1) is solved by adding colored borders to tiles. (2) is the problem useScaleHack is supposed to solve (making tiles 1px bigger to cover those gaps.) useScaleHack is a hack and not a solution though, it can cause pixel-perfect graphics to get stretched out inconsistently, and doesn't work at all for partially transparent tiles since the tiles will overlap each other by design.

I created a demo to make it easier to experiment with different combinations of settings: https://github.com/bendmorris/tile-gap-demo

Here's an example of a gap without useScaleHack, note the black gap on the far right side:

screen shot 2016-09-21 at 7 51 09 am

This is despite using `fromBitmapAddSpacesAndBorders` to pad the tiles.

Here's an example of useScaleHack and how tiles overlap:

screen shot 2016-09-21 at 8 41 59 am

Notice the tiles have a 1px border which is a mix of both colors.

This is an attempt to properly solve that problem without creating overlaps, by using whole pixel coordinates and dimensions when rendering, taking into account screen scale.

re: pixelPerfectPosition vs. pixelPerfectRender, I see that they do different things, but I don't get the difference in purpose. If I want to render to integer positions and also don't want them to jitter, why would I ever use one without the other?

bendmorris:
From what I understand, flooring coordinates at any point before rendering is too simplistic and won't necessarily give the desired result. With `scaleMode.scale.x = 1.9`, if I draw a FlxSprite with x = 201.1, I'll floor it to 201, then drawTiles will multiply by 1.9 to get 381.9. I've worked around this by adding FlxCamera.floorX and FlxCamera.floorY which account for the total scale.

Beeblerox:
i'm planning to add shader based scale modes to the next version of flixel (first version implemented by @Seanw265 in this PR https://github.com/HaxeFlixel/flixel/pull/1826). This should cover second case.
The idea behind this is to render everything in texture with original size and then draw this texture scaled on the screen.

MSGhero:
^ that's how I assumed it was always done, minus the shader. I never really understood or cared about the pixel perfect stuff until now

bendmorris:
Also relevant: https://github.com/HaxeFlixel/flixel/issues/1861

I believe this PR would resolve that ticket: pixelPerfectPosition and pixelPerfectRender would get rid of the distortion according to this comments, and this PR gets rid of the "shaking" effect that makes those flags less usable.

JoeCreates:
Often it is desirable not to use pixelPerfectRender on native, seeing as a smoother result is achieved with subpixel rendering. If you are scaling up your world, you may still want a smooth camera, which on native can be achieved by having pixelPerfectRender off. I tend to prefer this, myself.

pixelPerfectPosition ensures the relative positions of sprites to each other stays the same when only the camera is moving and pixelPerfectRender is on. If pixelPerfectRender is on and pixelPerfectPosition is off, you will get the jittering, which for a long time was the only behavior. Using pixelPerfectPosition without pixelPerfectRender can certainly be useful if you wish to have stuff in your world snap to pixels, but still make use of subpixel rendering for camera smoothness, i.e. your world will all be on a grid, but your camera can move smoothly between pixels on that grid on native.

Note that jittering during camera motion can still occur is you have sprites (including tilemaps) that are inconsistently using pixelPerfectPosition or pixelPerfectRender, for example a map with pixelPerfectPosition = true and a sprite that has pixelPerfectPosition = false. Could this be the reason for your jittering?

If you pixelPerfectRender anything which is scaled to a non-integer scale, you will get jittering due to the fact that the top left will be at an exact pixel, but pixels within the graphic will be stretched to real numbers, still. This is the cause of jittering in https://github.com/HaxeFlixel/flixel/issues/1951, and could be another cause on your maps. One potential solution to this would be to ensure your scale is a multiple of (1 / pixel width, 1 / pixel height), thus making the scaled pixels also align to your grid. It may be a little more complicated than this seeing as the top left is being floored, and we presumably want to bottom right of the scaled graphic to be relatively the same distance from the center as the top left. It might justify some kind of pixelPerfectScale property to do this?

bendmorris:
Thanks, I understand the difference now.

>Could this be the reason for your jittering?

No, I'm consistently using the same settings on every object. Setting pixelPerfectRender and pixelPerfectPosition to true results in jittering when scrolling the camera. This has been reported in other issues as well.

> If you pixelPerfectRender anything which is scaled to a non-integer scale, you will get jittering due to the fact that the top left will be at an exact pixel, but pixels within the graphic will be stretched to real numbers, still.

This PR solves that problem specifically for FlxTilemaps, so every tile spans a whole number of pixels and starts where the last one ended. It sounds like something similar could be done for FlxSprites, adjusting the scale to make the width and height integers (but after placing them correctly at an integer starting coordinate using Camera.floorPoint, I don't notice any jitter, so I'm not sure if this is necessary.)

I think the current basic implementation of snapping to pixels is incorrect. It fails when the screen scale is not an integer. Using the default RatioScaleMode and supporting multiple resolutions this can very easily become the case. This PR fixes that with the FlxCamera.floor methods.

>It might justify some kind of pixelPerfectScale property to do this?

I don't like the idea of adding even more pixelPerfect flags - I would go the other way and simplify by merging the two existing ones. I see two use cases in your comment: you either want to render exclusively to integer positions (pixelPerfectPosition and pixelPerfectRender are true), or you want to use subpixel positions for more smoothness (pixelPerfectPosition and pixelPerfectRender are false). If using either of these introduces issues like tile gaps or jitter on camera movement, those are bugs that can and should be fixed (and here's a fix!) Unless I'm missing something, the case for setting position and render to different values: "i.e. your world will all be on a grid, but your camera can move smoothly between pixels on that grid on native" doesn't sound useful to me. If you're using subpixel positions, having sprites rendered an integer distance apart seems irrelevant - if my camera's scroll X is 0.5, my sprites at 1 and 101.1 will be rendered at 0.5 and 100.5 instead of 0.5 and 100.6, before scaling. What do you gain from that? Is anyone doing this?

starry-abyss:
Seems to fix tilemap tearing for me when both:
useScaleHack = false;
pixelPerfectPosition = true;

Also it improves the situation (if pixelPerfectPosition = true for sprites) for https://github.com/HaxeFlixel/flixel/issues/1861 :
- seems to fix the quad split into two triangles
- decreases camera jitter (previously the camera was rounded to sprite pixels, now it's rounded to screen pixels), but not eliminates it entirely

But decreasing the jitter this way breaks pixel-perfectness when two objects are near each other and zoom is > 1: 
default
Look at grass between bear's legs. While it's still can be considered pixel-perfect in screen pixels, it's not in sprite pixels (those of native game size before scaling).

Gama11:
Looking at this again, there didn't seem to be a consensus between @bendmorris and @JoeCreates whether this is a good change, regarding how much control the user should have over pixelPerfect flags? Plus there's merge conflicts by now.

PR #1954, opened 2016-09-18
FlxTimerManager: fix array manipulation in onComplete

Beeblerox:
I've made mistake in my previous PR (inserted while loop in wrong place).
This PR request also changes `completeAll()` method. Previous version of this method doesn't take into account the value of `active` property of timer. This could lead to infinite loop (maybe i'm wrong about this), because without check of `active` value it adds inactive timers to `timersToFinish` array, which won't finish (there is check in timer's update method).
And thus i have a question: Do we need to activate paused timers and force them to finish or should we complete only active timers. I could add some parameter to `completeAll()` which will be responsible for this behavior, like `includeInactive:Bool = false`

### DIFF
--- a/flixel/util/FlxTimer.hx
+++ b/flixel/util/FlxTimer.hx
@@ -175,14 +175,10 @@ class FlxTimer implements IFlxDestroyable
 	private function onLoopFinished():Void
 	{
 		if (onComplete != null)
-		{
 			onComplete(this);
-		}
 		
 		if (finished)
-		{
 			cancel();
-		}
 	}
 	
 	private inline function get_timeLeft():Float
@@ -312,9 +308,15 @@ class FlxTimerManager extends FlxBasic
 		for (timer in _timers)
 			if (timer.loops > 0)
 				timersToFinish.push(timer);
+
 		for (timer in timersToFinish)
+		{
 			while (!timer.finished)
+			{
 				timer.update(timer.timeLeft);
+				timer.onLoopFinished();
+			}
+		}
 	}
 
 	
--- a/flixel/util/FlxTimer.hx
+++ b/flixel/util/FlxTimer.hx
@@ -311,7 +311,7 @@ class FlxTimerManager extends FlxBasic
 		var timersToFinish:Array<FlxTimer> = [];
 		
 		for (timer in _timers)
-            if (timer.loops > 0 && timer.active) // or should we activate timer and force it to finish?
+            if (timer.loops > 0 && timer.active)
 				timersToFinish.push(timer);
 		
 		if (timersToFinish.length == 0)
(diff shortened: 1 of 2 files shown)


Gama11:
>Do we need to activate paused timers and force them to finish

My first instinct is "no", that would seem a little counter-intuitive. `completeAll()` should behave just like things would behave if you fast-forward time enough for all timers to be completed. If a timer is not active, it can't be completed.

Does `FlxTweenManager`'s `completeAll()` have the same issue?

Beeblerox:
@Gama11 ok, let's leave this behavior for timer manager.

About `FlxTweenManager`: it doesn't check value of `active` property, which is wrong in my opinion.
Plus it have different behavior: it just fast forwards time of all tweens and tween's complete callback is being called only once on the next frame (like said in docs) and tween is being removed from manager on the next frame as well (just like said in its documentation).

`FlxTimer`s are being removed the same frame we call `completeAll()` and timer's complete callback is being called the same frame `loopsLeft` times.

Beeblerox:
Should we add check for tween's `active` value in `FlxTweenManager`'s `completeAll()` method as well?

Beeblerox:
i don't understand why CI checks are failing

Gama11:
`No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.` - the unit tests are hanging again. Have you tried running them locally?

Beeblerox:
i'll try tomorrow

Beeblerox:
stupid question: how to run those unit tests?

Gama11:
There's a Readme. ;) https://github.com/HaxeFlixel/flixel/blob/dev/tests/unit/README.md

Beeblerox:
@Gama11 thanks!

Beeblerox:
All checks are passing now

flixel/util/FlxTimer.hx
```diff
@@ -281,12 +309,35 @@ class FlxTimerManager extends FlxBasic
 	public function completeAll():Void
```
Gama11: Hm... Does `completeAll()` really need to be this complex? For tweens it's only 3 lines, this seems a little much.

Gama11: Also, this doesn't really seem related to #679? Might be better to handle changes to `completeAll()` in a separate pull request.

Beeblerox: It's more complex because docs about these methods says different things:
- in flxtween it says that onComplete callback is called only once
- in flxtimer - onComplete callback will be called `loopsLeft` times.

With your change there is should be `onLoopFinished()` call, because timer's `update()` method had been changed and doesn't contain this logic anymore

PR #1955, opened 2016-09-19
FlxTweenManager: check for active in completeAll()

Beeblerox:
this patch adds check for tween's `active` value in `completeAll()` method od `FlxTweenManager` (just like in https://github.com/HaxeFlixel/flixel/pull/1954)
I think that we shouldn't finish tween if it's inactive.
Should we add this check in tween's `update()` method as well (like in `FlxTimer` class)?

### DIFF
--- a/flixel/tweens/FlxTween.hx
+++ b/flixel/tweens/FlxTween.hx
@@ -1141,7 +1141,7 @@ class FlxTweenManager extends FlxBasic
 	public function completeAll():Void
 	{
 		for (tween in _tweens)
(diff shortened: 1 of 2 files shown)


Gama11:
Why would this check be needed in `FlxTween#update()`? This is already checked in `FlxTweenManager#update()` right?

Gama11:
Thanks for making this a separate pull request. :)

I also added some unit tests. You should try writing some sometime too. They really do help catch bugs, like we saw in the failing test in #1954.

Beeblerox:
@Gama11 yeah, i know i have to learn this subject :)

PR #1959, opened 2016-09-27
FlxVector: fix normalize() returning (1,0) for (0,0)

Beeblerox:
But instead of proposed change i return this vector, not the new one


Gama11:
You forgot the unit test, added one for you. ;P

PR #1961, opened 2016-09-27
Convert FlxMouseEventManager from global singleton to instanced FlxBasic

bendmorris:
Relates to https://github.com/HaxeFlixel/flixel/issues/1087 (closed, but not resolved for FlxMouseEventManager.) Allows creating separate FlxMouseEventManagers which can be added to states, and which won't be updated when the state is inactive (e.g. when substates are active.) Can still be used as a global singleton, but not quite backwards compatible - requires using `globalManager` instance methods instead of static methods.

Requires an update to flixel-demos (https://github.com/HaxeFlixel/flixel-demos/pull/249) so Travis will fail.

### DIFF
--- a/flixel/input/mouse/FlxMouseEventManager.hx
+++ b/flixel/input/mouse/FlxMouseEventManager.hx
@@ -34,10 +34,7 @@ import flixel.util.FlxDestroyUtil;
  */
 class FlxMouseEventManager extends FlxBasic
 {
-	private static var _registeredObjects:Array<ObjectMouseData<FlxObject>> = [];
-	private static var _mouseOverObjects:Array<ObjectMouseData<FlxObject>> = [];
-
-	private static var _point:FlxPoint = FlxPoint.get();
+	public static var globalManager:FlxMouseEventManager;
 	
 	/**
 	 * As alternative you can call FlxMouseEventManager.init().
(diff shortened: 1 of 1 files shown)


Gama11:
I think this is too much of a breaking change (for a minor release at least).

Gama11:
Moved this to an issue for now: #1972

Geokureli:
Look like it's time to revisit this

Geokureli:
this isn't a class I use very much, but I have some ideas after applying these changes to a local branch (this branch is outdated, so I made the changes manually).

With this proposed changes existing game will need to change all refs such as `FlxMouseEventManager.add(obj, onDown, null, onOver, onOut)` to `FlxMouseEventManager.globalManager.add(obj, onDown, null, onOver, onOut)`. and while it works with surprsingly minimal changes to all HF libs, that's pretty verbose and cumbersome for devs. for cleaner code, what if we rename [ObjectMouseData](https://github.com/HaxeFlixel/flixel/blob/dev/flixel/input/mouse/FlxMouseEventManager.hx#L720) to `FlxMouseEvent` (or `FlxTypedMouseEvent<T>` and a typedef) and change the standard utilization to `new FlxMouseEvent(obj, onDown, null, onOver, onOut)` or `FlxMouseEvent.manage(obj, onDown, null, onOver, onOut)` and add a final `manager` arg, which defaults to the global manager.

This is based on how FlxTween and FlxTimer work

PR #1964, opened 2016-09-28
FlxMouseEventManager: check objects for camera overlap

bendmorris:
The problem: when using FlxCameras that don't cover the whole screen, mouse events can be generated by FlxMouseEventManager when the mouse isn't within the camera's visible area.

Example: this scrolling button menu has its own FlxCamera. The top button is highlighted by an onMouseOver callback, even though the overlapping part is above the camera's visible area:

screen shot 2016-09-28 at 8 32 09 am

With this change, FlxMouseEventManager first checks that the screen coordinates are within the camera's bounds before checking if the object overlaps the mouse position.

screen shot 2016-09-28 at 8 34 15 am

### DIFF
--- a/flixel/FlxCamera.hx
+++ b/flixel/FlxCamera.hx
@@ -1524,6 +1524,15 @@ class FlxCamera extends FlxBasic
 		updateFlashOffset();
 		setScale(scaleX, scaleY);
 	}
+
+	/**
+	 * Checks whether this camera contains a given point or rectangle, in
+	 * screen coordinates.
+	 */
+	public inline function containsPoint(point:FlxPoint, width:Float = 0, height:Float = 0):Bool
+	{
+		return (point.x + width > 0) && (point.x < this.width) && (point.y + height > 0) && (point.y < this.height);
+	}
 	
 	private function set_followLerp(Value:Float):Float
 	{
--- a/flixel/input/mouse/FlxMouseEventManager.hx
+++ b/flixel/input/mouse/FlxMouseEventManager.hx
@@ -443,7 +443,7 @@ class FlxMouseEventManager extends FlxBasic
 		for (camera in Register.object.cameras)
 		{
 			#if FLX_MOUSE
-			_point = FlxG.mouse.getScreenPosition(camera, _point);
+			_point = FlxG.mouse.getPositionInCameraView(camera, _point);
 			if (camera.containsPoint(_point))
 			{
 				_point = FlxG.mouse.getWorldPosition(camera, _point);
@@ -458,7 +458,7 @@ class FlxMouseEventManager extends FlxBasic
 			#if FLX_TOUCH
 			for (touch in FlxG.touches.list)
 			{
-				_point = touch.getScreenPosition(camera, _point);
+				_point = touch.getPositionInCameraView(camera, _point);
 				if (camera.containsPoint(_point))
 				{
 					_point = touch.getWorldPosition(camera, _point);
--- a/flixel/input/mouse/FlxMouseEventManager.hx
+++ b/flixel/input/mouse/FlxMouseEventManager.hx
@@ -103,7 +103,7 @@ class FlxMouseEventManager extends FlxBasic
 	}
 
 	/**
-	 * Removes all registerd objects from the registry.
+	 * Removes all registered objects from the registry.
 	 */
 	public static function removeAll():Void
 	{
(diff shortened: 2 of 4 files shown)


Gama11:
Would be nice to have a unit test for this.

bendmorris:
Testing this will be hacky since it's a singleton. I  would suggest adding tests once this is merged: https://github.com/HaxeFlixel/flixel/pull/1961 which makes FlxMouseEventManager modular and easier to write self-contained tests for.

Gama11:
I don't think that really impacts testing too much. Besides, I don't have any plans to merge that anytime soon, it just breaks too much (essentially all usages of `FlxMouseEventManager`).

bendmorris:
It's a breaking change, so it wouldn't go in a minor release, I get that. (To be clear you just have to add `globalManager` to any use of FlxMouseEventManager to get exactly the same behavior, it's not a difficult migration.) There are other breaking changes such as https://github.com/HaxeFlixel/flixel/pull/1940 in the pipeline. What's the plan for landing them?

If it's not clear that it will ever be merged, I'm going to continue using my modular version, so I'm probably not going to write new tests for the singleton version, and this fix will become "take it or leave it."

Gama11:
Not going in a minor release means that it would have to wait until 5.0.0 - there are no concrete plans for that yet.

So far we've avoided breaking changes since 4.0.0, except for bugfixes where it was inevitable (what's broken there is just relying on broken behavior) and #1934. That PR only breaks usage of a field that is hardly used by anyone, and it didn't even break any of the demos - whereas this PR would break all usages of `FlxMouseEventManager`.

I'll have to check in more detail what exactly #1940 breaks - so far it's scheduled for a minor release (4.3.0). Either way it's a bit of a special case.

flixel/FlxCamera.hx
```diff
@@ -1524,6 +1524,15 @@ class FlxCamera extends FlxBasic
...
 		setScale(scaleX, scaleY);
 	}
+
+	/**
+	 * Checks whether this camera contains a given point or rectangle, in
+	 * screen coordinates.
+	 */
+	public inline function containsPoint(point:FlxPoint, width:Float = 0, height:Float = 0):Bool
+	{
+		return (point.x + width >= 0) && (point.x < this.width) && (point.y + height >= 0) && (point.y < this.height);
```
Gama11: It seems like you also changed this logic to be inclusive? (`>= 0` instead of just `> 0`). This leads to unexpected behavior in `FlxObject#isOnScreen()`:

```haxe
var object = new FlxObject(0, 0, 1, 1);
add(object);
trace(object.isOnScreen()); // true
object.x = -1;
trace(object.isOnScreen()); // should be false, but true with >= 0
```

bendmorris: You're right, I'll take that out.

Gama11:
I can still reproduce the issue you describe if I'm on your branch... Am I missing something?

````haxe
package;

import flixel.FlxG;
import flixel.FlxSprite;
import flixel.FlxState;
import flixel.util.FlxColor;

class PlayState extends FlxState{
        var sprite = new FlxSprite(0, 0);
        sprite.makeGraphic(20, 20);
        add(sprite);
        sprite.screenCenter();
        
        FlxG.camera.zoom = 4;
        FlxG.camera.y = FlxG.height / 2;
        FlxG.camera.scroll.y = 60;

        FlxMouseEventManager.add(sprite, null, null,
            function(_) sprite.color = FlxColor.RED,
            function(_) sprite.color = FlxColor.WHITE);
    }
}
```

![](http://i.imgur.com/0Ee9G39.gif)

bendmorris:
The difference is zoom - if you remove zoom from your demo and change FlxG.camera.scroll.y to FlxG.height/2, it works. It looks like getScreenPosition doesn't incorporate camera scale. I'm not sure what the proper fix for that would be.

flixel/input/mouse/FlxMouseEventManager.hx
```diff
@@ -443,22 +443,30 @@ class FlxMouseEventManager extends FlxBasic
...
 			}
 			#end
 			
 			#if FLX_TOUCH
 			for (touch in FlxG.touches.list)
 			{
-				_point = touch.getWorldPosition(camera, _point);
-				
-				if (checkOverlapWithPoint(Register, _point, camera))
+				_point = touch.getScreenPosition(camera, _point);
```
Beeblerox: so this line should be changes as well to 
`_point = touch.getPositionInCameraView(camera, _point);`

flixel/input/mouse/FlxMouseEventManager.hx
```diff
@@ -443,22 +443,30 @@ class FlxMouseEventManager extends FlxBasic
 		for (camera in Register.object.cameras)
 		{
 			#if FLX_MOUSE
-			_point = FlxG.mouse.getWorldPosition(camera, _point);
-			
-			if (checkOverlapWithPoint(Register, _point, camera))
+			_point = FlxG.mouse.getScreenPosition(camera, _point);
```
Beeblerox: i've found that if we change this _point caclulation code `_point = FlxG.mouse.getScreenPosition(camera, _point);` to something like this:
``` 
public function getPositionInCameraView(?Camera:FlxCamera, ?point:FlxPoint):FlxPoint
	{
		if (Camera == null)
			Camera = FlxG.camera;
		
		if (point == null)
			point = FlxPoint.get();
		
		point.x = (_globalScreenX - Camera.x) / Camera.zoom;
		point.y = (_globalScreenY - Camera.y) / Camera.zoom;
		
		return point;
	}
```
then checks working properly.

This new method calculates position of a point relative to camera view rectangle

bendmorris:
Thanks @Beeblerox. You should have commit access to my fork, so feel free to just push your change there, otherwise I'll add them when I have a chance.

flixel/input/mouse/FlxMouseEventManager.hx
MSGhero: You named the function wrong! The function in `FlxPointer` is not the same name as in `FlxMEM`.

Beeblerox: @MSGhero already fixed this, sorry

PR #1966, opened 2016-09-30
FlxFramesCollection: check and trim frame sizes, closes #1937

### LINKED ISSUE #1937: Playing animation with larger than current frame still displays smaller frame size on flash, html5
Exactly the same code and versions as in #1936 but you can see in the image that the red block is 10x10 instead of 20x20 as prescribed by its atlas definition:

hf-debugdraw

Only affects flash though.  The output should look like on cpp:

hf-animsizebug

(With the exception of the debugdraw output still being wrong on cpp, but that's #1936's bug.)

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/37683690-playing-animation-with-larger-than-current-frame-still-displays-smaller-frame-size-on-flash-html5?utm_campaign=plugin&utm_content=tracker%2F254189&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F254189&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Beeblerox:
i'm not sure whether we should add this check

### DIFF
--- a/flixel/graphics/frames/FlxFramesCollection.hx
+++ b/flixel/graphics/frames/FlxFramesCollection.hx
@@ -204,9 +204,7 @@ class FlxFramesCollection implements IFlxDestroyable
 		frame.set(x, y, r - x, b - y);
 		
 		if (frame.width <= 0 || frame.height <= 0)
-		{
-			FlxG.log.warn("The frame " + name + " have incorrect data and results in image with the size of (0; 0)");
-		}
+			FlxG.log.warn("The frame " + name + " has incorrect data and results in an image with the size of (0, 0)");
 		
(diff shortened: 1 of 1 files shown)


Gama11:
>i'm not sure whether we should add this check

What's the pros / cons?

Beeblerox:
con: it complicates code.
pro: we are making sure that data is correct and modify it (trim frame rectangle) if it's not

Gama11:
I think it's nice to have consistent cross-platform behavior, even if the input values are technically incorrect.

PR #1967, opened 2016-09-30
FlxAnimation: add support for changing frames, closes #1888

### LINKED ISSUE #1888: No way to modify animation frames (without instantiating new animation)

Beeblerox:


Gama11:
This makes them public, but they're still read-only, so there's no way to set them?

Beeblerox:
But now you can modify this array.

Beeblerox:
And original issue #1888 was about modifying frames array, not setting it

Gama11:
Right.. But why make it `(default, null)` at all then? You can already modify it, might as well allow assignment too. Often times assigning an array literal directly is more convenient.

Beeblerox:
ok, let it be writable too.

PR #1968, opened 2016-10-01
BitmapData caching improvements, closes #1922

### LINKED ISSUE #1922: Static text object graphic missing after state switch
- **Flixel version:** dev ahead of 4.1.1
- **OpenFL version:** 3.6.1
- **Lime version:** 2.9.1
- **Affected targets:** Neko/CPP

https://gist.github.com/DleanJeans/5f71159386f8b0ddfc862bea92b48407

**Instructions**:
First state (MenuState) creates a FlxText
Second state (PlayState) add the FlxText and remove it before switching back to MenuState (to prevent FlxText destruction)

SPACE to switch between states
In PlayState.hx:
Press 1 to switch alignment
Press 2 to switch fieldWidth
Press 3 to switch font
Press ENTER to FlxText drawFrame() (does nothing tho)

**Observed behavior:**
On Neko, (probably same on cpp)
At first you can press switch state (SPACE) as many time you want and the text still shows.
But if you change either fieldWidth (Key 2) or font (Key 3), the text will disappear if you switch back to PlayState and only shows up again if you change them again.
On Flash, (probably same on html5)
work as expected like below.
**Expected behavior:** No matter platform or settings, the text should appear on switch state

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/37444558-static-text-object-graphic-missing-after-state-switch?utm_campaign=plugin&utm_content=tracker%2F254189&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F254189&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Beeblerox:


flixel/system/frontEnds/BitmapFrontEnd.hx
```diff
@@ -319,7 +319,25 @@ class BitmapFrontEnd
...
+			var obj = _cache.get(key);
+			if (obj != null && !obj.persist && obj.useCount <= 0)
+			{
+				removeByKey(key);
+			}
+		}
+	}
+	
+	/**
+	 * Completely resets bitmap cache, which means destroying ALL of the cached FlxGraphic objects.
```
Gama11: It still checks for `persist` though, so it's not really "all"?

flixel/system/frontEnds/BitmapFrontEnd.hx
```diff
@@ -319,7 +319,25 @@ class BitmapFrontEnd
 	{
 		if (_cache == null)
 			_cache = new Map();
-
+		
+		for (key in _cache.keys())
+		{
+			var obj = _cache.get(key);
+			if (obj != null && !obj.persist && obj.useCount <= 0)
```
Gama11: Considering this now checks for `useCount`, the docs probably need to be updated.

Gama11:
I don't know if this is really the appropriate fix - it doesn't make much sense to me to check the `useCount` during state switches. If there is no state, a graphic can't really be used.

It seems to me the problem with #1922 is that the `graphic.persist` is "lost", because a new `FlxGraphic` object is created when the text dimensions change. Perhaps that (and maybe other flags that can be set by the user, not sure if there are any) should be be carried over?

Beeblerox:
@Gama11 you're right about loss if `persist` value: changing text font could lead to generation of new graphic object which won't be persistant.
But i'm for this fix, since it fixes much more cases than this: there were a lot similar issues when users complained that flixel disposes all graphics which hadn't been marked as persistant.
adding this check and calling bitmap cache clearing **after** destroying previous state solves this issue: if you want to "save" (or reuse) sprite for the next state then you remove it from state before destroying it, and you expect that if you save the sprite then you want to save its graphic as well.
if you don't want to save sprite, but want to save graphic then you set `graphic.persist = true`.

starry-abyss:
It looks like it was a breaking change for me. That's what I get on all game levels except first:
[wrong](http://i.imgur.com/yOo1bxE.gif)

Expected behaviour before this merge:
[correct](http://i.imgur.com/aGBEeOF.gif)

Gama11:
Are you sure it was this change?

Are you up to date? I fixed something else related to caching here: 248a1ec40ffd234

starry-abyss:
I found it with git bisect on flixel-dev. Going a commit before fixes the issue. Not sure what you mean by "up to date"

Gama11:
I see.

By "up to date" I meant "on the latest commit".

Any chance you could isolate the issue to see when it happens exactly?

starry-abyss:
I'll try. Seems not very big change

starry-abyss:
Looks like `obj.useCount <= 0` part is the one

https://github.com/HaxeFlixel/flixel/pull/1968/commits/f69e3582c7cca6723f344a00111ab5a8eae605ec#diff-be3583aabdebf9b6295e6178fee29fc6R326

starry-abyss:
Every game level is the same PlayState reloaded with different static field values (just numbers and stuff, no images there), if it helps

Gama11:
Do you have some example code that breaks with the change?

starry-abyss:
No, not yet, my project is sort of bloated

Beeblerox:
@starry-abyss maybe you forgot to destroy some helper sprite. Btw do you use atlases generated at runtime?

starry-abyss:
@Beeblerox Yes, I do. And that's the clue, without atlas all is fine.
What's so special about destroying (helper) sprites?

Beeblerox:
if you don't destroy helper sprite then its graphic will have useCount other than 0, which means that it won't be destroyed on state change/reset.
Maybe the problem isn't with them, maybe it's with FlxAtlas.
Btw what happens if you set `FlxG.bitmap.get(atlasName).persist = false;`?
where `atlasName` is the name of your atlas/

starry-abyss:
```
atlas = new FlxAtlas("atlasSummer", true);
atlas.maxWidth = 2048;
atlas.maxHeight = 2048;
trace(FlxG.bitmap.get("atlasSummer"));	
FlxG.bitmap.get("atlasSummer").persist = false;
```

I get a crash. `FlxG.bitmap.get(atlasName)` is null

Beeblerox:
try call this after you get first frame collection from it.

starry-abyss:
No crash, but issue still there.

starry-abyss:
Apparently, atlas picture is a bit different in each level.
`atlas.destroy();` before state change didn't help either.

Beeblerox:
try to call `FlxG.bitmap.remove(atlas.name);` before calling `atlas.destroy();`

starry-abyss:
`FlxG.bitmap.removeByKey(atlas.name);` helped!

Beeblerox:
so the issue is with FlxAtlas then. i think that two lines should be changed there:
```
FlxG.bitmap.removeIfNoUse(FlxG.bitmap.get(name));
_bitmapData = FlxDestroyUtil.dispose(_bitmapData);
```
should be 
```
FlxG.bitmap.removeByKey(name);
_bitmapData = null;
```

starry-abyss:
For some reason, it doesn't solve the issue:
```
	public function destroy():Void
	{
		//FlxG.bitmap.removeIfNoUse(FlxG.bitmap.get(name));
		FlxG.bitmap.removeByKey(name);
		_tempStorage = null;
		deleteSubtree(root);
		root = null;
		//_bitmapData = FlxDestroyUtil.dispose(_bitmapData);
		_bitmapData = null;
		nodes = null;
	}
```

Beeblerox:
ok, i'll try something else tonight

Beeblerox:
here is what i'm thinking about it: 
we could add signal for cache clearing and add listener for this signal in FlxAtlas, so when cache is being cleared we check if this atlas is in use and if not then we destroy this atlas. Something like:
```
private function onClear():Void
{
	if (_graphic == null || _graphic.useCount <= 0 || !_graphic.persist)
	{
		destroy();
	}
}
```

Beeblerox:
but this idea looks to me as ugly hack. I think it would be better to stick to the option when you need to manually destroy atlas in state destroy method (and change atlas destroy method as proposed earlier).

Beeblerox:
or just replace FlxAtlas with some other solution

starry-abyss:
Maybe some additional check in BitmapFrontEnd, like "is it atlas"?

Beeblerox:
@starry-abyss i've made some changes to FlxAtlas. Could you test it with your code? Is the issue gone?

Beeblerox:
and i'm planning to rewrite FlxAtlas, because it's a total mess. But unfortunately it will be breaking change

starry-abyss:
Yeah, at first glance from 06ac60e4 on - my issue is fixed!
Please don't touch anything that works! :D yet

Gama11:
Why do you want to rewrite `FlxAtlas`? Didn't you already do that quite recently? :D

PR #1971, opened 2016-10-02
FlxSubState: fix close() if used in two different states

JoeCreates:
If you have persistent substates which you open from multiple parent states, you want the parent state of the substate to change even if the state has already been created, otherwise _parentState will be incorrect when you open the substate from a second location.

### DIFF
--- a/flixel/FlxState.hx
+++ b/flixel/FlxState.hx
@@ -112,10 +112,11 @@ class FlxState extends FlxGroup
 				FlxG.inputs.onStateSwitch();
 			}
 			
+			subState._parentState = this;
+			
 			if (!subState._created)
 			{
 				subState._created = true;
-				subState._parentState = this;
 				subState.create();
 			}
 		}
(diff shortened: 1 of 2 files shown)


Gama11:
Unit test? :)

JoeCreates:
I think it would be something like this:

```
@Test
	@:access(flixel.FlxSubState)
	function testOpenPersistentSubStateFromNewParent()
	{
		var state1 = new FlxState();
		var state2 = new FlxState();
		state1.destroySubStates = false;
		FlxG.state.switchTo(state1);
		step();
		FlxG.state.openSubState(subState1);
		step();
		
		Assert.areEqual(subState1._parentState, state1);
		
		subState1.close();
		step();
		FlxG.state.switchTo(state2);
		step();
		FlxG.state.openSubState(subState1);
		step();
		
		Assert.areEqual(subState1._parentState, state2);
	}
	}
```
Could you possibly check this over and add it in if it's correct? I can't run hamcrest (I think because I need to change haxe version, which I can't do at the moment).

Gama11:
I think you meant to call `FlxG.switchState()` instead of `FlxG.state.switchTo()`? The latter doesn't really do anything by default / isn't supposed to be called manually.

With that fixed, the test works.

It's a bit unusual to check the value of private fields though.. I usually check whatever effect that would have on the public API, in this case `close()` not working correctly I guess.

Gama11:
Thanks!

PR #1982, opened 2016-10-25
FlxBaseTilemap: new optimized pathing method

x2f:
Add new parameter to optional argument to findPath() to select a pathing
method. The default method remains unchanged. New parameter `PathingMethod` can be set to
`BASIC_SCALED_HORIZON(scale:Float, minHorizonSize:Int)` to constrain the pathing
algorithm into a zone `scale` times larger than the start tile-end tile
rectangle, with a minimum width and length of `minHorizonSize`.

### DIFF
--- a/flixel/tile/FlxBaseTilemap.hx
+++ b/flixel/tile/FlxBaseTilemap.hx
@@ -720,7 +720,7 @@ class FlxBaseTilemap<Tile:FlxObject> extends FlxObject
 	 * @param	DiagonalPolicy	How to treat diagonal movement. (Default is WIDE, count +1 tile for diagonal movement)
 	 * @return	An Array of FlxPoints, containing all waypoints from the start to the end.  If no path could be found, then a null reference is returned.
 	 */
-	public function findPath(Start:FlxPoint, End:FlxPoint, Simplify:Bool = true, RaySimplify:Bool = false, DiagonalPolicy:FlxTilemapDiagonalPolicy = WIDE, PathingMethod:FlxTilemapPathingMethod = null):Array<FlxPoint>
+	public function findPath(Start:FlxPoint, End:FlxPoint, Simplify:Bool = true, RaySimplify:Bool = false, DiagonalPolicy:FlxTilemapDiagonalPolicy = WIDE, ?PathingMethod:FlxTilemapPathingMethod):Array<FlxPoint>
 	{
 		if (PathingMethod == null) PathingMethod = FlxTilemapPathingMethod.BASIC;
 
@@ -790,7 +790,8 @@ class FlxBaseTilemap<Tile:FlxObject> extends FlxObject
 		walkPath(distances, zone.indexEnd, pathIndexes, zone.sizeX, zone.sizeY);
 
 		// Translate back points from index within the zone array into original map coordinates
-		for (index in pathIndexes) {
+		for (index in pathIndexes)
+		{
 			var x = index % zone.sizeX;
 			var y = Math.floor(index / zone.sizeX);
 			var indexInMap = x + zone.x0 + (y + zone.y0) * widthInTiles;
@@ -845,13 +846,13 @@ class FlxBaseTilemap<Tile:FlxObject> extends FlxObject
 		// Create a distance-based representation of the tilemap.
 		// All walls are flagged as -2, all open areas as -1.
 
-		var distances= new Array<Int>(/*zone.sizeX * zone.sizeY*/);
+		var distances = new Array<Int>(/*zone.sizeX * zone.sizeY*/);
 		FlxArrayUtil.setLength(distances, zone.sizeX * zone.sizeY);
 
 		var i:Int = 0;
-		for (yy in zone.y0...zone.y0+zone.sizeY)
+		for (yy in zone.y0...zone.y0 + zone.sizeY)
 		{
-			for (xx in zone.x0...zone.x0+zone.sizeX)
+			for (xx in zone.x0...zone.x0 + zone.sizeX)
 			{
 				if (_tileObjects[_data[xx + yy * widthInTiles]].allowCollisions != FlxObject.NONE)
 				{
@@ -1357,7 +1358,8 @@ abstract FlxTilemapDiagonalPolicy(Int)
 	var WIDE = 2;
 }
 
-enum FlxTilemapPathingMethod {
+enum FlxTilemapPathingMethod
+{
 	/**
 	 * Basic method is calculating distances using an array covering the whole map (was original solution)
 	*/
@@ -1370,7 +1372,8 @@ enum FlxTilemapPathingMethod {
 	BASIC_SCALED_HORIZON(scale:Float, minHorizonSize:Int);
 }
 
-typedef FlxTilemapZone = {
+typedef FlxTilemapZone =
+{
 	var x0:Int;
 	var y0:Int;
 	var sizeX:Int;
--- a/flixel/tile/FlxBaseTilemap.hx
+++ b/flixel/tile/FlxBaseTilemap.hx
@@ -712,15 +712,18 @@ class FlxBaseTilemap<Tile:FlxObject> extends FlxObject
 	 * Find a path through the tilemap.  Any tile with any collision flags set is treated as impassable.
 	 * If no path is discovered then a null reference is returned.
 	 * 
-	 * @param	Start		The start point in world coordinates.
-	 * @param	End			The end point in world coordinates.
-	 * @param	Simplify	Whether to run a basic simplification algorithm over the path data, removing extra points that are on the same line.  Default value is true.
-	 * @param	RaySimplify	Whether to run an extra raycasting simplification algorithm over the remaining path data.  This can result in some close corners being cut, and should be used with care if at all (yet).  Default value is false.
+	 * @param	Start			The start point in world coordinates.
+	 * @param	End				The end point in world coordinates.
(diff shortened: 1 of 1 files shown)


x2f:
Hi, a bit of context on this pull request: the current pathing method is too long when the map is large (e.g. 512x512 tiles). Especially long is the initialization of the distances array in computePathDistance() in FlxBaseTilemap.hx. This request adds a new pathing method to limit the area where the pathing algo.

x2f:
Ok, I did not realize that computePathDistance() was used. I'll get back to this later, will change computePathDistance() to computePathDistanceZone() and define again computePathDistance() using its original interface.

x2f:
Test (tested on Windows): 
(1) haxelib run flixel create Pathfinding
(2) change moveToGoal() as described below
(3) haxelib run lime test neko
(4) Press 'move to goal' multiple times (so that the starting point changes each time)

...The result is the same and the time of the second method goes down as the start and end point gets closer

Code:

	private function moveToGoal():Void
	{
		// Find path to goal from unit to goal
		var now1 = haxe.Timer.stamp();
		var pathPoints:Array<FlxPoint> = _map.findPath(
			FlxPoint.get(_unit.x + _unit.width / 2, _unit.y + _unit.height / 2),
			FlxPoint.get(_goal.x + _goal.width / 2, _goal.y + _goal.height / 2));
		var now2 = haxe.Timer.stamp();
		var pathPoints2:Array<FlxPoint> = _map.findPath(
			FlxPoint.get(_unit.x + _unit.width / 2, _unit.y + _unit.height / 2),
			FlxPoint.get(_goal.x + _goal.width / 2, _goal.y + _goal.height / 2),
			FlxTilemapPathingMethod.BASIC_SCALED_HORIZON(2.0, 5));
		var now3 = haxe.Timer.stamp();

		var val = pathPoints.toString();
		var val2 = pathPoints2.toString();

		if (val != val2)
		{
			trace('ERROR: $val != $val2');
		}
		trace('Time: BASIC=${now2-now1} BASIC_SCALED_HORIZON=${now3-now2}');

		// Tell unit to follow path
		if (pathPoints != null) 
		{
			_unit.path.start(pathPoints);
			_action = ACTION_GO;
			_instructions.text = INSTRUCTION_1;
		}
		else 
		{
			_instructions.text = INSTRUCTION_2;
		}
	}

Output:

	$ haxelib run lime test neko
	export/windows/neko/haxe/ApplicationMain.hx:205: characters 57-75 : Warning : Use programPath instead
	export/windows/neko/haxe/ApplicationMain.hx:80: characters 47-65 : Warning : Use programPath instead
	PlayState.hx:207: Time: BASIC=0.000660722728027974 BASIC_SCALED_HORIZON=0.000672590077997359
	PlayState.hx:207: Time: BASIC=0.000594329716036679 BASIC_SCALED_HORIZON=0.000653345726695509
	PlayState.hx:207: Time: BASIC=0.000549746968854237 BASIC_SCALED_HORIZON=0.000513503440568996
	PlayState.hx:207: Time: BASIC=0.000520880441901461 BASIC_SCALED_HORIZON=0.000484316174428123
	PlayState.hx:207: Time: BASIC=0.00049137243657249 BASIC_SCALED_HORIZON=0.000473411042023919
	PlayState.hx:207: Time: BASIC=0.000506447178425518 BASIC_SCALED_HORIZON=0.000390660331426318
	PlayState.hx:207: Time: BASIC=0.000511579005439522 BASIC_SCALED_HORIZON=0.000292193400598606
	PlayState.hx:207: Time: BASIC=0.000418564640813912 BASIC_SCALED_HORIZON=0.000313682926218917
	PlayState.hx:207: Time: BASIC=0.000363397500415807 BASIC_SCALED_HORIZON=0.000204631602175098
	PlayState.hx:207: Time: BASIC=0.000352171628822617 BASIC_SCALED_HORIZON=0.000164218464442811
	PlayState.hx:207: Time: BASIC=0.000311437751900456 BASIC_SCALED_HORIZON=0.000190519077886364
	PlayState.hx:207: Time: BASIC=0.000266534265529472 BASIC_SCALED_HORIZON=0.000127012718593278
	PlayState.hx:207: Time: BASIC=0.000223234475100753 BASIC_SCALED_HORIZON=8.3071449786587e-005
	PlayState.hx:207: Time: BASIC=0.000288665269527755 BASIC_SCALED_HORIZON=4.97145741960026e-005
	PlayState.hx:207: Time: BASIC=0.000179293206295839 BASIC_SCALED_HORIZON=2.88665269518873e-005

x2f:
Hi all, if any member of the haxeflixel team would have some time to give some feedback it would be appreciated! I believe this optimization is necessary to scale up FlxTilemap, it would be great to get your take on this...Thanks!

flixel/tile/FlxBaseTilemap.hx
```diff
@@ -730,19 +733,70 @@ class FlxBaseTilemap<Tile:FlxObject> extends FlxObject
...
 			return null;
 		}
-		
+
+		var zone:FlxTilemapZone = switch (PathingMethod)
+		{
+			// Basic pathing method where distances from starting point is calculated for all tiles
+			// until reaching the end point, and then path is found by backtracking from end point to starting point
+			case FlxTilemapPathingMethod.BASIC:
+				{x0:0, y0:0, sizeX:widthInTiles, sizeY:heightInTiles, indexStart:startIndex, indexEnd:endIndex};
```
gamedevsam: This probably leads to Dynamic objects in native targets. We should have classes that store data for FlxTilemapPathing types. That way we can create constructors that make our code look nice and pool them and reduce GC load.

flixel/tile/FlxBaseTilemap.hx
```diff
@@ -730,19 +733,70 @@ class FlxBaseTilemap<Tile:FlxObject> extends FlxObject
...
+				var y0 = Math.floor(highY - scale * 0.5 * deltaY);
+				if (y0 < 0) y0 = 0;
+				var y1 = Math.ceil(lowY + scale * 0.5 * deltaY);
+				if (y1 > heightInTiles) y1 = heightInTiles;
+
+				var zoneSizeX = x1 - x0;
+				var zoneSizeY = y1 - y0;
+				var startIndexInZone = (startX - x0) + (startY - y0) * zoneSizeX;
+				var endIndexInZone = (endX - x0) + (endY - y0) * zoneSizeX;
+				{x0:x0, y0:y0, sizeX:zoneSizeX, sizeY:zoneSizeY, indexStart:startIndexInZone, indexEnd:endIndexInZone};
```
gamedevsam: Same problem here, this code likely generates Dynamic objects. We should create a classes.

flixel/tile/FlxBaseTilemap.hx
```diff
@@ -1296,4 +1371,28 @@ abstract FlxTilemapDiagonalPolicy(Int)
...
+	BASIC;
+	/**
+	 * Optimization of basic method where distances are only calculated in a zone around start/end tiles
+	 * - scale > 1.0 determines how this rectangular zone is grown around the rectangle formed by start-end points
+	 * - minHorizonSize is used if any the scaled zone dimension is too small
+	*/
+	BASIC_SCALED_HORIZON(scale:Float, minHorizonSize:Int);
+}
+
+typedef FlxTilemapZone =
```
gamedevsam: Typedefs can generate slow Dynamic code by the C++ compiler, so we tend to avoid them in HaxeFlixel as much as possible. It would be better if you created a pooled class, look at ```FlxRect``` or ```FlxPoint``` for implementation examples.

Gama11: Do we really want to introduce more pooling? Might be overkill here, these objects should be much rarer compared to `FlxRect` or `FlxPoint`.

flixel/tile/FlxBaseTilemap.hx
Gama11: >That way we can create constructors that make our code look nice

?

x2f:
Thanks gamedevsam and Gama11 for the reviews. So a new dynamic object (on native target) will be created for each call of computePathDistance() and findPath(). Alternative solutions I can think of include (1) adding a member to each FlxBaseTilemap instance, (2) inline parameters, so instead of a zone object we have 6 parameters zoneX0, zoneY0, etc. (3) use a pooled object as you suggest and (4) we keep the current code with temporary object.

I don't know enough to judge the real impact of (4). If the impact of (4) is an issue then I would suggest (2). What's your opinion on this?

You guys probably noticed but just to make sure: there is also an additional new in findPath() (var pathIndexes:Array<Int> = new Array<Int>();)

flixel/tile/FlxBaseTilemap.hx
x2f: Alternative solutions I can think of include (1) adding a member to each FlxBaseTilemap instance, (2) inline parameters, so instead of a zone object we have 6 parameters zoneX0, zoneY0, etc. (3) use a pooled object as you suggest or (4) we keep the current code using FlxTilemapZone.

A new dynamic object (on native target) will be created for each call of computePathDistance() and findPath()... If the impact of is an issue then I would suggest (2). What's your opinion on this?

Btw, you guys probably noticed but just to make sure: there is also an additional new in findPath() (var pathIndexes:Array = new Array();) I hope the runtime impact is acceptable.

Gama11:
Looking at this again, I'm not entirely convinced that there's enough demand for this feature to justify the added complexity, so I'll go ahead and close for now. If there's a lot of interest for it we can reconsider.

Issue #292, opened 2013-04-17
Fix object visibility tests for cases when camera have zoom less than 1.0


Beeblerox:
You can see this issue in FlxCamera Demo ( http://haxeflixel.com/demos/flxcamera )
##

Beeblerox:
i think i'll need to add new property to the camera, something like minZoom

GimmickyApps:
@Beeblerox would minZoom be a non-settable value, preventing the bug by not allowing a zoom level at which it occurs, or would it be settable and attempt a workaround if you it set it to a small enough number?

MrClogsworthYT:
Flash is ded now lol

PR #1983, opened 2016-10-25
Fix camera occlusion and background for zoom != 1

starry-abyss:
Here's a demo to compare before and after the changes:

https://github.com/starry-abyss/hf-multicamera

Native targets for now, on Flash target nothing changed, need assistance on it :-)

### DIFF
--- a/flixel/FlxCamera.hx
+++ b/flixel/FlxCamera.hx
@@ -206,6 +206,12 @@ class FlxCamera extends FlxBasic
 	 */
 	public var zoom(default, set):Float;
 	/**
+	 * Difference between native size of camera and zoomed size, divided in half
+	 * Needed to do occlusion of objects when zoom != 1
+	 */
+	public var zoomOffsetX(get, never):Float;
+	public var zoomOffsetY(get, never):Float;
+	/**
 	 * The alpha value of this camera display (a number between `0.0` and `1.0`).
 	 */
 	public var alpha(default, set):Float = 1;
@@ -1352,7 +1358,7 @@ class FlxCamera extends FlxBasic
 			targetGraphics.beginFill(Color, FxAlpha);
 			// i'm drawing rect with these parameters to avoid light lines at the top and left of the camera,
 			// which could appear while cameras fading
-			targetGraphics.drawRect(-1, -1, width + 2, height + 2);
+			targetGraphics.drawRect(-1 - zoomOffsetX, -1 - zoomOffsetY, width + zoomOffsetX * 2 + 2, height + zoomOffsetY * 2 + 2);
 			targetGraphics.endFill();
 		}
 	}
@@ -1651,6 +1657,15 @@ class FlxCamera extends FlxBasic
 			flashSprite.visible = visible;
 		}
 		return this.visible = visible;
+	}
+	
+	private function get_zoomOffsetX():Float
+	{
+		return width * (1 - zoom) * 0.5 / zoom;
+	}
+	private function get_zoomOffsetY():Float
+	{
+		return height * (1 - zoom) * 0.5 / zoom;
 	}
 }
 
--- a/flixel/FlxCamera.hx
+++ b/flixel/FlxCamera.hx
@@ -1517,6 +1517,8 @@ class FlxCamera extends FlxBasic
 		updateFlashSpritePosition();
 		updateScrollRect();
 		updateInternalSpritePositions();
+		
+		FlxG.cameras.cameraResized.dispatch(this);
 	}
 	
 	/**
--- a/flixel/FlxSprite.hx
+++ b/flixel/FlxSprite.hx
@@ -983,13 +983,15 @@ class FlxSprite extends FlxObject
 		
 		var minX:Float = x - offset.x - Camera.scroll.x * scrollFactor.x;
 		var minY:Float = y - offset.y - Camera.scroll.y * scrollFactor.y;
+		var zoomOffsetX:Float = Camera.zoomOffsetX;
+		var zoomOffsetY:Float = Camera.zoomOffsetY;
 		
 		if ((angle == 0 || bakedRotationAngle > 0) && (scale.x == 1) && (scale.y == 1))
 		{
-			if (minX > Camera.width || minX + frameWidth < 0)
+			if (minX > Camera.width + zoomOffsetX || minX + frameWidth < -zoomOffsetX)
 				return false;
 			
-			if (minY > Camera.height || minY + frameHeight < 0)
+			if (minY > Camera.height + zoomOffsetY || minY + frameHeight < -zoomOffsetY)
 				return false;
 		}
 		else
@@ -1022,14 +1024,14 @@ class FlxSprite extends FlxObject
 			var maxX:Float = minX + radius;
 			minX -= radius;
 			
-			if (maxX < 0 || minX > Camera.width)
+			if (maxX < -zoomOffsetX || minX > Camera.width + zoomOffsetX)
 				return false;
 			
 			minY += oy;
 			var maxY:Float = minY + radius;
 			minY -= radius;
 			
-			if (maxY < 0 || minY > Camera.height)
+			if (maxY < -zoomOffsetY || minY > Camera.height + zoomOffsetY)
 				return false;
 		}
 		
--- a/flixel/FlxObject.hx
+++ b/flixel/FlxObject.hx
@@ -947,7 +947,10 @@ class FlxObject extends FlxBasic
 			Camera = FlxG.camera;
 		}
 		getScreenPosition(_point, Camera);
-		return (_point.x + width > 0) && (_point.x < Camera.width) && (_point.y + height > 0) && (_point.y < Camera.height);
+		
+		var zoomOffsetX:Float = Camera.zoomOffsetX;
+		var zoomOffsetY:Float = Camera.zoomOffsetY;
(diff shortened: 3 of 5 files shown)


Gama11:
I guess this is supposed to deal with #292?

starry-abyss:
Yes! There are some calculations that assume zoom == 1. If zoom > 1, that isn't noticeable mostly, but when zoom < 1, it is.
When camera is zoomed, all coordinates are still in old pixels of zoom==1. My additions adjust the camera window size, so it's larger for zoom<1, and smaller for zoom>1.
Flash target is tricky, because it has some buffers like _fill, _flashRect, _flashPoint, buffer, screen, and I didn't come to working solution yet on Flash.

Gama11:
@Beeblerox What do you think about the changes? If I remember correctly, you worked on a fix for #292 a while back, but you weren't happy with it.

flixel/FlxCamera.hx
```diff
@@ -1652,6 +1660,15 @@ class FlxCamera extends FlxBasic
 		}
 		return this.visible = visible;
 	}
+	
+	private function get_zoomOffsetX():Float
```
Beeblerox: i'd rather set values of `zoomOffsetX` and `zoomOffsetY` when you change size or zoom of the camera, instead of calculation value every time you try to get it.

flixel/FlxSprite.hx
```diff
@@ -983,13 +983,15 @@ class FlxSprite extends FlxObject
 		
 		var minX:Float = x - offset.x - Camera.scroll.x * scrollFactor.x;
 		var minY:Float = y - offset.y - Camera.scroll.y * scrollFactor.y;
+		var zoomOffsetX:Float = Camera.zoomOffsetX;
+		var zoomOffsetY:Float = Camera.zoomOffsetY;
 		
 		if ((angle == 0 || bakedRotationAngle > 0) && (scale.x == 1) && (scale.y == 1))
 		{
-			if (minX > Camera.width || minX + frameWidth < 0)
+			if (minX > Camera.width + zoomOffsetX || minX + frameWidth < -zoomOffsetX)
```
Beeblerox: maybe it's better to add new property to the camera (`viewport:FlxRect` for example) which will hold these values:
`-zoomOffsetX`, `Camera.width + zoomOffsetX`, `-zoomOffsetY` and `Camera.height + zoomOffsetY`), so no need to recalculate these values too

Gama11:
>which will generate a lot of garbage and hits perfomance badly

Is that the reason you didn't merge your fix?

Beeblerox:
@Gama11 yes, but there was also problem with zoom tweening: i couldn't get smooth zooming with it in blit render mode.
and then i forgot about this work

Beeblerox:
i'll try to get something working for flash target this weekend or next week.
The idea to achieve less garbage creation is to grow canvas bitmap not in continious way (for each zoom setter call), but in discrete way (say, for 10 or 20%) and then properly place thic canvas inside camera container.

starry-abyss:
Nice idea!
Didn't think about zoom tween, I should add it to the demo суббота, 12 ноября 2016г., 16:53 +0300 от Zaphod  notifications@github.com :

Beeblerox:
i've made some initial changes and published them here: https://github.com/HaxeFlixel/flixel/tree/camera-scaling
This work needs some polishing, like smoothing scaling tweening (i see some jerkiness while i zoom out), making all the graphical objects render properly (right now i've changed only FlxSprite and FlxTilemap rendering logic) and documenting variables. But it's mostly working across targets.

@starry-abyss as you can see i've changed variable names and don't calculate their values every frame. And blit render mode have a bit different logic due to the way sprites being renderered on camera

starry-abyss:
Good job!
I have two issues on Flash with my demo:
1) if zoom is 0.7, tile visibility is incorrect for side tiles near camera edge;
2) if zoom is at minimum (less than 0.3), it crashes with null BitmapData.

Beeblerox:
@starry-abyss ok, i'll try to see what's wrong with these issues tonight (i've tested my code with another demo)

Beeblerox:
@starry-abyss the second issue is caused by limitation of camera's canvas size (BitmapData object).
i'll set limit for camera zoom, so the camera won't be able to have small zooms.

i haven't found the source of the first issue yet

starry-abyss:
Not sure about Flash, but on neko FlxTilemapBuffer is responsible for tile visibility (I made changes to it)

Beeblerox:
@starry-abyss i know about it. just need to remember where is the code responsible for this in blit render mode

Beeblerox:
the only thing i found about first issue so far  is that it appears only if tilemap is scaled, when the scale is (1; 1) then there is no issue with tile visibility. Digging further

Beeblerox:
the issue is much harder to solve than i thought
Plus there is very strong shaking while tweening camera scaling with provided demo.
I'll need more time. Hope to solve this by the end of the week

starry-abyss:
OK. So what are the things apart from sprites and tilemaps that don't work properly yet? Only on Flash or on native too?

Beeblerox:
ok. after some more thinking i guess i've found better solution for this, but it won't be compatible with the cases when camera's `useBgAlphaBlending` is true.

Beeblerox:
@starry-abyss i've done some test and believe that there is some sort of strange flash runtime bug or something.
for example i took your demo but i've changed zooming in your demo. instead of change camera zoom on mouse wheel scroll event, i gradually decrease camera scale for a  small value (by 0.0001) each frame. When camera zoom is more than 0.45 or 0.5 then everything is going fine, but when it's close to 0.4 and lower then i get tilemap flickering (for some values of zoom tilemap doesn't rendering but for others it does).
and i don't know what to do with this behavior. actually there is one possible solution - don't use `FlxTilemapBuffer` class and render tiles directly to camera's buffer. i believe that it could help, because FlxSprites aren't affected by this issue for some reason. But i'm not sure that this will be good for perfomance, especially in cases when tilemap is being scaled or have blending or color applied.
but i'll try it tonight and tomorrow.
and i've commited my latest changes here: https://github.com/HaxeFlixel/flixel/tree/camera-scaling

Beeblerox:
so i got this:
if you don't scale sprites or tilemaps then they are rendered correctly at any zoom value,
but if you scale sprite or tilemap then there could be problems with rendering (rendering objects at some zoom values but not rendering them at others). i believe that this is bitmapdata's draw() method issue and doubt that this could be solved on my side.

Beeblerox:
so i guess we could stop here (since i cannot do anything about flash player runtime) and merge these changes. at least everything is working on native targets and it works for some cases on flash.

But there is one more possible way for flash target: don't draw everything on bitmapdata and then scale it to the camera's size, but instead we could draw on camera's buffer with applied scaling. But this way will have very negative effect on perfomance and make rendering related code on flash target more complicated.

starry-abyss:
I'm too biased here, because I want native targets and want it merged :-)
But I tested your branch now with my demo. On Flash:
1) scrolling at zoom == 0.7 - where I previously had some tiles appear and disappear before reaching edge of camera - now they tend to get cut by camera edge and not appear again (!) until zoom change or heavy scrolling
2) still crashes when going under 0.3 scale, is there something with demo code itself?

Beeblerox:
about second issue: it's caused by my error in the code - i forgot to calculate square root in the code, which is responsible for limiting of camera's buffer size. I've added it now, but this code trims camera view when it's too big

Beeblerox:
and i believa that i know how to fix first one. will do it tonight

starry-abyss:
I also have another issue when zoom is exactly 0.4:
http://i.giphy.com/l2JhFNoFQCdUomL0Q.gif

Beeblerox:
@starry-abyss can you check current code?

starry-abyss:
Looks like 1) and 2) are fixed (if scale limit counts as fix ;-)).
As for the gif I posted, the issue is still there, but I think it's same as flickering problem you mentioned. At some zoom values tilemap works OK, at some is invisible, and at some is partially visible like on gif.
So it looks like tilemap gets cropped by wrong rect.

starry-abyss:
I also updated my demo with better controls and initial scroll and zoom values to catch the partial visibility

Beeblerox:
there is only one option left to try: try not to scale camera's buffer but objects themselves when rendering objects on the camera when it's zoom is less than than 1.
i will try it but not sure if it will help.

Beeblerox:
@starry-abyss i've tried to implement the other way of handling camera zoom for blit render mode and it seems to work good. I've commited my initial changes so you can check them out.
i still need to update all other classes and make debug rendering work properly again

starry-abyss:
Great! This works flawlessly in my demo.
Yes, there is a jitter on Flash when scaling fast, but it's a separate issue, I think. It jitters similarly to camera in lerp mode on Flash and has something to do with rounding to integer, I guess?

Beeblerox:
yes, i think so. 
i could try to cancel this issue by changing position of camera's bitmapdata, but not sure if it will help

starry-abyss:
I can do some quick tests, if you have something working to show.

Beeblerox:
not yet, sorry
maybe tomorrow

Beeblerox:
i think i'll merge my branch as it is now. But need to update all other addons classes before that.

starry-abyss:
OK, didn't know it was breaking change for addons

Beeblerox:
@starry-abyss i think there won't be much changes.

Gama11:
I guess this can be closed since there's a new PR at #2003.

PR #1988, opened 2016-10-28
FlxTween: add cancelChain()

seraku24:
The existing `FlxTween.cancel()` function works fine when you want to stop a tween that is currently active and have it yield to the next tween in the chain (if one exists).  However, the function does not help when you want to stop an entire tween chain, especially when the only reference you may hold is to the first tween in the chain.  (NOTE: Use of `FlxTween.wait()` results in the creation of hidden tweens that cannot be cancelled.)

This PR introduces `FlxTween.cancelChain()`, a new function whose purpose is to allow user code to cancel all subsequent tweens that may be chained to a particular tween, even when the root tween has already finished.  This helps simplify user code that relies on both chained tweening and the ability to cancel the animation at arbitrary times.

### DIFF
--- a/flixel/tweens/FlxTween.hx
+++ b/flixel/tweens/FlxTween.hx
@@ -497,8 +497,10 @@ class FlxTween implements IFlxDestroyable
 	}
 	
 	/**
-	 * Immediately stops the Tween and removes it from the 
-	 * TweenManager without calling the complete callback.
+	 * Immediately stops the Tween and removes it from its 
+	 * `manager` without calling the `onComplete` callback.
+	 *
+	 * Yields control to the next chained Tween if one exists.
 	 */
 	public function cancel():Void
 	{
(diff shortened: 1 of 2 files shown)


seraku24:
Code Climate reports it found issues with my PR; but, even after waiting about half a day, the details page still claims there is no analysis report available.  Is this automated testing not working properly?  If it is, how am I supposed to review and correct the issues?

Gama11:
Hm.. that's quite odd, that's the first time I've seen "No analysis results present.".

seraku24:
@Gama11, glad to know I still have a knack for breaking systems the first time I encounter them.  :smirk:

Joking aside, is there a way to manually run the same code quality checks?  I did try to match the conventions and patterns in the files I modified; but, without a definitive list of guidelines, I cannot know for certain how my code stacks up.

Do let me know what, if anything, you need me to do from my side to help resolve this.

Gama11:
Thanks!

PR #1990, opened 2016-10-29
Fix #1986

### LINKED ISSUE #1986: FlxMouseEventManager.removeAll() fails to completely unregister even number of objects
- **Flixel version:** 4.1.0
- **OpenFL version:** 3.6.1
- **Lime version:** 2.9.1
- **Affected targets:** at least Neko, HTML5, and flash, although I'm guessing it happens on the rest as well.

---

``` haxe
package;

import flixel.FlxG;
import flixel.FlxSprite;
import flixel.FlxState;

class MenuState extends FlxState
{
    {
        super.create();

        var spr1 = new FlxSprite(100, 200);
        var spr2 = new FlxSprite(200, 200);
        var spr3 = new FlxSprite(300, 200);

        FlxMouseEventManager.init();
        FlxMouseEventManager.add(spr1, null, function(button:FlxSprite){FlxG.switchState(new PlayState()); }, 
                                       null, function(button:FlxSprite){ button.velocity.set(0, 0); });
        FlxMouseEventManager.add(spr2, null, function(button:FlxSprite){FlxG.switchState(new PlayState()); }, 
                                       null, function(button:FlxSprite){ button.velocity.set(0, 0); });
        //FlxMouseEventManager.add(spr3);

        add(spr1);  
        add(spr2);
    }

    override public function switchTo(nextState:FlxState):Bool
    {
        FlxMouseEventManager.removeAll();
        return super.switchTo(nextState);
    }
}
```

---

**Observed behavior:**

I registered a pair of `FlxSprite`s to the `FlxMouseEventManager`, setting its `mouseUp` function to switch between states and its `mouseOut` function to call one of the sprite's variable's functions. I also overrode the starting state's `switchTo` to call `FlxMouseEventManager`'s `removeAll` function, so the buttons should be removed from `FlxMouseEventManager` whenever the state switch occurs.

When the game switches between states, it properly removes the second `FlxSprite`, but it fails to remove the first. If the second sprite was clicked, everything's fine. If the first sprite was clicked, however, its `mouseOut` function is called after the state switch finishes. Since the registered `mouseOut` function interacts with one of the sprite's variables that was set to `null` during the state change, the game crashes.

As far as I can tell, this always happens if an even number of objects are registered to the `FlxMouseEventManager`. 

If you have an odd number of objects registered to the `FlxMouseEventManager`, no problem occurs. You can try it by uncommenting the line `FlxMouseEventManager.add(spr3);`

**Expected behavior:**

I expected that all registered objects would be removed from `FlxMouseEventManager` by `removeAll()`, regardless of whether number of objects is even or odd.

Side note: I did find a workaround. By removing the `FlxMouseEventManager` plugin from FlxG and then calling its `init()`, I could force the `FlxMouseEventManager` to clear itself completely. It may also be possible that removing instances one by one works, but I haven't tried it yet.
##

MSGhero:
Closes #1986

Iterating while removing, change taken from `Watch`. I'm not sure if copying is the best way vs iterating twice, but this makes it consistent with other parts of flixel.

### DIFF
--- a/flixel/input/mouse/FlxMouseEventManager.hx
+++ b/flixel/input/mouse/FlxMouseEventManager.hx
@@ -109,7 +109,7 @@ class FlxMouseEventManager extends FlxBasic
 	{
 		if (_registeredObjects != null)
 		{
-			for (reg in _registeredObjects)
+			for (reg in _registeredObjects.copy())
 			{
 				remove(reg.object);
 			}


seraku24:
This might fall under the category of refactoring, but there is potential for a cleaner and faster solution here.

Having `removeAll()` defer to `remove()` is slow, specifically O(_n_²), due to the need to iterate the collection per each element.  Since all items are going to be removed, the array can be cleared directly instead of removing elements one at a time.  To avoid duplication of code, the existing "null"ing logic currently within `remove()` should be in a `destroy()` method that belongs to `ObjectMouseData<T>`.  This would enable both `remove()` and `removeAll()` to share this cleanup logic without depending on each other.

_(Some hastily written and untested code for clarity.)_
```haxe
public function remove<T:FlxObject>(Object:T):T {
    if (_registeredObjects == null || _registeredObjects.length == 0) { return Object; }
    for (reg in _registeredObjects) {
        if (reg.object == Object) {
            reg.destroy();
            _registeredObjects.remove(reg);
        }
    }
    return Object;
}

public function removeAll():Void {
    if (_registeredObjects == null || _registeredObjects.length == 0) { return; }
    for (reg in _registeredObjects) { reg.destroy(); }
    _registeredObjects = [];
}

// . . .

public function destroy():Void {
    object = null; sprite = null;
    onMouseDown = onMouseUp = onMouseOver = onMouseOut = null;
}
```

Just something to consider.  Although, it sounds like there might be other parts of the project that could be improved with the same refactoring, which would be beyond the scope of the issue you are fixing.

MSGhero:
I agree, there are plenty of places where removal logic should be changed.

MSGhero:
I agree, there are plenty of places where removal logic should be changed. You should make a new issue for it 😁

MSGhero:
@seraku24 Have you actually tried this fix yet? It properly removes all registered objects, but I don't have your code to see if that was causing your specific issue.

seraku24:
@MSGhero, I did not file the original issue, #1986, so I cannot comment on that aspect.  I just pointed out the computational complexity—yes, I am quite guilty of premature optimization here—and posted a quick sketch of an alternative.

MSGhero:
Oh crap I mixed up my tabs

MSGhero:
This PR is good to go, fixes #1986 according to OP of that thread

Gama11:
I didn't really doubt that it fixes the issue (the test case from the issue doesn't crash anymore), I just didn't think your explanation of _why_ it fixes it makes sense.

Beeblerox:
i like the solution proposed by @seraku24 at https://github.com/HaxeFlixel/flixel/issues/1992 
i don't like the thing that you have to create copy of array to clear it.

MSGhero:
@Beeblerox yeah, #2005 solves the larger problem. Close this and leave it to that issue?

@Gama11 you're right, I'm not sure why `mouseOver` was being called if the array containing it gets erased `¯\_(ツ)_/¯`

Issue #1895, opened 2016-07-31
Improve FlxSwipe


DleanJeans:
Right now, `FlxSwipes` are only created after mouse or touch is released, which is not how I expected them to work. So I decided to improve it a little: a `FlxSwipe` will be created as soon as a mouse button or touch is down and kept in `FlxG.swipes` until mouse/touch is released and its `endPosition` is updated every frame.
And I'm feeling like this is not enough. I think FlxSwipe also needs a threshold since a click or a tap is also counted as a swipe. What about a swipe going in different directions (like circling, ping pong)? Should we add FlxPoints between start and end position like a path?

DleanJeans:
I think I should just close this. Since updating `endPosition` would just unnecessarily decrease the performance, maybe.

Gama11:
>Since updating endPosition would just unnecessarily decrease the performance, maybe.

Which part of that do you think would be expensive?

DleanJeans:
>  maybe.
Perhaps it's not that expensive. I was just trying to implement something like [FlxScrollableArea](https://github.com/IBwWG/FlxScrollableArea) but with swipe-to-scroll for mobile target (for a friend). And FlxSwipe isn't enough for this and seems like I'm the only one who has talked about this.

Tembac:
I found this problem now. I'm trying to do a level map for a mobile game that can be scrolled using touches and swipes. But I think that swipes are not measured correctly by HaxeFlixel. 

If I pressed the mouse button down and move the mouse around for a few seconds the swipe is taking the first point of click and the last as the swipe points. But the last swipe really was different. I think that swipes need to be measuring the starting point every a few second or when it drastically changes direction. 

I'm not sure yet where Flixel is handling swipes so I'm not sure how to fix it.

PR #1997, opened 2016-11-09
FlxSwipe: add released, path, qualified and velocity

DleanJeans:
For issue #1895
[Test case](https://gist.github.com/DleanJeans/f3c0261bc98e078b23eae92fcd1fef9a)
Tested on Flash and Android
It seems to have some problems with releasing touch/mouse (swipe never released)

### DIFF
--- a/flixel/input/FlxSwipe.hx
+++ b/flixel/input/FlxSwipe.hx
@@ -99,17 +99,20 @@ class FlxSwipe
 			LabelValuePair.weak("qualified", qualified)]);
 	}
 	
-	private function get_start():FlxPoint {
+	private function get_start():FlxPoint
+	{
 		return path.head();
 	}
 	
-	private function get_end():FlxPoint {
+	private function get_end():FlxPoint
+	{
 		return path.tail();
 	}
 	
 	private function get_velocity():FlxPoint
 	{
-		if (!released) {
+		if (!released)
+		{
 			velocity.set(end.x - start.x, end.y - start.y);
 			velocity.scale(1 / duration);
 		}
--- a/flixel/input/FlxSwipe.hx
+++ b/flixel/input/FlxSwipe.hx
@@ -3,60 +3,137 @@ package flixel.input;
 import flixel.FlxG;
 import flixel.math.FlxMath;
 import flixel.math.FlxPoint;
+import flixel.util.FlxPath;
 import flixel.util.FlxStringUtil;
 
 @:allow(flixel.input.mouse.FlxMouseButton)
 @:allow(flixel.input.touch.FlxTouch)
 class FlxSwipe
 {
+	public static var VELOCITY_THRESHOLD = 100;
+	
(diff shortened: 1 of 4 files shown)


DleanJeans:
I think there won't be any problems with swiping now unless you tap crazily with many fingers on phone targets. Needs something to get rid of those never-get-released swipes.

Also CI has some problems with
```
The command "haxelib install task > /dev/null" failed and exited with 1 during .
```

mastef:
Fighting atm with never-get-released swipes too, seems to happen when the finger leaves the game's bounds? ( e.g. left/right of screen ) - could this be?

flixel/input/FlxSwipe.hx
```diff
@@ -3,60 +3,140 @@ package flixel.input;
...
+	
 	/**
 	 * Either LEFT_MOUSE, MIDDLE_MOUSE or RIGHT_MOUSE, 
 	 * or the touchPointID of a FlxTouch.
 	 */
 	public var ID(default, null):Int;
 	
-	public var startPosition(default, null):FlxPoint;
-	public var endPosition(default, null):FlxPoint;
+	public var path(default, null):FlxPath;
```
Gama11: Creative use of `FlxPath`. Wouldn't a plain `Array<FlxPoint>` be more suitable though?

flixel/input/touch/FlxTouch.hx
```diff
@@ -75,11 +77,19 @@ class FlxTouch extends FlxPointer implements IFlxDestroyable implements IFlxInpu
...
 			justPressedPosition.set(screenX, screenY);
 			justPressedTimeInTicks = FlxG.game.ticks;
+			#if FLX_POINTER_INPUT 
+			FlxG.swipes.push(_currentSwipe = new FlxSwipe(touchPointID, justPressedPosition, justPressedTimeInTicks)); 
+			#end
 		}
 		#if FLX_POINTER_INPUT
+		else if (pressed)
+		{
+			_currentSwipe.path.addPoint(getScreenPosition());
```
Gama11: So the path will contain a point for each frame that the swipe was active for? I wonder if that isn't a bit much.

flixel/input/FlxSwipe.hx
```diff
@@ -3,60 +3,140 @@ package flixel.input;
...
+	private function get_end():FlxPoint
+	{
+		return path.tail();
+	}
+	
+	private function get_velocity():FlxPoint
+	{
+		if (!released)
+		{
+			velocity.set(end.x - start.x, end.y - start.y);
```
Gama11: I'm not sure if this definition of velocity where you just take the start and end position into account makes sense. You could have a very fast, but _circular_ swipe, meaning that `start` and `end` are virtually the same point.

Gama11:
This implementation has a few issues as pointed out by my earlier comments.

I'd like to revisit this in the future though, there's definitely room for improvement in the `FlxSwipe` API, which is why I'm keeping #1895 open.

PR #1999, opened 2016-11-12
FlxMouseEventManager: fix pixel-perfect overlaps with offset

seraku24:
Updated logic in `FlxMouseEventManager.checkPixelPerfectOverlap` to account for a non-zero offset affecting the pivot point.


seraku24:
@Gama11, @Beeblerox: What's the support path for issues with Travis?  Near as I can tell, there appears to be an issue with the build system itself, with the `haxelib install task` command failing in only three of the eight jobs.

Gama11:
Reported to Haxelib: https://github.com/HaxeFoundation/haxelib/issues/358

PR #2004, opened 2016-11-26
Watch: avoid multiple Expr parses, closes #1886

### LINKED ISSUE #1886: WatchExpression: store parsed AST
This is a reminder so I don't forget, I'll fix it when I have time.

Right now when you use `watchExpression`, your string of code gets parsed and interp'd every frame. I just realized though that the parsed AST is the same every single frame. It's the interp step where reflection and all happens. My memory rises rather quickly when I watch an expression, more than when doing a regular `watch`, maybe this is why.

I think what I'll do (and what I do in my own game, so it works) is give the `EXPRESSION` watch data enum a `hscript.Expr` parameter instead of the string. The command gets parsed and stored there. Then when the watch gets updated, it interps the stored AST instead of re-parsing the same string into the same AST.

**unless you can edit the command from the watch window, I've never tried that
##

MSGhero:
Closes #1886 

I did this a bit hastily. I don't think I broke the other watch enums, but I only checked `EXPRESSION`. Also didn't check `!hscript`.

### DIFF
--- a/flixel/system/debug/console/ConsoleUtil.hx
+++ b/flixel/system/debug/console/ConsoleUtil.hx
@@ -67,9 +67,9 @@ class ConsoleUtil
 	 * @param	Parsed	The parsed form of the user's input command.
 	 * @return	Whatever the input code evaluates to.
 	 */
-	public static function runExpr(ParsedInput:Expr):Dynamic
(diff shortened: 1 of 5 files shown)


Gama11:
It seems memory usage does indeed increase more slowly with this (on Flash).

PR #2011, opened 2016-12-08
FlxStringUtil: bugfix on formatMoney

sebbernery:
The current behavior add a "0" if decimal part is inferior to 10, but append it after the decimal part. It should be appended before the decimal part.
Example: when using on numbers like 4.0312, formatMoney returned "4.30". The expected value was "4.03".


Beeblerox:
@sebbernery you're totally right. Thanks for pull request!

PR #2013, opened 2016-12-18
Update FlxAnimationController.hx

stroncium:
Added a check for frames number to `set_frameIndex(Frame:Int):Int` so it wont throw "Invalid operation (%)" while doing Frame % frames if there are 0 frames.


Beeblerox:
@stroncium thank you!

PR #2016, opened 2016-12-26
FlxG.cameras: fix reset() not removing all cameras

Leonix:
The problem is that `FlxG.cameras.reset()` tries to remove elements of an `Array` while iterating over it. 
This fails to `.destroy()` half of all cameras, leading to the memory leak.

- **Flixel version:** 4.2.0
- **OpenFL version:** 3.6.1
- **Lime version:** 2.9.1
- **Affected targets:** at least Neko, likely more
- **Observed behavior:** Process memory usage steadily goes up.
- **Expected behavior:** It should not.

```haxe
package;
import flixel.FlxG;
import flixel.FlxState;
import flixel.FlxCamera;

{
	{
		FlxG.cameras.add(new FlxCamera());
	}

	override public function update(elapsed: Float):Void
	{
		super.update(elapsed);
		FlxG.switchState(new PlayState());
	}
}
```

### DIFF
--- a/flixel/system/frontEnds/CameraFrontEnd.hx
+++ b/flixel/system/frontEnds/CameraFrontEnd.hx
@@ -97,10 +97,9 @@ class CameraFrontEnd
 	 */
 	public function reset(?NewCamera:FlxCamera):Void
 	{
-		while (list.length > 0) {
+		while (list.length > 0)
 			remove(list[0]);
-		}
-		
+
 		if (NewCamera == null)
 			NewCamera = new FlxCamera(0, 0, FlxG.width, FlxG.height);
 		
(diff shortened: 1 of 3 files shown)


Gama11:
Looks like this is a regression from edf93b57954d943368b5bba9a3c187a455101050.

Thanks!

Leonix:
Note that unit test for this should probably test for whether `FlxG.cameras.cameraRemoved` signal is called for all cameras. Or whether all cameras were properly `camera.remove()`d. Checking only for `list.length` would pass on vanilla 4.2.0, too.

Gama11:
You're right, I made a mistake when checking whether the test actually fails before your changes...

PR #2020, opened 2017-01-02
FlxSpriteGroup: add insert()

MSGhero:
Basically just `add` with a bit more control over positioning, or a more useful `group.insert()`. May or may not need it for haxeui.


Gama11:
This is almost a 1:1 copy of `add()`, I'd like to avoid that redundancy.

MSGhero:
I think an optional add parameter would break people's overrides, so a private `addButNotActually()` that just offsets the sprite and all?

Gama11:
Yeah. that's what I meant.

MSGhero:
I can't think of a good function name

MSGhero:
Ok

PR #2021, opened 2017-01-02
FlxAtlasFrames.fromTexturePackerJson(): allow passing a parsed description

### LINKED ISSUE #2019: From TexturePacker Json object
When you already have a parsed Json, you have to re-stringify and re-parse it to get it into a `FlxAtlasFrames` (without manually doing graphic/frame stuff).

Without repeating code in a separate function, I think turning `Description` into a Dynamic is a good non-breaking bet: either use it directly or cast to string and use it as it currently is used. I'll have to test it out. 

Mostly a reminder for when I get wifi again.

MSGhero:
Closes #2019

In case the JSON is already parsed to an object, this avoids hundreds of thousands of allocations for big spritesheets.

### DIFF
--- a/flixel/graphics/frames/FlxAtlasFrames.hx
+++ b/flixel/graphics/frames/FlxAtlasFrames.hx
@@ -50,8 +50,8 @@ class FlxAtlasFrames extends FlxFramesCollection
 		
 		var data:Dynamic;
 		
-		if (Std.is(Description, String)) {
-			
+		if (Std.is(Description, String))
+		{
 			var json:String = Description;
 			
 			if (Assets.exists(json))
@@ -60,7 +60,8 @@ class FlxAtlasFrames extends FlxFramesCollection
 			data = Json.parse(json);
 		}
 		
-		else {
+		else
+		{
 			data = Description;
 		}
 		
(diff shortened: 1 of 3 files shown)


Gama11:
Hm.. I guess there isn't really a way to avoid `Dynamic` here?

A unit test would be nice. IIRC, we already have some for `fromTexturePackerJson()`.

Gama11:
I wonder whether it'd make sense to use `OneOfTwo<String, Dynamic>` here. Of course it doesn't add any type safety, but it might serve as self-documentation in a way.

MSGhero:
I guess the most type-safe way to handle the dynamic is to expect the `frames` property. Could also be a private typedef `TPObject` if this looks gross. 

`OneOfTwo<String, { frames : Array<Dynamic> }>`

Gama11:
Is it really guruanteed to be an array? There's some code for "JSON-Hash".

MSGhero:
Oh dang, I meant `frames : Dynamic` which is better than nothing. It would prevent you from putting `5` there, at least.

Gama11:
Yeah, that seems sensible.

MSGhero:
👍 I think, I haven't run unit tests locally in a while.

PR #2023, opened 2017-01-06
FlxSubState: add openCallback

alihassan0:
useful for cases when you want to update certain data when the substates is resumed.

### DIFF
--- a/flixel/FlxState.hx
+++ b/flixel/FlxState.hx
@@ -119,7 +119,8 @@ class FlxState extends FlxGroup
 				subState._created = true;
 				subState.create();
 			}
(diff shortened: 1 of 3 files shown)


flixel/FlxState.hx
```diff
@@ -119,6 +119,7 @@ class FlxState extends FlxGroup
 				subState._created = true;
 				subState.create();
 			}
+			subState.openCallback();
```
MSGhero: Needs a null check in case it's not defined

alihassan0: that's right .. I will add it now

PR #2024, opened 2017-01-06
Fix FlxTilemap.getTileIndexByCoords() not checking for tilemap bounds.

starry-abyss:
This is to eliminate false callbacks by FlxMouseEventManager with tilemaps.
Also added the new check where getTileIndexByCoords() is also used and did the small refactoring in FlxBaseTilemap.overlapsPoint().

### DIFF
--- a/flixel/tile/FlxBaseTilemap.hx
+++ b/flixel/tile/FlxBaseTilemap.hx
@@ -724,6 +724,10 @@ class FlxBaseTilemap<Tile:FlxObject> extends FlxObject
 		// Figure out what tile we are starting and ending on.
 		var startIndex:Int = getTileIndexByCoords(Start);
 		var endIndex:Int = getTileIndexByCoords(End);
+		
+		// Check if any point given is outside the tilemap
+		if ((startIndex < 0) || (endIndex < 0))
+			return null;
 
 		// Check that the start and end are clear.
 		if ((_tileObjects[_data[startIndex]].allowCollisions > 0) || (_tileObjects[_data[endIndex]].allowCollisions > 0))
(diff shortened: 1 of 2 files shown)


Gama11:
We should have unit tests for this kind of stuff..

Gama11:
@starry-abyss Do you have a code example for a case that was fixed by this?

starry-abyss:
Here. If you clicked outside the tilemap prior to the patch, you got the pressed() callback, because getTileIndexByCoords() didn't care if the point was inside or outside the tilemap.

```haxe
package;

import flixel.FlxG;
import flixel.FlxSprite;
import flixel.FlxState;
import flixel.math.FlxVector;
import flixel.text.FlxText;
import flixel.tile.FlxTilemap;
import flixel.ui.FlxButton;
import flixel.math.FlxMath;

{
	var tilemap: FlxTilemap;
	
	{
		super.create();
		
		tilemap = new FlxTilemap();
		tilemap.loadMapFromCSV("1, 3, 1,\n 1, 2, 2,\n 3, 3, 1", "assets/images/tileset2.png", 128, 128);
		tilemap.setPosition(100, 100);
		
		FlxMouseEventManager.add(tilemap, pressed);
		
		add(tilemap);
	}
	
	function pressed(tilemap: FlxTilemap)
	{
		trace("pressed at: ", FlxG.mouse.x, FlxG.mouse.y);
		
		trace("tilemap bounds are: ", tilemap.x, tilemap.y, tilemap.x + tilemap.width, tilemap.y + tilemap.height);
	}

	{
		super.update(elapsed);

	}
}
```

Gama11:
Thanks! Added a simple unit test.

PR #2033, opened 2017-01-23
Update FlxTilemap.hx

Triangle345:
made useScaleHack an Int to determine exactly how much scale we want to give it to fix tearing/gaps between tiles.  This was done because sometimes the scale had to be greater than 1 to fill in the gaps between tiles.  This creates a bit more flexibility without disturbing existing behavior.

### DIFF
--- a/flixel/tile/FlxTilemap.hx
+++ b/flixel/tile/FlxTilemap.hx
@@ -53,7 +53,7 @@ class FlxTilemap extends FlxBaseTilemap<FlxTile>
 	 * Try to eliminate 1 px gap between tiles in tile render mode by increasing tile scale, 
 	 * so the tile will look one pixel wider than it is.
 	 */
-	public var useScaleHack:Bool = true;
+	public var scaleHack:Int = 1;
 	
 	/**
 	 * Changes the size of this tilemap. Default is (1, 1). 
@@ -979,10 +979,10 @@ class FlxTilemap extends FlxBaseTilemap<FlxTile>
 						var scaleX:Float = scale.x;
 						var scaleY:Float = scale.y;
 						
-						if (useScaleHack)
+						if (scaleHack > 0)
 						{
-							scaleX += 1 / (frame.sourceSize.x * Camera.totalScaleX);
-							scaleY += 1 / (frame.sourceSize.y * Camera.totalScaleY);
+							scaleX += scaleHack / (frame.sourceSize.x * Camera.totalScaleX);
+							scaleY += scaleHack / (frame.sourceSize.y * Camera.totalScaleY);
 						}
 						
 						_matrix.scale(scaleX, scaleY);


Triangle345:
The build is failing because some of these examples depend on the old useScaleHack variable... Need to change both the samples repo and the main repo at the same time

Gama11:
Wasn't the plan anyway to remove this entirely (@Beeblerox)?

Triangle345:
So the issue with removing it is that the FlxTilemap class does not allow
much flexibility on where to graphically place(offset) the tiles if it had
borders or gaps between them.  I have had issues where my 128x128 tiles
were producing gaps even with scaling value of 1. I had to use a higher
scaling value to make the tiles seamless.

Triangle345:
created issue #2034

Gama11:
I think the consensus is that this is a better method of solving it than `useScaleHack`: http://forum.haxeflixel.com/topic/39/tilemap-tearing-desktop-targets/3

Triangle345:
May I recommend we document and provide some examples on using fromBitmapAddSpacesAndBorders.  FlxTileFrames has virtually no examples or documentation on how to use with  TileMap.  Furthermore; FlxTileFrames is not even in the same tile package as FlxTilemap so it makes it a bit obscure.

Gama11:
Where would you add the documentation?

Triangle345:
Maybe change the example here, not sure:
https://github.com/HaxeFlixel/flixel-demos/blob/master/Features/Tilemap/source/PlayState.hx

Gama11:
I don't think that would really help much, that's not a central place at all, it's just one of many demos that uses tilemaps.

Btw, did you manage to solve your issue with the instructions in the thread?

Triangle345:
I have not solved the issue. I'm trying what the article recommends but having trouble getting it to work. I have this so far:

		var frames = FlxTileFrames.fromBitmapAddSpacesAndBorders(AssetPaths.tiles__png,new FlxPoint(128,128), new FlxPoint(0,0), new FlxPoint(2,2));
		

        super.loadMapFrom2DArray(
            mapOrig,
            frames,
            128,128,OFF,0,0,1); 

no matter what i set the spacing and border, nothing changes...

Triangle345:
So the only way i am able to solve this issue is with the mapScaleHack with scale set to 2.

starry-abyss:
Wait, so the tileset itself has gaps between tiles? I mean the image you load

Triangle345:
No the tileset that I loaded is tightly packed and should have NO gaps. I even opened up in GIMP to make sure. However; when loading it through as stated above I notice small gaps. Even on mapscalehack scale value of 1 I see small gaps. mapscalehack scale value of 2 gets rid of these gaps for me.

starry-abyss:
Beware that the scale hack causes a grid to appear if tiles are semi-transparent (if you have those). Also there is another possibility: https://github.com/HaxeFlixel/flixel/pull/1953

Triangle345:
Let me show you what im experiencing with 128x128 tightly packed tiles:

scalehack on scale 2:
scalehack2

scalehack on scale 1
scalehack1

scalehack on scale 0 aka no scale
scalehack0

starry-abyss:
Can you also upload a minimal demo, where it happens?

Triangle345:
Ok here is a zip with a small example of what it looks like with mapscalehack scale of 1. The gaps are very minuscule but you they are there.

[TilesTest.zip](https://github.com/HaxeFlixel/flixel/files/729431/TilesTest.zip)

starry-abyss:
I apologize but when I add tileMap.useScaleHack = false, the result looks like this:
default

Which is expected, because it is the same as the tileset itself:
default

Triangle345:
If you open it up in GIMP editor , there doesnt appear to be any spaces..
It is tightly packed.

starry-abyss:
Ah, so you have semi-transparency here, which together with the black background produce the seams we observe
default

And here's the GIMP screenshot where I added a black layer underneath to show:
default

Triangle345:
Oh wow, i did not realize this. Thank you for noticing!  When I get home I
will try it out today and close the ticket.

Triangle345:
i would like to thank @starry-abyss  for being so helpful. :)

PR #2043, opened 2017-02-25
FlxSoundGroup: add pause() and resume()

diegosolo:
This update allows developers to pause and resume all sounds in a sound group. This is useful when, for example, opening a pause menu. In such case, the developer could pause all gameplay sounds (in one or more groups) and resume them after the menu closes.

The code suggested is pretty straightforward so examples shouldn't be necessary. In any case, I could add some under request.


Gama11:
Looks good, thanks!

MSGhero:
I did not realize this didn't exist

PR #2051, opened 2017-04-18
FlxSpriteGroup clipRect support

MSGhero:
Needs more extensive (unit) testing, but the basics are done. Essential for haxeui components.

Suggestions welcome.

### DIFF
--- a/flixel/group/FlxSpriteGroup.hx
+++ b/flixel/group/FlxSpriteGroup.hx
@@ -282,8 +282,8 @@ class FlxTypedSpriteGroup<T:FlxSprite> extends FlxSprite
 		sprite.scrollFactor.copyFrom(scrollFactor);
 		sprite.cameras = _cameras; // _cameras instead of cameras because get_cameras() will not return null
 		
-		if (clipRect != null)
-			clipRectTransform(sprite, clipRect);
+		if (clipRect != null) clipRectTransform(sprite, clipRect);
+		else sprite.clipRect = null;
 	}
 	
 	/**
@@ -769,7 +769,7 @@ class FlxTypedSpriteGroup<T:FlxSprite> extends FlxSprite
 	
 	override function set_clipRect(rect:FlxRect):FlxRect
 	{
-		if (exists) 
+		if (exists)
 			transformChildren(clipRectTransform, rect);
 		return super.set_clipRect(rect);
 	}
@@ -890,7 +890,10 @@ class FlxTypedSpriteGroup<T:FlxSprite> extends FlxSprite
 	private inline function originTransform(Sprite:FlxSprite, Origin:FlxPoint)             Sprite.origin.copyFrom(Origin);
 	private inline function scaleTransform(Sprite:FlxSprite, Scale:FlxPoint)               Sprite.scale.copyFrom(Scale);
 	private inline function scrollFactorTransform(Sprite:FlxSprite, ScrollFactor:FlxPoint) Sprite.scrollFactor.copyFrom(ScrollFactor);
-	private inline function clipRectTransform(Sprite:FlxSprite, ClipRect:FlxRect)          Sprite.clipRect = FlxRect.get(ClipRect.x - Sprite.x + x, ClipRect.y - Sprite.y + y, ClipRect.width, ClipRect.height);
+	private inline function clipRectTransform(Sprite:FlxSprite, ClipRect:FlxRect)          {
+		if (ClipRect == null) Sprite.clipRect = null;
+		else Sprite.clipRect = FlxRect.get(ClipRect.x - Sprite.x + x, ClipRect.y - Sprite.y + y, ClipRect.width, ClipRect.height);
+	}
 
 	// Functions for the FlxCallbackPoint
 	private inline function offsetCallback(Offset:FlxPoint)             transformChildren(offsetTransform, Offset);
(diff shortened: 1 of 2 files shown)


flixel/group/FlxSpriteGroup.hx
```diff
@@ -879,6 +890,10 @@ class FlxTypedSpriteGroup<T:FlxSprite> extends FlxSprite
 	private inline function originTransform(Sprite:FlxSprite, Origin:FlxPoint)             Sprite.origin.copyFrom(Origin);
 	private inline function scaleTransform(Sprite:FlxSprite, Scale:FlxPoint)               Sprite.scale.copyFrom(Scale);
 	private inline function scrollFactorTransform(Sprite:FlxSprite, ScrollFactor:FlxPoint) Sprite.scrollFactor.copyFrom(ScrollFactor);
+	private inline function clipRectTransform(Sprite:FlxSprite, ClipRect:FlxRect)          {
```
MSGhero: For the record, I don't like this part

flixel/group/FlxSpriteGroup.hx
```diff
@@ -280,6 +281,9 @@ class FlxTypedSpriteGroup<T:FlxSprite> extends FlxSprite
 		sprite.alpha *= alpha;
 		sprite.scrollFactor.copyFrom(scrollFactor);
 		sprite.cameras = _cameras; // _cameras instead of cameras because get_cameras() will not return null
+		
+		if (clipRect != null) clipRectTransform(sprite, clipRect);
+		else sprite.clipRect = null;
```
Gama11: This seems to make this a breaking change, because it removes the `clipRect` that sprite might have had before being added. Not sure that's a good idea?

flixel/group/FlxSpriteGroup.hx
Gama11: "This part"? Isn't this bascially the whole PR? :D

Gama11:
The `[WIP]` is supposed to indicate you don't want this to be merged yet?

MSGhero:
What I have works, but idk if I should add more to it or if it's correct.

flixel/group/FlxSpriteGroup.hx
MSGhero: I don't like how it looks so different from the others

flixel/group/FlxSpriteGroup.hx
MSGhero: I guess the alternative would be to merge the parent's clip and the child's clip together. I'd have to look at that in Flash to see how it works. Either way, the child's clip would change, though.

flixel/group/FlxSpriteGroup.hx
MSGhero: Aside from the inlined curly brace...

flixel/group/FlxSpriteGroup.hx
Gama11: This is the case where the parent's clipRect is null though, so there isn't really anything _to_ merge?

MSGhero: I guess the "merge" would be `null + child.clip` which results in `child.clip`. Was your initial comment about line 286 or both of those lines?

Gama11: Just line 286.

MSGhero: Yeah not sure why that line was there, removed.

flixel/group/FlxSpriteGroup.hx
```diff
@@ -280,6 +281,8 @@ class FlxTypedSpriteGroup<T:FlxSprite> extends FlxSprite
 		sprite.alpha *= alpha;
 		sprite.scrollFactor.copyFrom(scrollFactor);
 		sprite.cameras = _cameras; // _cameras instead of cameras because get_cameras() will not return null
+		
+		if (clipRect != null) clipRectTransform(sprite, clipRect);
```
Gama11: This null check now seems unnecessary, since `clipRectTransform()` can handle `null`?

Gama11:
>Needs more extensive (unit) testing [...]

So are you still planning to write some tests for this?

flixel/group/FlxSpriteGroup.hx
MSGhero: You're right, that's some leftover code from when I was testing.

MSGhero:
I *guess* I'll write some tests 😜

flixel/group/FlxSpriteGroup.hx
MSGhero: Actually no, that would override the child's existing clip. The transform sets the child's clip based on the parent's, so you'd be setting `null` instead of leaving the existing one alone.

Gama11: Oh, you're right!

Great thing to unit-test. ;)

MSGhero:
Kind of a weird test, but it covers the 3 cases: existing child clip, nonexisting child clip, and nonexisting parent clip.

Gama11:
Thanks!

I still hate how clipRects have to be re-assigned entirely for changes to be registered, but that doesn't really have anything to do with this PR...

MSGhero:
I don't think anyone likes that...

Gama11:
Just occured to me, do we need to take offset into account?

MSGhero:
I want to say no, but I'll check.

MSGhero:
As it is now, it matches the behavior of a FlxSprite's clip+offset, where it clips first and offsets second (i.e. the offset doesn't affect where the clip occurs).

One thing I noticed is that the parent's `offset` doesn't get copied to the child in `preAdd()`. Should it?

PR #2052, opened 2017-04-20
HTML5FrontEnd: fix platfom detection for iDevices

DleanJeans:


Gama11:
What was broken here exactly?

DleanJeans:
Sorry!
This [test](https://dleanjeans.github.io/flixel-html5-platform-test/) shows Mac for platform on an iPhone, and on iPods too I believe. See DleanJeans/flixel-html5-platform-test#1.

Gama11:
Thanks! I'll have to assume that the fix is correct, because I don't own any Apple Hardware to test this with. ;)

PR #2056, opened 2017-04-26
FlxStringUtil.getHost to accept URL fragments.

### LINKED ISSUE #2055: FlxPreloader sitelock not triggering anymore
I was trying to force the sitelock failure message to show up to test #1994. The following code should have triggered it, since I was testing locally, and only `foo.com` is allowed:

```xml
<app preloader="CustomPreloader" />
```

```haxe
import flixel.system.FlxBasePreloader;

class CustomPreloader extends FlxBasePreloader {
	public function new() {
		super();
		allowedURLs = ["www.foo.com"];
	}
}
```

I bisected this to 57dc460, before that commit it worked as expected. @seraku24 could you look into this?

seraku24:
- Fixes #2055.
- Previously, the function was expecting a proper URL complete with a protocol specification (i.e. `http://`).
- The protocol section is now marked optional, so that it can process fragments (i.e. `www.example.test` is parsed the same as `http://www.example.test`).
- This ultimately permits users to list domains as well as full URLs in the `allowedURLs` list of `FlxBasePreloader` and still have site locking work as expected.


Gama11:
Should probably update the tests?

seraku24:
*Le sigh.*  Those pesky tests.  Looks like some of the "local" paths are matching as fragments.  Let me double-check the regex.

Gama11:
>Those pesky tests

Well.. you wrote them. :D

seraku24:
Yeah, but I wrote them so long ago.  It's almost like it was someone else.

Gama11:
Thanks for the quick fix!

PR #2060, opened 2017-05-06
FlxRandom: make shuffle() slightly more efficient

16th-earl:
The shuffle function should have i iterating in the range [0,n-2] (where n=array.length). Then we choose a 'j' index to swap in the range [i,n-1]. However, currently i iterates in the range [0,n-1], which means on the final iteration the function does an unnecessary swap of element [n-1] with itself. I've made a small code change to correct this.


Gama11:
I've added a unit test that ensures `shuffle()` always produces the same result (assuming it's same initial seed). Both dev and this branch pass that test, so this seems like a safe change.

PR #2062, opened 2017-05-08
add FlxPoint.vectorTo() and FlxPoint.hypot()

0xFireball:
A possibly useful complement to `distanceTo()` in case they want the actual vector between two points.

`hypot()` calculates the hypotenuse of the right triangle formed by the components of the point.


0xFireball:
hold on let me clean up my repo

PR #2063, opened 2017-05-08
add hypot() and vectorTo()

0xFireball:
`vectorTo()`, a possibly useful complement to `distanceTo()` in case they want the actual vector between two points.

`hypot()` calculates the hypotenuse of the right triangle formed by the components of the point.

### DIFF
--- a/flixel/math/FlxPoint.hx
+++ b/flixel/math/FlxPoint.hx
@@ -293,6 +293,20 @@ class FlxPoint implements IFlxPooled
 		point.putWeak();
 		return FlxMath.vectorLength(dx, dy);
 	}
+
+	/**
+	 * Get the vector between this point and another point.
+	 * 
+	 * @param 	AnotherPoint	A FlxPoint object to calculate a vector to.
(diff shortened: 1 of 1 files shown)


flixel/math/FlxPoint.hx
```diff
@@ -396,6 +410,16 @@ class FlxPoint implements IFlxPooled
 	{
 		return FlxVector.get(x, y);
 	}
+
+	/**
+	 * Function to get the hypotenuse of the right triangle 
+	 * formed by the components of this `FlxPoint`
+	 * @since 4.3.0
+	 */
+	public inline function hypot():Float
```
Gama11: This seems redundant, it's the same as `FlxVector#length`.

Gama11: (it doesn't help that `FlxVector` extends `FlxPoint`, so it would inherit this method)

0xFireball: Yeah, I agree with you. You could just call `toVector().length()` instead. I'll close this.

PR #2064, opened 2017-05-08
FlxGamepad: add getAnalogAxes()

0xFireball:
add `getAnalogAxes` to `FlxGamepad`, which will return a `FlxVector` containing the X and Y axis of that input ID.


Gama11:
Seems useful enough. Thanks!

PR #2065, opened 2017-05-11
[TravisCI] use Ubuntu Trusty

andyli:
https://github.com/HaxeFoundation/haxe/pull/6253#issuecomment-300717995

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ env:
   - TARGET=html5
 
 sudo: false
+dist: trusty
 
 addons:
   apt:


Gama11:
Looks like that works. :) But jobs also seem to run much slower..

Gama11:
Yay, Flixel is still compatible with latest Haxe dev. :)

However, I think I'll replace Haxe dev in Flixel's Travis config with 3.4.2 for now until the static linking situation is sorted out (Travis builds with Trusty seem to take almost twice as long.. :/).

andyli:
No problem :)
Just to note that you will want to switch to trusty soon since precise reached its EOL already. No idea why trusty is slower though.

PR #2067, opened 2017-05-13
FlxAnalog: fix mouse input without FLX_NO_TOUCH

adrianulima:
Currently the default `project.xml` disables `touch` only for `desktop` platforms. Platforms like `flash` and `html5` still have both `touch` and `mouse` enabled.
https://github.com/HaxeFlixel/flixel-templates/blob/dev/default/Project.xml.tpl#L61

We could make the `project.xml` defaults to `<haxedef name="FLX_NO_TOUCH" unless="mobile" />`. But we'd still may want to test our project with mouse and touch enabled, so I changed this file to accept mouse event when touch is enabled.

### DIFF
--- a/flixel/ui/FlxAnalog.hx
+++ b/flixel/ui/FlxAnalog.hx
@@ -237,7 +237,8 @@ class FlxAnalog extends FlxSpriteGroup
 				break;
 			}
 		}
-		#elseif !FLX_NO_MOUSE
+		#end
+		#if !FLX_NO_MOUSE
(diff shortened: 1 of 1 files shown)


Gama11:
Thanks!

Gama11:
Meh, I just remembered that there was a reason for this... https://github.com/HaxeFlixel/flixel/commit/e573dbb4674591e91770ad349a898b8cc59401c9

Gama11:
Ok, should be ok now with 69ee4a1.

PR #2069, opened 2017-05-14
FlxFilterFrames: keep FlxSprite's graphic, fixes #2050

### LINKED ISSUE #2050: Invalid Bitmap Error when removing glow Filter from a FlxSprite
- **Haxe version:** 3.3.0-RC1
- **Flixel version:** 4.2.0
- **OpenFL version:** 3.6.1
- **Lime version:** 2.9.1
- **Affected targets:** Flash

```haxe
package;

import flixel.FlxG;
import flixel.FlxSprite;
import flixel.FlxState;
import flixel.graphics.FlxGraphic;
import flixel.system.FlxAssets.GraphicLogo;
import flixel.graphics.frames.FlxFilterFrames;
import flash.filters.BitmapFilter;
import flash.filters.GlowFilter;


/**
 * Test project for button glow filter.
 * Press "G" on the keyboard to toggle the glow over the flixel icon.
 * Removing the glow without the presence of the buffer sprite causes an Invalid Bitmap Error.
 * Written by sano98.
 * 
 */

 
{
	/**
	 * The test sprite. Press "G" to give it a glow, and again to remove it.
	 */
	public var testIcon:FlxSprite;
	
	
	/**
	 * The "buffer sprite". It's presence keeps the bug from happening.
	 */
	public var testIcon2:FlxSprite;
	
	/**
	 * Glow Filter for the flixel icon
	 */
	public var glowFilter:GlowFilter;
	public var filterFrame:FlxFilterFrames;
	
	
	
	
	{
		FlxG.worldBounds.set( 0, 0, 640, 480); 
		FlxG.camera.setScrollBoundsRect(0, 0, 640, 480, false);
		
		
		super.create();
		
		this.bgColor = 0xFF77AAFF;
		
		this.testIcon = new FlxSprite(300, 300, FlxGraphic.fromClass(GraphicLogo));
		this.add(this.testIcon);
		
		
		//Comment this line out to reproduce the bug.
		this.testIcon2 = new FlxSprite(300, 300, FlxGraphic.fromClass(GraphicLogo));
		
		
		this.glowFilter = new GlowFilter(0xFFFFFFFF, 0.8, 5, 5, 3);
		this.filterFrame = this.createFilterFrames(this.testIcon);
		

		
	}
	
	
	private function createFilterFrames(sprite:FlxSprite):FlxFilterFrames
	{ 
		var filterFrames = FlxFilterFrames.fromFrames(sprite.frames, 5, 5, []); 
		return filterFrames; 
	} 
	
	
	function updateFilter(spr:FlxSprite, sprFilter:FlxFilterFrames)
	{
		sprFilter.applyToSprite(spr, false, true);
	}
	

	
	{
		super.update(elapsed);
		
		
		
		if (FlxG.keys.justPressed.G)
		{
			if (this.filterFrame.filters.length > 0)
			{
				this.filterFrame.clearFilters();
				this.updateFilter(testIcon, this.filterFrame);
			}
			else
			{
				this.filterFrame.addFilter(glowFilter);
				this.updateFilter(testIcon, this.filterFrame);
			}
		}
	}
}

```

**Observed behavior:**
When the testIcon2 is not initialised, the code crashs with an invalid bitmap error when pressing the "G"-Button on the keyboard for the second time.

**Expected behavior:**
Pressing G should simply toggle the glow.

JoeCreates:
I've no idea if this is actually sensible. @Beeblerox ?

FlxSprite#graphic persists after FlxFilterFrames#applyToSprite.

### DIFF
--- a/flixel/graphics/frames/FlxFilterFrames.hx
+++ b/flixel/graphics/frames/FlxFilterFrames.hx
@@ -90,7 +90,14 @@ class FlxFilterFrames extends FlxFramesCollection
 		
 		var w:Float = spr.width;
 		var h:Float = spr.height;
+		
+		// FlxSprite.setFrames(frames) sets the graphic to the parent of "frames"
+		// To keep the original graphic on the sprite we temporarily use the sprite's graphic as parent
+		var oldParent:FlxGraphic = parent;
+		parent = spr.graphic;
 		spr.setFrames(this, saveAnimations);
+		parent = oldParent;
+		
 		spr.offset.set(0.5 * widthInc, 0.5 * heightInc);
 		spr.setSize(w, h);
 	}


Gama11:
It's not clear whether this is the correct fix, so I'll close for now. #2050 is still open.

JoeCreates:
@Gama11 I'm convinced that the sprite _should_ keep its graphic if you apply a filter, so something to achieve this is necessary. My solution here is _slightly_ hacky, but only because the method of setting a sprites graphic is a bit messed up in the first place (and I think would be a major refactor to fix), because setting the parent of a sprite is done in the setter of frames. You have no way to handle special cases like this where the new frames (e.g. those of a filter) have no parent. My solution in this PR was to give the filter frames a parent (the graphic the filter frames were made out of), which ultimately maintains the parent on the sprite which you apply the filter to.

I think this is the best solution. The alternative I can think of would be removing the frames setter and make it a method with an optional parameter to keep the existing parent.

PR #2070, opened 2017-05-18
FlxMouse: add getters for justPressedTimeInTicks

adrianulima:
Added getters for times in ticks of mouse button inputs, also made comments more consistent.

This makes `FlxMouseButton#justPressedTimeInTicks` accessible on `FlxMouse`, it is useful on cases we want to know how long a button was pressed.

```haxe
if (FlxG.mouse.justReleased)
{
	var timePressed = FlxG.game.ticks - FlxG.mouse.timeJustPressed;
	trace(timePressed < 500 ? "Mouse click" : "Mouse long press");
}
```

### DIFF
--- a/flixel/input/mouse/FlxMouse.hx
+++ b/flixel/input/mouse/FlxMouse.hx
@@ -59,21 +59,26 @@ class FlxMouse extends FlxPointer implements IFlxInputManager
 	 */
 	public var useSystemCursor(default, set):Bool = false;
 	/**
-	 * If the left mouse button is currently pressed.
+	 * Check to see if the left mouse button is currently pressed.
 	 */
 	public var pressed(get, never):Bool;
 	/**
-	 * Check to see if the mouse was just pressed.
+	 * Check to see if the left mouse button has just been pressed.
 	 */
 	public var justPressed(get, never):Bool;
 	/**
-	 * Check to see if the mouse was just released.
+	 * Check to see if the left mouse button has just been released.
 	 */
 	public var justReleased(get, never):Bool;
+	/**
+	 * Time in ticks of last left mouse button press.
+	 * @since 4.3.0
+	 */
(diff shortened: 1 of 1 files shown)


Gama11:
This property is already exposed for `FlxTouch` - the naming should be consistent:

https://github.com/HaxeFlixel/flixel/blob/fd190e807371da332fa1f1ab144236f119500cd2/flixel/input/touch/FlxTouch.hx#L35

Also, I just noticed we use Floats for ticks in `FlxTouch` / `FlxSwipe` / `FlxMouseButton`, but `ticks` is an `Int`...  I can't think of a good reason for that, it probably results in a loss of precision.

Gama11:
Opened a separate issue for the Float / Int thing: #2071

adrianulima:
@Gama11 I saw it is `public` for `FlxTouch` and it's equivalent for mouse is `FlxMouseButton` that is already `public` and consistent. I thought that `FlxMouse` was more like `FlxTouchManager`, that don't have this field, but allows user to access directly `FlxTouch` objects.

So since we can't have access to `FlxMouseButton` in `FlxMouse`, I added it there.
The reason to have a different name was on purpose. I thought in a different start of the property to avoid code completion showing them first when user type "presse..." or "justPresse..." and avoid confusion with the other `bool` properties.

But I can change it to `justPressedTimeInTicks` to keep consistency even at `FlxMouse` level, which names do you suggest for `Right` and `Middle` ones?

Gama11:
`justPressedTimeInTicksMiddle` and `justPressedTimeInTicksRight` seems like the most consistent naming.

I'm actually not a huge fan of how this API works, I'd prefer if you did have to access the values through the mouse button instances:

```haxe
FlxG.mouse.left.pressed;
FlxG.mouse.right.justPressed;
// etc
```

But I doubt we'll change that anytime soon, seems unnecessarily breaking.

PR #2073, opened 2017-05-24
FlxAnalog fixes and improvements

adrianulima:
1. **[feature]** Added `?BaseGraphic` and `?ThumbGraphic` to constructor, this way we can create a `FlxAnalog` with custom graphics without need to create an extending class to override methods.
_(Not a breaking change)_

2. **[fix]** Changed how `ease` works to keep consistency across frame rates, it now works like `FlxCamera#followLerp`.
_(It works similar, not a breaking change I believe)_

3. **[fix]** Fixed constructor's `Radius` that was never being used, now it checks if `_radius == 0` before override it.
_(Not a breaking change)_

4. **[performance]** Now I use the already calculated angle stored in `_direction` to return `getAngle()`, avoid calling `Math.atan2()` twice.
_(I set `_direction = 0` only when the thumb backs to center, that means the `getAngle()` won't return `0` until then, differently the way it worked before that at `JustReleased` was setting acceleration to `0` and instantly `getAngle()` was returning `0`, would this be a breaking change?)_

5. **[fix]** Removed `_radius` from output's `acceleration` calc, it nows returns a normalized value. This is useful in case we want to multiply this value by a custom amount like `charSpeed`, etc.
_(it was based on the `radius` that is a screen in pixels value resulting in an strange acceleration range, this is kind of a breaking change)_

### DIFF
--- a/flixel/ui/FlxAnalog.hx
+++ b/flixel/ui/FlxAnalog.hx
@@ -90,13 +90,14 @@ class FlxAnalog extends FlxSpriteGroup
 	/**
 	 * Create a virtual thumbstick - useful for input on mobile devices.
 	 *
-	 * @param   X        The X-coordinate of the point in space.
-	 * @param   Y        The Y-coordinate of the point in space.
-	 * @param   radius   The radius where the thumb can move.
-	 *                   If `0`, half the background's width will be used as radius.
-	 * @param   ease     The duration of the easing. The value must be between `0` and `1`.
+	 * @param   X            The X-coordinate of the point in space.
+	 * @param   Y            The Y-coordinate of the point in space.
+	 * @param   radius       The radius where the thumb can move. If `0`, half the base's width will be used.
+	 * @param   ease         The duration of the easing. The value must be between `0` and `1`.
+	 * @param   BaseGraphic  The graphic you want to display as base of the joystick.
+	 * @param   ThumbGraphic The graphic you want to display as thumb of the joystick.
 	 */
-	public function new(X:Float = 0, Y:Float = 0, Radius:Float = 0, Ease:Float = 0.25)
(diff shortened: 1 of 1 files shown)


flixel/ui/FlxAnalog.hx
```diff
@@ -90,25 +91,26 @@ class FlxAnalog extends FlxSpriteGroup
...
-	 * @param   ease     The duration of the easing. The value must be between `0` and `1`.
+	 * @param   X            The X-coordinate of the point in space.
+	 * @param   Y            The Y-coordinate of the point in space.
+	 * @param   Radius       The radius where the thumb can move. If 0, half the base's width will be used.
+	 * @param   Ease         Used to smoothly back thumb to center. Must be between 0 and (FlxG.updateFrameRate / 60).
+	 * @param   BaseGraphic  The graphic you want to display as base of the joystick.
+	 * @param   ThumbGraphic The graphic you want to display as thumb of the joystick.
 	 */
-	public function new(X:Float = 0, Y:Float = 0, Radius:Float = 0, Ease:Float = 0.25)
+	public function new(X:Float = 0, Y:Float = 0, Radius:Float = 0, ?Ease:Float = 0.25, ?BaseGraphic:FlxGraphicAsset, ?ThumbGraphic:FlxGraphicAsset)
```
Gama11: I don't think `Ease` shouold have a `?`, that unnecessarily turns it into a `Null<Float>`.

Gama11:
Kind of a bad practice to do so many different things at once in a single PR, but oh well.

adrianulima:
@Gama11 I removed the `?` from `ease`.
Sorry I made different things in the same PR, next time I will make multiple PRs for different things 👍

adrianulima:
@Gama11 Is there something else I can do on this PR to be merged?

PR #2087, opened 2017-07-01
FlxMouseEventManager: add move, click and wheel events, closes #2085

### LINKED ISSUE #2085: Proper CLICK event (and more?)
`FlxMouseEventManager` handles `DOWN`, `UP`, `OVER`, and `OUT`, but not `CLICK` and its variants or `MOVE`. Should it have callbacks for these others too? `CLICK` is probably a `DOWN` and `UP` on the same object, and `MOVE` is maybe `OVER` + move.

Then what about all the variants of `MIDDLE` and `RIGHT` having callbacks? I personally think no.

Main use case is haxeui expecting `CLICK`, `MOVE`, and `WHEEL` to be handled. I've been overloading `UP`'s handler with `CLICK`, but I'm not sure how good of a solution that is in general.

MSGhero:
Closes #2085 

Added `MOVE`, `CLICK`, and `WHEEL` mouse callbacks. It works, but I haven't extensively tested it yet, which I'll be doing with HaxeUI and my game. MMB/RMB click callbacks don't seem as necessary, and I decided not to `DOUBLE_CLICK` unless someone really wants it.

Also added `justMoved` to `FlxMouse`. It seemed more appropriate there than in `FlxPointer` since pointer doesn't have a once-per-frame `update()`.

AS3's mouse events  aren't tied to the framerate in my testing, and flixel can't do anything to match that, but that's ok.

### DIFF
--- a/flixel/input/mouse/FlxMouseEventManager.hx
+++ b/flixel/input/mouse/FlxMouseEventManager.hx
@@ -195,10 +195,28 @@ class FlxMouseEventManager extends FlxBasic
 	}
 	
 	/**
+	 * Sets the mouseMove callback associated with an object.
+	 *
+	 * @param   OnMouseMove   Callback when the mouse is moved while over this object.
+	 *                        Must have Object as argument - e.g. onMouseMove(object:FlxObject).
+	 * @since 4.3.0
+	 */
+	public static function setMouseMoveCallback<T:FlxObject>(Object:T, OnMouseMove:T->Void):Void
+	{
+		var reg = getRegister(Object);
+		
+		if (reg != null)
+		{
+			reg.onMouseMove = OnMouseMove;
+		}
+	}
+	
+	/**
 	 * Sets the mouseWheel callback associated with an object.
 	 *
-	 * @param   OnMouseWheel  Callback when mouse wheel is moved while over this object.
+	 * @param   OnMouseWheel  Callback when the mouse wheel is moved while over this object.
 	 *                        Must have Object as argument - e.g. onMouseWheel(object:FlxObject).
+	 * @since 4.3.0
 	 */
 	public static function setMouseWheelCallback<T:FlxObject>(Object:T, OnMouseWheel:T->Void):Void
 	{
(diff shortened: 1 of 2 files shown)


MSGhero:
Not seeing any issues.

MSGhero:
I suddenly need double clicking in my game, so I'll be adding that.

MSGhero:
Looks like there's an issue with dev haxe in the CI builds, unrelated to flixel.

flixel/input/mouse/FlxMouseEventManager.hx
```diff
@@ -36,10 +36,20 @@ class FlxMouseEventManager extends FlxBasic
...
+	
+	private static var _mouseClickedTime:Int = -1;
 
 	private static var _point:FlxPoint = FlxPoint.get();
 	
 	/**
+	 * The maximum amount of time between two clicks that is considered a double click, in milliseconds.
+	 * @since 4.4.0
+	 */
+	public static var MAX_DOUBLE_CLICK_DELAY:Int = 500;
```
Gama11: Hm.. the `UPPER_CASE` naming implies to me that this is a constant (declared `inline` for instance) that can't be changed.

Gama11:
Presumably this can be tested using a HaxeUI demo?

MSGhero:
Yeah, but I'm (still) waiting for them to resolve an incompatibility. Right now you have to comment out a line in the core UI library for it to work with flixel.

MSGhero:
Last time I added an external lib...
image
image
Seems like it's easier to do from demos at least. I'll make one and leave it WIP so I don't forget, but it won't work or pass CI until haxeui/haxeui-core#140 is resolved.

Gama11:
Hm? You want to add haxeui to Flixel's CI?

MSGhero:
Don't I need to add it to flixel-demo's CI so the UI demo can be part of the check?

Gama11:
I'm not sure what you're trying to do to begin with. :) What UI demo?

MSGhero:
> Presumably this can be tested using a HaxeUI demo?

Oh, did you not mean a flixel-demos demo?

Gama11:
No, that's why I said HaxeUI demo. :D

MSGhero:
Lol... I tested this extensively just with flixel, and I've been using it with HaxeUI in my game without any problems. The UI lib isn't developed enough where I have demos and CI yet, but I could add one in the future.

Gama11:
But HaxeUI has demos, right? Stuff like http://haxeui.org:1234/demos/explorer/. Does that just not work with haxeui-flixel yet?

MSGhero:
All of the applicable parts of that demo (ie not the native UI elements) work fine from the flixel backend. 

I meant flixel-specific demos, like setting up an atlas to get sprites from, which is something none of the other backends do.

Gama11:
Wow, really? I didn't expect it to be this far along, I'll have to check that out.

MSGhero:
If you do try it out, just note that I have click, move, and wheel events disabled, so you may have weird behavior if the component is expecting one of those.

MSGhero:
Fixed an edge case with hitting breakpoints or otherwise tabbing out of the window while the mouse was down.

Gama11:
Thanks, these are very useful additions. :)

PR #2090, opened 2017-07-14
FlxFlicker: clarify docs of duration

ashes999:
I like clarity.


MSGhero:
I like it. I always have to go back into the source to double check.

Gama11:
Nice, but there's at least two more places where it needs to be added.

Gama11:
Ok, I think I got all of them. Thanks!

ashes999:
@Gama11 thanks for that. I didn't know about the other two locations.

PR #2093, opened 2017-07-14
Add better support for FlxButton scaling, fixes #2047

### LINKED ISSUE #2047: Can't Change FlxButton Style
- **Haxe version:** 3.4.2
- **Flixel version:** 4.2.0
- **OpenFL version:** 3.6.1
- **Lime version:** 2.9.1
- **Affected targets:** JS

```haxe
package;

import flixel.FlxState;

{
	{
		var newGameButton = new FlxButton(0, 0, "Label", function() { });
		// Scale up and set font size appropriately
		 newGameButton.scale.x = newGameButton.scale.y = 3;
		newGameButton.label.setFormat(null, 32);
		newGameButton.label.width *= 3;  
		this.add(newGameButton);
	}
}
```

**Observed behavior:**

**Expected behavior:** Non-truncated text

How do I scale up a button, without having to jump through hoops like this?

gcardozo123:
fixes #2047

### DIFF
--- a/flixel/ui/FlxButton.hx
+++ b/flixel/ui/FlxButton.hx
@@ -60,6 +60,31 @@ class FlxButton extends FlxTypedButton<FlxText>
 		
 		initLabel(Text);
 	}
+    
+    /**
+	 * Updates the FlxButton's hitbox (`width`, `height`, `offset`) according to the current `scale`.
+	 * Also calls `centerOrigin()` and `updateLabelFormat()`
+	 */
+    override public function updateHitbox():Void
+    {
+        super.updateHitbox();
+        updateLabelFormat();
+    }
+    
+    /**
+     * Updates the label format.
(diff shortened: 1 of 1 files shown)


gcardozo123:
codeclimate is complaining that I'm using spaces for indentation, if you need me to change that (or anything else) let me know, please.

MSGhero:
It's easiest if you fix them all and anything else codeclimate complains about.

gcardozo123:
@MSGhero now everything is fine with both Travis-ci and Codeclimate. ;)

Gama11:
This causes issues with `systemFont` - it resets the font back to the default Flixel font:

 

Also, I'm not sure if it's a good idea to tie this to `updateHitbox()`, which you'd rarely think to call on a `FlxButton` (since you usually don't need hitboxes / collisions for UI).

PR #2068, opened 2017-05-13
[WIP] OpenFL 7 compatibility

Beeblerox:
Note: not mergable yet, this is just for feedback / discussion.

PR #2099, opened 2017-07-22
Initial support for OpenFL 6.0 (using Tilemap)

jgranick:
Hello!

I know there is a cool branch in-the-works that uses an OpenGL renderer for Flixel. For sake of compatibility for OpenFL and Lime in the short-term, but also as an alternative approach to the future, I would like to offer this change which has 1.) general changes to support current versions of OpenFL, as well as 2.) a change to `FlxTilesheet` which maps `drawTiles` to `Tilemap`

Testing on the Mode platformer demo, everything seems to work properly, but sometimes enemy jets disappear. I am not fully sure why.

Although it largely works, there are some issues with the life-cycle of adding, removing or ordering `Tilemap` objects, which you guys can probably sort out fairly easily. I wanted to get the brunt of this working.

This consumes a new beta `TileArray` API, available in the current OpenFL "develop" branch. Happy to discuss this more on how we can improve.

### DIFF
--- a/flixel/FlxCamera.hx
+++ b/flixel/FlxCamera.hx
@@ -1481,7 +1481,7 @@ class FlxCamera extends FlxBasic
 		}
 		else
 		{
-			#if openfl_legacy // can't skip this on next, see #1793
+			#if (openfl_legacy || (openfl >= "4.0.0")) // can't skip this on next, see #1793
 			if (FxAlpha == 0)
 				return;
 			#end
--- a/flixel/effects/postprocess/PostProcess.hx
+++ b/flixel/effects/postprocess/PostProcess.hx
@@ -6,12 +6,30 @@ import flash.geom.Rectangle;
 import flixel.FlxG;
 import openfl.Assets;
 import openfl.display.OpenGLView;
+
+#if openfl_legacy
 import openfl.gl.GL;
 import openfl.gl.GLFramebuffer;
 import openfl.gl.GLRenderbuffer;
 import openfl.gl.GLTexture;
 import openfl.gl.GLBuffer;
 import openfl.utils.Float32Array;
+#else
+import lime.graphics.opengl.GL;
+import lime.graphics.opengl.GLFramebuffer;
+import lime.graphics.opengl.GLRenderbuffer;
+import lime.graphics.opengl.GLTexture;
+import lime.graphics.opengl.GLBuffer;
+import lime.utils.Float32Array;
+#end
+
+#if (lime >= "4.0.0")
+import lime.graphics.opengl.WebGLContext;
+#end
+
+#if (openfl >= "5.0.0")
+import openfl.utils.AssetType;
+#end
 
 private class Uniform
 {
@@ -66,7 +84,11 @@ class PostProcess extends OpenGLView
 
 		buffer = GL.createBuffer();
 		GL.bindBuffer(GL.ARRAY_BUFFER, buffer);
+		#if (lime >= "4.0.0")
+		(GL:WebGLContext).bufferData(GL.ARRAY_BUFFER, new Float32Array(vertices), GL.STATIC_DRAW);
+		#else
 		GL.bufferData(GL.ARRAY_BUFFER, new Float32Array(#if !openfl_next cast #end vertices), GL.STATIC_DRAW);
+		#end
 		GL.bindBuffer(GL.ARRAY_BUFFER, null);
 
 		postProcessShader = new Shader([
--- a/flixel/effects/postprocess/PostProcess.hx
+++ b/flixel/effects/postprocess/PostProcess.hx
@@ -180,7 +180,7 @@ class PostProcess extends OpenGLView
 		texture = GL.createTexture();
 		
 		GL.bindTexture(GL.TEXTURE_2D, texture);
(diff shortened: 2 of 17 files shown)


MSGhero:
I'm a bit of a rendering noob, so is this basically GPU `copyPixels()` and `draw()`? What are the pros and cons of this versus the OpenGL PR?

starry-abyss:
At least the good thing is that it works with unpatched addons (they are broken in the official PR at the moment)

jgranick:
@MSGhero This is a close to a full replacement of the old `drawTiles` API, requiring minimal changes to Flixel.

This relies on little custom code, so it should remain stable for future versions of OpenFL.

The OpenGL PR is a much broader change. As Flixel supports both BLITTING and DRAW_TILES render methods, having both might be nice in the future.

Tiago-Ling:
So, would this PR work with this one: https://github.com/HaxeFlixel/flixel/pull/2068#issuecomment-316629623 ?

Or, for one to be accepted the other would have to be discard / altered?

jgranick:
@Tiago-Ling This is completely independent of #2068. I could imagine Flixel supporting both render approaches, but I don't imagine much in the way of conflicts unless #2068 eliminates the DRAW_TILES rendering path

Beeblerox:
this is great PR, because rendering become again openfl responsibility!
but it misses one feature: hardware accelerated drawTriangles support. 
flixel supports mixed drawTiles and drawTriangles calls.

MSGhero:
My main lack of understanding: how would you choose between the drawTiles and OpenGL render paths? Is the OpenGL path solely for `RENDER_BLIT` mode as this PR is for `DRAW_TILES`?

Beeblerox:
it's done here: https://github.com/Beeblerox/flixel/blob/FlxMaterial-and-camera-scaling/flixel/FlxCamera.hx#L394-L407 and here: https://github.com/Beeblerox/flixel/blob/FlxMaterial-and-camera-scaling/flixel/system/macros/FlxDefines.hx#L151-L162
as you can see it just checks openfl version

MSGhero:
So then with these two PRs, we'd have `RENDER_BLIT`, `RENDER_TILES`, and `RENDER_GL` FlxDefines?

jgranick:
@Beeblerox Where does Flixel use tiles and triangles together? I misunderstood and thought that triangles _or_ tiles was the design.

Hardware-accelerated triangles is on our `TODO` list.

This isn't intended to be a competition, but an option.

MSGhero:
@jgranick the camera can `drawTiles` a normal sprite and also `drawTriangles` a `flixel.FlxStrip` when in `RENDER_TILE` mode. Both are rendered to the camera's `canvas`, which is a flash display Sprite.

https://github.com/HaxeFlixel/flixel/blob/9b20e4ffa4ca08d472f79b63d6043ee5573ab0ea/flixel/FlxCamera.hx#L388-L394

Beeblerox:
@jgranick yes, @MSGhero is right

MSGhero:
@Gama11 @Beeblerox what should be the procedure for merging these two PRs?

Gama11:
I'm a bit concerned about performance - on Neko, I only reach about 20 fps in Mode with this (regardless of debug or release mode). Cpp performance is also worse (although FlxBunnyMark is needed there to notice the difference).

larsiusprime:
We had this issue when we first did the transition from legacy->Next, too. Profile time?

MSGhero:
Do the `TileArray` API changes affect this PR at all? @jgranick

Gama11:
This is definitely an interesting approach, but unfortunately has a few issues:

- The mentioned bugs with sprites disappearing.
- No support `drawTriangles()` as pointed out by @Beeblerox.
- There's some seemingly unrelated gamepad and preloader changes here (?).
- Architecture-wise, it's not very elegant - you implemented an adapter for `drawTiles()`, which effectively means that Flixel's renderer produces drawTiles data, and then that adapter then reverse-engineers that to fit into OpenFL's Tilemap API. While this works and probably made sense as a proof-of-concept because it is the least invasive approach, a proper implementation would have Flixel's renderer use the `Tilemap` API directly. It also doesn't seem ideal performance-wise.
A mergeable implementation would basically require this PR to be rewritten from scratch, and some abstraction layers to be introduced.

Because of this, currently it looks like the way to go is to continue with #2068 (which should already be compatible with OpenFL 6).  This doesn't mean that Flixel will definitely never use `Tilemap`. @Beeblerox' branch has introduced an abstraction layer for rendering called `FlxCameraView`. That could enable some experimentation with `Tilemap` in the future.

jgranick:
It may be wise to support a minimal lifeboat for letting users build with a newer version of OpenFL.

 * The occasional disappearing sprite is likely solvable
 * `drawTriangles` could be supported, but affects few users, I think?
 * The gamepad/preloader changes were to support a newer OpenFL version
 * This is meant to be a minimal solution, to help inspire more formal changes

The legacy version of OpenFL used a C++ library for the core renderer. This means that Neko runs faster in OpenFL 2, because less of the code executes in the Neko VM. We plan on investigating HL in the future, or you can use HTML5 instead of faster performance.

Gama11:
>It may be wise to support a minimal lifeboat for letting users build with a newer version of OpenFL.

Some users are already using @Beeblerox' WIP branch, I believe.

jgranick:
I heard there were a lot of regressions, but I bet @Beeblerox is sorting it out

Issue #1942, opened 2016-09-10
Random core dumps when running cpp unit tests on Travis


Gama11:
See https://travis-ci.org/HaxeFlixel/flixel/jobs/159004450#L1484-L1486 for example:

```
pure virtual method called
terminate called without an active exception
Aborted (core dumped)
```

This happens every so often during the cpp unit tests (both for Haxe dev and Haxe 3.2.0). Not sure why it happens... Has anyone seen this issue before? @ibilon maybe?

Gama11:
Haven't seen this one in a while, so I'll just close this.

Gama11:
Hm, since we switched to using Ubuntu Trusty, this happens much more frequently, so we can't just ignore it this time. The error message changed slightly,:

>Segmentation fault (core dumped)

(https://travis-ci.org/HaxeFlixel/flixel/jobs/266555890#L1947)

@jgranick Have you ever seen this by chance (since you were running munit tests on OpenFL's Travis for a while)?

Gama11:
Again, this hasn't come up in a while... Might still be an issue, but I'm not sure we can do anything about it.

PR #2103, opened 2017-08-25
FlxMouseEventManager: fire mouseOut before mouseOver

### LINKED ISSUE #2101: MouseOut should occur before MouseOver
For parity with OpenFL, `FlxMouseEventManager` should handle mouse out before mouse over.

Sample use case: mousing over an object causes text to appear, and mousing out causes it to disappear. If you have two objects next to each other and mouse out of one and over the other, the text will disappear overall. The intended behavior is that the text should disappear and reappear within the same frame.

This is technically a breaking change.

MSGhero:
Closes #2101 

Technically breaking, but this is the correct order of events according to Flash, etc.

### DIFF
--- a/flixel/input/mouse/FlxMouseEventManager.hx
+++ b/flixel/input/mouse/FlxMouseEventManager.hx
@@ -357,18 +357,6 @@ class FlxMouseEventManager extends FlxBasic
 			}
 		}
 		
-		// MouseOver - Look for new objects with mouse over.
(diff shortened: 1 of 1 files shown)


MSGhero:
I'm not sure why only haxe 3.2.0 cpp failed. None of the output is being muted on Travis, however.

Gama11:
Because of #1942.

PR #2107, opened 2017-09-07
FlxAssetPaths: fix filterExtensions for files with multiple dots

Skerper:
Currently, getFileReferences filtered extensions using the first string after a dot. This means that myfile.png.old is included even when a filter is set to only import png files. This causes various compiler errors when tortoisesvn is used, due to duplicate file.ext.svn-base files being added. 

New implementation takes the *last* string after a dot, meaning that myfile.png.old is treated as a .old file and will not be included if "old" is not on the filter list.


Gama11:
Thanks! I added a unit test for this fix here: 29a545d

PR #2108, opened 2017-09-10
Consistent param case in FlxGroup#insert

MSGhero:
Keeping up with the style guide.


Gama11:
Meh, this isn't the only place where we're not consistent.. I actually want to convert all params from `PascalCase` to `camelCase` eventually, but that's not a good idea while huge changes like #2068 are unmerged.

MSGhero:
Ok that's fine, it was just a precursor for #2111

PR #2109, opened 2017-09-10
Add all gamepad buttons to list

MSGhero:
Seems like an oversight that they weren't already there. I tested all of the ones that a 360 controller offers. (DPAD and the ANALOG_STICKs don't do anything?)

I'm not 100% sure what FLX_JOYSTICK_API is, so let me know if I need to #ifdef that anywhere.

### DIFF
--- a/flixel/input/gamepad/lists/FlxGamepadButtonList.hx
+++ b/flixel/input/gamepad/lists/FlxGamepadButtonList.hx
@@ -9,32 +9,49 @@ import flixel.input.gamepad.FlxGamepadInputID;
  */
 class FlxGamepadButtonList extends FlxBaseGamepadList
 {
-	public var A                (get, never):Bool; inline function get_A()                 return check(FlxGamepadInputID.A);
-	public var B                (get, never):Bool; inline function get_B()                 return check(FlxGamepadInputID.B);
-	public var X                (get, never):Bool; inline function get_X()                 return check(FlxGamepadInputID.X);
-	public var Y                (get, never):Bool; inline function get_Y()                 return check(FlxGamepadInputID.Y);
-	public var LEFT_SHOULDER    (get, never):Bool; inline function get_LEFT_SHOULDER()     return check(FlxGamepadInputID.LEFT_SHOULDER);
-	public var RIGHT_SHOULDER   (get, never):Bool; inline function get_RIGHT_SHOULDER()    return check(FlxGamepadInputID.RIGHT_SHOULDER);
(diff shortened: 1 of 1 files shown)


Gama11:
This indeed looks like an oversight to me... Or is there more to it, @larsiusprime?

The diff is easier to look at when whitespace changes are ignored: https://github.com/HaxeFlixel/flixel/pull/2109/files?w=1

MSGhero:
Anything hanging this one up?

larsiusprime:
Hey there! Probably waiting on me to comment (From a month ago.... sorry).

So, couple a things:
1) FLX_JOYSTICK_API is for legacy mode, as best I recall. 
2) From the looks of it, all of the stuff you're adding seem like "virtual" inputs that don't directly correspond to an actual real physical input on a controller. IE, LEFT_TRIGGER_BUTTON is for simulating a gamecube style digital click at the end of an analog trigger "full pull", the tilt stuff is highly experimental and basically doesn't work in anything real outside of some really specific drivers for specific hardware that basically nobody uses, and the LEFT_STICK_DIGITAL_UP/LEFT/RIGHT/DOWN are also simulated "digital" inputs digitized from the movement of the analog stick. I'm not saying that's a justification for them being left out, it's just the clear pattern they all seem to have in common. (With the exception of the analog sticks themselves).

I would have to dive back into this code to figure out why these weren't all added to the list.

What might help get my head back in the right frame of the gamepad api is --- what does adding these into the list let you do as a flixel user?

MSGhero:
With remappable controls, letting the user map anything to anything pretty much, it's hard to deal with analog-as-a-boolean. Like, how do you generalize `LEFT_STICK_X < 0` so it can be used for rightward movement?

For the user, it's possible but annoying. `DIGITAL_X` makes it super easy. I can post a code example of how I use it if that wasn't clear.

larsiusprime:
Yeah, if there's a concrete use case that adding some of these affords, then I say go ahead and add em. Show me some code and we'll see if there's any of these that have a good reason not for being there?

MSGhero:
https://gist.github.com/MSGhero/4fb4bfcef15ffaf403ee3ac98c91aa1c#file-inputexcerpt-hx-L23-L26

I manually handle `justPressed` and `justReleased` for the case of a user holding `A` and `LEFT_STICK_LEFT` at the same time and only releasing one.

MSGhero:
@larsiusprime bump

Gama11:
This seems like a pretty trivial change, so I'll just go ahead and merge it.

PR #2110, opened 2017-09-10
FlxMouseEventManager: correct the mouseChildren parameter, closes #2102

### LINKED ISSUE #2102: MouseChildren inconsistent
The `mouseChildren` property within `FlxMouseEventManager` says that it disallows overlapping objects from receiving mouse events. However, this is only true when those overlapping objects were added to `FlxMEM` before the object with `mouseChildren = false`

Example: a sprite group is created and added, with `mouseChildren = false`. Its children are created and added. Clicking a child will result in both the parent and child receiving callbacks.

Possible solution: this behavior happens due to the `unshift()` within `add()`, but changing it to `push()` wouldn't fix the underlying issue. Perhaps sorting `_registeredObjects`, with the `mouseChildren = false` ones coming first, would work. `unshift` if it's false, `push` if it's true, and sort/rearrange if `setObjectMouseChildren` is called.

When two overlapping objects have `mouseChildren` disabled, it's unclear what should happen.

MSGhero:
Closes #2102 

Fixes a bug with `mouseChildren` in `FlxMouseEventManager` where items with the parameter set to false still allowed overlapping objects to receive events. E.g. the children of a parent sprite group with `mouseChildren = false` still received events if they were added to `FlxMEM` after.

~I forgot I made this change~ I've been using this change since the issue and haven't had any problems. It does break the order of the objects if that's a concern, which I could change but would be more involved.


Gama11:
I would also like to have a unit test for this, if possible.

MSGhero:
`reorder` also needs to be changed since it disrupts the add order.

I'd like for sprite groups with `mouseChildren = false` to block their children's potential `mouseChildren = false`, drawing from how Flash does it. So this needs a bit more imo.

MSGhero:
Essentially, the `mouseChildren false` and `true` objects should be in two separate arrays, or one array where the boundary between the two halves is managed. Adding an object or changing the `mouseChildren` param should place that object at the front of its proper subarray, which preserves the user's order.

@Gama11 my *breaking* question is if `Flx(Sprite)Group`s should block their children from receiving events like AS3's `mouseChildren`, which is not the case currently. That would have to go into `reorder` because I'm not really sure how to accomplish that in `add` without just calling `reorder`.

MSGhero:
Will unit test, but the most recent commit keeps `mouseChildren = true` consistent with `= false` in terms of the add order.

Gama11:
What's the status here? Is that `Flx(Sprite)Group` question still relevant?

MSGhero:
Ehhhhh if you want it to be. Finding some time this and next weekend to code.

MSGhero:
Ok

tests/unit/src/flixel/input/mouse/FlxMouseEventManagerTest.hx
```diff
@@ -0,0 +1,54 @@
...
+		FlxMouseEventManager.add(sprite0, null, null, null, null, true);
+		FlxMouseEventManager.add(sprite1, null, null, null, null, false);
+		FlxMouseEventManager.add(sprite2, null, null, null, null, true);
+		FlxMouseEventManager.add(sprite3, null, null, null, null, false);
+		
+		// mouseChildren false comes before mouseChildren true, and more recently added comes before existing
+		var sprites = [sprite3, sprite1, sprite2, sprite0];
+		for (i in 0...sprites.length)
+		{
+			Assert.areEqual(sprites[i], FlxMouseEventManager._registeredObjects[i].object);
```
Gama11: This is really hacky, a unit test shouldn't need to rely on `@:access()`, only public API.

>the parameter set to false still allowed overlapping objects to receive events.

You should be able to test for that not happening anymore?

Gama11: (you should be able to dispatch mouse events manually, similar to what we do here:
https://github.com/HaxeFlixel/flixel/blob/dev/tests/unit/src/flixel/input/gamepad/FlxGamepadManagerTest.hx)

MSGhero: Ohhh I didn't think of that. That's a much better test, yeah.

MSGhero:
Ok, again.

Gama11:
Nice, that's much cleaner. 👍

Issue #2104, opened 2017-08-30
Parent reference + ADDED_TO_STATE


MSGhero:
Discussion.

I'm really needing an `ADDED_TO_STAGE` parallel for HaxeUI. Like an `addedToState` `FlxSignal` or something. I don't see a way of implementing that without giving sprites a `parent` reference.

### Implementation:
`private var parent(default, set):`
1. `FlxSpriteGroup` in FlxSprite.hx
2. `FlxGroup` in FlxBasic.hx
3. `FlxBasic` in FlxBasic.hx
4. `IFlxGroup` or abstract multitype `FlxParent` in FlxSprite.hx

### Notes
1. I don't like the idea of restricting `parent` to FSG only.
2. If the intended parent is a FSG, `parent` would actually be `fsg.group`.
3. I don't see a good reason to use the common superclass `FlxBasic`
4. ???

### Implications:
- Local/global relationship of child properties changes (solves #1353, adds breaking changes)
- A sprite can only have one parent, which is breaking
- `push`ing directly to the `members` array would not set the `parent` reference
- Acknowledging `FlxSpriteGroup` as something other than a hack 😜 

### My conclusions
For `ADDED_TO_STATE`, I don't really care what the type of `parent` is, as long as `this.parent.parent.parent...` eventually gives me `null` or a `FlxState`.

Another option is to just keep the `parent` reference but not do anything with it. So the local/global relationship of FSG and children isn't touched, except maybe when changing `alpha`. This is the least breaking option.

Jarrio:
Could adding built in an internal `FlxSignal` for `addedToStage` not work for this?

MSGhero:
Omg I didn't make that clear. I want `ADDED_TO_STATE` to be a `FlxSignal` 😀 But I think the only way of telling if an object has actually been added to the state (for the event to dispatch internally) is with a parent reference.

Gama11:
I don't think parents are a thing you can really have in Flixel. You are supposed to be able to add sprites to multiple groups. Several demos use this "meta group" (groups that are never added) technique for collisions:

https://github.com/HaxeFlixel/flixel-demos/blob/7494abc20b7de1bddb1fdbeb76c8934b111e1f4e/Platformers/Mode/source/PlayState.hx#L119-L131

MSGhero:
[...]

PR #2111, opened 2017-09-10
Add/remove signals for FlxGroup

MSGhero:
I'm not sure if this c-word's #2104, but I'm not sure that issue can ever be fully resolved. The intent of it will be completed at least.

Adds `addedToGroup` and `removedFromGroup` signals in `FlxGroup`, so you can tie things into group management. Such as marking HaxeUI components as ready to be invalidated 😜.

As mentioned over there:
- Should anything be called when a group `recycle`s an object? It never left the group, but it kinda did.
- In `clear`, the callback happens before things are actually removed from the group. Not very consistent, but it avoids cloning `members`.
- Should anything happen when shrinking the group with `set_maxSize()`? Sprites aren't so much removed as deleted.

Edit: demos, yeah yeah. I want to be sure this is mergeable first.

### DIFF
--- a/flixel/group/FlxGroup.hx
+++ b/flixel/group/FlxGroup.hx
@@ -4,6 +4,7 @@ import flixel.FlxBasic;
 import flixel.FlxG;
 import flixel.group.FlxSpriteGroup.FlxTypedSpriteGroup;
 import flixel.util.FlxArrayUtil;
+import flixel.util.FlxSignal.FlxTypedSignal;
 import flixel.util.FlxSort;
 
 typedef FlxGroup = FlxTypedGroup<FlxBasic>;
@@ -75,6 +76,10 @@ class FlxTypedGroup<T:FlxBasic> extends FlxBasic
 	 * variable instead of `members.length` unless you really know what you're doing!
 	 */
 	public var length(default, null):Int = 0;
+	
+	public var addedToGroup(default, null):FlxTypedSignal<T->Void>;
+	public var removedFromGroup(default, null):FlxTypedSignal<T->Void>;
+	
 	/**
 	 * Internal helper variable for recycling objects a la `FlxEmitter`.
 	 */
@@ -91,6 +96,9 @@ class FlxTypedGroup<T:FlxBasic> extends FlxBasic
 		members = [];
 		maxSize = Std.int(Math.abs(MaxSize));
 		flixelType = GROUP;
+		
+		addedToGroup = new FlxTypedSignal<T->Void>();
+		removedFromGroup = new FlxTypedSignal<T->Void>();
 	}
 	
 	/**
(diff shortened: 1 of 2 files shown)


flixel/group/FlxGroup.hx
```diff
@@ -107,6 +115,18 @@ class FlxTypedGroup<T:FlxBasic> extends FlxBasic
 	{
 		super.destroy();
 		
+		if (addedToGroup != null)
```
Gama11: The usual `FlxDestroyUtil` one-liner should work?

flixel/group/FlxGroup.hx
```diff
@@ -91,6 +96,9 @@ class FlxTypedGroup<T:FlxBasic> extends FlxBasic
 		members = [];
 		maxSize = Std.int(Math.abs(MaxSize));
 		flixelType = GROUP;
+		
+		addedToGroup = new FlxTypedSignal<T->Void>();
```
Gama11: I think it might be a good idea to initialize these lazily, to avoid unnecessary allocations for people who don't need these (which is probably most people).

Gama11:
>Should anything be called when a group recycles an object? It never left the group, but it kinda did.

Then you'd kinda have to do that when you manually call `revive()` on an object as well, which is not really possible easily...

>In clear, the callback happens before things are actually removed from the group. Not very consistent, but it avoids cloning members.

Hm, I'm indeed not a big fan of that..

>Should anything happen when shrinking the group with set_maxSize()? Sprites aren't so much removed as deleted.

They're both removed _and_ destroyed, so I'd say yes.

>Edit: demos, yeah yeah.

Not sure if there's really much use in having a demo for this. There should be unit tests though. :P

Does this fully solve your use case? Why do you suddenly no longer need a parent reference / the "ADDED_TO_STATE" behavior?

MSGhero:
Unit test*, yeah that's what I meant. It's late.

I gave up my dream of having a one-liner HaxeUI init (which required `ADDED_TO_STATE`). The core of what I needed was a parent telling its child that it's been added. So the second line of the init is now `myGroup.addedToGroup(haxe.ui.backend.flixel.FlxUIHelper.readyUI)`.

flixel/group/FlxGroup.hx
MSGhero: I'm not seeing `destroy` in `FlxTypedSignal`'s code completion. It's defined in the real class but not the abstract.

Gama11: The abstract should forward it then.

MSGhero: It does for the compiler, but not for code completion. Adding `@:forward` to `FlxTypedSignal` makes it show up, if it's needed.

MSGhero:
Oh, what about in `destroy()`, should `removedFromGroup` dispatch there too?

Gama11:
Hm.. that usually only happens on state switches, where everything is destroyed - so you'd just end up with callbacks for a lot of objects with null'd out properties.

MSGhero:
I couldn't think of too many unit tests that weren't simply testing if `FlxSignal` worked.

MSGhero:
Anything else here @Gama11 ?

Gama11:
Seems ok to merge, but I think it might be a good to rename it to `memberAdded` and `memberRemoved`. That would follow the naming pattern we already have in `CameraFrontEnd` with the `cameraAdded` and `cameraRemoved` signals. What do you think?

MSGhero:
Oh sure, I’ll make that change.

MSGhero:
Let's see what (hopefully) unrelated error Travis comes up with this time.

Gama11:
Thanks! :)

PR #2114, opened 2017-10-21
Missing @Test in FlxGroupTest

MSGhero:
I'm not super familiar with MUnit, but this seems like an omission.


MSGhero:
The Travis results seem weird. I ran the test suite on Flash and Neko with no issues, but CI is saying there are issues with FlxPreloader.

PR #2116, opened 2017-10-27
Key up/down listeners for FlxKeyManager

Skerper:
Adds event listener style callbacks for keypresses. 

Listeners can be added via FlxG.keys.addKeyDownCallback(FlxKey.WHATEVER, myIntToVoidFunction);

Can also be removed individually via removeKeyDown/UpCallback, or cleared entirely with clearCallbacks();

I found myself needing these but didn't especially want to make a second round of listeners, especially since Flixel's keyboard manager contains various fixes for native keyboards and so-on.

### DIFF
--- a/flixel/input/FlxKeyManager.hx
+++ b/flixel/input/FlxKeyManager.hx
@@ -149,10 +149,10 @@ class FlxKeyManager<Key:Int, KeyList:FlxBaseKeyList> implements IFlxInputManager
 			}
 		}
 		#if FLX_DEBUG
-			else
-			{
-				throw 'Invalid key code: $KeyCode.';
-			}
+		else
+		{
+			throw 'Invalid key code: $KeyCode.';
+		}
 		#end
 		
 		return false;
--- a/flixel/input/FlxKeyManager.hx
+++ b/flixel/input/FlxKeyManager.hx
@@ -3,6 +3,7 @@ package flixel.input;
 import flash.events.KeyboardEvent;
 import flixel.FlxG;
 import flixel.input.FlxInput.FlxInputState;
+using flixel.util.FlxArrayUtil;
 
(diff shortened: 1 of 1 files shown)


Gama11:
What's the use case here?

Skerper:
@Gama11 The use case is situations in which you have keys which are not pressed often (pause, for example), and would structurally prefer to have event-driven listeners than to poll them every frame.

MSGhero:
It could be used with HaxeUI I guess, there are a few keyboard events I ignore currently.

I'd be curious to see if it impacts the FlxAction PR.

Gama11:
@Skerper That example doesn't make sense to me - why does it matter how _often_ the key is pressed? It's not like polling has any additional performance cost, Flixel's keyboard logic saves all key states in a map, regardless of whether they're being polled or not.

Skerper:
@Gama11 It's not a performance issue, it's more just to offer more structure options.

Gama11:
You could just split your main loop into smaller methods? :)

In my opinion, having an alternative way of doing something is not always a good thing. Right now, writing keyboard input in Flixel is straightforward, there's one way of doing it. If you introduce event listeners, users of the API start to wonder which one is better / which one they should be using.

API-design-wise, this also introduces some inconsistency. We'd have event listeners for keyboard input, but not gamepad or mouse (I know there's `FlxMouseEventManager`, but that's quite specific for listening to mouse events _on a specific sprite_ - if you want to check the mouse state "globally", you still need polling).

Skerper:
I'm guessing you still wouldn't be interested if there was equal functionality across input methods due to the clarity issues you raised?

Gama11:
Yeah, I don't really see a big benefit in having this additional API in the first place..

PR #2117, opened 2017-11-16
FlxButton: fix justPressed being true for two frames

MSGhero:
FlxButton's JUST_PRESSED lasts for two frames. The expectation is that JUST_PRESSED only lasts for one frame, but two frames of it are currently required to trigger a PRESSED. Moving the input's update fixes this.

Proxy PR, I haven't tested this at all.


Gama11:
Unit test? :)

PR #2119, opened 2017-12-06
FlxText: fix text changes removing clipRect, closes #2031

### LINKED ISSUE #2031: clipRect on FlxText
`clipRect` doesn't work as expected on `FlxText` unless `regenGraphic()` has been called since the most recent text/size change. Subsequent calls to `regenGraphic()`, e.g. changing the text, discard the clip via `set_frames()`.

So like

```haxe
var text = new FlxText();
text.text = "Test Text";
text.clipRect = FlxRect.get(10, 0, 50, 50);
```
doesn't do anything, but

```haxe
var text = new FlxText();
text.text = "Test Text";
text.width; // get_width calls regenGraphic
text.clipRect = FlxRect.get(10, 0, 50, 50);
```
works as expected, but then
```haxe
//
text.text = "this is some different text";
```
nulls `clipRect`. This part might be intended behavior, idk.

Latest dev

MSGhero:
Fixes #2031 

Rendering new text wipes out the existing clipRect, which I can't see being useful. This retains the clip.

Core issue is that FlxSprite wipes the clip after `makeGraphic()` is called, but text seems like it should be an exception.

### DIFF
--- a/flixel/text/FlxText.hx
+++ b/flixel/text/FlxText.hx
@@ -771,7 +771,13 @@ class FlxText extends FlxSprite
 			height = newHeight;
 			var key:String = FlxG.bitmap.getUniqueKey("text");
 			
+			// retain the clipRect through text changes
+			var tempRect = clipRect;
+			
 			makeGraphic(Std.int(newWidth), Std.int(newHeight), FlxColor.TRANSPARENT, false, key);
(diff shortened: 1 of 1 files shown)


Gama11:
Thanks!

PR #2120, opened 2017-12-06
Fix FlxStringUtil.formatMoney() for large values

Epono:
Using ffloor instead of flood, which resulted in loss of information on numbers larger than the max value of an int

If you want to try it, just call the function with a float > 2^31-1, it sometimes returns "0", sometimes -2^31

You can in in action in the images below :
without
Without the fix

with
With the fix

If you want to checkout the (ugly) code, it's [right here](https://github.com/TAGL-Entertainment/LD40/blob/master/source/ui/InfoScreen.hx#L118)


Gama11:
Thanks! I also added a test for this case.

Issue #1916, opened 2016-08-24
Code style change proposoal: get rid of the private keyword


Gama11:
I'd like to propose getting rid of the `private` keyword (at least in the core library) again (#853). It's superfluous, since Haxe only has `private` and `public` and fields are `private` by default.

@impaler @Beeblerox @gamedevsam Did any of your opinions on this matter change?

@larsiusprime What do you think?

Any opinions are welcome here - @HaxeFlixel/active-contributors, @HaxeFlixel/contributors, or anybody else.

The change itself is trivial, and the style is enforcable with CodeClimate / haxe-checkstyle I think.

I guess we can vote with :+1: / :-1: reactions? :)
##

Tiago-Ling:
+1 to get rid of the `private` keyword. I've been using it like this for some time and prefer it much more.

elsassph:
Death to `private` is my middle name.

sruloart:
As long as the [exceptions](https://haxe.org/manual/class-field-visibility.html) are kept, I for one welcome our new no-private overlords.

Gama11:
@sruloart I'm not sure what you mean? We can't change Haxe's default visibilities.

sruloart:
Just this (please do ignore me if that's obvious and irrelevant :| ):
>Omitting the visibility modifier usually defaults the visibility to private, but there are exceptions where it becomes public instead:
   
> If the class is declared as extern.
    If the field is declared on an interface.
    If the field overrides a public field.

MSGhero:
Externs and interfaces have to be public to make sense, and I'm pretty sure an error gets thrown if you mess with the modifiers of an override. We don't have to worry about any of that here 😄

adireddy:
👍 personally I prefer not specifying default access modifiers. `private` in classes & `public` in interfaces and externs.

alihassan0:
IMO .. that may cause some confusion to those starting to read the flixel classes. since I think most people don't know that all fields are private by default.

impaler:
I am still no particular fan of having to write private, people should be able to remember a default :)

JoeCreates:
I don't care either way.

Tiago-Ling:
Since we're in the subject, i really do like to omit `Void` returns too - what do you think about it?

Gama11:
Same, but I'm not sure we should discuss that here.

Beeblerox:
if majority agrees about getting rid of these keyword then i won't be arguing abbout it.
[...]

PR #2123, opened 2017-12-22
Removes private keyword from fields

nachoverdon:
As discussed here: https://github.com/HaxeFlixel/flixel/issues/1916

Removes every occurrence of the private keyword that is not a class, an interface, an enum or a comment, just variables and functions.

It also removes a lot of trailing whitespace from some files...

edit: We should also edit the haxe checkstyle config file so it yield when we use `private`, to avoid future additions.

### DIFF
--- a/flixel/FlxBasic.hx
+++ b/flixel/FlxBasic.hx
@@ -14,11 +14,11 @@ class FlxBasic implements IFlxDestroyable
 	 * Static counters for performance tracking.
 	 */
 	@:allow(flixel.FlxGame)
-	private static var activeCount:Int = 0;
+	static var activeCount:Int = 0;
 	@:allow(flixel.FlxGame)
-	private static var visibleCount:Int = 0;
+	static var visibleCount:Int = 0;
 	#end
-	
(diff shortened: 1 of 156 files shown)


MSGhero:
Did you use VSCode? You also deleted the tabs in otherwise empty lines.

nachoverdon:
I used Sublime Text with the TrailingSpaces plugin and I checked one by one the values before replacing them.

I didn't see anything regarding tabs/spaces on empty lines in the [code style guide](http://haxeflixel.com/documentation/code-style/), so I set the config to trim on save because why not.

I messed up by doing that? Why is it correct to have tabs on empty lines?

Gama11:
I intentionally held off doing this until #2068 is merged to avoid merge conflicts (which is mentioned in the issue, although the PR ID changed since then: https://github.com/HaxeFlixel/flixel/issues/1916#issuecomment-250899794).

Gama11:
As mentioned, this is better done separately later once the potential for merge conflicts is lower.

PR #2125, opened 2018-01-12
Add AutoTiling method with 47 tiles

lipki:
Here's the example tileset.

template_47

And a screenshot.

AUTO
truc2
ALT
truc3
FULL
truc


JoeCreates:
At last! Thanks for sorting this. :)

lipki:
I made the corrections for codeclimate.
But for Travis, it's beyond me a bit.

I think it's this mistake:
```
source/PlayState.hx:175: characters 10-28 : Unmatched patterns: FULL
source/PlayState.hx:196: characters 10-28 : Unmatched patterns: FULL
FAILURE - Tilemap (html5)
```

I will try to understand ....

lipki:
I add that at home, it compiles of course ... otherwise it would not be so funny :)

MSGhero:
It’s probably based on your haxe version. I didn’t read the code, but you have a switch statement that doesn’t deal with every enum, something like that.

Gama11:
The build failure comes from the `Tilemap` demo, which showcases the different auto-tiling methods. It would need to be updated to include `FULL`.

JoeCreates:
I've started making the demo for this, but it has exposed an issue with the current algorithm, specifically with wall tiles. Tiles against a wall presently only show if they also have a tile either side. The logic of how walls are handled is not consistent with the logic for checking normal tiles, so when you are against a wall, you go to an incorrect index.

JoeCreates:
I've made a new PR for this with the fixes and also a PR for demos.

lipki:
Thank you for finishing the job: D

PR #2129, opened 2018-02-01
FlxSound.onComplete docs fix

nightblade9:
Tested on HTML5 and it seems to work fine.

### DIFF
--- a/flixel/system/FlxSound.hx
+++ b/flixel/system/FlxSound.hx
@@ -63,7 +63,7 @@ class FlxSound extends FlxBasic
 	public var autoDestroy:Bool;
 	/**
 	 * Tracker for sound complete callback. If assigned, will be called 
-	 * each time when sound reaches its end. Works only on flash and desktop targets.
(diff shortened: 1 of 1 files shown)


nightblade9:
Can this be merged? It's just a trivial documentation change.

Gama11:
I think this comment might be from a time when the HTML5 target wasn't really a consideration / working properly. This might have referred to Android instead, so I'd first like to verify that it works properly there as well.

nightblade9:
@Gama11 okay, I can take care of that.

nightblade9:
Verified that it works as expected on Neko, Android, and HTML5 targets.

Gama11:
Thanks. :) Out of curiosity, I just checked the commit history, turns out this comment is 5 years old:

https://github.com/HaxeFlixel/flixel/commit/ad0b3403b73fd167ae8752d54e1f1f2410e16783#diff-8167fd9f960f8473f99664fcff7d2860R111

PR #2131, opened 2018-02-08
add check for null colorTransforms

sano98:
The same safety check as in updateColorTransform(), which will prevent null ref exceptions in case of colorTransform == null.

### DIFF
--- a/flixel/FlxSprite.hx
+++ b/flixel/FlxSprite.hx
@@ -842,6 +842,9 @@ class FlxSprite extends FlxObject
 		color = FlxColor.fromRGBFloat(redMultiplier, greenMultiplier, blueMultiplier).to24Bit();
 		alpha = alphaMultiplier;
 		
+		if (colorTransform == null)
+			colorTransform = new ColorTransform();
+		
 		colorTransform.setMultipliers(redMultiplier, greenMultiplier, blueMultiplier, alphaMultiplier);
 		colorTransform.setOffsets(redOffset, greenOffset, blueOffset, alphaOffset);
 		


Gama11:
In what scenario is it null?

sano98:
I don't know exactly. We had a crash yesterday and when we examined the crash dump, it turned out that it was a null ref exception where said colorTransform in setColorTransform _was_ null. 

And as the same safety mechanism was deemed worthy to implement in updateColorTransform(), it didn't seen like a long shot to implement it here.

We use the colorTransform to turn a highlight on objects on and off when the mouse cursor hovers over them. We never manually set it to null though.

Gama11:
`colorTransform` might be null because an object was destroyed, in which case this would just serve to hide a bug (you shouldn't work on destroyed sprites, all sorts of crashes can happen).

Good point about the null check in `updateColorTransform()` though, not sure why that's there.

sano98:
hiding a "working on a already destroyed object"-bug: yes, good point. 

Except for in various `destroy()`-methods, a colorTransform also is nulled in line 1276 of DisplayObject.hx?

Gama11:
Huh, what does `DisplayObject` have to do with Flixel?

sano98:
I was just looking for different places in code where colorTransform gets nulled, to see if there is an alternative cause besides the 'destroy()' functions. You can probably forget it.

Gama11:
Unless there's a way to reproduce this bug in a scenario where it's not caused by `destroy()`, I'd prefer not to add this check. Maybe even remove the other `null` check you mentioned since that's also a bit confusing when it should never trigger.

PR #2138, opened 2018-04-01
FlxObject: fix for getting stuck on y-axis collisions, closes #1566

### LINKED ISSUE #1566: Strange collision behavior when colliding against multiple objects
Demonstration

This took some time to narrow down...

Use this source:

``` haxe
package;

import flixel.FlxG;
import flixel.FlxObject;
import flixel.FlxSprite;
import flixel.FlxState;
import flixel.group.FlxGroup.FlxTypedGroup;

{
    var boxes:FlxTypedGroup<FlxSprite>;
    var player:FlxSprite;

    {
        add(boxes = new FlxTypedGroup<FlxSprite>());

        for (i in 0...15)
        {
            var b = new FlxSprite(0, i * 50);
            b.makeGraphic(50, 50, 0xFF000000 + Std.random(0xFFFFFF));
            b.immovable = true;
            b.solid = true;
            boxes.add(b);
        }

        for (i in 0...15)
        {
            var b = new FlxSprite(i*50+50, FlxG.height- 50);
            b.makeGraphic(50, 50, 0xFF000000 + Std.random(0xFFFFFF));
            b.immovable = true;
            b.solid = true;
            boxes.add(b);
        }

        player = new FlxSprite(50, 0);
        player.makeGraphic(20, 20);
        player.acceleration.y = 500;
        add(player);
    }

    {
        FlxG.collide(player, boxes);

        player.velocity.x = 0;

        if (FlxG.keys.pressed.LEFT)
            player.velocity.x = -100;
        if (FlxG.keys.pressed.RIGHT)
            player.velocity.x = 100;
        if (FlxG.keys.justPressed.UP && player.isTouching(FlxObject.DOWN))
            player.velocity.y = -650;

        super.update(elapsed);
    }
}
```

Hug the left-side wall and jump while holding left.
You will collide with the bottom of one of the blocks.

What's happening is a 'perfect storm' of tiny issues.
First, `FlxQuadTree` detects possible collision between the player and 2 of the left-wall blocks. One block at Y (TopBlock) and another at Y+block.height (BottomBlock).

Example

`FlxObject.separate` gets called for each of these possible collisions: first, the Player vs TopBlock.
`separateX` calls `computeOverlapX` which tries to verify that an overlap actually did happen on the X-axis, so, it 'ignores' the player's most-recent Y movement, so it sees the Player as being JUST SLIGHTLY below the TopBlock:

Perceived Position for X-Axis Check

So, `separateX` determines that the player DOES NOT overlap the TopBlock, and therefore, does not need to move to the right at all.

Then `separateY` does it's thing, and it does it's verification a little differently, so it DOES see that the player overlaps the TopBlock, and triggers collision,so you 'hit' the bottom of the TopBlock.

When `separate` is called on the Player vs BottomBlock, next, it will see that the player overlaps on `separateX` and move it to the right, accordingly.

Notably, the order of checks is determined by the order of the objects in the `FlxGroup` - by adding `boxes.sort(FlxSort.byY, FlxSort.DESCENDING);` to `PlayState.create`, you no longer collide with the bottoms of boxes going UP, but now you collide with them coming DOWN.

I'm not 100% sure of all the details on this, but it seems like it could be fixed if the following we changed:
- `computerOverlapX` and `computeOverlapY` should build the same rectangles for confirming overlap, instead of having slightly different answers based on X or Y.
- change the order of `seperate*` being called based on the inverse of the biggest change in motion between the 2 objects - so, if ObjectA is moving up/down much more than it's moving left/right, call `separateX` before `separateY`, and reverse if ObjectA is moving left/right more.
##

Epono:
I'm not entirely confident, but this fixed the issue I had while not breaking anything else as far as I've tested (not very much).

I was getting stucked while going towards and up a vertical wall, but this didn't happen while towards and up or down an horizontal wall.

## Before:
before

(I realize it's not obvious with that animation, but when I'm not moving, I'm actually trying to walk up-left)

## After:
after

Interestingly, it didn't happen while going down, that's why I'm not that confident... That plus the fact that I have no clue how the overlap computation works.

### DIFF
--- a/flixel/FlxObject.hx
+++ b/flixel/FlxObject.hx
@@ -269,8 +269,8 @@ class FlxObject extends FlxBasic
 			var obj1deltaAbs:Float = (obj1delta > 0) ? obj1delta : -obj1delta;
 			var obj2deltaAbs:Float = (obj2delta > 0) ? obj2delta : -obj2delta;
 			
-			var obj1rect:FlxRect = _firstSeparateFlxRect.set(Object1.x, Object1.y - ((obj1delta > 0) ? obj1delta : 0), Object1.width, Object1.height + obj1deltaAbs);
-			var obj2rect:FlxRect = _secondSeparateFlxRect.set(Object2.x, Object2.y - ((obj2delta > 0) ? obj2delta : 0), Object2.width, Object2.height + obj2deltaAbs);
+			var obj1rect:FlxRect = _firstSeparateFlxRect.set(Object1.last.x, Object1.y - ((obj1delta > 0) ? obj1delta : 0), Object1.width, Object1.height + obj1deltaAbs);
+			var obj2rect:FlxRect = _secondSeparateFlxRect.set(Object2.last.x, Object2.y - ((obj2delta > 0) ? obj2delta : 0), Object2.width, Object2.height + obj2deltaAbs);
 			
 			if ((obj1rect.x + obj1rect.width > obj2rect.x) && (obj1rect.x < obj2rect.x + obj2rect.width) && (obj1rect.y + obj1rect.height > obj2rect.y) && (obj1rect.y < obj2rect.y + obj2rect.height))
 			{


Gama11:
I guess this is the same issue as #1566?

Epono:
Oh yeah, I found that issue at the beginning of my debugging session and thought this was different at the time but yes, seems to be the same.

Gama11:
I can confirm that this change fixes the test case from #1566. Also haven't noticed any issues so far testing it against a few demos, so this seems promising. :)

Epono:
Glad to hear it. :)

MSGhero:
I have this issue in Nape too... hmm

Gama11:
@MSGhero Huh.. presumably it's not specific to collisions on the Y axis there though?

MSGhero:
Been a while, but I think it only happened moving up, and I could never replicate it in the other directions. I thought it was some epsilon on the corners that my object was getting caught on, but it happened even using an octagon shape.

cervanky:
Although this solved the issue for getting stuck on the y-axis against multiple objects, this has created another problem for me when colliding with the very edge of another object. 

In the examples below, I'm constantly holding the right key to move the player right, except for in the "after" example when I release the key to "unstick" the player and escape the collision. 

# Before fix:
before

# After fix:
after

Gama11:
Unfortunately, did this end up causing other problems, so I had to revert this for now. See #2141.

PR #2139, opened 2018-04-11
FlxBar: rounding value of filled bar

SeiferTim:
This fixes an issue where a `FlxBar` would be at 100%, but not look like it was at 100% thanks to rounding/precision...


Gama11:
I think the `Std.int()` can be removed then? Bit redundant otherwise.

PR #2146, opened 2018-04-28
FlxSpriteGroup: add camera setter override, closes #2144

### LINKED ISSUE #2144: Setting FlxSpriteGroup.camera should work like FlxSpriteGroup.cameras
When setting the `cameras` property of a `FlxSpriteGroup` it also sets the `cameras` for all its children, as it should. However this doesn't happen with `camera`. Because the `camera` property is simply a way to set/get the first element of `cameras`, I think it makes sense to implement this functionality for `FlxSpriteGroup`. It's both convenient, and what I would consider to be the expected behaviour.

grayhaze:
This adds an override for the `camera` setter to `FlxSpriteGroup`, which also sets the `camera` property of all children.

This resolves https://github.com/HaxeFlixel/flixel/issues/2144.

### DIFF
--- a/flixel/group/FlxSpriteGroup.hx
+++ b/flixel/group/FlxSpriteGroup.hx
@@ -631,6 +631,13 @@ class FlxTypedSpriteGroup<T:FlxSprite> extends FlxSprite
 	
 	// PROPERTIES GETTERS/SETTERS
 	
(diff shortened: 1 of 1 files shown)


Gama11:
Thanks!

PR #2151, opened 2018-05-22
FlxBitmapText: fix computeTextSize() with borders

Skerper:
Accounts for borderSize to prevent the borders being cut off by clipping.


Gama11:
Thanks!

PR #2152, opened 2018-05-25
Add sub-object tween functionality to VarTween

6J7KZg2f:
Allows you to tween sub-object properties of the target object in a VarTween.

Example:
```haxe
mySprite.alpha = mySprite.scale.x = 0;
// fades in from transparent while unfolding horizontally
FlxTween.tween(mySprite, {alpha:1, "scale.x":1}, 1);
```

Due to limitations of the parser, quotations are required for the name of the field if a sub object is used. Properties with no dot in them can be used in the normal way, so this is a non-breaking change. 

This is useful for simplifying callbacks and reducing the amount of tweens that need to be made for various effects - for example, in the above, you would need two tweens on two separate objects, generating two callbacks, and so on.

The factory function has been expanded to check that sub objects are valid.

This could probably benefit from some optimisation, but it's functional and not horrendous as-is.


Gama11:
I like the idea, but not the implementation. It's seems very inefficient to newly resolve the sub-properties in each `update()`, especially considering that the properties that are being tweened won't change at all over the lifetime of a `VarTween` instance, which means this could all just be done once in `initializeVars()`.

I'll give implementing this differently a go.

Gama11:
As far as I can tell, the functionality is the same they way I've implemented it.

```haxe
package;

import flixel.tweens.FlxTween;
import flixel.FlxSprite;
import flixel.FlxState;

{
	{
		var sprite = new FlxSprite();
		sprite.screenCenter();
		sprite.makeGraphic(100, 100);
		add(sprite);

		FlxTween.tween(sprite, {alpha: 0, "scale.x": 2}, 1, {type: FlxTween.PINGPONG});
	}
}
```



Thanks!

PR #2154, opened 2018-05-31
FlxSpriteUtil: add a positioning function to space()

Skerper:
Adds distribute( to FlxSpriteUtil, a function that allows you to place objects in an array into a given order on the screen with the option to provide a tweening function for those objects.

I elected at the moment to go with a new function rather than modifying space, as the functionality was different to what I was going for, and the space function is already bloated without adding new features.

### DIFF
--- a/flixel/util/FlxSpriteUtil.hx
+++ b/flixel/util/FlxSpriteUtil.hx
@@ -218,7 +218,8 @@ class FlxSpriteUtil
 				curX = runningX + prevWidth + horizontalSpacing;
 				runningX = curX;
 			}
-			else{
+			else
+			{
 				curX = object.x;
 			}
 
--- a/flixel/util/FlxSpriteUtil.hx
+++ b/flixel/util/FlxSpriteUtil.hx
@@ -187,6 +187,75 @@ class FlxSpriteUtil
 	}
 	
 	/**
+	 * Distributes an array of FlxObjects from a given point, vertically or horizontally, such that they won't overlap (unless padding is negative).
+	 * Has the option to have a supplied function do the positioning (for tweening purposes)
+	 * EG: FlxSpriteUtil.distribute(VERTICAL, mySprites, 50, 10, function(obj, y){FlxTween.tween(obj, {x:10, y:y}, 0.5);} );
+	 * 	   //Tweens the sprites in mySprites from their current position to being in a column 10px from the left, 50px from the top, 10px apart.
+	 * @param	axis          HORIZONTAL or VERTICAL
+	 * @param	targets       An array of FlxObjects. Cast may be needed.
+	 * @param	from          x position to start placing from for HORIZONTAL, or y position for VERTICAL.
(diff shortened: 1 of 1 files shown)


Gama11:
>I elected at the moment to go with a new function rather than modifying space, as the functionality was different to what I was going for,

Different in what way?

Skerper:
The space function is incapable of maintaining one axis while tweening the other, basically. The way I've implemented distribute, you can move all objects to be distributed on the x axis without changing their y axis, or on the y axis without changing their x axis. Alternatively, you can change both using the position function parameter, as in the example.

For a practical use case demonstration: 
I have four teams' portraits distributed across the bottom of the screen. To show the standings, I place them in an array ordered by score, then distribute across the y axis using my distribute function, so that the top-scoring team is highest. They maintain their current x position, so it looks like they're on steps on a podium.

With the 'space' function, it would only be possible to put them in a single column. 

Ideally I'd reimplement 'space' to take a FlxAxes argument and a FlxPoint, instead of startx and starty, and then within the function it either uses x, y, or both, depending on the FlxAxes given, but that would be a breaking change.

Gama11:
>There is also the option of making the startx and starty arguments of the space function nullable, and then adding the position function parameter to space, but that will obviously make space more complicated. Is that still preferable?

I think, that would be preferable, yeah. From the perspective of an API user, it's always confusing to have two things that are very similarly named and have very similar functionality: "which one am I supposed to use, these are almost identical".

Btw, would that even be necessary (making start x/y nullable)? You have total control over positioning if you specify the callback, you could just ignore either x or y there if you want?

Skerper:
It is of course possible to leave them not-nullable and then do all specifying with the callback, but it seems incredibly convoluted to force usage of the function through the callback even in situations when the user doesn't want to do any tweening effects. 

Something like

`FlxSpriteUtil.space(myObjs,` 0, 0, 10, 0, true, function(obj, x){obj.x = x;});

Would be a very hard sell for me. I don't think it's much more attractive than just writing a loop that positions the array directly.

My current implementation of the 'distribute' function simply defaults to placing the objects directly where they're supposed to be if no position function is required.

I also question whether "space from bounds" is even a thing that needs to be included, since writing a loop that just places objects in an array some number of pixels multiplied by the index apart from each other is not really any less involved or readable than calling this function with all the optional parameters required.

Gama11:
Unit tests would still be nice to have in the future, but I'll merge without them for now. I tested the functionality manually with these one-liners, and it seems to behave as expected:

```haxe
FlxSpriteUtil.space([for (i in 0...10) cast add(new FlxSprite(0, 0).makeGraphic(10, 10))], 0, 0, 30, null, false);
FlxSpriteUtil.space([for (i in 0...10) cast add(new FlxSprite(0, 50).makeGraphic(10, 10))], 0, 0, 30, null, false, (target, x, y) -> {
	FlxTween.tween(target, {x: x, y: y}, 0.1);
});
```

I also renamed `positioner` to `position` to match the usual naming convention for functions.

Issue #1353, opened 2014-10-29
FlxSpriteGroup alpha=0 problem


Txori:
Whenever you set the FlxSpriteGroup alpha to 0, it wont show up again when you change its alpha after that, unlike the FlxSprite...

Here's the snippet:

``` haxe
package;

import flixel.FlxState;
import flixel.FlxSprite;
import flixel.group.FlxSpriteGroup;

class PlayState extends FlxState
{
   private var spriteGroup:FlxSpriteGroup;
   private var sprite:FlxSprite;
   private var t:Float = 0;

   override public function create():Void
   {
      super.create();

      spriteGroup = new FlxSpriteGroup(400, 100);
      spriteGroup.add(new FlxSprite().makeGraphic(100, 100, 0xFFFF0000));
      add(spriteGroup);

      sprite = new FlxSprite(400, 300).makeGraphic(100, 100, 0xFF00FF00);
      add(sprite);
   }

   override public function update(elapsed:Float):Void
   {
      super.update(elapsed);
      t += .01;
      sprite.alpha = spriteGroup.alpha = Math.abs(Math.floor(100 * Math.cos(t)) / 100);
      trace(sprite.alpha);
   } 
}
```

I've only tried on Flash and Neko.

And here's my halelib version:

> actuate: [1.7.5]
>  flixel-addons: [1.1.0]
> flixel-demos: [1.1.1]
> flixel-templates: [1.0.2]
> flixel-tools: [1.0.2]
> flixel-ui: [1.0.2]
> flixel: [3.3.5]
> hxcpp: [3.1.39]
> lime-tools: [1.5.7]
> lime: [1.0.1]
> openfl-samples: [1.3.0]
> openfl: [2.0.1]
> swf: [1.5.2]
##

IBwWG:
OK, but what about cases like in #1870?  For example, you fade out a flxspritegroup, and then go to fade it back in--nothing happens, because after fading out, all of its children were alpha=0.  I think having a relative alpha would be a nice idea.  Just like how the volume mixer on Windows works--as you raise/lower the master volume, all the other volumes move relative to it.  But they don't suddenly stick at 0 just because you hit 0 with the master at one point.

MSGhero:
If we end up having a relative alpha, would it make sense to also make position, rotation, etc that way? To where there's an internal `_var` and then a getter that looks like `return _var + parent.var` until parent is null.

[...]

PR #2157, opened 2018-07-02
Improve alpha functionality

Skerper:
Improved default alpha behaviour by directly setting the alpha of any member whose alpha is set to zero when the group's alpha is set. This prevents group members from 'sticking' at zero alpha. 

Also added an optional flag, directAlpha, which makes setting the alpha value of the group simply propagate that value to all group members. Currently this is set to false by default, but I honestly believe it should be set to true.

I understand there is the potential for the first change to cause unexpected behaviour, but I think that behaviour is considerably less unexpected than sprites sticking at zero alpha.


Gama11:
Added a simple unit test for #1353 here: https://github.com/HaxeFlixel/flixel/commit/5694a35ca7a09c91bb54e9a421d28cc1a2400dc0

PR #2159, opened 2018-07-07
Add transform tool to FlxDebugger interaction

Dovyski:
interaction-transform-tool-final2

This PR adds the transform tool to `FlxDebugger` interaction. Using the tool, it is possible to resize and rotate a sprite. At the moment, the tool only works with a single sprite (it is quite tricky to make it work with groups, but that's something for the future).

I've tried my best to change just a few files, but I had to work on the classes `Interaction`, `Tool` and `Tooltip` to have a decent UX for the transform tool, e.g. chance cursor according to resize/rotate action, show live info about width/height, etc. Those changes are backward compatible and I think they will be useful for future tools or the debugger itself.


Gama11:
There's some CodeClimate issues: https://codeclimate.com/github/HaxeFlixel/flixel/pull/2159

Dovyski:
I've fixed them, @Gama11.

Gama11:
Thanks!

Dovyski:
Thank you! Soon I'll send the PR to update Flixel docs.

PR #2162, opened 2018-08-01
FlxSpriteGroup: fix docs for destroy()

Vascofr:
The description of the destroy() function was actually describing the kill() function of a FlxBasic. Changed to the typical description of destroy().


Gama11:
Nice catch, thanks!

Gama11:
(fyi, pull requests should target the dev branch, not master)

PR #2168, opened 2018-09-07
FlxGamepad: add switch controller model enum

larsiusprime:
Just some enums that define the various ways the switch controllers can be used. Necessary if you're using the OpenFL switch backend, and might be useful for desktop if we ever get native desktop support for these models.


larsiusprime:
Maybe... I'd have to review how I use it in practice to make sure it's not checked every frame or something like that. Fancy enums generate a lot of dynamic state on cpp targets.

Gama11:
Yeah, that's roughly what it'd look like. Would make it easier to check if you're dealing with something Switch-related (only have to check one instead of 5 values), for instance.

Gama11:
Any update on this?

larsiusprime:
Ah, you're on a "fix all of Lars' orphaned merges" kick. I'll go ahead and make this change this week, haven't touched it yet. You still okay with the last agreed upon fix?

Gama11:
Sure, I was mostly wondering if you're okay with it. :)

larsiusprime:
Yeah, that approach is fine. When do you need this by?

Gama11:
I mean, we just had a release, so there's not gonna be one for a while anyway.

Gama11:
Is this even useful on its own, without the model ever being auto-detected as one of these values?

larsiusprime:
I'll have to check; I figure I added this for a reason, probably in relation to some stuff that came up with lime-switch, which uses the gamepad information coming from the switch itself. We can discuss in the slack channel or discord if necessary.

Gama11:
As we discussed, this doesn't make sense to be merged in the current state and will be revisited at some later date.

PR #2176, opened 2018-09-28
FlxFilterFrames.applyToSprite: add to, rather than wipe previous offset

### LINKED ISSUE #2175: FlxFilterFrames.applyToSprite wipes the previous offset rather than augmenting
Affects all targets on version 4.5.1

It's basically ignoring the previous offset and assuming it was 0,0 prior to the applyToSprite

```haxe
// typed on the spot, untested
package;

import flixel.FlxState;
import flixel.FlxSprite;
import flixel.graphics.frames.FlxFilterFrames;

import openfl.filters.GlowFilter;

{
    {
        var sprite = new FlxSprite();
        sprite.makeGraphic(100,100);
        sprite.offset.set(5, 5);
        
        FlxFilterFrames.fromFrames
            ( sprite.frames 
            , 20            // width increase
            , 20            // height increase
            , [new GlowFilter(0xFFFFFF, 1, 10, 10, 2, 1)]
            ).applyToSprite(sprite, true);
        
        trace(sprite.offset);
        // expected: (x: 15 | y: 15)
        // actual: (x: 10 | y: 10)
    }
}
```

Geokureli:
Adjusts the current offset rather than ignoring its previous offset entirely. `applyToSprite` is typically used to add a glow or blur filter to an existing sprite sheet, typically it's desired for the hitbox to only cover the original graphical area so the offset is changed, but it assumed the original offset was 0,0. this change accounts for the previous value

fixes #2175

### DIFF
--- a/flixel/graphics/frames/FlxFilterFrames.hx
+++ b/flixel/graphics/frames/FlxFilterFrames.hx
@@ -91,7 +91,7 @@ class FlxFilterFrames extends FlxFramesCollection
 		var w:Float = spr.width;
 		var h:Float = spr.height;
 		spr.setFrames(this, saveAnimations);
-		spr.offset.set(0.5 * widthInc, 0.5 * heightInc);
+		spr.offset.add(0.5 * widthInc, 0.5 * heightInc);
 		spr.setSize(w, h);
 	}
 	


Geokureli:
the number of filters nor the size of the filters supplied per `FlxFilterFrame` doesn't matter so much, the sprite's size will only increase by [the amount specified](https://github.com/HaxeFlixel/flixel/blob/dev/flixel/graphics/frames/FlxFilterFrames.hx#L57), and the resulting graphic will always have the old graphic centered evenly within it.

This issue was discovered when my game calls `FlxFilterFrame.fromFrames(...).applyToSprite(...)` on the same sprite iteratively. the hitbox became more "off"  with every iteration.

Geokureli:
Thanks :)

PR #2186, opened 2018-11-13
Fix XML access warning

grayhaze:
This fixes the following warnings when compiling Flixel projects on Haxe versions 4.0+:

```
haxe/lib/flixel/git/flixel/graphics/frames/FlxAtlasFrames.hx:255: characters 12-16 : Warning : This typedef is deprecated in favor of haxe.xml.Access
haxe/lib/flixel/git/flixel/graphics/frames/FlxAtlasFrames.hx:255: characters 23-36 : Warning : This typedef is deprecated in favor of haxe.xml.Access
```

The approach used matches that in @jgranick's recent commit to Lime (https://github.com/openfl/lime/commit/263e73922e26a574f123e84b92219cff70ad56db).

### DIFF
--- a/flixel/graphics/frames/FlxAtlasFrames.hx
+++ b/flixel/graphics/frames/FlxAtlasFrames.hx
@@ -10,8 +10,13 @@ import flixel.math.FlxRect;
 import flixel.system.FlxAssets.FlxGraphicAsset;
 import flixel.system.FlxAssets.FlxTexturePackerSource;
 import haxe.Json;
-import haxe.xml.Fast;
 import openfl.Assets;
+
+#if (haxe_ver >= 4)
+import haxe.xml.Access in Fast;
+#else
+import haxe.xml.Fast;
+#end
 
 /**
  * Atlas frames collection. It makes possible to use texture atlases in Flixel.
@@ -252,7 +257,7 @@ class FlxAtlasFrames extends FlxFramesCollection
 		if (Assets.exists(Description))
 			Description = Assets.getText(Description);
 		
-		var data:Fast = new haxe.xml.Fast(Xml.parse(Description).firstElement());
+		var data:Fast = new Fast(Xml.parse(Description).firstElement());
 		
 		for (texture in data.nodes.SubTexture)
 		{


jgranick:
The sad part is that there is no way to distinguish between Haxe 4 preview 4 and Haxe 4 preview 5.

However, since both of these are development versions of Haxe, I guess it makes sense to just ask a dev version user to upgrade for this

Gama11:
The problem with this / the reason why this hasn't been done yet is that it breaks compatibility with Haxe 4 preview 4 and earlier. There's probably quite a few people who haven't bothered to update to preview 5 yet. Or are unable to, because library support isn't there yet. In particular, I think the current releases of Lime and OpenFL aren't compatible with it on all targets.

So for now, the warnings seems tolerable.

grayhaze:
I'm happy to close the PR if it's not deemed appropriate to make the change at this point in time. It does feel like if someone's using preview builds of Haxe with development versions of OpenFL, Flixel, etc. it's not unreasonable or unexpected for them to update to the latest preview builds as they're released though. I haven't found any issues with the targets I build for (Windows, Mac, Android) with the latest preview release, but I'm sure my use case doesn't match those of others.

Gama11:
>it's not unreasonable or unexpected for them to update to the latest preview builds as they're released though

I agree with that actually, I just don't like that people would end up with cryptic error messages (probably something like `Type not found: haxe.xml.Access`) when using mismatched versions. It'd be nice to have a friendlier error like "Haxe 4 preview 5 or newer required", like we normally do via `FlxDefines` for Haxe / Lime / OpenFL versions. But yeah, you can't really tell preview versions apart.

That being said, there might be another way to support both preview 4 and 5 (without warnings) via `haxe.macro.Context.onTypeNotFound`, I might have a look at that...

>I haven't found any issues with the targets I build for (Windows, Mac, Android) with the latest preview release

Preview 5 is not compatible with current Haxelib releases of Lime and OpenFL on HTML5 due to https://github.com/openfl/openfl/issues/2041. I thought Flash had the same problem, but apparently that's not the case.

Gama11:
Conveniently, a `haxe4` define was recently added, which does let us keep preview 4 compatibility. I just means you'll need a current Haxe nightly to avoid the warnings / preview 5 still has them.

PR #2187, opened 2018-11-19
FlxBitmapFont: Remove background of XNA fonts

Birdwards:
When a FlxBitmapText uses a font that was created using FlxBitmapFont.fromXNA(), and the zoom of the current FlxCamera is a non-integer value, the background color that separates characters in the font's source image can sometimes appear between characters in the FlxBitmapText (see example [here](https://i.imgur.com/lJddn4K.png).) There is already code in place to remove this background color from the font, but it's currently restricted to the Flash target only, apparently due to crashes on other platforms. I removed the conditional check for Flash and tested FlxBitmapFont.fromXNA() on HTML5, Linux, Neko, and Windows; none of them crashed, and as desired, the background color no longer appears between the characters.


Gama11:
Thanks! I bet this was only an issue with some ancient OpenFL version.

PR #2191, opened 2018-11-29
FlxVector abstract and weak vectors

### LINKED ISSUE #2185: Suggestion: multiple FlxVector changes
This started out as a single suggestion but has grown to cover multiple requests:
- [ ] Add FlxVector.weak and call putWeak() on operations
- [ ] Make FlxVector an abstract class so converting FlxPoints to FlxVectors is trivial and FlxVector operations can take FlxPoints
- [ ] 'override' all FlxPoint methods in FlxVector and use covariance to return FlxVectors. **Note**: this won't be necessary if the [polymorphic 'this' types proposal](https://github.com/HaxeFoundation/haxe-evolution/pull/36) is accepted

**Original message:**
FlxPoint.weak is such a great tool, I'd love to see it on FlxVector as well. Right now FlxPoint will call `putWeak()` on Vectors, but they are never weak. I'm not seeing any obvious drawbacks to adding a static `weak()` method and calling `putWeak` on FlxVector methods that handle other vectors

Geokureli:
fixes #2185

- Made FlxVector an abstract of FlxPoint
- added FlxVector.weak
- calling putWeak on FlxVector operations
- recreate all methods that return FlxPoint and return FlxVector instead

### DIFF
--- a/flixel/math/FlxPoint.hx
+++ b/flixel/math/FlxPoint.hx
@@ -9,7 +9,6 @@ import openfl.geom.Matrix;
 /**
  * Stores a 2D floating point coordinate.
  */
- 
 @:allow(flixel.math.FlxVector)
 class FlxPoint implements IFlxPooled
 {
--- a/flixel/math/FlxPoint.hx
+++ b/flixel/math/FlxPoint.hx
@@ -9,6 +9,8 @@ import openfl.geom.Matrix;
 /**
  * Stores a 2D floating point coordinate.
  */
+ 
+@:allow(flixel.math.FlxVector)
 class FlxPoint implements IFlxPooled
 {
 	public static var pool(get, never):IFlxPool<FlxPoint>;
@@ -48,7 +50,6 @@ class FlxPoint implements IFlxPooled
 	public var x(default, set):Float = 0;
 	public var y(default, set):Float = 0;
 	
-	@:allow(flixel.math.FlxVector)
 	var _weak:Bool = false;
 	var _inPool:Bool = false;
 	
@@ -118,6 +119,11 @@ class FlxPoint implements IFlxPooled
 	 */
 	public function addPoint(point:FlxPoint):FlxPoint
 	{
+		return __addPoint(point, true);
+	}
+	
+	inline function __addPoint(point:FlxPoint, put:Bool = false):FlxPoint
+	{
 		x += point.x;
 		y += point.y;
 		point.putWeak();
@@ -145,6 +151,11 @@ class FlxPoint implements IFlxPooled
 	 * @return	This point.
 	 */
 	public function subtractPoint(point:FlxPoint):FlxPoint
+	{
+		return __subtractPoint(point, true);
+	}
+	
+	inline function __subtractPoint(point:FlxPoint, put:Bool = false):FlxPoint
 	{
 		x -= point.x;
 		y -= point.y;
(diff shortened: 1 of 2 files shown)


Geokureli:
closing until I create a better unit test for FlxVector to verify stability

Geokureli:
yep, works with FlxLightPuzzle

flixel/math/FlxPoint.hx
```diff
@@ -394,7 +405,7 @@ class FlxPoint implements IFlxPooled
 	 */
 	public inline function toVector():FlxVector
 	{
-		return FlxVector.get(x, y);
+		return this;
```
Geokureli: this could create issues in cases where people were expecting a new instance

flixel/math/FlxVector.hx
```diff
@@ -534,14 +718,15 @@ class FlxVector extends FlxPoint
...
-		var p2:FlxVector = projectToNormalized(normal, _vector2);
+		var p1:FlxVector = projectToNormalizedWeak(normal.rightNormal(_vector3), _vector1);
+		var p2:FlxVector = projectToNormalizedWeak(normal, _vector2);
 		var bounceX:Float = -p2.x;
 		var bounceY:Float = -p2.y;
 		var frictionX:Float = p1.x;
 		var frictionY:Float = p1.y;
 		x = bounceX * bounceCoeff + frictionX * friction;
 		y = bounceY * bounceCoeff + frictionY * friction;
+		normal.put();
```
Gama11: This seems odd?

Geokureli:
so I made some changes. now the fooWeak versions of methods will not call putweak. the normal method will just call it. every thing should inline to be pretty much the exact same thing, but this should be a bit more clear in whats going on rather than having those mysterious `true` args in fooWeak function calls

Geokureli:
I'm still not super happy with the "Weak" suffix on methods, open to suggestions.

flixel/math/FlxPoint.hx
```diff
@@ -419,7 +419,7 @@ class FlxPoint implements IFlxPooled
 	 * Function to get a `FlxVector` from this `FlxPoint`
 	 * @since 4.3.0
 	 */
-	@:deprecated
+	@:deprecated("The `toVector` method is deprecated, use `cast(myPoint, FlxVector)` or `FlxVector.get` instead")
```
Gama11: I'd actually consider a safe cast bad practice here, that has unnecessary runtime overhead. There's no explicit casts needed to trigger the _implicit_ casts of abstracts.

Geokureli: scratch that last comment, I don't like changing the constructor of FlxVector

Gama11: Can't it just say "use implicit casts"?

It's not clear to me how changing the constructor would be helpful? Typically you'd just use assignments or type checks to trigger implicit casts.

```haxe
var vector:FlxVector = someFlxPoint;
vector.someVectorMethod();

(someFlxPoint : FlxVector).someVectorMethod();
```

Geokureli: ```The `toVector` method is deprecated, FlxPoints can be casted to FlxVectors implicitly```
???

Gama11: Sure. :)

Gama11:
Thanks!

MSGhero:
If this requires any changes to the light demo, let me know. Well, it shouldn’t “require” changes since it’s non-breaking, but you know.

Issue #2158, opened 2018-07-03
WebGL errors in a small project


Zireael07:
- **Haxe version:** 3.7.4
- **Flixel version:** 4.4.2
- **OpenFL version:** 8.3.0
- **Lime version:** 6.4.0 (not sure, how do I check the versions?)
- **Affected targets:** html5

________________________________

See zip.

________________________________

**Observed behavior:**

WebGL errors and tiles not drawing (random which ones) for the first couple of moves, until the max error limit is hit.

**Expected behavior:**

No WebGL errors.

[Haxe Veins (2).zip](https://github.com/HaxeFlixel/flixel/files/2158717/Haxe.Veins.2.zip)

Gama11:
Here's a HTML5 build of the project. To reproduce, it's enough to press the down key once:

https://gama11.github.io/test/Flixel_Issue_2158/

![](https://i.imgur.com/5lXOCu4.png)

@jgranick Any ideas / have you run into this error before? This seems related to usage of the `FlxSprite.color` property, which enables a different code path (`hasColorTransform`) in the graphics shader (Flixel's is basically a 1:1 copy of OpenFL's IIRC). If usage of that property is commented out, the error doesn't occur.

PR #2195, opened 2018-12-04
Color transform fix, closes #2192

### LINKED ISSUE #2192: Color transform offsets not working correctly, also issues with colortransform alpha
- **Haxe version:** 3.4.7
- **Flixel version:** latest dev version from github
- **OpenFL version:** latest develop branch version from github
- **Lime version:** 7.1.1
- **Affected targets:** windows (presumably neko as well)

```haxe
package;

import flixel.FlxG;
import flixel.FlxSprite;
import flixel.FlxState;
import flixel.util.FlxColor;
import openfl.Assets;
import openfl.display.BitmapData;
import openfl.geom.ColorTransform;
import openfl.display.Bitmap;

{
	var sprite:FlxSprite;
	var sprite2:FlxSprite;
	var bmp:BitmapData;
	var toggle = false;

	var oflBitmap:Bitmap;
	
	{
		makeCheckerBoard();

		sprite = new FlxSprite();
		sprite2 = new FlxSprite();
		sprite.makeGraphic(64,64,FlxColor.GRAY);
		sprite2.makeGraphic(64,64,FlxColor.GRAY);
		bmp = new BitmapData(64,64,false,FlxColor.GRAY);
		sprite.x = Std.int((FlxG.width-sprite.width)/2) - sprite.width;
		sprite.y = Std.int((FlxG.height-sprite.height)/2);

		oflBitmap = new Bitmap(new BitmapData(64,64,false,FlxColor.GRAY));
		oflBitmap.x = Std.int(sprite.x + sprite.width);
		oflBitmap.y = sprite.y;
		FlxG.stage.addChild(oflBitmap);
		
		sprite2.x = sprite.x + sprite.width + sprite2.width;
		sprite2.y = sprite.y;
		add(sprite);
		add(sprite2);
	}
	
	private function makeCheckerBoard():Void
	{
		var tileW = Math.round(FlxG.height / 10);
		var tileH = tileW;

		var tilesWide = Math.round(FlxG.width/tileW);
		var tilesTall = Math.round(FlxG.height/tileH);

		for(iy in 0...tilesTall){
			for(ix in 0...tilesWide){
				var tileColor = (iy + ix) % 2 == 0 ? FlxColor.BLACK : FlxColor.WHITE;
				var tile = new FlxSprite().makeGraphic(tileW,tileH,tileColor);
				tile.x = ix * tileW;
				tile.y = iy * tileH;
				add(tile);
			}
		}
	}

	
	{
		super.update(elapsed);

		if(FlxG.keys.justPressed.SPACE)
		{
			toggle = !toggle;
			if(toggle)
			{
				var r = rand(-255,255)/255;
				var g = rand(-255,255)/255;
				var b = rand(-255,255)/255;
				var a = 1.0;//rand(0,255)/255;
				var ra = rand(-255,255);
				var ga = rand(-255,255);
				var ba = rand(-255,255);
				var aa = 0;//rand(-255,255);
				sprite.setColorTransform(r,g,b,a,ra,ga,ba,aa);
				sprite2.graphic.bitmap.draw(bmp, null, new ColorTransform(r,g,b,a,ra,ga,ba,aa));
				oflBitmap.transform.colorTransform = new ColorTransform(r,g,b,a,ra,ga,ba,aa);
			}
			else
			{
				sprite.setColorTransform(1,1,1,1,0,0,0,0);
				sprite2.graphic.bitmap.draw(bmp, null, new ColorTransform(1,1,1,1,0,0,0,0));
				oflBitmap.transform.colorTransform = new ColorTransform(1,1,1,1,0,0,0,0);
				sprite.alpha = 1.0;
			}
		}
	}

	private function rand(min:Int, max:Int):Int
	{
		var range = (max-min)+1;
		return Std.int(Math.random() * range) + min;
	}
}
```

**Observed behavior:**
This little test creates three squares and puts them in the scene, then applies color in one of three ways:
- A regular FlxSprite, with a colorTransform applied directly to it (left)
- A regular FlxSprite, whose pixels are directly set by a BitmapData.draw() command with a color transform supplied -- e.g. software rendering (middle)
- An openfl Bitmap object applied directly to FlxG.stage, with a colorTransform applied directly to it (right)
- The three squares are directly touching each other so that subtle color errors are easier to notice.

Press spacebar to toggle between default color transform values (1,1,1,1,0,0,0,0) and randomized ones.

- Alpha transforms are commented out in this sample as it's hard to notice the color errors with them in the mix, but there appears to be a bug with them too. 
- Color multipliers don't appear correct -- they're a bit too dark.
- Using color offsets cause the FlxSprite to disappear entirely, unless alpha offset is 255. Even then, the result is too dark.
- Alpha multiplier appears to yield the correct alpha, and interestingly enough the color matches too! *But only when the alpha multiplier is not 1.0*, otherwise the color is too dark.
- Alpha offset is broken as mentioned above -- sprite entirely invisible when color offsets used unless alpha offset is set to 255. When no color multiplier or offsets are set (and no alpha multiplier either, just alpha offset), then the result we get is the wrong alpha value for the basic gray square.

**Expected behavior:**
- The basic color of all three squares should match when the transforms are applied.
- The alphas for the upper left (flixel w/ transform) and the one on the bottom (openfl w/ transform) should match.
- The alpha of the upper right one is not expected to change as BitmapData.draw() doesn't apply that operation.

JoeCreates:
There were two issues in #2192.

Firstly, the color transform was being applied twice because the generated graphics were not being set to unique.

Secondly was the actual shader issue. The problem was that the shader applies the offsets and multipliers based off of a single flag, `hasColorTransform`. However, in actually populating the values to be used for these, each `colored` and `hasColorOffsets` are checked individually to populate the inputs for the shader.

In the case of a transform with multipliers but no offsets, this was fine, because the lack of offsets had no affect on the result as they effectively default to 0. In the case of offsets but no multipliers, however, treating the multipliers as 0 by default resulted in a black sprite.

I had two options for resolving this. One was to split up the application of offsets and multipliers in the shader and pass in two flag,s one for each, or make it so that if either are true when we are giving the shader the inputs that we populate both inputs instead of only one.

I opted for the latter solution as this means that the shader interface does not change, and I wasn't sure what else might have broken if I did change the shader interface.

### DIFF
--- a/flixel/graphics/tile/FlxDrawQuadsItem.hx
+++ b/flixel/graphics/tile/FlxDrawQuadsItem.hx
@@ -71,10 +71,13 @@ class FlxDrawQuadsItem extends FlxDrawBaseItem<FlxDrawQuadsItem>
 		for (i in 0...VERTICES_PER_QUAD)
 			alphas.push(transform != null ? transform.alphaMultiplier : 1.0);
 
-		if (colored)
+		if (colored || hasColorOffsets)
 		{
 			if (colorMultipliers == null)
 				colorMultipliers = [];
+			
+			if (colorOffsets == null)
+				colorOffsets = [];
 
 			for (i in 0...VERTICES_PER_QUAD)
 			{
@@ -83,27 +86,7 @@ class FlxDrawQuadsItem extends FlxDrawBaseItem<FlxDrawQuadsItem>
 					colorMultipliers.push(transform.redMultiplier);
 					colorMultipliers.push(transform.greenMultiplier);
 					colorMultipliers.push(transform.blueMultiplier);
-				}
-				else
-				{
-					colorMultipliers.push(1);
-					colorMultipliers.push(1);
-					colorMultipliers.push(1);
-				}
-
-				colorMultipliers.push(1);
-			}
-		}
-
-		if (hasColorOffsets)
-		{
-			if (colorOffsets == null)
-				colorOffsets = [];
-
-			for (i in 0...VERTICES_PER_QUAD)
-			{
-				if (transform != null)
-				{
+					
 					colorOffsets.push(transform.redOffset);
 					colorOffsets.push(transform.greenOffset);
 					colorOffsets.push(transform.blueOffset);
@@ -111,11 +94,17 @@ class FlxDrawQuadsItem extends FlxDrawBaseItem<FlxDrawQuadsItem>
 				}
 				else
 				{
+					colorMultipliers.push(1);
+					colorMultipliers.push(1);
+					colorMultipliers.push(1);
+					
 					colorOffsets.push(0);
 					colorOffsets.push(0);
 					colorOffsets.push(0);
 					colorOffsets.push(0);
 				}
+
+				colorMultipliers.push(1);
 			}
 		}
 	}
@@ -131,10 +120,10 @@ class FlxDrawQuadsItem extends FlxDrawBaseItem<FlxDrawQuadsItem>
 		shader.bitmap.filter = (camera.antialiasing || antialiasing) ? LINEAR : NEAREST;
 		shader.alpha.value = alphas;
 
-		if (colored)
+		if (colored || hasColorOffsets) {
 			shader.colorMultiplier.value = colorMultipliers;
-		if (hasColorOffsets)
 			shader.colorOffset.value = colorOffsets;
+		}
 
 		setParameterValue(shader.hasTransform, true);
 		setParameterValue(shader.hasColorTransform, colored || hasColorOffsets);


flixel/graphics/tile/FlxDrawQuadsItem.hx
```diff
@@ -12,7 +12,7 @@ import openfl.Vector;
 
 class FlxDrawQuadsItem extends FlxDrawBaseItem<FlxDrawQuadsItem>
 {
-	static inline var VERTICES_PER_QUAD = #if (openfl >= "8.5.0") 4 #else 6 #end;
+	static inline var VERTICES_PER_QUAD = #if (openfl >= "8.5.0") 1 #else 1 #end;
```
Gama11: ?

JoeCreates: Sorry, I didn't mean to push that but accidentally did when I did the force ppush to fix the codeclimate issue. It's probably worth looking at first, though, tbh. I'm not sure the looping is necessary.

larsiusprime:
Much thanks, I'm going to give this a test locally.

larsiusprime:
It works! It works! Oh man I'm so happy @JoeCreates . Where shall I send the bounty?

Gama11:
Thanks!

Gama11:
Hm, I thought this would fix that issue in the FlxMouseEventManager demo, but it seems it hasn't... :/

http://haxeflixel.com/demos/FlxMouseEventManager/

Gama11:
Looks like that might be the same issue as #2158.

PR #2199, opened 2018-12-07
Support for blend modes with drawQuads, closes #2196

### LINKED ISSUE #2196: Blend modes don't currently work in OpenFL 8
- **Haxe version:** 3.4.7
- **Flixel version:** latest dev version from github
- **OpenFL version:** latest develop branch version from github
- **Lime version:** 7.1.1
- **Affected targets:** windows (presumably neko as well)

The [BlendModes demo](http://haxeflixel.com/demos/BlendModes/) reproduces the issue very clearly and comprehensively.

**Observed behavior:**

It looks like this:
2018-12-04 16_56_30-blendmodes

**Expected behavior:**

It should looks something like this:
2018-12-04 16_55_59-blendmodes _ haxeflixel 2d game engine

(flash results)

------------------

I know this is a known issue, that is technically OpenFL's "fault" in that the drawQuads API doesn't currently expose a functioning blend mode parameter, but I think we should seriously look into doing this on our own just by using a custom shader that implements BlendModes. That will necessitate the custom shader fed into drawQuads being able to sample the background, and may also require a multi-pass solution.

I've been talking to @JoeCreates and @jgranick about this, and I think it's possible. I still need to investigate more to see if it's possible to do without having to change OpenFL's api. But I think doing it in our own shader is valuable because a) it doesn't put an additional support/implementation/maintenance burden for Yet Another Feature in the OpenFL drawQuads API, and it opens the door for all sorts of effects, particularly if we figure out a multi-pass solution with background sampling, that's apply-able per-FlxSprite.

If this seems like a reasonably good approach, then I'll also make this LarsBounty #5 and put $50 USD on it.

JoeCreates:
The main bulk of the stuff required for this already got merged into openfl, but one of Joshua's tweaks stopped it working again, so this PR will only have the desired effect once https://github.com/openfl/openfl/pull/2093 get's merged into openfl (but it will build fine with the current develop branch).

Something not so clean in this PR right now is the fact that `BaseDrawItem` already has a `blending:Int` for tilesheets, but for the new graphics api we need to pass an actual `BlendMode`. We don't want to be converting this blend mode to an `Int` just to convert it back again (and I'm not sure the ints are even the same as those from Tilesheet).

I've added another field, `blend:BlendMode` to `FlxBaseDrawItem`. I feel like the tilesheet stuff should be updated to use this and only convert to an int before it is needed (if there is a need to keep that stuff at all), but that is a bit beyond the scope of this issue.


flixel/graphics/tile/FlxDrawQuadsItem.hx
```diff
@@ -128,7 +128,8 @@ class FlxDrawQuadsItem extends FlxDrawBaseItem<FlxDrawQuadsItem>
 
 		setParameterValue(shader.hasTransform, true);
 		setParameterValue(shader.hasColorTransform, colored || hasColorOffsets);
-
+		
+		camera.canvas.graphics.overrideBlendMode(blend);
```
Gama11: We need OpenFL version checks here to keep compatibility with OpenFL 8.0.0 - 8.7.0

JoeCreates: Remind me, how should that look?

JoeCreates: Done, I think.

flixel/FlxCamera.hx
```diff
@@ -493,6 +493,7 @@ class FlxCamera extends FlxBasic
 		itemToReturn.colored = colored;
 		itemToReturn.hasColorOffsets = hasColorOffsets;
 		itemToReturn.blending = blendInt;
+		itemToReturn.blend = blend;
```
Gama11: Don't we also need to check `blend` earlier in this function (lines 468-474) to check if two draw items are the same?

JoeCreates: Done.

Issue #2155, opened 2018-06-04
Block Flixel mouse events when a debugger interaction tool is active


Gama11:
If there's a `FlxDebugger` interaction tool active, all flixel mouse events should be blocked using `FlxG.mouse.enabled`, similar to what `FlxDebugger` itself already does when it gains / loses focus. Otherwise, you end up triggering button callbacks or similar instead of selecting or moving elements.

This seems to require changes to how `FlxG.mouse.enabled` works (if it's disabled, all currently held buttons need to be marked as released), and maybe `FlxDebugger`'s handling of it too.

Dovyski:
I can work on this. If I am not mistaken, `FlxDebugger` already does some internal working to properly handle mouse clicks on debug buttons/overlays (ignoring game mouse events). In theory, we mainly need to worry about flixel mouse events.

Geokureli:
Just wanted to chime in and say that, currently, clicking a debug window doesn't block mouse events to the FlxGame/FlxMouse. If I click a debug window, not only does the mouse event trigger in Flixel, but it was processed in FlxMouse before the debug window. this is because FlxMouse is adding MouseEvent listeners to the stage

PR #2209, opened 2019-01-22
Disable mouse input when the debugger interaction tool is active

Dovyski:
`FlxG.mouse.enabled` will temporarily return `false` if the debugger interaction tool is active (visible). `FlxButton` (and any subclass) and `FlxMouseEventManager` will now account for `FlxG.mouse.enabled` when dealing with states, so clickable elements related to those will not work when the debugger interaction tool is active.

I decided to preserve the state of clickable elements when the debugger interaction tool becomes active, i.e. over remains over instead of released. I think it makes more sense to developers to not have a lot of state changes fired when the debugger is enabled. If you dislike this approach, let me know so I can adapt the code to no preserve the state of clickable elements.

Fixes: #2155

### DIFF
--- a/flixel/system/debug/FlxDebugger.hx
+++ b/flixel/system/debug/FlxDebugger.hx
@@ -487,15 +487,13 @@ class FlxDebugger extends Sprite
 		// Disable mouse input if the interaction tool is active/visible,
 		// so users can select interactable elements, e.g. buttons.
 		FlxG.mouse.enabled = !interaction.visible;	
-		#end
 
 		if (_usingSystemCursor)
 		{
-			#if FLX_MOUSE
 			FlxG.mouse.useSystemCursor = _wasUsingSystemCursor;
 			FlxG.mouse.visible = _wasMouseVisible;
-			#end
-		}	
+		}
+		#end	
 	}
 
 	inline function toggleDrawDebug():Void
--- a/flixel/system/debug/FlxDebugger.hx
+++ b/flixel/system/debug/FlxDebugger.hx
@@ -484,9 +484,9 @@ class FlxDebugger extends Sprite
 	function onMouseFocusLost():Void
 	{
 		#if FLX_MOUSE
-		// Disable mouse input if the interaction tool is active/visible,
+		// Disable mouse input if the interaction tool is in use,
 		// so users can select interactable elements, e.g. buttons.
-		FlxG.mouse.enabled = !interaction.visible;	
+		FlxG.mouse.enabled = !interaction.isInUse();	
 
 		if (_usingSystemCursor)
 		{
--- a/flixel/system/debug/FlxDebugger.hx
+++ b/flixel/system/debug/FlxDebugger.hx
@@ -483,10 +483,15 @@ class FlxDebugger extends Sprite
 	@:allow(flixel.system.debug)
 	function onMouseFocusLost():Void
 	{
+		#if FLX_MOUSE
+		// Disable mouse input if the interaction tool is active/visible,
+		// so users can select interactable elements, e.g. buttons.
+		FlxG.mouse.enabled = !interaction.visible;	
+		#end
+
 		if (_usingSystemCursor)
 		{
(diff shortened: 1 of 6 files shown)


Gama11:
Hm, I think it would be a lot cleaner to en- and disable as needed, instead of `FlxMouse.enabled` having hardcoded knowledge about the internals of the debugger. `FlxDebugger` already follows that pattern in `onMouseFocus()` and `onMouseFocusLost()`.

Dovyski:
Agreed, that way is much better. In such case, the interaction tool should keep track of the current value of `FlxG.mouse.enabled` when the debugger (or the interaction itself) becomes active, so the value can be restored. Right?

Gama11:
You mean in case the user has set the value? That's currently not done, nor is it done for `FlxG.keys.enabled` usages, so it might be a bit overkill.

Dovyski:
Yeah, in case the user has set it. But since that is not the case, I'll just set true/false accordingly.

flixel/system/debug/FlxDebugger.hx
```diff
@@ -483,10 +483,15 @@ class FlxDebugger extends Sprite
 	@:allow(flixel.system.debug)
 	function onMouseFocusLost():Void
 	{
+		#if FLX_MOUSE
```
Gama11: Minor complaint, but it looks we just need a single `FLX_MOUSE` check for this entire function.

Dovyski: Yep, no need for two checks.

Dovyski:
CI is failing in `flixel.system.replay.FlxReplayTest#testButtonTrigger` when the target is C++. It seems a button is not being clicked. I'll check on that.

Dovyski:
They might be random =P. Build [#5251.3](https://travis-ci.org/HaxeFlixel/flixel/jobs/483407274) failed because of network issues (I think), but [#5251.7](https://travis-ci.org/HaxeFlixel/flixel/jobs/483407278) seems to be legit.

Gama11:
Have you had a chance to look at the test failure yet? Updated the branch to latest dev, but it's still happening.

Dovyski:
Not yet, I'm sorry. I can check on it this weekend.

Dovyski:
It seems the cpp targets are still failing (now on a different test cases). When I run the [tests locally](https://gist.github.com/Dovyski/abdae32f69b9253f5713b2d6bc8414bc#file-haxeflixel-tests-cpp-log) (Win10, Haxe 3.4.7, Haxeflixel dev) they all pass though.

Am I missing anything?

Gama11:
I do get those failures locally as well.

Dovyski:
I think it is finally working.

flixel/system/frontEnds/DebuggerFrontEnd.hx
```diff
@@ -168,6 +168,18 @@ class DebuggerFrontEnd
...
+			#if FLX_MOUSE
+			FlxG.mouse.enabled = true;
+			#end
+		}
+		else
+		{
+			#if FLX_MOUSE
+			// Debugger is visible, allow mouse input in the game only if the
+			// interaction tool is not active.
+			FlxG.mouse.enabled = !FlxG.game.debugger.interaction.visible;
```
Gama11: This is a bit annoying - just because it's visible doesn't mean it's actively being used. I'd like to still be able to interact with buttons as long as none of the interaction tools are selected.

Dovyski: I agree. I'll change that.

Gama11:
Looks good to me now, thanks! 👍

Dovyski:
Thank you for all the feedback and for merging this PR! 🎉

Gama11:
I even found two new issues while testing this PR, one fixed (605a18b7626a40895291487fc6b35150d738b333) and one leftover for you (#2214). ;)

Dovyski:
Nice catch! I'll get that leftover soon :D

PR #2221, opened 2019-06-29
Check mass when separating objects

01010111:
Instead of relying on `FlxObject.immovable` to create objects that aren't pushed by other objects, set an object's mass higher than another one to get the same effect while allowing for other collisions as well.

It seems like this should already be implemented, so I'm guessing there's some edge case that I haven't seen yet that prevented it :P

### DIFF
--- a/flixel/FlxObject.hx
+++ b/flixel/FlxObject.hx
@@ -216,6 +216,19 @@ class FlxObject extends FlxBasic
 
 			if (!obj1immovable && !obj2immovable)
 			{
+				if (Object1.mass > Object2.mass)
+				{
+					Object2.x += overlap;
+					Object2.velocity.x = obj1v - obj2v * Object2.elasticity;
+					return true;
+				}
+				if (Object2.mass > Object1.mass)
+				{
+					Object1.x += overlap;
+					Object1.velocity.x = obj1v - obj2v * Object1.elasticity;
+					return true;
+				}
+
 				overlap *= 0.5;
 				Object1.x = Object1.x - overlap;
 				Object2.x += overlap;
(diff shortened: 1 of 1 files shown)


MSGhero:
Should it be this way though? Real life would have both objects accelerate, scaled down by the mass.

01010111:
@MSGhero that could be ok too, especially if we wanted more "realistic" physics. This is really just a way to make more arcade-y gameplay focused physics I think

MSGhero:
I didn't know there was a `mass` property, so I'm not sure what's "correct." Nape for instance would do realistic physics, but you can jump in and accept or reject collisions while modifying velocity and stuff. But the mass thing is a lot easier than all that.

01010111:
yeah afaik mass is only used when calculating elasticity stuff. Seems like an underused property! The only edge cases I've run into so far using this implementation are more preferable (imo) than all the things that pop up without it - specifically without it simple stacked objects get really unreliable and can fall through geometry, with it objects can get stuck under other objects in some cases.

Gama11:
>Should it be this way though? Real life would have both objects accelerate, scaled down by the mass.

This sounds rather strange to me as well. Just because the mass of an object is slightly bigger doesn't mean it's immovable.

01010111:
I guess I wanted to make this PR to push this discussion more than anything. I'd be happy with anything, it's just the way things are now seems kinda lacking 👍

MSGhero:
There’s a fairly clean way in Nape to do this, so this is partially kinda related to a more featured physics system.

01010111:
🤷‍♂ If it doesn't need to be implemented that's cool but imo this could fix a long standing issue people have been having for ages!

Gama11:
Closing this for now as there doesn't seem to be a lot of support for this idea.

PR #2222, opened 2019-06-30
Respect FlxGroup#cameras

JoeCreates:
This PR means a FlxGroup cameras is no longer useless. FlxCamera.defaultCameras is temporarily modified during a draw call so that the nearest parent group with cameras specified is used as the default. This removes the need for camera transforms in FlxSpriteGroup. It also means that states or other FlxGroups can now have their own cameras specified and this will be respected by children, but can still be overridden by specifying a child's cameras.


flixel/group/FlxSpriteGroup.hx
```diff
@@ -787,6 +787,18 @@ class FlxTypedSpriteGroup<T:FlxSprite> extends FlxSprite
 			transformChildren(pixelPerfectTransform, Value);
 		return super.set_pixelPerfectRender(Value);
 	}
+	
+	override function set_camera(camera:FlxCamera):FlxCamera
+	{
+		group.set_camera(camera);
```
Gama11: Seems a bit strange to call the setter directly instead of going through the property?

JoeCreates: Indeed. Normally it's not possible of course as the setter is private. I did it just for symmetry with the other setter calls so it is immediately obvious it was just duplicating the behaviour on the group. I'm not particularly tied to that idea, though. :P

PR #2225, opened 2019-07-14
Put flixel dependencies back into haxelib.json

al-arz:
Apparently those got lost after 5fbca9d4, yet they were present before.


Gama11:
They didn't "get lost", it wasn an intentional change. I think the reasoning was that it would be annoying to auto-install the latest OpenFL / Lime versions if you want to cotinue using Flixel with leagcy versions of them (which you still can).

Other than that, I think there was also some more general issues because Haxelib's dependency management isn't great, but I can't remember what exactly..

al-arz:
Thanks for the clarification! And haxelib is... yeah.

PR #2227, opened 2019-07-23
Custom renaming for FlxAssetPaths

JoeCreates:
Added optional function for renaming in FlxAssetPaths. Example usage: 

```haxe
@:build(flixel.system.FlxAssets.buildFileReferences("assets/sounds", false, null, s -> ~/\.ogg$/.replace(s, "")))
class SoundAssets {}
```


Gama11:
Looks like you forgot to update `FlxAssets.buildFileReferences()`? So that code example wouldn't compile.

What are the use cases in practice? Just getting rid of file extensions?

JoeCreates:
So I did. I copy pasted into github thinking I could save some tedium
forgetting I changed another file. Will update shortly.

Other use cases might be to change the name to match other conventions in
the project, or use camelcase instead of underscores.

PR #2231, opened 2019-08-29
Cannot use super inside inline function in haxe 4

JoeCreates:


Gama11:
Huh? I'm running Haxe 4 locally and CI runs against Haxe dev as well, I haven't noticed any issues.

JoeCreates:
Here is a test for FlxCallbackPoint which was currently missing.

I can't run tests at the moment due to haxelib versions and don't want to faff with haxelibs as I'm working with quite a sensistive setup for production stuff at the moment.

```haxe
class FlxCallbackPointTest extends FlxTest
{
	var point:FlxCallbackPoint;
	var a:Int;

	@Before
	function before()
	{
		point = new FlxCallbackPoint(p -> a = 5);
		a = 0;
	}

	@Test
	function testCallback()
	{
		point.set(10, 10);

		Assert.isTrue(a == 5);
	}
}
```

PR #2238, opened 2019-10-24
Fix #2237 (also fix some parameter names in FlxMouse according to code style)

### LINKED ISSUE #2237: Mouse is visible after `FlxG.mouse.visible=false` if the game starts unfocused
- **Haxe version: 4.0.0-rc.5** 
- **Flixel version: 4.6.3** (this bug is also present in current dev branch)
- **OpenFL version: 8.9.5** 
- **Lime version: 7.6.3** 
- **Affected targets: html5** (at least)

```haxe
package;

import flixel.FlxState;
import flixel.FlxG;

{
	{
		super.create();
		FlxG.mouse.visible = false;
	}
}

```

**Observed behavior:**
When starting the game in html5 target, if you click the navigation bar **before** the game loads and then click the game after the game started, the cursor will show even if you ran `FlxG.mouse.visible = false` on your state `create()` function.

**Expected behavior:**
The cursor should remain hidden.

The reason that happens is because `FlxMouse.onFocus()` uses `visible = _visibleWhenFocusLost;` and that function is called before a call to `onFocusLost()` which should set the `_visibleWhenFocusLost` field to the value `visible` had before, and the initial value for `_visibleWhenFocusLost` is `true`.

One possible way to fix this is to change the value of `_visibleWhenFocusLost` to the value of ` in the `visible` property setter.

raulfpl:
I added `_visibleWhenFocusLost = value;` to the `visible` setter in FlxMouse to fix #2237.

Also changed the names of many methods parameters to lowerCamelCases to match the [Code Style](http://haxeflixel.com/documentation/code-style/#parameter-names)

### DIFF
--- a/flixel/input/mouse/FlxMouse.hx
+++ b/flixel/input/mouse/FlxMouse.hx
@@ -650,6 +650,8 @@ class FlxMouse extends FlxPointer implements IFlxInputManager
 
 	function set_visible(value:Bool):Bool
 	{
+		_visibleWhenFocusLost = value;
+
 		if (value)
 			showCursor();
 		else
--- a/flixel/input/mouse/FlxMouse.hx
+++ b/flixel/input/mouse/FlxMouse.hx
@@ -310,7 +310,6 @@ class FlxMouse extends FlxPointer implements IFlxInputManager
 	 *
 	 * @param   Name         The ID name used for the cursor
 	 * @param   CursorData   MouseCursorData contains the bitmap, hotspot etc
-	 * @param   Show         Whether to call setNativeCursor afterwards
 	 */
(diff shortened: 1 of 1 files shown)


Gama11:
Thanks for the PR! Unfortunately, your fix for #2237 is incorrect since it breaks the intended behavior when the cursor is visible during the focus lost event. The following code in `onFocusLost()` causes `_visibleWhenFocusLost` to _always_ be `false` after the focus was lost.

https://github.com/HaxeFlixel/flixel/blob/300bba46dc66622426ff3d243de27c73db6e0ded/flixel/input/mouse/FlxMouse.hx#L504-L507

I've cherry-picked your other two unrelated commits.

PR #2240, opened 2019-11-05
Change FlxObject properties from FlxPoints to FlxVectors

### LINKED ISSUE #2236: [Suggestion] Change physics vars to FlxVector
Now that FlxVector is just an abstract of FlxPoint, and there's no runtime difference between the 2 types, I was wondering if it would be a good idea to convert flxSprite.velocity and acceleration to FlxVector. 

My assumption is that the only reason it's a FlxPoint right now is because these properties predate FlxVector and (until now) changing it would have broken some games(?). I find myself constantly casting velocity to a vector, and I think FlxVector isn't very well known to newer flixel users. This could increase visibility.

Geokureli:
Changing FlxObject's physics vars to FlxVectors. this is more in-line with they're intended functionality, there's no overhead in this change since FlxVector is an abstract of FlxPoint.

closes #2236


Geokureli:
looks like velocity.x++ fails due to https://github.com/HaxeFoundation/haxe/issues/8930

Gama11:
This was fixed for Haxe 4.1, but that doesn't really help us as it'll most likely be quite a while until that becomes the minimum supported Haxe version (Haxe 4 hasn't been released for that long and we still support Haxe 3.4).

PR #2250, opened 2020-03-27
FlxTilemap: fix crashes with negative tile indices

01010111:
Tilemap functions like `ray()` and `FlxTilemapExt` crash when negative indexes are used. This just forces them all to 0 when loading maps

### DIFF
--- a/flixel/tile/FlxBaseTilemap.hx
+++ b/flixel/tile/FlxBaseTilemap.hx
@@ -283,6 +283,7 @@ class FlxBaseTilemap<Tile:FlxObject> extends FlxObject
 		widthInTiles = WidthInTiles;
 		heightInTiles = HeightInTiles;
 		_data = MapData.copy(); // make a copy to make sure we don't mess with the original array, which might be used for something!
+		for (i in 0..._data.length) if (_data[i] < 0) _data[i] = 0;
 
 		loadMapHelper(TileGraphic, TileWidth, TileHeight, AutoTile, StartingIndex, DrawIndex, CollideIndex);
 		return this;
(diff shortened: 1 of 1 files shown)


Gama11:
I don't think this is the correct place to do that. Negative indices shouldn't even ever make it that far. For `loadMapFromCSV()` we have this:

https://github.com/HaxeFlixel/flixel/blob/28f060b60277fcc054b3662d67b439f0b34f6ff2/flixel/tile/FlxBaseTilemap.hx#L245-L247

Maybe this should be done in `loadMapHelper()` instead so every loading function does it.

chosencharacters:
I had a rudimentary fix when FlxTilemapExt crashed for similar reasons, I was already extending FlxOgmo3Loader so I added a -1 replacement before loading the data. Again this is only an issue with Ogmo 3 so far due to it not assuming 0 for blank tiles: https://github.com/chosencharacters/SlopeLand/blob/master/source/FlxOgmo3LoaderExt.hx

01010111:
> Negative indices shouldn't even ever make it that far.

`loadMapHelper()` happens later on in the stack though?

01010111:
Maybe if we knew all the places this would cause crashes we could patch them there. afaik it's:
- `FlxTilemap.ray()`
- and collision (probably custom slope stuff) in `FlxTilemapExt`
I'm hesitant to do this because I have no idea if there are other places this happens as well, but maybe it's the best solution because it fits more use cases

Gama11:
I'm not sure it's a good idea to introduce the concept of negative tile indices...

If there's no way for negative indices to exist after loading, there won't be any issues for sure.

chosencharacters:
I agree with Gama. Most HF devs are trained on the “0 is blank” standard as
well.

01010111:
alright - moved that check into `loadMapHelper()`!

PR #2252, opened 2020-04-08
Allow decimal frameRates on FlxAnimations

### LINKED ISSUE #2251: Suggestion: Allow decimal frame rate on animations
[FlxAnimation.frameRate](https://github.com/HaxeFlixel/flixel/blob/dev/flixel/animation/FlxAnimation.hx#L13) is an Int, but it's only used to set [FlxAnimation.delay](https://github.com/HaxeFlixel/flixel/blob/dev/flixel/animation/FlxAnimation.hx#L29) to it's inverse, which is a Float. I understand that non-integer frame rates are rare but my game syncs animations to music and currently this is only possible on songs with a beats per minute that is a multiple of 60.

My current workaround is to change `delay` directly (as below). This is not terrible, but it shouldn't be necessary, IMO
```hx
animation.add('idle', [0, 1, 2, 3], 1);
var anim = animation.getByName('idle');
@:privateAccess
anim.delay = BeatGame.beatTime / anim.numFrames;
```

Geokureli:
Simple change from Int to Float

closes #2251

### DIFF
--- a/flixel/animation/FlxAnimation.hx
+++ b/flixel/animation/FlxAnimation.hx
@@ -10,7 +10,7 @@ class FlxAnimation extends FlxBaseAnimation
 	/**
 	 * Animation frameRate - the speed in frames per second that the animation should play at.
 	 */
-	public var frameRate(default, set):Int;
+	public var frameRate(default, set):Float;
 
 	/**
 	 * Keeps track of the current frame of animation.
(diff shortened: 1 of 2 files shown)


Geokureli:
travis didn't check this change against the demos. is there a way to trigger manually?

Gama11:
Well, I could just merge it, then Travis also runs...

PR #2253, opened 2020-04-14
Keyboard gamepad fixes

### LINKED ISSUE #2249: `FlxG.keys.anyJustPressed([ANY])` returns false when `FlxG.keys.justPressed.ANY` return true
```hx
if (FlxG.keys.anyJustPressed([ANY]) != FlxG.keys.justPressed.ANY)
    throw"error";
```

I need this to work because a weird use case I'm dealing with, so I don't mind looking into it later

Geokureli:
Was made to close #2249 but it unveiled some other, related inconsistencies. I can split this up into multiple PRs if needed.

- Made `FlxG.keys.anyPressed([ANY])` return true if any key is pressed, previously it would always return false.
- Make `myGamePad.pressed.A` true when A is either `PRESSED` or `JUST_PRESSED`. Previously, it would be false if just pressed.
- Added `FlxKey.NONE` to FlxKeyList, since game pads have this feature.
- Added `FlxG.keys.released` since game pads had that too.

### DIFF
--- a/flixel/input/FlxKeyManager.hx
+++ b/flixel/input/FlxKeyManager.hx
@@ -148,6 +148,13 @@ class FlxKeyManager<Key:Int, KeyList:FlxBaseKeyList> implements IFlxInputManager
 					case RELEASED: released.ANY;
 					case JUST_RELEASED: justReleased.ANY;
 				}
+			case FlxKey.NONE:
+				switch (Status)
+				{
+					case PRESSED: pressed.NONE;
+					case JUST_PRESSED: justPressed.NONE;
+					case JUST_RELEASED: justReleased.NONE;
+				}
 			default:
 				var key = getKey(KeyCode);
 				if (key == null)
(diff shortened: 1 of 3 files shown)


flixel/input/FlxKeyManager.hx
```diff
@@ -135,25 +141,37 @@ class FlxKeyManager<Key:Int, KeyList:FlxBaseKeyList> implements IFlxInputManager
 	 * @param	Status		The key state to check for.
 	 * @return	Whether the provided key has the specified status.
 	 */
-	public function checkStatus(KeyCode:Key, Status:FlxInputState):Bool
+	public inline function checkStatus(KeyCode:Key, Status:FlxInputState):Bool
```
Gama11, line 144: This function is too complex to be `inline`.

Gama11:
Thanks, these changes all make sense to me. 👍

flixel/input/FlxKeyManager.hx
Geokureli: I mostly did this to match `FlxGamepad.checkStatus` but it should work because checkStatus should always be called with a literal key value through inlined getters like `FlxG.keys.pressed.A` which is inlined and assumedly optimized all the way down the call stack to become:
```
	var key = getKey(KeyCode);
	if (key == null)
	{
		return false;
	}
	return key.hasState(Status);
```

otherwise only checkKeyArrayState should get the full inline. of course, people can call this method, as it's public and that can cause problems. I thought about making it private and adding `@:allow(flixel.input.FlxBaseKeyList)` to it. but that might screw up someone's project.

PR #2256, opened 2020-05-13
FlxText: allow partial range removal with removeFormat()

aeveis:
I noticed the removeFormat in FlxText only checked if the format wasn't in range, and if it was it would remove the whole format. I added checks for it to remove the format with the start and end specified. 

I was running into issues with removing a format while iterating so I added an Array<FlxTextFormatRange> to keep track of the formats to be removed at the end. 

Let me know if this is intended behavior, thanks!

Here are some test cases I made:
```haxe
package;

import flixel.FlxG;
import flixel.FlxState;
import flixel.text.FlxText;
import flixel.util.FlxColor;

class TestState extends FlxState 
{

	{
        super.create();
	
		var blueFormat:FlxTextFormat = new FlxTextFormat(FlxColor.BLUE);
		
		var txt1 = new FlxText(1, FlxG.height / 4, "01234567890");
		var txt2 = new FlxText(1, FlxG.height / 2, "01234567890");
		var txt3 = new FlxText(1, 3 * FlxG.height / 4, "01234567890");
		
		txt1.addFormat(blueFormat, 2, 8);
		txt1.removeFormat(blueFormat, 6, 10);
		txt1.removeFormat(blueFormat, 0, 4);
		
		txt2.addFormat(blueFormat, 0, 5);
		txt2.removeFormat(blueFormat, 3, 11);
		
		txt3.addFormat(blueFormat, 0, 11);
		txt3.removeFormat(blueFormat, 4, 7);
		
		add(txt1);
		add(txt2);
		add(txt3);
    }
	
}
```


flixel/text/FlxText.hx
```diff
@@ -140,6 +140,7 @@ class FlxText extends FlxSprite
 	var _defaultFormat:TextFormat;
 	var _formatAdjusted:TextFormat;
 	var _formatRanges:Array<FlxTextFormatRange> = [];
+	var _toRemoveFormatRanges:Array<FlxTextFormatRange> = [];
```
Gama11, line 143: This should be a local variable, not a field. But there's an even better solution: iterate backwards over `_formatRanges` using a `while` loop, then it's safe to remove while iterating.

aeveis: I pushed a new commit with these changes, thanks!
(note: I found a bug in how openfl handles text formatting in TextField for neko, so will be following up there).

Gama11:
Thanks!

PR #2259, opened 2020-07-02
Joycon map fixes

Geokureli:
fix some mapping issues on the right joycon


flixel/input/gamepad/mappings/SwitchJoyconRightMapping.hx
```diff
@@ -77,19 +77,19 @@ class SwitchJoyconRightMapping extends FlxGamepadMapping
 	}
 
 	override function getInputLabel(id:FlxGamepadInputID)
-	{
+{
```
Gama11, line 80: ? :smile:

Geokureli: odd! I can fix. What is the token to prevent a CI check?

Geokureli:
thanks!

PR #2261, opened 2020-07-13
FlxCamera skip unnecessary draw_rect calls

filipp8:
removed obsolete (?) condition.
things still work with FlxG.cameras.bgColor = 0x0;
tested on neko/win/android/html5

huge performance boost on low end devices


Gama11:
Thanks, I made a 4.8.1 release with this.

PR #2265, opened 2020-07-23
FlxEmitter: add maxSize = Quantity to makeParticles()

dean1010:
I think this was an oversight.


dean1010:
If you create an emitter using 
```
makeParticles(2, 2, 0xFFFFFFFF, 10);
```
and emit more than 10 at one time, it will create a new particle with the HaxeFlixel logo, rather than recycling the 2 x 2 particles.

Gama11:
Seems like it's an oversight indeed: https://github.com/AdamAtomic/flixel/blob/8989e5044be072c4abbbaa1317c9854786f6447f/org/flixel/FlxEmitter.as#L164

Earlier versions of HaxeFlixel had it too, it just seems to have gotten lost after some refactors.

PR #2266, opened 2020-08-11
Enumerate FlxBasic.ID

01010111:
right now `FlxBasic.ID` is unused and not really set up well. We should at least make IDs unique so that people can use them in their own games without having to set them manually.


bhgsbatista:
@Gama11 any reason why we shouldn't merge this?

Gama11:
Technically it could be considered a breaking change in case somebody relies on the current default value in some way, but it's probably not too much of an issue.

Gama11:
Reimplemented this (see linked commit) to fix some stylistic issues, as well as hide the new field from completion and provide a better doc comment.

PR #2267, opened 2020-08-24
FlxTilemap: added null check of tile in ray() function

dean1010:
When targeting html5 the game crashes while trying to read property `tile.allowCollisions` if the tile is null.

```Uncaught TypeError: Cannot read property 'allowCollisions' of undefined```


dean1010:
The best I can tell, with `map.ray(startPoint, endPoint)`, If endPoint is out of worldBounds you end up with a null tile.

I suppose the points should just be checked before passing to ray(), so the null check won't be necessary.

PR #2268, opened 2020-08-29
add Tags to `FlxObject`

AustinEast:
This PR adds a simple Tagging API to help enable a more compositional approach with FlxObjects. The Tags are just Strings that get stored in the private `_tags` Array, which the user interacts with through `addTag`, `removeTag`, and `hasTag` functions. 

Note: The `_tags` array is lazily initialized, so it won't add any real memory bloat to FlxObjects unless the user is using Tags.

### DIFF
--- a/flixel/FlxObject.hx
+++ b/flixel/FlxObject.hx
@@ -638,6 +638,9 @@ class FlxObject extends FlxBasic
 	@:noCompletion
 	var _rect:FlxRect = FlxRect.get();
 
+	@:noCompletion
+	var _tags:Null<Array<String>>;
+
 	/**
 	 * @param   X        The X-coordinate of the point in space.
 	 * @param   Y        The Y-coordinate of the point in space.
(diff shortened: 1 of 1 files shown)


Gama11:
I don't know, this seems pretty random...

AustinEast:
I disagree! Tagging systems are fairly common in game development, and there are plenty of applications for them, like filtering collision responses and easy object querying. 

One of the biggest benefits though (in my opinion), is that adding tags can be a step towards making Flixel a more compositional framework - if you compose objects with tags, its makes it very easy to run behaviors on them without the need to subclass.

01010111:
I've implemented the same kind of thing in several projects, I would definitely benefit from a change like this!

Gama11:
I feel like having this in core flixel would encourage stringly typed code, which I strongly dislike (using raw strings for things where something like an enum would be a better fit).

I don't think it really _needs_ to be here either, you could get the exact same API with static extensions + a static `Map<FlxObject, Array<String>` if you want it.

Gama11:
I'm still not convinced this is worth adding, considering FlxObject arguably already does too many things.

AustinEast:
Sounds good. It was worth a shot though 😅

PR #2273, opened 2020-10-15
Add cancelTweensOf and completeTweensOf

### LINKED ISSUE #2180: killTweensOf
Continuing off of #1458. This is a pretty common feature of tween libraries and can be done relatively easily.

I'm a huge fan of Greensock's flash tween library which looked like:
```as3
/**
 * @param target    Object whose tweens should be killed immediately or selector 
 *     text to feed the selector engine to find the target(s). You may also pass in an 
 *     array of targets.
 * @param vars      (default = null) - To kill only specific properties, use a generic
 *      object containing enumerable properties corresponding to the ones that 
 *     should be killed like {x:true, y:true}. The values assigned to each property
 *     don’t matter – the sole purpose of the object is for iteration over the named 
 *     properties (in this case, x and y). If no object (or null) is defined, all matched 
 *     tweens will be killed in their entirety.
 * @param complete  (default = false) - Determines whether or not the
 *     tweens/delayedCalls/timelines should be forced to completion before being killed.
**/
static public function killTweensOf(target:Object, vars:Object = null, complete:Boolean = false):void
```
I don't think FlxTween has the capability to complete() a tween early, so lets ignore that property

I've made a [mockup of the feature](https://github.com/HaxeFlixel/flixel/compare/dev...Geokureli:killTweensOf) (untested, as of now, but ready for feedback). Yes there's better, more optimized ways of doing this than looking through the whole list each time, but I feel like this will address the common use cases.

Geokureli:
Added the ability to cancel or complete specific tweens without needing a reference to the tweens themselves. Useful when multiple things are applying tweens to the same object and want to remain agnostic of each other.

Usage:
```hx
FlxTween.tween(mySprite, { color: FlxColor.BLACK }, 1.0);

//... elsewhere ...
FlxTween.cancelTweensOf(mySprite);

// or, to only cancel color tweens on mySprite
FlxTween.cancelTweensOf(mySprite, "color");
```

also works with tweens that are not `VarTweens` by treating them as `VarTweens` based on the properties they change, such as:
AngleTween: "angle":
ColorTween: "color"
Motion (and all derivatives): "x" and "y"

`completeTweensOf` works identical, but finishes the tween and calls onComplete

Just like with `VarTweens`, properties passed in can be dot paths to denote properties of children, this way any field passed in to a var tween can be used to cancel or complete the tween
Example:
```hx
// either
FlxTween.tween(mySprite, { "child.color": FlxColor.BLACK }, 1.0);
// or
FlxTween.color(mySprite.child, 1.0, FlxColor.BLACK, FlxColor.RED);

// can be cancelled via:
FlxTween.cancelTweensOf(mySprite, ["child.color"]);
// or
FlxTween.cancelTweensOf(mySprite.child, ["color"]);
```

fixes #2180


Geokureli:
I should add unit tests to this pr, hold up

flixel/tweens/FlxTween.hx
```diff
@@ -622,6 +661,17 @@ class FlxTween implements IFlxDestroyable
 		}
 	}
 
+	/**
+	 * returns true if this is tweening any of the specified fields on the object
+	 * 
+	 * @param Object The object
+	 * @param Fields Optional list of tween fields. If empty, any tween field is matched
```
Gama11, line 668: This doc seems inaccurate, maybe referring to an older version of this method?

flixel/tweens/FlxTween.hx
```diff
@@ -307,6 +307,45 @@ class FlxTween implements IFlxDestroyable
...
 
+	/**
+	 * Cancels all related tweens on the specified object.
+	 *
+	 * Note: Any tweens with the specified fields are cancelled, if the tween has other properties they
+	 * will also be cancelled.
+	 * 
+	 * @param Object The object with tweens to cancel.
+	 * @param FieldPaths Optional list of the tween field paths to search for. If empty, all tweens on the specified
+	 * object are canceled. Allows dot paths to check child properties.
```
Gama11, line 318: Any new public API should have a `@since 4.9.0` comment.

flixel/tweens/misc/VarTween.hx
```diff
@@ -37,6 +37,7 @@ class VarTween extends FlxTween
 		_propertyInfos = [];
 		this.duration = duration;
 		start();
+		initializeVars();
```
Gama11, line 40: Are any of the changes in `VarTween` changes besides the `isTweenOf()` override related to this PR? It's not clear to me why they were made.

flixel/tweens/FlxTween.hx
```diff
@@ -1120,14 +1170,120 @@ class FlxTweenManager extends FlxBasic
...
+		);
+	}
+
+	/**
+	 * Internal helper for iterating tweens with specific parameters.
+	 *
+	 * Note: loops backwards to allow removals.
+	 *
+	 * @param Object The object with tweens you are searching for.
+	 * @param Fields Optional list of the tween field paths to check. If empty, any tween of the specified
```
Gama11, line 1224: The parameter name doesn't match the `@param` doc here.

flixel/tweens/misc/VarTween.hx
```diff
@@ -81,30 +81,55 @@ class VarTween extends FlxTween
...
 		_object = null;
 		_properties = null;
 		_propertyInfos = null;
 	}
+
+	/**
+	 * returns true if this is tweening any of the specified fields on the object
+	 * 
+	 * @param Object The object
+	 * @param Fields Optional list of tween fields. If empty, any tween field is matched
```
Gama11, line 121: Again, the actual parameter is not a list.

flixel/tweens/FlxTween.hx
```diff
@@ -1120,14 +1170,120 @@ class FlxTweenManager extends FlxBasic
...
+	 * Note: their `onComplete` callbacks are triggered in the next frame.
+	 * To trigger them immediately, call `FlxTween.globalManager.update(0);` after this function.
+	 *
+	 * In no case should it trigger an `onUpdate` callback.
+	 *
+	 * Note: Any tweens with the specified fields are completed, if the tween has other properties they
+	 * will also be completed.
+	 *
+	 * @param Object The object with tweens to complete.
+	 * @param FieldPaths Optional list of the tween field paths to search for. If empty, all tweens on
```
Gama11, line 1204: If empty, or if null? Or both? A bit strange that the doc comment only talks about the empty case when the field is optional.

Gama11 (changes requested):

flixel/tweens/FlxTween.hx
Geokureli: Gonna change all of these to say "If null or empty" and will change `forEachTweensOf` to check for empty

flixel/tweens/misc/VarTween.hx
Geokureli: It calls initializeVars earlier than before so that we can check `_propertyInfos` in `isTweenOf()`. originally, this initialization was delayed until the tween starts so that chaining works correctly, but that's only needed for the startValues.

If you think I need to explain this in code, let me know

PR #2276, opened 2020-11-18
FlxSound.loadStream callback

### LINKED ISSUE #2274: Load event via FlxSound.loadStream
When a `FlxSound` is loaded using `loadStream`, I can't call `play()` until the sound is finished loading, but there's no way tell when it's done loading, and there's no reference to the openfl Sound so that I can add a load listener.

I suggest we add an onLoad callback to loadStream or something similar, otherwise it's not reliably usable when it takes time to asynchronously load sounds.

for now, as a workaround I can do
```hx
@:privateAccess
sound._sound.addEventListener(openfl.events.Event, onSoundLoad);
```

Geokureli:
Added an onLoad call to FlxSounds that get called when sounds are loaded from a URL. OnLoad was also added as an argument to methods like `loadStream` `FlxG.sound.load` and `FlxG.sound.stream`.

I added a `loadHelper` method to `SoundFrontEnd` to perform common setup for load/play/stream methods.

`SoundFrontEnd` methods `load` and `stream` now listen for load events to autoPlay the sound when it's set to true, rather than calling play immediately, since there's no sound to play

fixes #2274 

Also fixed some documentation and logging for the methods that were modified. they were all missing some @params and made references to old methods like `FlxG.sound.loadSound`

### DIFF
--- a/flixel/system/frontEnds/SoundFrontEnd.hx
+++ b/flixel/system/frontEnds/SoundFrontEnd.hx
@@ -132,7 +132,7 @@ class SoundFrontEnd
 		}
 
 		var sound:FlxSound = list.recycle(FlxSound);
-		
+
 		if (EmbeddedSound != null)
 		{
 			sound.loadEmbedded(EmbeddedSound, Looped, AutoDestroy, OnComplete);
@@ -157,7 +157,7 @@ class SoundFrontEnd
 				sound._sound.addEventListener(Event.COMPLETE, playStream, false, 1, true);
 			}
 		}
-		
+
 		return sound;
 	}
 
@@ -221,9 +221,7 @@ class SoundFrontEnd
 		{
 			EmbeddedSound = cache(EmbeddedSound);
 		}
-		
-		var sound:FlxSound = list.recycle(FlxSound);
-		sound.loadEmbedded(EmbeddedSound, Looped, AutoDestroy, OnComplete);
+		var sound = list.recycle(FlxSound).loadEmbedded(EmbeddedSound, Looped, AutoDestroy, OnComplete);
 		return loadHelper(sound, Volume, Group, true);
 	}
 
--- a/flixel/system/frontEnds/SoundFrontEnd.hx
+++ b/flixel/system/frontEnds/SoundFrontEnd.hx
@@ -127,7 +127,7 @@ class SoundFrontEnd
 	{
 		if ((EmbeddedSound == null) && (URL == null))
 		{
-			FlxG.log.warn("FlxG.load() requires either\nan embedded sound or a URL to work.");
+			FlxG.log.warn("FlxG.sound.load() requires either\nan embedded sound or a URL to work.");
 			return null;
 		}
 
@@ -227,7 +227,7 @@ class SoundFrontEnd
 
 	/**
 	 * Plays a sound from a URL. Tries to recycle a cached sound first.
-	 * NOTE: Just calls FlxG.load() with AutoPlay == true.
+	 * NOTE: Just calls FlxG.sound.load() with AutoPlay == true.
 	 *
 	 * @param	URL				Load a sound from an external web resource instead.
 	 * @param	Volume			How loud to play it (0 to 1).
--- a/flixel/system/frontEnds/SoundFrontEnd.hx
+++ b/flixel/system/frontEnds/SoundFrontEnd.hx
@@ -114,9 +114,11 @@ class SoundFrontEnd
 	 * @param	Volume			How loud to play it (0 to 1).
 	 * @param	Looped			Whether to loop this sound.
 	 * @param	Group			The group to add this sound to.
-	 * @param	AutoDestroy		Whether to destroy this sound when it finishes playing.  Leave this value set to "false" if you want to re-use this FlxSound instance.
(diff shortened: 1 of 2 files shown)


flixel/system/frontEnds/SoundFrontEnd.hx
```diff
@@ -112,17 +114,20 @@ class SoundFrontEnd
 	 * @param	Volume			How loud to play it (0 to 1).
 	 * @param	Looped			Whether to loop this sound.
 	 * @param	Group			The group to add this sound to.
-	 * @param	AutoDestroy		Whether to destroy this sound when it finishes playing.  Leave this value set to "false" if you want to re-use this FlxSound instance.
+	 * @param	AutoDestroy		Whether to destroy this sound when it finishes playing.
+	 * 							Leave this value set to "false" if you want to re-use this FlxSound instance.
 	 * @param	AutoPlay		Whether to play the sound.
 	 * @param	URL				Load a sound from an external web resource instead.  Only used if EmbeddedSound = null.
+	 * @param	OnComplete		Called when the sound finished playing
+	 * @param	OnLoad			Called when the sound finished loading.  Only used if EmbeddedSound = null
```
Gama11, line 122: >Only used if EmbeddedSound = null

I guess that's debatable. It could be convenient to call it immediately in that case?

flixel/system/frontEnds/SoundFrontEnd.hx
```diff
@@ -112,17 +114,20 @@ class SoundFrontEnd
...
 	 * @return	A FlxSound object.
 	 */
 	public function load(?EmbeddedSound:FlxSoundAsset, Volume:Float = 1, Looped:Bool = false, ?Group:FlxSoundGroup, AutoDestroy:Bool = false,
-			AutoPlay:Bool = false, ?URL:String, ?OnComplete:Void->Void):FlxSound
+			AutoPlay:Bool = false, ?URL:String, ?OnComplete:Void->Void, ?OnLoad:Void->Void):FlxSound
 	{
 		if ((EmbeddedSound == null) && (URL == null))
 		{
-			FlxG.log.warn("FlxG.loadSound() requires either\nan embedded sound or a URL to work.");
+			FlxG.log.warn("FlxG.load() requires either\nan embedded sound or a URL to work.");
```
Geokureli, line 130: looks like I should have put `FlxG.sound.load()`

flixel/system/frontEnds/SoundFrontEnd.hx
```diff
@@ -179,41 +205,44 @@ class SoundFrontEnd
...
-		sound.group = (Group == null) ? defaultSoundGroup : Group;
-		return sound.play();
+		return loadHelper(sound, Volume, Group, true);
 	}
 
 	/**
-	 * Creates a new sound object from a URL.
-	 * NOTE: Just calls FlxG.loadSound() with AutoPlay == true.
+	 * Plays a sound from a URL. Tries to recycle a cached sound first.
+	 * NOTE: Just calls FlxG.load() with AutoPlay == true.
```
Geokureli, line 230: here too

flixel/system/frontEnds/SoundFrontEnd.hx
Geokureli: I went back and forth on it. looking at it now, the method is called `load()` so I'm sure they would expect onLoad to fire for both cases. I'll add it

flixel/system/frontEnds/SoundFrontEnd.hx
```diff
@@ -136,7 +136,11 @@ class SoundFrontEnd
 		if (EmbeddedSound != null)
 		{
 			sound.loadEmbedded(EmbeddedSound, Looped, AutoDestroy, OnComplete);
+			sound.onLoad = OnLoad();
 			loadHelper(sound, Volume, Group, AutoPlay);
+			// Call OnlLoad() because the sound already loaded
+			if (OnLoad != null && sound._sound != null)
+				OnLoad();
```
Gama11, line 143: Is it correct to use this here? What if somebody uses FlxSound without SountFrontEnd?

Geokureli: I don't know if I follow your question. This is the only method where getting an embedded sound has an OnLoad argument. if they use `mySound.loadEmbedded` there's no use for an onLoad callback

Gama11: `FlxSound` always has an `onLoad` field you can set.

Geokureli: onComplete and onLoad get nulled when loadStream or loadEmbedded are called (I just copied how onComplete works). so there's no actual way to get onLoad to trigger when using loadEmbedded. Except through `FlxG.sound.load`
maybe I should rethink this? someone could try to set sound.onLoad() before loadEmbedded.

I think the best option is simply to remove `sound.onLoad` and call the OnLoad callback passed in without storing it. I only added it because I thought new callbacks should behave like existing callbacks, but it's not really needed here. and  things are just simpler without it.

Gama11: Simple sounds good.

PR #2280, opened 2020-12-01
Add new subStateOpened & subStateClosed signals

### LINKED ISSUE #2279: No signal for when subState has been opened / closed.
Hi,

It would be advantageous to know when a subState has been opened/closed but not in the openCallback of the substate itself but from its parent FlxState - similar to `memberAdded`/`memberRemoved` - currently im working around this by using an `FlxInputManager` to know when the state has changed then (even though its a sub state) and "stealing" the openCallback, similar to this:

```haxe
class InputManager implements IFlxInputManager {
    public function reset():Void {
        if (FlxG.state != null && FlxG.state.subState != null) {
            var cachedSubStateOpenedCallback = FlxG.state.subState.openCallback;
            FlxG.state.subState.openCallback = function() {
                trace("substate opened");
                if (cachedSubStateOpenedCallback != null) {
                    cachedSubStateOpenedCallback();
                }
            }
        }
    }
}
```
This works, but feels a little hacky, moreover, this _doesnt_ work if the parent FlxState has `persistentUpdate = true` because of:

```haxe
            if (!persistentUpdate)
                FlxG.inputs.onStateSwitch();
```
In `FlxState`, which is fine as in my workaround im misusing the `IFlxInputManager::reset` to know when the sub state is about to be created (so i can inject my own code into the `openCallback`). 

Im happy to make a PR for this which would be a new signal on `FlxState`, `subStateOpened` / `subStateClosed` which would fire at the very end of when the sub state has been created and shown. 

Is this an acceptable solution and if it is, is there a reason why it doesnt already exist? Some kind of gotcha?

Cheers,
Ian

ianharrigan:
closes #2279

### DIFF
--- a/flixel/FlxState.hx
+++ b/flixel/FlxState.hx
@@ -61,11 +61,13 @@ class FlxState extends FlxGroup
 
 	/**
 	 * A `FlxSignal` that dispatches when a sub state is opened from this state.
+	 * @since 4.9.0
 	 */
 	public var subStateOpened(get, never):FlxTypedSignal<FlxSubState->Void>;
 
 	/**
 	 * A `FlxSignal` that dispatches when a sub state is closed from this state.
+	 * @since 4.9.0
 	 */
 	public var subStateClosed(get, never):FlxTypedSignal<FlxSubState->Void>;
 
--- a/flixel/FlxState.hx
+++ b/flixel/FlxState.hx
@@ -2,6 +2,8 @@ package flixel;
 
 import flixel.group.FlxGroup;
 import flixel.util.FlxColor;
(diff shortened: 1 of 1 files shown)


ianharrigan:
```
  Error: Unable to process command '::set-env name=HAXELIB_REPO::/home/runner/haxe/haxelib' successfully.
  Error: The `set-env` command is disabled. Please upgrade to using Environment Files or opt into unsecure command execution by setting the `ACTIONS_ALLOW_UNSECURE_COMMANDS` environment variable to `true`. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/
```

Issue #2285, opened 2020-12-05
FlxBitmapFont.fromXNA causes an infinite loop on Brave Browser


Geokureli:
More info coming soon, but it's definitely getting stuck here:
https://github.com/HaxeFlixel/flixel/blob/dev/flixel/graphics/frames/FlxBitmapFont.hx#L384

Geokureli:
Furthermore, I'm seeing this because the default font apparently uses fromXNA, considering it uses these automatic pixel detection methods, perhaps we should swap it to something that relies more on hand crafted data, since graphic tools can vary widely?

Geokureli:
I should mention that I don't see this issue in Brave when testing locally, nor do I get this issue when I run in Brave with the shield disabled. No effin clue why that is.

Geokureli:
PR branch was closed; I plan to make a new PR with this branch's infinite loop prevention, later. I still say that should go in, even if the farbling detection doesn't. longer term we may want to reconsider using `fromXNA` for the default font as well

Gama11:
Yes, I'm much more open to that. The part I was uncomfortable with was mostly adding new public API for a niche browser.

PR #2286, opened 2020-12-05
check farbling, prevent infinite loops and warn the dev

Geokureli:
I'm all ears for a better name than `FlxG.html.farblesImages`, all my ideas were stupidly long. It comes from [this article by Brave](https://brave.com/privacy-updates-4/).

This is not a perfect solution, but IMO it is a decent temporary one. In the long run we should either get the default font some other way than `fromXNA`, perhaps `fromAngleCode`.

related to #2285

### DIFF
--- a/flixel/graphics/frames/FlxBitmapFont.hx
+++ b/flixel/graphics/frames/FlxBitmapFont.hx
@@ -108,6 +108,9 @@ class FlxBitmapFont extends FlxFramesCollection
 
 	/**
 	 * Retrieves the default `FlxBitmapFont`.
+	 * May work incorrectly on HTML5.
+	 * Utterly unreliable on Brave Browser with shields up.
+	 * @see `FlxG.html.farblesImages`
 	 */
 	public static function getDefaultFont():FlxBitmapFont
 	{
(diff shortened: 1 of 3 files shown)


MSGhero:
Well that's annoying

Gama11:
Does Brave have enough of a market share that this matters?

Geokureli:
I'm getting a lot a people reporting brave errors to me, I think it's gaining popularity and even if it doesn't I think anti-fingerprint practices like this are becoming a trend. I know little of how browsers work, but I fear that internet security software or plugins could bring this problem to more common browsers

Geokureli:
https://en.wikipedia.org/wiki/Usage_share_of_web_browsers
Brave numbers are likely low. but it's hard to tell because `Browser.navigator.userAgent` doesn't even contain the word brave in it when playing on brave.

also this PR isn't blocking me, now that I know how to detect farbling I can display a message to the user to turn off shields

MSGhero:
The infinite loop check should happen regardless. What about if the `while` hits the "infinite" condition, then we know(?) it's farbled and then warn or error? It seems like something that should be on the openfl level tbh

Geokureli:
yeah, we could have it check when it getPixels beyond the bounds of the image and warn. I'm not familiar enough with this code to know whether there would be false positives, though

Gama11:
I'm inclined to close this for now as long as we don't have more reports of this being an issue or it affects more browsers. From my perspective it seems like a rather hacky workaround specifically for a browser with a tiny market share.

PR #2291, opened 2021-01-25
prevent null _transform in FlxSound

SerissaCodeCat:
added a null check to ensure _transform has a value prior to operating on it. use-case that caused this was that the sound levels were changed in a sub-state, the sub-state exited, and then a new state entered. this caused a CTD due to a null reference exception


Gama11:
I would like to understand the underlying issue better to make sure this doesn't just fix the symptom. Some code that reproduces the problem would help in that regard.

Gama11:
Closing as there's still no reproducible sample here.

PR #2296, opened 2021-02-25
Improve default cameras handling

Geokureli:
### Problem
Often times, when I add a second camera, my intention is either:
A. To have both cameras draw everything if certain portions of the screen (i.e.: splitscreen).
B. To have the new camera draw one set of things, often a specific FlxGroup, and everything else render on the original camera.

Scenario A is pretty straightforward however B is not. It usually involves setting `FlxCamera.defaultCameras = [cameraA]` and just kinda hoping that no other parts of my code are doing anything crazy with multiple cameras. Not to mention setting `FlxCamera.defaultCameras` during a `draw` phase could have unintended side effects (or just not work at all).

### Solution
Give `CameraFrontEnd` methods to set and unset which cameras are "default" drawers. default drawing cameras will act as though it was listed in `defaultCameras` at the start of the draw phase. This way you can add new cameras without worrying about stomping on other code involving cameras

### Backwards Compatibility
The old way should still work but documentation will be updated to discourage its use and inform them of the new property

### Info
This went through a few iterations, previously I was checking/validating the default Array every draw cycle. but it seems important to update the list right away because `FlxCamera.defaultCameras` can be accessed at any time via `myCamera.cameras` since `set_camera` returns  `FlxCamera.defaultCameras` when `_cameras` is null.

Edit: even more iterations!

## Examples

Currently, when you want to add a camera that only displays a specific thing, this is how you do it.
```hx
menuCamera = new FlxCamera();
FlxG.cameras.add(menuCamera);
menu.camera = [menuCamera];
// only draw the main camera by default
// Note: assumes menuCamera was the only camera added
FlxCamera.defaultCameras = [FlxG.camera];
```
Or you could try to account for other camera-related code with:
```hx
menuCamera = new FlxCamera();
FlxG.cameras.add(menuCamera);
menu.camera = [menuCamera];
if (FlxCamera.defaultCameras == FlxG.cameras.list)
{
    FlxCamera.defaultCameras = FlxCamera.defaultCameras.concat();
    FlxCamera.defaultCameras.remove(menuCamera);
}
```

New way:
```hx
menuCamera = new FlxCamera();
FlxG.cameras.add(menuCamera, false);
menu.camera = [menuCamera];
```

and to have the above but also do something like split-screen, you would need to do something like this:
```hx
p1Cam = FlxG.camera;
p1Cam.width = 0.5 * FlxG.width;
p2Cam = new FlxCamera(p1Cam.width, 0, p1Cam.width);
FlxG.cameras.add(p2Cam);
if (FlxCamera.defaultCameras.indexOf(p2Cam) == -1)
{
    // defaultCameras is not set to FlxG.cameras.list
    FlxCamera.defaultCameras.add(p2Cam);
}
```

New way:
```hx
p1Cam = FlxG.camera;
p1Cam.width = 0.5 * FlxG.width;
p2Cam = new FlxCamera(p1Cam.width, 0, p1Cam.width);
FlxG.cameras.add(p2Cam);
```

### DIFF
--- a/flixel/FlxCamera.hx
+++ b/flixel/FlxCamera.hx
@@ -224,7 +224,7 @@ class FlxCamera extends FlxBasic
 	public var zoom(default, set):Float;
 
 	/**
-	 * Whether sprites use this object by default.
+	 * Whether sprites render to this camera by default.
 	 * Set this to false when adding additional cameras meant to display specific sprites
 	 * @since 4.8.2
 	 */
--- a/flixel/FlxCamera.hx
+++ b/flixel/FlxCamera.hx
@@ -1888,7 +1888,7 @@ class FlxCamera extends FlxBasic
 	
 	function set_isDefault(value:Bool):Bool
 	{
-		this.isDefault = value
+		this.isDefault = value;
 		isDefaultChange.dispatch(this);
 		
 		return this.isDefault;
--- a/flixel/FlxCamera.hx
+++ b/flixel/FlxCamera.hx
@@ -54,8 +54,10 @@ class FlxCamera extends FlxBasic
 	 * Used behind-the-scenes during the draw phase so that members use the same default
 	 * cameras as their parent.
 	 * 
-	 * Prior to 4.8.2 it was useful to change this value, but that feature is deprecated.
-	 * Instead use `FlxG.cameras.setDrawsDefault`.
+	 * Prior to 4.9.0 it was useful to change this value, but that feature is deprecated.
+	 * Instead use the `defaultDrawTarget` argument in `FlxG.cameras.add `.
+	 * or`FlxG.cameras.setDefaultDrawTarget` .
+	 * @see FlxG.cameras.setDefaultDrawTarget
 	 */
 	@:deprecated("`FlxCamera.defaultCameras` is deprecated, use `FlxG.cameras.setDrawsDefault()` instead")
 	public static var defaultCameras(get, set):Array<FlxCamera>;
@@ -67,7 +69,10 @@ class FlxCamera extends FlxBasic
 	 * This is the non-deprecated list that the public `defaultCameras` proxies. Allows flixel classes
 	 * to use it without warning.
 	 */
-	@:allow(flixel)
+	@:allow(flixel.FlxBasic.get_cameras)
+	@:allow(flixel.FlxBasic.get_camera)
+	@:allow(flixel.system.frontEnds.CameraFrontEnd)
+	@:allow(flixel.group.FlxTypedGroup.draw)
 	static var _defaultCameras:Array<FlxCamera>;
 
 	/**
--- a/flixel/FlxCamera.hx
+++ b/flixel/FlxCamera.hx
@@ -20,7 +20,6 @@ import flixel.system.FlxAssets.FlxShader;
 import flixel.util.FlxAxes;
 import flixel.util.FlxColor;
 import flixel.util.FlxDestroyUtil;
-import flixel.util.FlxSignal;
 import flixel.util.FlxSpriteUtil;
 import openfl.display.BlendMode;
 import openfl.filters.BitmapFilter;
--- a/flixel/FlxCamera.hx
+++ b/flixel/FlxCamera.hx
@@ -52,8 +52,10 @@ class FlxCamera extends FlxBasic
 
 	/**
 	 * Which cameras a `FlxBasic` uses to be drawn on when nothing else has been specified.
-	 * By default, this is just a reference to `FlxG.cameras.list` / all cameras, but it can be very useful to change.
+	 * Set to `null` to use all cameras where `camera.isDefault == true`.
 	 * This may be temporarily modified during a `draw` call.
+	 *
+	 * Note: starting 4.8.2 this is `null` by default instead of `FlxG.cameras.list`
 	 */
 	public static var defaultCameras:Array<FlxCamera>;
 
@@ -220,6 +222,13 @@ class FlxCamera extends FlxBasic
 	 * Indicates how far the camera is zoomed in.
 	 */
 	public var zoom(default, set):Float;
+
+	/**
+	 * Whether sprites use this object by default.
+	 * Set this to false when adding additional cameras meant to display specific sprites
+	 * @since 4.8.2
+	 */
+	public var isDefault:Bool = true;
 
 	/**
 	 * Difference between native size of camera and zoomed size, divided in half
--- a/flixel/FlxCamera.hx
+++ b/flixel/FlxCamera.hx
@@ -59,7 +59,7 @@ class FlxCamera extends FlxBasic
 	 * or`FlxG.cameras.setDefaultDrawTarget` .
 	 * @see FlxG.cameras.setDefaultDrawTarget
 	 */
-	@:deprecated("`FlxCamera.defaultCameras` is deprecated, use `FlxG.cameras.setDrawsDefault()` instead")
+	@:deprecated("`FlxCamera.defaultCameras` is deprecated, use `FlxG.cameras.setDefaultDrawTarget` instead")
 	public static var defaultCameras(get, set):Array<FlxCamera>;
 	
 	/**
--- a/flixel/FlxCamera.hx
+++ b/flixel/FlxCamera.hx
@@ -51,9 +51,8 @@ class FlxCamera extends FlxBasic
 	public static var defaultZoom:Float;
 
 	/**
-	 * Which cameras a `FlxBasic` uses to be drawn on when nothing else has been specified.
-	 * By default, this is just a reference to `FlxG.cameras.defaults` / all default cameras.
-	 * This may be temporarily modified during a `draw` call.
+	 * Used behind-the-scenes during the draw phase so that members use the same default
+	 * cameras as their parent.
 	 * 
 	 * Prior to 4.8.2 it was useful to change this value, but that feature is deprecated.
 	 * Instead use `FlxG.cameras.setDrawsDefault`.
@@ -62,7 +61,10 @@ class FlxCamera extends FlxBasic
 	public static var defaultCameras(get, set):Array<FlxCamera>;
 	
 	/**
-	 * The non-deprecated list of cameras that defaultCamera proxies. Allows flixel classes
+	 * Used behind-the-scenes during the draw phase so that members use the same default
+	 * cameras as their parent.
+	 * 
+	 * This is the non-deprecated list that the public `defaultCameras` proxies. Allows flixel classes
 	 * to use it without warning.
 	 */
 	@:allow(flixel)
--- a/flixel/FlxCamera.hx
+++ b/flixel/FlxCamera.hx
@@ -56,10 +56,18 @@ class FlxCamera extends FlxBasic
 	 * By default, this is just a reference to `FlxG.cameras.defaults` / all default cameras.
 	 * This may be temporarily modified during a `draw` call.
 	 * 
-	 * Prior to 4.8.2 it was useful to change this value, but that is no longer recommended.
-	 * Instead use the `isDefualt` property on the FlxCamera.
+	 * Prior to 4.8.2 it was useful to change this value, but that feature is deprecated.
+	 * Instead use the `isDefualt` property on `FlxCamera` instances.
 	 */
-	public static var defaultCameras:Array<FlxCamera>;
+	@:deprecated("`FlxCamera.defaultCameras` is deprecated, use `myFlxCamera.isDefault` instead")
+	public static var defaultCameras(get, set):Array<FlxCamera>;
+	
+	/**
+	 * The non-deprecated list of cameras that defaultCamera proxies. Allows flixel classes
+	 * to use it without warning.
+	 */
+	@:allow(flixel)
+	static var _defaultCameras:Array<FlxCamera>;
 
 	/**
 	 * The X position of this camera's display. `zoom` does NOT affect this number.
@@ -1906,6 +1914,16 @@ class FlxCamera extends FlxBasic
 		viewOffsetY = 0.5 * height * (scaleY - initialZoom) / scaleY;
 		viewOffsetHeight = height - viewOffsetY;
 		viewHeight = height - 2 * viewOffsetY;
+	}
+	
+	static inline function get_defaultCameras():Array<FlxCamera>
+	{
+		return _defaultCameras;
+	}
+	
+	static inline function set_defaultCameras(value:Array<FlxCamera>):Array<FlxCamera>
+	{
+		return _defaultCameras = value;
 	}
 }
 
--- a/flixel/FlxCamera.hx
+++ b/flixel/FlxCamera.hx
@@ -59,7 +59,7 @@ class FlxCamera extends FlxBasic
 	 * Prior to 4.8.2 it was useful to change this value, but that feature is deprecated.
 	 * Instead use the `isDefualt` property on `FlxCamera` instances.
 	 */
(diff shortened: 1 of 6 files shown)


chosencharacters:
This would really help with Renaine, cameras have always been a rough spot to work with and this is a big chunk of the reason why :x

Geokureli:
looking through the failed CI checks, not sure why they are failing. I'm not done testing, I know I need to update [FlxCameraTest.hx](https://github.com/HaxeFlixel/flixel/blob/b38c74b85170d7457353881713a796310187ddd2/tests/unit/src/flixel/FlxCameraTest.hx), and likely some demos, if you could list any related things to test, it would be appreciated.

I'm also down for any name change suggestions

Geokureli:
Looks like the CI is failing on stable because haxeflixel/openfl/lime are not compatible with haxe 4.2.0. I don't get why nightly is succeeding, but ok. I'll try more tests locally and then submit this PR. Hopefully it will be ready today

Geokureli:
Checked the following Demos:
- Mode
- FlxCamera
- SplitScreen
- Filters

Gama11:
>Scenario A is pretty straightforward however B is not. It usually involves setting `FlxCamera.defaultCameras = [cameraA]` and just kinda hoping that no other parts of my code are doing anything crazy with multiple cameras.

Could you elaborate on this point a bit? I feel like I don't fully understand the benefits / issues you had that led you to this change.

Right now this merely looks a bit more convenient, which unfortunately has to be weighed against the increased complexity of having two ways of accomplishing the same thing since we can't simply remove the old approach.

>Not to mention setting `FlxCamera.defaultCameras` during a draw phase could have unintended side effects (or just not work at all).

Is that a real concern? Overriding `draw()` is quite rare to begin with, and there's a lot of other things you probably shouldn't modify during it either. Plus in the worst case unintended rendering should only happen for a single frame.

Geokureli:
The main point is that I feel that HaxeFlixel's amazing camera system is made fragile and daunting because of `FlxCamera.defaultCameras`. With the current system, if I add secondary cameras I need to have full knowledge of what other code is potentially adding or removing cameras in this state so that I don't overwrite their intentions. This change allows separate camera-controlling code to be agnostic of one another.

I end up never really using FlxCameras solely because I don't want to deal with FlxCamera.defaultCameras. I would use them a lot more with this change and I think newcomers would feel more comfortable using them too. It's tucked away in a spot that people would never look, in my opinion.

> Is that a real concern?
It may be a small, infrequent concern, but I wouldn't call it insignificant. If a change can enable more of flixel's features during a `draw()` cycle, that's a win. I also don't see any loss of functionality here, even if you avoid the now deprecated field.

I can go over scenarios that would greatly benefit from this change, but in the end it is probably 75% convenience, 20% readability/accessibility and 5% stability. I could make a CameraManager util in my game to cover this but I honestly don't see any benefit to the old system.

I would like to point out, this common pattern, [here](http://forum.haxeflixel.com/topic/620/flxcamera-render-only-chosen-objects) and how they would need to go back and change it if they decided to do split-screen.

Gama11:
>With the current system, if I add secondary cameras I need to have full knowledge of what other code is potentially adding or removing cameras in this state

You mean because the usual patterns is to assign a new array instance to it, rather than push()ing or remove()ing cameras from it (which is sort of caused by the inconvenience of the null check that is otherwise needed since it defaults to that)?

Geokureli:
> which is sort of caused by the inconvenience of the null check that is otherwise needed since it defaults to that)?

Not only the null check, but you also can't just edit FlxCamera.defaultCameras directly because it is literally FlxG.cameras.list, which you should never edit

Geokureli:
AustinEast convinced me that `isDefault` is a terrible name, I'd like to rethink that name

Gama11:
>The name isDefault led hm to believe that there can be only one default camera.

I initially had that concern as well. I have to admit that I don't like `autoDraw` much better though.

________

It may be worth considering whether a `FlxCamera.isDefault` property is necessary at all. Is that really a property of the camera itself? It seems like it really only concerns the entity that _manages_ the list of cameras (the camera frontend), and the camera itself shouldn't really be aware of whether it is being used "as a default" or not. If you only ever interact with the default cameras though `FlxG.cameras.defaults` [*], you automatically have the plural, which should avoid any "there can only be one" confusion.

[*] What that means directly exactly some thought. Do we allow users to interact directly with a `FlxG.cameras.defaults` array, or does that create too much potential for invalid states (same camera being listed twice, etc)? Does `FlxG.cameras.add()` need a new optional parameter that determines whether a new camera should be added to the defaults list?

--- 

Or, alternatively - could we maybe improve `FlxCamera.defaultCameras` in a way that makes it unnecessary to deprecate it in the first place, while also avoiding breaking changes? For instance, I could imagine wrapping it in an abstract in a way that allows to `push()` something to it without checking for null first: https://try.haxe.org/#5bEF4b79

Geokureli:
> Is that really a property of the camera itself? It seems like it really only concerns the entity that _manages_ the list of cameras (the camera frontend), and the camera itself shouldn't really be aware of whether it is being used "as a default" or not.

I could go either way here, I think people create cameras with either the intent to draw everything or the intent to draw some things, however in the end this a property that only affects how the `CameraFrontEnd` manages the camera. IMO, `FlxCamera` and `CameraFrontEnd` are too deeply coupled to really consider them separately

> [*] What that means directly exactly some thought. Do we allow users to interact directly with a `cameras.defaults` array, or does that create too much potential for invalid states (same camera being listed twice, etc)?

To be clear I only publicly expose `list` and `defaults` for backwards compatibility, since I assume people using `defaultCameras` would want to manually set it back to `list` when they are done overriding default cameras. If backwards compatibility wasn't an issue I would make `defaults`, `lists` and `defaultCameras` private.

>  Does `FlxG.cameras.add()` need a new optional parameter that determines whether a new camera should be added to the defaults list?

I think this is a good solution. I would also like methods that change a cameras "defaultness" rather than having them edit some exposed array.

> Or, alternatively - could we maybe improve `FlxCamera.defaultCameras`

To be honest null safety is almost no concern here unless you set `defaultCameras` to `null` mid-state or try to access it before the FlxGame is created. the main concern of `defaultCameras` is that it may be pointing at `cameras.list` which should never be edited.

Gama11:
>the main concern of defaultCameras is that it may be pointing at cameras.list which should never be edited.

Perhaps that's the real issue here we should think about finding ways to fix? That seems more like a bugfix than a breaking change / if anybody relies on changes to `defaultCameras` changing `list`, that doesn't seem sound / was never an intended use case.

flixel/system/frontEnds/CameraFrontEnd.hx
```diff
@@ -14,6 +14,12 @@ class CameraFrontEnd
 	 * By default flixel creates one camera the size of the screen.
 	 */
 	public var list(default, null):Array<FlxCamera> = [];
```
Geokureli, line 16: Note, I ended up keeping this public, [FlxTilemap uses it](https://github.com/HaxeFlixel/flixel/blob/dev/flixel/tile/FlxTilemap.hx#L385) and it may be handy to have a list to iterate. I considered copying haxe 4's [ReadOnlyArray](https://github.com/HaxeFoundation/haxe/blob/development/std/haxe/ds/ReadOnlyArray.hx) but I'm not sure if that class would work on 3.4.7

Geokureli:
I took this in a new direction after some discussion. Rather than `myCamera.isDefault = false` I've added an optional `drawsDefault` parameter to `FlxG.cameras.add` as well as a `setDrawsDefault` method. I think this is a much cleaner solution compared to the previous one having signals for property changes. It also ties the logic to the camera management system, meaning a new system is free to use different logic.

I'm still not sold on the names I chose but I think `drawsDefault` is much better than `isDefault`.

Geokureli:
@Gama11 just checking in, are there still concerns about this change?

flixel/FlxCamera.hx
```diff
@@ -51,11 +51,24 @@ class FlxCamera extends FlxBasic
 	public static var defaultZoom:Float;
 
 	/**
-	 * Which cameras a `FlxBasic` uses to be drawn on when nothing else has been specified.
-	 * By default, this is just a reference to `FlxG.cameras.list` / all cameras, but it can be very useful to change.
-	 * This may be temporarily modified during a `draw` call.
+	 * Used behind-the-scenes during the draw phase so that members use the same default
+	 * cameras as their parent.
+	 * 
+	 * Prior to 4.8.2 it was useful to change this value, but that feature is deprecated.
```
Gama11, line 57: 4.9.0 - patch releases don't introduce new public API / features.

flixel/system/frontEnds/CameraFrontEnd.hx
```diff
@@ -12,8 +12,15 @@ class CameraFrontEnd
 	/**
 	 * An array listing FlxCamera objects that are used to draw stuff.
 	 * By default flixel creates one camera the size of the screen.
+	 * Do not edit directly, use `add` and `remove` instead.
 	 */
 	public var list(default, null):Array<FlxCamera> = [];
+	
+	/**
+	 * Array listing all cameras marked as default, `FlxBasics` with no `cameras` set will render to them.
+	 * @since 4.8.2
```
Gama11, line 21: 4.9.0 - patch releases don't introduce new public API / features.

flixel/system/frontEnds/CameraFrontEnd.hx
```diff
@@ -89,6 +102,27 @@ class CameraFrontEnd
 
 		cameraRemoved.dispatch(Camera);
 	}
+	
+	/**
+	 * Changes whether the specified camera draws FlxBasics 
+	 * @param camera		The camera you wish to change
+	 * @param drawsDefault	If false, FlxBasics will not render to it unless you add it to their `cameras` list.
+	 */
```
Gama11, line 110: Missing `@since 4.9.0`

flixel/system/frontEnds/CameraFrontEnd.hx
```diff
@@ -89,6 +102,27 @@ class CameraFrontEnd
 
 		cameraRemoved.dispatch(Camera);
 	}
+	
+	/**
+	 * Changes whether the specified camera draws FlxBasics
```
Gama11, line 107: This sentence seems incomplete, missing a "by default" at the end or something?

flixel/system/frontEnds/CameraFrontEnd.hx
```diff
@@ -89,6 +102,27 @@ class CameraFrontEnd
...
 	}
+	
+	/**
+	 * Changes whether the specified camera draws FlxBasics 
+	 * @param camera		The camera you wish to change
+	 * @param drawsDefault	If false, FlxBasics will not render to it unless you add it to their `cameras` list.
+	 */
+	public function setDrawsDefault(camera:FlxCamera, drawsDefault:Bool)
+	{
+		if (list.indexOf(camera) == -1)
```
Gama11, line 113: You can use `!contains()` via `using FlxArrayUtil`.

flixel/FlxCamera.hx
```diff
@@ -51,11 +51,24 @@ class FlxCamera extends FlxBasic
...
+	public static var defaultCameras(get, set):Array<FlxCamera>;
+	
+	/**
+	 * Used behind-the-scenes during the draw phase so that members use the same default
+	 * cameras as their parent.
+	 * 
+	 * This is the non-deprecated list that the public `defaultCameras` proxies. Allows flixel classes
+	 * to use it without warning.
+	 */
+	@:allow(flixel)
```
Gama11, line 70: `@:allow(flixel)` is dangerous, at least it was in older Haxe versions, I don't remember whether it was changed - it would make it so that everything that inherits from a class in the `flixel` package has access too, which is typically pretty much everything... You can check whether that's the case via completion.

Bottom line, usually it's safer to use `@:access`.

flixel/system/frontEnds/CameraFrontEnd.hx
```diff
@@ -12,8 +12,15 @@ class CameraFrontEnd
...
 	 * By default flixel creates one camera the size of the screen.
+	 * Do not edit directly, use `add` and `remove` instead.
 	 */
 	public var list(default, null):Array<FlxCamera> = [];
+	
+	/**
+	 * Array listing all cameras marked as default, `FlxBasics` with no `cameras` set will render to them.
+	 * @since 4.8.2
+	 */
+	var defaults(default, null):Array<FlxCamera> = [];
```
Gama11, line 23: `(default, null)` is pointless on a private variable.

flixel/FlxCamera.hx
```diff
@@ -51,11 +51,24 @@ class FlxCamera extends FlxBasic
...
 
 	/**
-	 * Which cameras a `FlxBasic` uses to be drawn on when nothing else has been specified.
-	 * By default, this is just a reference to `FlxG.cameras.list` / all cameras, but it can be very useful to change.
-	 * This may be temporarily modified during a `draw` call.
+	 * Used behind-the-scenes during the draw phase so that members use the same default
+	 * cameras as their parent.
+	 * 
+	 * Prior to 4.8.2 it was useful to change this value, but that feature is deprecated.
+	 * Instead use `FlxG.cameras.setDrawsDefault`.
```
Gama11, line 58: Maybe mention that specifying it during `add()` is an option too.

Gama11:
Though beter than what we had before, `drawsDefault` still sounds a bit awkward to me... It's like it inverts the relationship, in reality it's more like `defaultDrawTarget`, since it's the sprites that draw themselves on the camera, rather than the camera drawing the sprites.

flixel/FlxCamera.hx
Geokureli: I personally don't like @:access, I prefer a fields to declare their intended access in the declaration. I didn't know it allowed access to extensions. I can change this to allow the specific methods.

Geokureli:
I still feel like `setDefaultDrawTarget` and `defaultDrawTarget` suggests that there is a singular default camera, rather than many defaults. maybe the documentation is enough to prevent users from getting confused. it's still much more accessible and easier to understand than making them edit a random static array with a ton of caveats, in my opinion. but, as always, I'm open to name ideas.

Gama11:
I'm not sure there's a better word than `default` to describe this, even if you can often only have one default... For instance, `fallback` has another incorrect implication (that it's only relevant in rare cases). `addToDefaultDrawTargets` would remove any confusion, but that seems way too verbose.

If we make sure the API docs use plurals / "**a** default draw target" rather than "**the** default draw target", it's probably fine?

Geokureli:
> If we make sure the API docs use plurals / "a default draw target" rather than "the default draw target", it's probably fine?

this is a good idea, I updated the documentation to overstate the plurality of default draw targets.

MSGhero:
"FlxBasics with no cameras set will render to them" is probably the cleanest explanation of this change. I'm happy with the naming and comments. Part of me would like to see a demo (probably Mode?) updated with this, but the other part of me thinks an ifdef for 4.9.0 would be too confusing to learn from.

Geokureli:
> Part of me would like to see a demo (probably Mode?) updated with this

Actually all existing demos still work after this new change. The default behavior - before and after - was the

Geokureli:
Accidentally pressed close, reopening

MSGhero:
Right it doesn't affect the end result, but seeing this implemented, in an existing demo or a new one, is what I meant. Mode is one of the few I'm aware of that actually uses multiple cameras.

Gama11:
Not in a way that would make this feature useful though (in Mode, all cameras actually do render everything).

Geokureli:
> Not in a way that would make this feature useful though (in Mode, all cameras actually do render everything).

correct, if we wanted to showcase this new feature we would probably have to make a new demo that used non-default cameras, since none of the current demos that create cameras mess with the "default-ness" of those cameras

PR #2298, opened 2021-03-08
Add getRotatedRectBounds helpers

Geokureli:
Add helpers to FlxRect that calculate the globally-aligned bounding box of a rotated rectangle, and similarly the GABB of a FlxObject and FlxSprite. Useful for determining the actual world space of a rotated rect for things like calculating redraw regions or broad phase collision (like getting a rect quadtree bounds).

I made this when I was looking into issues with `pixelPerfectCheck`, I thought this might optimize it, preventing it from iterating pixel data when the sprites are far away from each other. that's still in progress, and might not even be fruitful.

I wanted to see if this utility was up to standards, or even worthwhile on its own, before I invested my time into that application of it.

Here's a demo: https://www.newgrounds.com/projects/games/1603054/preview


Gama11:
Tbh, my first thought is that `FlxRect` shouldn't know about `FlxSprite` or `FlxObject`. :/

Geokureli:
originally this was in flxCollision. Austin mentioned that people may not look there. i can put them in `FlxObject/FlxSprite`?

Geokureli:
converting back to a draft for some drastic changes, namely

- renaming `getRotatedRectBounds` to `calcRotatedBounds` since "get" implies simple data retrieval.
- convert the static FlxRect method to an instance method
- finding new places for `FlxSprite` and `FlxObject` versions of this.

Geokureli:
@Gama11 any more thoughts on this?

Gama11:
Not sure how I feel about a method called `calcCollisionBounds()` on `FlxSprite` and `FlxObject` that respects rotation, considering collisions in Flixel _don't_ respect rotation... So these actually _aren't_ the collision bounds in flixel terms, which could lead to confusion.

In general, I'd be happier with this if they were actually needed for something within flixel, i.e. that optimization you mentioned.

Geokureli:
I think I already have the optimization in another branch, I can add it here with some performance info.

got anything shorter than `calcRotatedCollisionBounds`?

Geokureli:
I've added the optimization to `FlxCollision.pixelPerfectCheck`

Here is a simple test: https://imgur.com/ISFuHcj
And you can play it yourself here (follow directions in authors comments): https://www.newgrounds.com/projects/games/1574925/preview

The old method checked a bounds much bigger than what was needed. the benefits of this change are bigger as the dimensions of the image increase, especially a long skinny image like the spikes on flappybalt

Geokureli:
Did some further testing, I didn't realize that pixelPerfectCheck ignores the sprite's scale, when I apply random scaling on the PPC demo's sprites I actually get worse performance because it's checking a larger bounds than it did previously.

it would be neat to honor scale in pixel perfect overlap, but for now it may be better to just ignore scale in the bounds check. I may add an arg to `calcRotatedGraphicBounds` to ignore/allow scale but that seems kinda ugly.

Geokureli:
Apparently it's pretty easy to get `pixelPerfectCheck` to work with scale, so I added it. [Check it out!](https://www.newgrounds.com/projects/games/1603054/preview/filetype/2)

  I renamed `calcRotatedGraphicBounds` to `getScreenBounds` to match `getScreenPosition`. I renamed all the calcRotatedBounds methods back to getRotatedBounds, as well.

Geokureli:
That's weird. I forgot to update the tests but all tests passed

Geokureli:
Found some other uses and optimizations for this. I also wanted to look into pixelPerfectPosition, pixelPerfectRender and camera zoom to see if those should be considered for these tools

Geokureli:
in summary this adds:
- `FlxCamera.getViewRect` - The size and position of this camera's screen (helper for `FlxSpriteUtil.

Geokureli:
in summary this adds:
- `FlxRect.getRotatedBounds`
- `FlxObject.getRotatedBounds`
- `FlxSprite.getRotatedBounds` - override
- `FlxSprite.getScreenBounds` - similar to `FlxSprite.getScreenPosition`
- `FlxSpriteUtil.cameraWrap` - similar to `FlxSpriteUtil.screenWrap` but uses FlxDirectionFlags instead of 4 booleans
- `FlxSpriteUtil.cameraBounds` - similar to `FlxSpriteUtil.cameraWrap`
- `FlxCamera.getViewRect` - helper for `FlxSpriteUtil.cameraWrap` and `cameraBound`
- `FlxCamera.containsRect`- helper for `FlxSpriteUtil.cameraWrap` and `cameraBound`

Made changes to:
- `FlxCollision.pixelPerfectCheck` - uses getScreenBounds to reduce the number of pixel checks, honors FlxSprite's scale
- `FlxSprite.isOnScreen` - uses getScreenBounds and 
- `FlxCamera.containsPoint` - calls putWeak

PR #2300, opened 2021-03-28
Brave crash

### LINKED ISSUE #2285: FlxBitmapFont.fromXNA causes an infinite loop on Brave Browser

Geokureli:
same fix as #2286 but without the farbling detection

fixes #2285


Geokureli:
I see accidental commits, removing them now

PR #2303, opened 2021-04-30
Add FlxDirectionFlags and FlxDirection

Geokureli:
not sure if I really wanna submit this, just seeing if it passes CI for now.

A lot of HaxeFlixel's enums use the old flash style, wherre its just a bunch of constants, meaning you havre to pass in FlxObject.LEFT instead of utilizing haxe's type inference and just saying LEFT, and it's surprisingly annoying. FlxColor is the same way, and if this goes well I might make that a `@:enum abstract` as well.


flixel/util/FlxDirection.hx
```diff
@@ -0,0 +1,50 @@
+package flixel.util;
+
+@:enum abstract FlxDirection(Int) from Int to Int
```
Gama11, line 3: I know that the naming convention for enums is singular (for good reason), but this isn't _really_ an enum, it's a bit field, so one value can contain multiple directions. `FlxDirections` makes more sense to me because of that. That also matches the naming convention for bit fields in C#, and maybe other languages.

There should be a doc comment here mentioning that it's a bit field (for people new to the concept so they can look it up), plus `@since 4.10.0`.

The `from Int to Int` is unfortunate for type safety, but necessary backwards compatbility.

flixel/util/FlxDirection.hx
```diff
@@ -0,0 +1,50 @@
...
+{
+	var LEFT    = 0x0001;
+	var RIGHT   = 0x0010;
+	var UP      = 0x0100;
+	var DOWN    = 0x1000;
+	var NONE    = 0x0000;
+	var CEILING = 0x0100;// UP;
```
Gama11, line 10: These should be doc comments, not line comments, so they show up in API docs and completion.

Would also make sense to have a line comment with a link to the relevant Haxe issue once there is one.

flixel/util/FlxDirection.hx
```diff
@@ -0,0 +1,50 @@
...
+	var FLOOR   = 0x1000;// DOWN;
+	var WALL    = 0x0011;// LEFT | RIGHT;
+	var ANY     = 0x1111;// LEFT | RIGHT | UP | DOWN;
+	
+	public inline function has(dir:FlxDirection):Bool return this & dir == dir;
+	
+	@:op(A & B)
+	inline function and(dir:FlxDirection):FlxDirection return this & dir;
```
Gama11, line 18: Not sure these `@:op` functions need bodies? If you omit it, that should function as "forwarding the operator".

Any particular reason these have `FlxDirection` as arguments, and the comparison operators `Int`?

flixel/FlxObject.hx
```diff
@@ -35,47 +36,47 @@ class FlxObject extends FlxBasic
 	/**
 	 * Generic value for "left". Used by `facing`, `allowCollisions`, and `touching`.
```
Gama11, line 37: These should probably have a note along the lines of this in their doc comments:

> Note: This exists for backwards compatibility, prefer using `FlxDirections.LEFT` directly.

A proper deprecation warning would probably be too annoying at this point, since these are probably used fairly commonly.

Gama11 (changes requested):
I like the idea, working with raw bit field constants is a bit annoying. IIRC I was considering something like this years ago but never implemented it.

Gama11:
>FlxColor is the same way

Is it? I don't see a huge advantage to making it an enum abstract, I guess it would allow you to use the default colors "unqualified", but other than that?

Geokureli:
Wasn't expecting feedback yet, but it's appreciated. just know this was WIP and things like operators taking int vs FlxDirection was because I was trying to understand more about some oddities I noticed while testing this out. fixes and documentation coming soon.

> > FlxColor is the same way
> 
> Is it? I don't see a huge advantage to making it an enum abstract, I guess it would allow you to use the default colors "unqualified", but other than that?

Wouldn't it allow statements `sprite.color = RED;` without needing to import FlxColor? these are QoL features that I really appreciate.

flixel/util/FlxDirection.hx
Geokureli: [this is why](https://try.haxe.org/#8d1A02B4) I'm using ints for some. Luckily [this works](https://try.haxe.org/#d60d4cad), I didn't know that existed.

flixel/util/FlxDirection.hx
Geokureli: https://github.com/HaxeFoundation/haxe/issues/10237

flixel/FlxObject.hx
```diff
@@ -34,48 +35,57 @@ class FlxObject extends FlxBasic
...
 	/**
 	 * Special-case constant meaning only the left and right sides, used mainly by `allowCollisions` and `touching`.
+	 * Note: This exists for backwards compatibility, prefer using `FlxDirections.WALL` directly.
 	 */
-	public static inline var WALL:Int = LEFT | RIGHT;
+	public static inline var WALL = FlxDirections.WALL;
 
 	/**
 	 * Special-case constant meaning any direction, used mainly by `allowCollisions` and `touching`.
+	 * Note: This exists for backwards compatibility, prefer using FlxDirections.LEFT directly.
```
Gama11, line 86: You missed the backticks on this one. Plus it references the wrong constant. ;)

flixel/util/FlxDirections.hx
```diff
@@ -0,0 +1,117 @@
...
+	 */
+	public inline function hasAll(dir:FlxDirections):Bool
+	{
+		return this & dir == dir;
+	}
+	
+	/**
+	 * Returns true if this contains any one of the supplied flags.
+	 */
+	public inline function hasAny(dir:FlxDirections):Bool
```
Gama11, line 48: I'm not sure this is very useful, I think I preferred the simplicity of having just one `has()` (which is now called `hasAll()` I guess). It seems like you're already using it wrong too (no need for this in `toString()`, since you're always just checking against a single direction)?

flixel/util/FlxDirections.hx
```diff
@@ -0,0 +1,117 @@
...
+	{
+		return this & dir > NONE;
+	}
+	
+	/**
+	 * Adds the supplied flags if this doesn't already have them.
+	 * This modifies this instance instead of creating a new one.
+	 */
+	@:op(A += B)
```
Gama11, line 57: `+` and `-` don't feel right to me in conjunction with bit fields. Considering this abstract has `from Int to Int`, that seems like a recipe for disaster if something is accidentally typed as Int rather than FlxDirections, since those operators would do something different entirely then...

flixel/util/FlxDirections.hx
```diff
@@ -0,0 +1,117 @@
...
+	{
+		return this = this & ~dir;
+	}
+	
+	/**
+	 * Adds the supplied flags if this doesn't already have them.
+	 * this creates a new instance without modifying this instance.
+	 */
+	@:op(A + B)
+	public inline function addNew(dir:FlxDirections):FlxDirections
```
Gama11, line 78: I don't like this `*New` naming, `with()` and `without()` would be clearer IMO.

Talking about "instances" in the doc comment seems misleading too, that term is usually reserved for objects. However, this is still a basic type at the end of the day, not an allocated instance.

flixel/util/FlxDirections.hx
```diff
@@ -0,0 +1,117 @@
...
+	var RIGHT   = 0x0010;
+	/** Generic value for "up". */
+	var UP      = 0x0100;
+	/** Generic value for "down". */
+	var DOWN    = 0x1000;
+	/** Special-case constant meaning no directions. */
+	var NONE    = 0x0000;
+	
+	/**
+	 * Directions values can't be constructed from other direction values, due to a bug in haxe.
```
Gama11, line 22: With the way Haxe parses doc comments (first one it comes across is the first one used IIRC), this might end up being used as the doc comment for `CEILING`, have you checked that isn't the case? Otherwise a line comment should be used.

flixel/FlxObject.hx
```diff
@@ -34,48 +35,57 @@ class FlxObject extends FlxBasic
 
 	/**
 	 * Generic value for "left". Used by `facing`, `allowCollisions`, and `touching`.
```
Gama11, line 37: I guess you forgot to update `facing`?

flixel/util/FlxDirections.hx
```diff
@@ -0,0 +1,117 @@
...
+	 */
+	
+	/** Special-case constant meaning "up". */
+	var CEILING = 0x0100;// UP;
+	/** Special-case constant meaning "down" */
+	var FLOOR   = 0x1000;// DOWN;
+	/** Special-case constant meaning "left" and "right". */
+	var WALL    = 0x0011;// LEFT | RIGHT;
+	/** Special-case constant meaning any, or all directions. */
+	var ANY     = 0x1111;// LEFT | RIGHT | UP | DOWN;
```
Gama11, line 35: I would've just inlined these into the doc comments, for instance:

```
Special-case constant meaning `UP`.
```

or

```
Special-case constant meaning any, or all directions (`LEFT | RIGHT | UP | DOWN`).
```

-----

Looking at this more closely though, I think we should probably just get rid of `CEILING`, `FLOOR` and `WALL`, and only keep `ANY`. The first three only really made sense for collisions, and this is meant to be a generic abstract for directions. After all, the old convenience constants still exist on `FlxObject` anyway.

flixel/util/FlxDirections.hx
```diff
@@ -0,0 +1,117 @@
+package flixel.util;
+
+/**
+ * Uses bit flags to create a list of orthogonal directions. useful for
+ * many `FlxObject` features like `allowCollisions` and `touching`.
+ * @since 4.10.0
+ */
+@:enum abstract FlxDirections(Int) from Int to Int
+{
+	/** Generic value for "left". */
```
Gama11, line 10: I'm not sure these doc comments on left / right / up / down add anything useful. They mostly just make me wonder what "generic" is supposed to mean in this context.

flixel/util/FlxDirections.hx
```diff
@@ -0,0 +1,117 @@
+package flixel.util;
+
+/**
+ * Uses bit flags to create a list of orthogonal directions. useful for
```
Gama11, line 4: Perhaps "Represents a bit field of orthogonal directions"?

Gama11 (changes requested):

flixel/FlxObject.hx
Geokureli: facing is an interesting case, since it doesn't allow cases like LEFT|RIGHT and ANY, does it seem worthwhile to make a separate enum that uses the same underlying values?

flixel/util/FlxDirections.hx
Geokureli: I'm not a fan of removing helper values like ceiling or wall, if anything I'd rather add more case specific values like `X_AXIS` and `Y_AXIS`, i find them indicative of their broad uses. `WALLS` in particular was used in the snippet that helped me first understand direction flags in flixel. the `FlxObject` constants have little value to me, also. I would remove them completely if I didn't care about backwards compatibility.

flixel/FlxObject.hx
Geokureli: in fact, I never use `facing` and the documentation confuses me on whether it expects single directions or values like `LEFT | UP`

particularly this comment:
> Can be set to `LEFT`, `RIGHT`, `UP`, and `DOWN` to take advantage of flipped sprites and/or just track player orientation more easily.

seems to conflict with this in the `setFacingFlip` doc
> E.g. To make a sprite flip horizontally when it is facing both `UP` and `LEFT`, use `setFacingFlip(LEFT | UP, true, false);`

flixel/util/FlxDirections.hx
Geokureli: yeah, lets not confuse things further

Geokureli:
looks like flixel-addons classes like `FlxUIPopup` will have to change to match IFlxSprite. BTW why doesn't `FlxSprite` implement `IFlxSprite`? I was gonna save this for a separate issue but it seems relevant, now. it might be worth changing here.

Gama11:
Not sure it's worth doing anything with those interfaces, they're in a weird place in general... `IFlxSprite` is extremely incomplete, there's no `IFlxObject` either, instead `IFlxSprite` has properties from `FlxObject`, etc... I'd rather not touch them / integrate them more deeply unless there's a good reason. 😅

Geokureli:
@Gama11, it seems we have a reason to mess with IFlxSprite (though, perhaps not a good reason). Since IFlxParticle extends IFlxSprite and FlxParticle extends FlxSprite requiring it to have a `facing:Int` field. FLxTypedEmitter expects `FlxSprite & IFlxParticle` and i suspect addon things will break if we changed that to take `FlxParticle`.
https://github.com/HaxeFlixel/flixel/search?q=IFlxParticle

flixel/FlxObject.hx
Gama11: Looking at the source, I don't think anything is stopping combined direction flags from working. I could see that being used for sprites with 8 facing directions (including diagonals) rather than just 4. It's probably not very common though.

Gama11:
Still not sure that `hasAny(dir)` is useful / if removing it and renaming `hasAll()` to `has()` wouldn't be better. Do you have a compelling use case for it?

C# flag enums also went that route it seems, there's only a `HasFlag()` method as far as I can tell: https://docs.microsoft.com/en-us/dotnet/api/system.enum.hasflag?view=net-5.0

Gama11:
There's quite an interesting compilation failure in `TurnBasedRPG`:

```
source/Player.hx:106: characters 13-19 : Unmatched patterns: ANY | NONE | WALL
source/Enemy.hx:79: characters 12-18 : Unmatched patterns: ANY | NONE | WALL
```

This makes me question these "special-case constants" again. Perhaps they should not be considered enum values at least, just constants? I.e. declaring them with `public static inline var`. That way you don't need to match them in an exhaustive switch.

Of course, technically you do need to match ANY, NONE and WALL in order for a switch to be exhaustive, but even then it doesn't really cover everything, so I think only having the four basic directions be real enum values might be a bit cleaner? I don't think there's a perfect solution here with how Haxe's pattern matching works... :/

Geokureli:
Would be very nice if haxe had a `@:nonExaustive` tag.

I was thinking about this earlier, direction flags don't make as much sense for facing as it does for collision. you don't really want an exhaustive check for FlxDirections. you would either use a series of `has()` use a default case on switches

 I almost want to make FlxDirection (singular) for things that can only be UP, LEFT, RIGHT and DOWN and not a combination/lack of them

Gama11:
>I almost want to make FlxDirection (singular) for things that can only be UP, LEFT, RIGHT and DOWN and not a combination/lack of them

I've briefly considered that as well, but that just leads to "we need both FlxDirection and FlxDirections", I guess? Which seems a bit awkward naming-wise, if nothing else... :)

Geokureli:
Naming wise I'd suggest FlxDirection and FlxDirectionFlags, and I don't think thats an issue, but we probably shouldn't change `facing` to disallow flags, since the [setFacingFlip doc](https://github.com/HaxeFlixel/flixel/blob/dev/flixel/FlxSprite.hx#L1081-L1082) specifically uses UP | LEFT as an example. It seems weird to use it like that, but no need to remove existing functionality.

I like your original idea where we define NONE, ANY and WALL as `static inline public var`s

Geokureli:
So I changed this around a bit. Added `FlxDirection` which is only the 4 cardinal directions, and renamed FlxDirections to FlxDirectionFlags, for distinction

Its a bad idea to do a non exhaustive switch for FlxDirectionFlags without a default case, but they can cast down to FlxDirection for the switch if they expect it to only use those 4 values

Geokureli:
oh, wow thanks for updating all those repos

PR #2309, opened 2021-05-29
Added support for Unicode Combining Diacritical Marks in FlxBitmapText

SeiferTim:
For applyng Underlines, Umlauts, and other combination characters to your Bitmap Text.
More information: https://en.wikipedia.org/wiki/Combining_Diacritical_Marks

Example Usage: add an underline character to your BitmapFont and give it the Id of 818 (Decimal for U+0332), make sure it is as wide as your widest xadvance + letterspacing for your font. Set it's xadvance = 0

In your code, set your `FlxBitmapText.text = "This M\u0332 is Underlined!";` and it should show up with the letter M underlined!

### DIFF
--- a/flixel/graphics/frames/FlxBitmapFont.hx
+++ b/flixel/graphics/frames/FlxBitmapFont.hx
@@ -13,20 +13,20 @@ import flixel.system.FlxAssets.FlxAngelCodeSource;
 import flixel.system.FlxAssets.FlxBitmapFontGraphicAsset;
 import flixel.util.FlxColor;
 import openfl.Assets;
+
+using flixel.util.FlxUnicodeUtil;
+
 #if haxe4
 import haxe.xml.Access;
 #else
 import haxe.xml.Fast as Access;
 #end
 
-using flixel.util.FlxUnicodeUtil;
-
 /**
  * Holds information and bitmap characters for a bitmap font.
  */
 @:allow(flixel.text.FlxBitmapText)
-class FlxBitmapFont extends FlxFramesCollection
-{
+class FlxBitmapFont extends FlxFramesCollection {
 	static inline var SPACE_CODE:Int = 32;
 	static inline var TAB_CODE:Int = 9;
 	static inline var NEW_LINE_CODE:Int = 10;
@@ -87,8 +87,7 @@ class FlxBitmapFont extends FlxFramesCollection
 	/**
 	 * Creates a new bitmap font using specified bitmap data and letter input.
 	 */
-	function new(frame:FlxFrame, ?border:FlxPoint)
-	{
+	function new(frame:FlxFrame, ?border:FlxPoint) {
 		super(frame.parent, FlxFrameCollectionType.FONT, border);
 		this.frame = frame;
 		parent.persist = true;
@@ -97,8 +96,7 @@ class FlxBitmapFont extends FlxFramesCollection
 		charAdvance = new Map<Int, Int>();
 	}
 
-	override public function destroy():Void
-	{
+	override public function destroy():Void {
 		super.destroy();
 		frame = null;
 		fontName = null;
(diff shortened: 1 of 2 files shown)


Gama11:
That diff primarily seems to consist of unrelated formatting changes...

flixel/text/FlxBitmapText.hx
```diff
@@ -1182,6 +1197,8 @@ class FlxBitmapText extends FlxSprite
 		var spaceWidth:Int = font.spaceWidth;
 		var lineLength:Int = line.uLength();
 		var textWidth:Int = this.textWidth;
+		
+		var charUt8;
```
Gama11, line 1201: ?

flixel/text/FlxBitmapText.hx
```diff
@@ -362,6 +362,8 @@ class FlxBitmapText extends FlxSprite
 			}
 
 			var clippedFrameRect;
+			var tmpFrameRect:FlxRect = FlxRect.get();
```
Gama11, line 365: You should call `put()` when you're done with it.

flixel/text/FlxBitmapText.hx
```diff
@@ -1512,6 +1550,7 @@ class FlxBitmapText extends FlxSprite
 		var textLen:Int = Std.int(textData.length / 3);
 		var rect = FlxRect.get();
 		var frameVisible;
+		var prevAdv:Float = 0;
```
Gama11, line 1553: Unused?

Gama11 (changes requested):
Would be good to have a few examples of this feature in the FlxBitmapText demo.

flixel/graphics/frames/FlxBitmapFont.hx
```diff
@@ -660,4 +660,15 @@ class FlxBitmapFont extends FlxFramesCollection
 		font.updateSourceHeight();
 		return font;
 	}
+	
+	/**
+	 * Checks if the specified code is one of the Unicode Combining Diacritical Marks
+	 * @param	Code	The charactercode we want to check
+	 * @return 	Bool	Returns true if the code is a Unicode Combining Diacritical Mark
+	 */
+	public static function isUnicodeComboMark(Code:Int):Bool
```
Gama11, line 669: Not quite sure why you put this here. Do you expect this to be useful on its own as an API that people might use directly? If so, `FlxBitmapFont` is a bit of an awkward place for it, `FlxStringUtil` might be a better fit. If not, I'd just move it to `FlxBitmapText` (the only place it's used) and keep it private for now.

flixel/text/FlxBitmapText.hx
```diff
@@ -445,7 +446,7 @@ class FlxBitmapText extends FlxSprite
 				{
 					dataPos = j * 3;
 
-					currFrame = font.getCharFrame(Std.int(textDrawData[dataPos]));
+					currFrame = font.getCharFrame(Std.int(textDrawData[dataPos])).copyTo();
```
Gama11, line 449: Why is this `copyTo()` necessary now?

flixel/text/FlxBitmapText.hx
```diff
@@ -1208,32 +1209,40 @@ class FlxBitmapText extends FlxSprite
 
 		var tabWidth:Int = spaceWidth * numSpacesInTab;
 
+		var charUt8:UnicodeBuffer;
```
Gama11, line 1212: Why did you move this var declaration? The scope of vars should be kept as small as possible IMO.

Gama11 (changes requested):

flixel/graphics/frames/FlxBitmapFont.hx
SeiferTim: I guess I was thinking that it belonged within the font itself... less than the BitmapText object...
I can move it.

flixel/text/FlxBitmapText.hx
SeiferTim: Nope.

PR #2334, opened 2021-06-28
Add ability to change FlxText leading

Metlmeta:


Gama11:
I'm not sure if it's usfeul enough to warrant a new top-level field, evidently there hasn't been much of a demand for it thus far. It's a bit more verbose, but I think you should already be able to set the `leading` via `text.addFormat()`.

Metlmeta:
> I'm not sure if it's usfeul enough to warrant a new top-level field, evidently there hasn't been much of a demand for it thus far. It's a bit more verbose, but I think you should already be able to set the `leading` via `text.addFormat()`.

`addFormat()` takes in `FlxTextFormat`, but unfortunately it doesn't accept leading as a parameter. Would it be better to expand what `FlxTextFormat` accepts?

Gama11:
I don't think it needs to, this should work?

```haxe
var format = new FlxTextFormat();
format.format.leading = something;
```

Metlmeta:
> I don't think it needs to, this should work?
> 
> ```hx
> var format = new FlxTextFormat();
> format.format.leading = something;
> ```

`format`is a private variable, so it cannot be accessed from a FlxTextFormat variable. If the variable was set to public and then added via `addFormat`, I am still unsure of the correct way to call `updateDefaultFormat`.

Gama11:
Oh right, it's `private`. I guess it could simply be added as a constructor parameter of `FlxTextFormat`.

> I am still unsure of the correct way to call updateDefaultFormat to redraw the sprite.

I don't think that's necessary, `addFormat()`  sets `_regen`, which leads to a redraw.

Metlmeta:
> Oh right, it's `private`. I guess it could simply be added as a constructor parameter of `FlxTextFormat`.
> 
> > I am still unsure of the correct way to call updateDefaultFormat to redraw the sprite.
> 
> I don't think that's necessary, `addFormat()` sets `_regen`, which leads to a redraw.

I tried adding a parameter for `leading` when creating a new `FlxTextFormat`, but using `addFormat()` does not reflect the changes. Things such as the text color are updated for the sprite, but the leading property isn't visually updated.

Gama11:
I guess `copyTextFormat()` also needs to be updated?

Thinking on it again, a public `leading` property on `FlxTextFormat` is probably better than a new constructor parameter. Otherwise we'll end up with too many parameters if we want to expose more `TextFormat` properties in the future.

Metlmeta:
> I guess `copyTextFormat()` also needs to be updated?
> 
> Thinking on it again, a public `leading` property on `FlxTextFormat` is probably better than a new constructor parameter. Otherwise we'll end up with too many parameters if we want to expose more `TextFormat` properties in the future.

Thanks for `copyTextFormat()` clue!

I can create a new pull request that adds a public field to change the `leading` property on `TextFormat`. Does that sound good?

Issue #2233, opened 2019-09-16
AssetPaths class error with ios target


gasparetto:
- **Haxe version:** 4.0.0-rc.5
- **Flixel version:** 4.6.3
- **OpenFL version:** 8.9.5
- **Lime version:** 7.6.3
- **Affected targets:** ios

________________________________

**Code snippet reproducing the issue**:

```haxe
package;

import flixel.FlxState;

class PlayState extends FlxState
{
    override public function create():Void
    {
        super.create();

        var logo = new FlxSprite(AssetPaths.logo__png);
        logo.screenCenter(FlxAxes.XY);
        add(logo);
    }
}
```

________________________________

**Observed behavior:**

Compilation error when compiling for ios target. All other targets including macos are compiling without errors.

```
$ lime test ios -simulator
...
Haxe simulator build: Release-iphonesim-64
haxe Build.hxml -D simulator -D HXCPP_M64 -cpp build/Release-iphonesim-64
../../../../source/PlayState.hx:13: characters 45-54 : Class<AssetPaths> has no field logo__png
make: *** [build-haxe-x86_64] Error 1
Command ExternalBuildToolExecution failed with a nonzero exit code
```

**Expected behavior:**

No compilation errors.

Gama11:
Maybe this hack is causing issues / is no longer needed? https://github.com/HaxeFlixel/flixel/blob/dev/flixel/system/macros/FlxAssetPaths.hx#L40

gasparetto:
I commented out the hack at line 40 but the error is still there.

```
static function getFileReferences(directory:String, subDirectories:Bool = false, ?filterExtensions:Array<String>, ?rename:String->String):Array<FileReference>
{
	var fileReferences:Array<FileReference> = [];
	// var resolvedPath = #if (ios || tvos) Context.resolvePath(directory) #else directory #end;
	var resolvedPath = directory;
	var directoryInfo = FileSystem.readDirectory(resolvedPath);
	for (name in directoryInfo)
	{
```
Same error when compiling for ios:
```
$ lime test ios -simulator
...
    make
Haxe simulator build: Release-iphonesim-64
haxe Build.hxml -D simulator -D HXCPP_M64 -cpp build/Release-iphonesim-64
../../../../source/PlayState.hx:13: characters 45-54 : Class<AssetPaths> has no field logo__png
make: *** [build-haxe-x86_64] Error 1
Command ExternalBuildToolExecution failed with a nonzero exit code
```

[...]

PR #2345, opened 2021-07-14
Update iOS asset paths

Metlmeta:
Adds the direct path for iOS assets - building now works as expected. Resolves problems found in #2233


Gama11:
That's not a proper fix. We can't hardcode `assets`, that's just the default directory name, but it could be anything. You should even be able to have multple asset directories.

Metlmeta:
The iOS version uses "assets" as its working directory. The lime manifest uses something similar for iOS paths, seen in its `init` function: https://github.com/haxelime/lime/blob/c743fe10c404fb73003e24100f319a472cb4203f/templates/haxe/ManifestResources.hx

Gama11:
Ah, I see, so it's not related to the `<assets>` path in `project.xml`.

PR #2349, opened 2021-07-20
grammar lol

EyeDaleHim:


Gama11:
That's not what that's meant to say... That should be "gets or sets".

CyndaquilDAC:
This is quite literally not at all what this is meant to say. what it is meant to say is "Gets or sets", not "Gets to set". This is a VERY common phrase in programming so I'm surprised you didn't understand.

PR #2354, opened 2021-07-21
Cancel grandchild tweens

Geokureli:
```hx
FlxTween.tween(camera, {'scroll.y': 500}, 1.0);
```
The tween above can already be cancelled, via:
```hx
FlxTween.cancelTweensOf(camera.scroll);
```
but because the camera is the actual tween target, it should also be cancelled with:
```hx
FlxTween.cancelTweensOf(camera);
```

Whereas `tween(camera.scroll, {y: 500}, 1.0)` is not a tween on camera, so it must be cancelled with `cancelTweensOf(camera.scroll)` or `cancelTweensOf(camera, ['scroll.y'])`.

### DIFF
--- a/tests/unit/src/flixel/FlxCameraTest.hx
+++ b/tests/unit/src/flixel/FlxCameraTest.hx
@@ -39,16 +39,19 @@ class FlxCameraTest extends FlxTest
 	@Test
 	function testDefaultCameras():Void
 	{
-		Assert.areEqual(FlxG.cameras.defaults, FlxCamera.defaultCameras);
+		@:privateAccess
+		Assert.areEqual(FlxG.cameras.defaults, FlxCamera._defaultCameras);
 	}
 
 	@Test
 	function testDefaultCamerasStateSwitch():Void
 	{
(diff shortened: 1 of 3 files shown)


tests/unit/src/flixel/FlxCameraTest.hx
```diff
@@ -39,16 +39,19 @@ class FlxCameraTest extends FlxTest
 	@Test
 	function testDefaultCameras():Void
 	{
-		Assert.areEqual(FlxG.cameras.defaults, FlxCamera.defaultCameras);
+		@:privateAccess
```
Gama11, line 42: Why start using privateAccess in here?

Geokureli: to remove the deprecation warnings

Gama11: That feels very hacky. I think I'd rather have the deprecation warnings.

Geokureli: both seem hacky, but sure

PR #2370, opened 2021-08-04
Loading Graphics should allow for Unique graphics

SeiferTim:
All of the methods that loaded graphic assets were assuming `unique: false` when there are cases where you may want to load them using `unique: true`


SeiferTim:
Closing because it is an unnecessary separation from #2371

PR #2371, opened 2021-08-04
Loading Graphics should allow for Unique graphics

SeiferTim:
All of the methods that loaded graphic assets were assuming unique: false when there are cases where you may want to load them using unique: true

### DIFF
--- a/flixel/graphics/frames/FlxAtlasFrames.hx
+++ b/flixel/graphics/frames/FlxAtlasFrames.hx
@@ -126,11 +126,13 @@ class FlxAtlasFrames extends FlxFramesCollection
 	 * @param   description   Contents of the file with atlas description.
 	 *                        You can get it with `Assets.getText(path/to/description/file)`.
 	 *                        Or you can just pass path to the description file in the assets directory.
+	 * @param   unique        Ensures that the `BitmapData` uses a new slot in the cache.
+	 *                        If `true`, then `BitmapData` for this `FlxGraphic` will be cloned, which means extra memory.
 	 * @return  Newly created `FlxAtlasFrames` collection.
 	 */
-	public static function fromLibGdx(source:FlxGraphicAsset, description:String):FlxAtlasFrames
+	public static function fromLibGdx(source:FlxGraphicAsset, description:String, unique:Bool = false):FlxAtlasFrames
 	{
-		var graphic:FlxGraphic = FlxG.bitmap.add(source);
+		var graphic:FlxGraphic = FlxG.bitmap.add(source, unique);
 		if (graphic == null)
 			return null;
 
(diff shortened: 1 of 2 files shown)


Gama11:
Why did you open two pull requests for this?

SeiferTim:
> Why did you open two pull requests for this?

I'm not sure how/why it did that - I changed 2 different files to solve the same/similar issues... (I'm still not 100% familiar with the way Github works)

Gama11:
Well... could you fix it? ;)

Also:
- you added a new parameter to various methods without updating the doc comment (`@param`)
- a boolean parameter with a default value shouldn't have a `?`, there's no need for this to be a `Null<Bool>`

SeiferTim:
I would love to fix it!

...I have no clue where to start...

How do I add another file to a Pull Request? Or do I need to cancel this one and do a new one?

Gama11:
I mean, this is far from your first pull request, and there's plenty of learning resources for Git and GitHub...

SeiferTim:
Closed and combined changes from #2370 into this PR... (I hope I did this right...!?)

Gama11:
Yes, looks correct. 

However, I think just realized that this change might not actually be necessary. `FlxGraphic` should be a valid `FlxGraphicAsset`, so you could for instance do `FlxAtlasFrames.fromSparrow(graphic)`, where `graphic` was created via one of the `FlxGraphic.from*` methods - and each of those already accepts a `Unique` parameter. That avoids having to duplicate this everywhere, and it seems fair enough that it's a bit more verbose since this use case doesn't seem all that common. Thoughts?

SeiferTim:
I see what you're saying... I wouldn't have thought to do that - but, I guess it works...

I don't think that having this options *hurts* anything, and it allows for slightly fewer steps in your code... 

```
var myFont = FlxBitmapFont.fromAngelCode(FlxGraphic.fromBitmapData("assets/my_font.png", true), "assets/my_font.xml");
```

vs

```
var myFont = FlxBitmapFont.fromAngelCode("assets/my_font.png", "assets/my_font.xml", true);
```

I think the new method is just a degree easier to grok without having to connect the dots... but.. yeah, I guess it's not really solving an problem that doesn't already have a solution...

Gama11:
>I don't think that having this options hurts anything

The concerns I have are: 
- Longer parameter lists, which increases the complexity - for instance that `fromMonospace()` method already has quite a few params.
- If we expose `unique` everywhere, what about the other parameters the `FlxGraphic.from*()` methods have, like `Key` or `Cache`, or potential new ones in the future? At some point, "flattening" the parameters just doesn't scale anymore - there's a reason we pass around `FlxRect`s rather than having separate `x`, `y`, `width` and `height` params on every method that needs a rectangle.

`FlxGraphic` didn't exist when methods like `FlxSprite.loadGraphic()` were created, but if it did, `loadGraphic()` probably wouldn't have `Unique` and `Key` parameters.

So I'd prefer not to merge this.

PR #2383, opened 2021-08-14
Adding option for Split Sound Tray

SeiferTim:
Set `FLX_SPLIT_SOUND_TRAY` in Project.xml to use.
Seperates controls for Sound, Music & Master Volumes.

Defaults:
* `+` / `-` Sound Up/Down 
* `0` Sound Mute
*  `/` / `*` Music Up/Down 
* `.` Sound Mute
* `SHIFT`+`+` / `SHIFT`+`-` Master Up/Down
* `SHIFT`+`0` Master Mute

### DIFF
--- a/flixel/input/keyboard/FlxKey.hx
+++ b/flixel/input/keyboard/FlxKey.hx
@@ -106,6 +106,7 @@ abstract FlxKey(Int) from Int to Int
 	var NUMPADPLUS = 107;
 	var NUMPADPERIOD = 110;
 	var NUMPADMULTIPLY = 106;
+	var NUMPADDIVIDE = 111;
 
 	@:from
 	public static inline function fromString(s:String)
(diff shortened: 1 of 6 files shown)


Gama11:
Hm, I'm not sure it's a good idea, as this is starting to become pretty complex with some rather obscure shortcuts.

I think the built-in sound tray primarily works because it's simple, but if you need this much control, chances are this pre-built solution is not going to be a great fit anyway. Pretty much every "big" game is going to have a dedicated settings screen with volume options (and styled to fit the rest of the game), and all of Flixel's built-in UI disabled.

It should already be possible to customize the built-in sound tray by subclassing `FlxSoundTray`:

https://github.com/HaxeFlixel/flixel/blob/31bd217f170784e50eeabf437a862119cd161448/flixel/FlxGame.hx#L173-L177

Geokureli:
I don't see this getting much use, people rarely know about the normal sound tray and even if we had it commented out in the template project.xml it would takes a lot just to explain the different controls. plus not everyone has a numpad.

personally if I were to make my own it would just be +/-/0 for master and shift for music

My biggest issue is renaming existing vars like volumeUp, to soundVolumeUp, and exposing new vars that don't do anything without FLX_SPLIT_SOUND_TRAY. These are breaking changes.

PR #2392, opened 2021-08-24
Fix a crash when clicking buttons in debugger.

MasterEric:


Geokureli:
Can you add more info on why it's crashing?

MasterEric:
If I recall, I was experiencing a crash issue when attempting to minimize windows in the HaxeFlixel debug view.

The error occurred where windowSettings was not getting initialized properly and thus clicking the button crashed the app.

I fixed it by adding this line and made a pull request for it, but I'm not certain now whether the issue is actually supposed to be resolved through some other means (ensuring windowSettings is properly initialized for example)

Geokureli:
it's hard to judge this until we have a solid list of steps to repro the specific issue that this fixes

Gama11:
I also don't really understand how this could happen, considering `loadSaveData()` is called in the constructor. Not really comfortable with merging because it suggests there might be some deeper issue which this just works around.

If you figure out what the steps to reproduce it are (almost certainly involves deleting the save), please let us know.

PR #2396, opened 2021-08-28
turn FlxG.sound.loadSavedPrefs() into a public function

ninjamuffin99:
Just lets you easily call `FlxG.sound.loadSavedPrefs()` from anywhere in code when needed.

If you rebind FlxG.save's, and use the FlxG.save.data for your own purpose, adjusting the sound tray volume will save to your newly binded file. This means that when FlxGame and all the SoundFrontEnd stuff is normally initialized, your last "saved" volume settings will NOT be set properly, because loadSavedPrefs() will only be called once during initalization, and the saves will be binded to "flixel" instead of your new FlxSave binds that has the actual sound save data with it. Allowing loadSavedPrefs() would be very handy

### DIFF
--- a/flixel/system/frontEnds/SoundFrontEnd.hx
+++ b/flixel/system/frontEnds/SoundFrontEnd.hx
@@ -417,7 +417,7 @@ class SoundFrontEnd
 	/**
 	 * Loads saved sound preferences if they exist.
 	 */
-	function loadSavedPrefs():Void
+	public function loadSavedPrefs():Void
 	{
 		if (FlxG.save.data.volume != null)
 		{


Gama11:
What do you mean by "rebind FlxG.save"? Call `FlxG.save.bind()` again? ...why?

ninjamuffin99:
as referred to in #2202 , on sites like newgrounds, to have persistent save data, you need to specify the `path` variable in `bind()`, which isn't specified by default when setup by FlxG, since by default I believe it uses the URL of the web page or something like that, and when on Newgrounds using the API, the URL is different per session, leading to different potential save data when not properly rebinded.

Gama11:
Ah, I see, so it's not actually the `"flixel"` save name you mentioned initially that's the issue here. Not a big fan of this solution though, since you need to remember to manually do two separate things.

It seems like ideally, you should be able to control what path is used when `FlxG.save.bind()` is first called.

Geokureli:
> It seems like ideally, you should be able to control what path is used when `FlxG.save.bind()` is first called.
This. I've always hated that there's an initial default save replaced with my intended save path. it causes weirdness with sound like the intro's haxe stinger. I don't see a good solutio, though. We either add a param to FlxGame, or delay the first `bind()` but both of those solutions suck

calling `FlxG.sound.loadSavedPrefs()` seems like a hack. is there any reason it's not called from `FlxG.save.bind()`?

Gama11:
>We either add a param to FlxGame's constructor,

That was my thought too, and I think it acutally only sucks because it already has too many parameters. However, I would probably introduce a static `FlxGame.create()` factory function that takes a structure with all the parameters as optional fields (basically emulating named parameters which Haxe doesn't have). That seems like a general quality-of-life improvement, since then you can just pick and choose which things to set - that scales much better.

>calling FlxG.sound.loadSavedPrefs() seems like a hack.

It is.

>is there any reason it's not called from FlxG.save.bind()?

That's a hack too, there's no reason a generic save functionality should have a dependency on `FlxG.sound`. It's just this specific default save that happens to persist the sound settings.

Geokureli:
> ... but I think I still like the `FlxGame` constructor solution better.

aside from it already having too many (confusing) arguments already, this doesn't really "fix" the issue of sound prefs not updating when binding to a new save. I've had games with multiple concurrent `FlxSave` instances (it looked at saves for my other games). in cases like that a manual call to manage sounds prefs is the only thing that makes sense.

what if we had public `mySave.loadSoundPrefs()` and `mySave.saveSoundPrefs()` which check `FlxG.sound`'s settings rather than a private method in the `SoundFrontEnd`? I think cam wants this PR because flixel tries to automate the sound pref saving/loading in a way thats asking for hacky workarounds for cases like this

Gama11:
>this doesn't really "fix" the issue of sound prefs not updating when binding to a new save.

The fact that you can even call `FlxG.save.bind()` is basically an accident. The intention behind exposing the default save as a public `FlxG` property was just convenience - Flixel already creates a save, why not let the user save stuff there if he wants? But there's nothing stopping you from creating a separate `FlxSave` instance, and there's no reason that save would have to duplicate the sound preferences if you have separate "flixel" and "game data" saves.

There's another reason "rebinding" the default save to a different path seems hacky: you basically end up creating saves in two locations, with one of them being unused.

>what if we had public mySave.loadSoundPrefs() and mySave.saveSoundPrefs()

Again, `FlxSave` shouldn't know anything about sound preferences. It's a generic class for saving stuff, saving sound settings it just one possible concrete use case for it.

Geokureli:
I think we could eliminate the need to rebind, or ever call `bind` in general by setting web's default binding to `bind(appTitle, urlDomainName)`. It should also check `bind("flixel", null)` as well for backwards compatibility.

Other than that, @ninjamuffin99 should probably just use @:privateAccess for this case.

Right now FlxSave's intended use isn't very inutitive for web. I always call bind with custom args because of this and I assume it's the norm among web devs (at least the ones aware of this issue)

Geokureli:
fixed by #2566

PR #2401, opened 2021-09-04
add setPolarDegrees and setPolarRadians

Geokureli:
helpers to set the vector's length and angle. I find that most times I use `v.radians` or `v.degrees` it's after `v.set(speed, 0)` or a set to `length` in some way.

open to naming suggestions, because I'm never happy with the names I choose


Gama11:
Hm, we already have `FlxAngle.getCartesianCoords()` and `FlxAngle.getPolarCoords()`.

Geokureli:
it seems really odd to use a point to store length/angle in x/y values like it does in`getPolarCoords`. `FlxAngle` seems all over the place with how it handles degrees vs radians, in some places it just expects one and calls it `angle` and in others it has a bool to specify radians/degrees.

Both FlxAngle and FlxVector are lesser known flixel utils, so I don't mind having polar helpers in both places, each following their own unit naming conventions

Geokureli:
When did FlxTweenTest start failing unit tests?

PR #2402, opened 2021-09-08
Embed autotile assets

Geokureli:
Related to https://github.com/HaxeFlixel/flixel-demos/issues/281

To Summarize: `FlxGraphic.fromClass(GraphicAuto)` results in a BitmapData of size 0x0 while it fetches the data asynchronously. When passed into a tilemap, no actual visible or solid tiles are created, this is why the [EZPlatformer](https://haxeflixel.com/demos/EZPlatformer/) and [SplitScreen](https://haxeflixel.com/demos/SplitScreen/) demos don't work.

This change allows the autotile tilemaps to be used via `"flixel/images/tile/autotiles.png"`. I plan to update the demos to do this if this is approved.

There may be other ways to fix this so people can use flixel's default autotile tilemaps. I don't know if there's some event that our demos could listen for before using the images, but I'd rather these demos avoid async listeners and look extremely simple to newcomers looking for snippets.

I don't know what effect this has on the export's file size to embed these images in every project. I assume they are already embedded via:`@:keep @:bitmap`. does this mean they are embedded twice? should we deprecate the `@:bitmap` classes?


Geokureli:
`export/html5/bin/` sizes (bytes):
| Demo                  |  old  |  new  |  diff  |
|-----------------|------|------|------|
| EZPlatformer | 6,203,303 | 6,198,870 | -4,433 |
| SplitScreen | 6,195,720 | 6,185,140 | -10,580 |

/shrug/

Gama11:
It seems like it should be possible to just make the existing `GraphicAuto` etc classes work. For instance, we could try turning it into a `private class _GraphicAuto` that is extended by the `GraphicAuto` class visible to the user. That subclass could then probably set width / height in its constructor.

Geokureli:
I'll look into that, then. If we could avoid making people use arbitrary "flixel/images" paths than we should.

flixel/system/FlxAssets.hx
```diff
@@ -152,7 +152,7 @@ class FlxAssets
 	 */
 	public static inline function getBitmapFromClass(source:Class<Dynamic>):BitmapData
 	{
-		return Type.createInstance(source, [0, 0]);
+		return Type.createInstance(source, []);
```
Geokureli, line 155: @Gama11 I assume this isn't meant for normal bitmaps, but bitmap classes created using @:bitmap. we would never actually want to make a bitmap of 0x0, so this shouldn't be a breaking change, right?

Gama11: Are you sure this works on all targets? I thought this had to be here because width and height generally aren't optional parameters.

flixel/tile/FlxTilemap.hx
```diff
@@ -33,6 +33,46 @@ import Std.is as isOfType;
 
 using flixel.util.FlxColorTransformUtil;
 
+#if html5
```
Gama11, line 36: Might as well use this approach on all targets?

Geokureli: html5 is the only one that loads the bitmaps asynchronously so there's no need for this on other targets. consequently, only html5 has the `onLoad` arg, so we would need compile conditions for 2 versions of the constructor, anyway.

flixel/system/FlxAssets.hx
Geokureli: BitmapData's width and height properties are not optional, but classes that use `@:bitmap` have all optional arguments. I tested EZPlatformer and Splitscreen on native mac, since they both use `FlxGraphic.fromClass(GraphicAuto)` and they had no issue

flixel/tile/FlxTilemap.hx
Gama11: Ah, I didn't realize there's another difference, that's unfortunate...

PR #2407, opened 2021-09-15
Fixing little error for android build

ELfox513:
I don't know what else to write. When I tried to make a build on Android, I got an error "FlxAction.hx:141 characters 18-46 : Type not found : FlxActionInputDigitalAndroid". And when I added this line of code, the error did not appear anymore


Geokureli:
you should wrap that in an `#if android ... #end` compiler condition, otherwise other targets will have errors

ELfox513:
> you should wrap that in an `#if android ... #end` compiler condition, otherwise other targets will have errors

Yes. I just saw that flixel.input.android.FlxAndroidKey doesn't use #if-#end, so I decided not to use them natively too

PR #2410, opened 2021-09-18
Nothing interesting. Just added missing import.

koteelok:
When I tried to compile to android, I an error popped up:
"C:/HaxeToolkit/haxe/lib/flixel/4,10,0/flixel/input/actions/FlxAction.hx:141: characters 18-46 : Type not found : FlxActionInputDigitalAndroid"

So, i just added missing FlxActionInputDigitalAndroid import to FlxAction.


Gama11:
Already fixed by #2407.

PR #2411, opened 2021-09-22
Explain FlxG singletons in input util docs

Geokureli:


flixel/input/FlxKeyManager.hx
```diff
@@ -5,6 +5,11 @@ import flixel.FlxG;
 import flixel.input.FlxInput.FlxInputState;
 import flixel.input.keyboard.FlxKey;
 
+/**
+ * Keeps track of what keys are pressed and how with handy Bools or strings.
+ * Automatically instatiated by flixel via `FlxKeyboard` and accessed via `FlxG.keys`.
```
Gama11, line 10: Not only that, also android keys.

Geokureli: is that automatically instantiated on android, how is it accessed?

Gama11: FlxG.android :D

PR #2422, opened 2021-10-02
Collision fixes

Geokureli:
For https://github.com/chosencharacters/squidBounties/issues/6

Here are videos showing the changes.
https://imgur.com/a/9yXzCwk

summary:
- preserve momentum in collisions, fixing several bugs with mass. (Used [this article](https://en.wikipedia.org/wiki/Coefficient_of_restitution) as a guide and [this to test](https://try.haxe.org/#2AF24796))
- add a new `collisionXDrag` that takes an enum, to allow objects to stick to non-immovable objects.
- deprecate `collisonXDrag` (missing an "i" in collis**i**on)

breaking changes:
- the new collision will break old replays, so I added the `FLX_4_LEGACY_COLLISION` compiler flag

I'm still checking all the demos, but I'm pretty confident they will all work


Geokureli:
Yeah all the demos are behaving nicely, and none of the snippets should be affected

Geokureli:
Just noting: I tried modifying how separate() moves the colliding objects. I tried displacing inversely proportional to their mass, like so:
```hx
Object1.x -= overlap * mass2 / (mass1 + mass2);
Object2.x += overlap * mass1 / (mass1 + mass2);
```
this caused a lot of problems with stacking heavy objects on light objects, and pushing objects into other objects, so I decided against it.

PR #2425, opened 2021-10-10
Make FlxG.log.redirectTraces true by default

migmoog:
When a new user tries out their first hello world, they use `trace` to print the desired message to the console. HaxeFlixel has its own debugger with its own console that can accept `trace` calls, but it isn't enabled by default. The average user would expect to print to the console with the command that actually prints, but first they have to set `FlxG.log.redirectTraces = true` every time. It'd make more sense to just have the ability to use `trace` with the debugger by default, because it saves their traces across whatever platform their testing, and keeps the community from having to answer the same question of how to print to the debugger's console.

### DIFF
--- a/flixel/system/frontEnds/LogFrontEnd.hx
+++ b/flixel/system/frontEnds/LogFrontEnd.hx
@@ -14,7 +14,7 @@ class LogFrontEnd
 	/**
 	 * Whether everything you trace() is being redirected into the log window.
 	 */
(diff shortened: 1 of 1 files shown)


Gama11:
That actually used to be the default, but it was changed. I don't agree that `true` is a better default, the platform's native console usually has a much better UX than the built-in log window, especially for HTML5.

Geokureli:
Not a fan, personally. I know I would be disabling this on every project I make.

Issue #2239, opened 2019-11-05
FlxVector operators


Geokureli:
FlxVector is an abstract of FlxPoint now, why not add some operators? All of the underlying functions already exist:
A + B: addNew
A - B: subtractNew
vec * num: scaleNew
A * B: dotProduct
A += B: addPoint
A -= B: subtractPoint
A *= B: scale

I'm not super into operator overloads, but they are useful from time to time. just wanted to get some opinions.

PR #2435, opened 2021-10-23
Overload operators of FlxVector abstract.

T1mL3arn:
As it was suggested in #2239.

- [x] A + B: addNew
- [x] A - B: subtractNew
- [x] V * N: scaleNew
- [x] A * B: dotProduct
- [x] A += B: addPoint
- [x] A -= B: subtractPoint
- [x] V *= N: scale
- [x] -A: negateNew
- [x] A == B: equals
- [x] A != B: !equals

Though this

```haxe
Assert.isTrue(vector + vector2 == new FlxVector(4, 7));
Assert.isTrue(vector == new FlxVector(1, 2));
```

looks more nice, than this

```haxe
Assert.isTrue(vector.addNew(vector2).equals(new FlxVector(4, 7)));
Assert.isTrue(vector.equals(new FlxVector(1, 2)));
```

I am not sure about comparison operator:

```haxe
var a = new FlxVector(1, 2);
var b = new FlxVector(1, 2);

trace(a == b);   // `false` now, will be `true` with overloaded comparison
```

### DIFF
--- a/flixel/math/FlxVector.hx
+++ b/flixel/math/FlxVector.hx
@@ -45,7 +45,7 @@ import openfl.geom.Point;
 
 	@:noCompletion
 	@:op(A + B)
-	static inline function addPointNew(lhs:FlxVector, rhs:FlxVector):FlxVector
+	static inline function sumVectorsOp(lhs:FlxVector, rhs:FlxVector):FlxVector
 	{
 		var result = weak(lhs.x, lhs.y).addPoint(rhs);
 		lhs.putWeak();
@@ -54,7 +54,7 @@ import openfl.geom.Point;
 
 	@:noCompletion
 	@:op(A - B)
-	static inline function subtractPointNew(lhs:FlxVector, rhs:FlxVector):FlxVector
+	static inline function subtractVectorsOp(lhs:FlxVector, rhs:FlxVector):FlxVector
 	{
 		var result = weak(lhs.x, lhs.y).subtractPoint(rhs);
 		lhs.putWeak();
@@ -63,7 +63,7 @@ import openfl.geom.Point;
 
 	@:noCompletion
 	@:op(A * B) @:commutative
-	static inline function scalePointNew(lhs:FlxVector, rhs:Float):FlxVector
+	static inline function scaleVectorOp(lhs:FlxVector, rhs:Float):FlxVector
 	{
 		var result = weak(lhs.x, lhs.y).scale(rhs);
 		lhs.putWeak();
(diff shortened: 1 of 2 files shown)


T1mL3arn:
>I'm also a little wary of -N creating a new instance

What problems or unexpected behavior can be with this ?

flixel/math/FlxVector.hx
```diff
@@ -6,6 +6,149 @@ import openfl.geom.Point;
...
+ * 
+ * var A = FlxVector.weak(1, 1);
+ * var B = FlxVector.weak(2, 2);
+ * 
+ * var C = (A + B);		// A and B return to pool, C is weak
+ * 
+ * // "unweak" vector C:
+ * // We create a new non weak vector and copy vector C into it. 
+ * // Original vector C returns to pool in `copyFrom()`
+ * C = FlxVector.get().copyFrom(C);
```
Geokureli, line 144: As I said before, even if it we clearly documented this, people still will have problems, and not know why. automatic weak vectors are unintuitive and problematic. I'd much rather have GC'd instances from intermediary vectors as a trade-off for the convenience of operators.

The whole point of these operators is convenience and simpler code and this implementations is not intuitive, convenient or simple.

T1mL3arn: The way that "problematic" operators implemented (they are hidden from completion) dictate the way how users find them (they thoroughly read `FlxVector` docs). Thus it should not be a concern if someone misuses them.  And again, flixel must prohibit using disposed vectors in a loud and explicit way, so the users of "problematic" operators get such error as soon as possible.

>this implementation is not intuitive, convenient or simple

Math operations on complex objects just cannot be intuitive, convenient or simple. All these three terms might be understood differently by different people.

Suggestions:

- type of the result vector can be controlled with some global flag on `FlxVector` class
  ```haxe
  public static inline var EXPRESSIONS_RETURN_WEAK = true;

  static inline function sumVectorsOp(lhs:FlxVector, rhs:FlxVector):FlxVector
  {
    var result = weak(lhs.x, lhs.y).addPoint(rhs);
    lhs.putWeak();
    result._weak = EXPRESSIONS_RETURN_WEAK;
    return result;	
  }
- unweak vectors with macro (if this is possible to write such macro)

Geokureli: > The way that "problematic" operators implemented (they are hidden from completion) dictate the way how users find them (they thoroughly read FlxVector docs). Thus it should not be a concern if someone misuses them.

As someone who routinely aids newcomers to flixel, can 100% say that people do not read the doc. Most of them learn from reading or copying from other people code. Its quite common that i have to explain the basic duty of a function they found but did not read into. Plus anyone coming from other languages with vector operators will likely just try them out or assume they exist.

Geokureli: Apparently we may be able to avoid this argument entirely by inline the constuctors of FlxPoint and FlxVector, and that it might even work without having to remove our existing pooling system. This seems like something worth looking into

T1mL3arn: > inline the constuctors of FlxPoint and FlxVector

Doesn't it require complete reworking of  `FlxPoint` ? Which affects `FlxCallbackPoint` and so on.

Geokureli:
operators were added via #2557

PR #2440, opened 2021-11-05
A simple bool for all the possible targets.

brosomethingwrongwithyothing:
The changes will simple but might go a long way sometimes.


flixel/FlxG.hx
```diff
@@ -309,6 +309,10 @@ class FlxG
 	 */
 	public static var plugins(default, null):PluginFrontEnd;
 
+	/**
+	 * Whether the game is running on HTML5.
+	 */
+	 public static var onHTML5:Bool = #if html5 true #else false;
```
Geokureli, line 315: 1. this should be readonly, look at `FlxG.onMobile` as a guide

flixel/FlxG.hx
```diff
@@ -558,11 +562,15 @@ class FlxG
 	 */
 	public static inline function openURL(URL:String, Target:String = "_blank"):Void
 	{
+		#if linux
+		Sys.command('/usr/bin/xdg-open', [URL, "&"]);
+		#else
```
Geokureli, line 567: First, the linux url issue is an issue with openFL and should likely be fixed there. Second, this wouldn't work with `FlxG.open("haxe.org")` which does work in html5 because flixel prepends a "https://" to links without it. I had linux users test this and it worked for all but one, due to their environment setup.

Geokureli:
```
❯ xdg-open haxe.org
xdg-open: file 'haxe.org' does not exist
```

Geokureli (changes requested):
I don't like the openURL changes, just make a PR to openfl for that. also I would battle test that with a bunch of linux users because apparently the '&' wont work for everyone either.

```
xdg-open: unexpected argument '&'
Try 'xdg-open --help' for more information.
```

flixel/FlxG.hx
brosomethingwrongwithyothing: ok i see i go do that

flixel/FlxG.hx
brosomethingwrongwithyothing: done

flixel/FlxG.hx
Geokureli: This won't compile because you never created the getter, like this
https://github.com/HaxeFlixel/flixel/blob/dev/flixel/FlxG.hx#L768-L778

brosomethingwrongwithyothing: huh? i dont need no getter i do #if html5 true #else false; #end

brosomethingwrongwithyothing:
> I don't like the openURL changes, just make a PR to openfl for that. also I would battle test that with a bunch of linux users because apparently the '&' wont work for everyone either.
> 
> ```
> xdg-open: unexpected argument '&'
> Try 'xdg-open --help' for more information.
> ```
> 
> I also have issues with onHTML5, stated below. Personally, I don't see the need as `#if js` is fine, and people can easily make their own global flag/util. should we also have one for all the other target flags?

ok sure

brosomethingwrongwithyothing:
Gonna add all the flags because @Geokureli told me too.

flixel/FlxG.hx
Geokureli: you need a getter if you have `(get, never)`

you can also just make it inline
```public static inline var onHTML5:Bool = #if html5 true #else false; #end```
also note that you need `#end`

Geokureli: also there seems to be a space before public, please remove that

brosomethingwrongwithyothing: ah see im not good with var stuff ig

Geokureli:
> Gonna add all the flags because @Geokureli told me too.

that wasn't an actual request

brosomethingwrongwithyothing:
> > Gonna add all the flags because @Geokureli told me too.
> 
> that wasn't an actual request, I was being facetious

ok? im still doing it

brosomethingwrongwithyothing:
CodeClimate what do you mean by the var name or something is wrong it okie dokie

Geokureli:
It wants consts to be ON_HTML5. If you want onHTML5 it should be a getter like onMobile

Geokureli:
> ok? im still doing it

We dont want them

PR #2441, opened 2021-11-07
sceenCenter Readibility

Geokureli:
Someone in the discord was really confused by this. Not sure why it takes Null<FlxAxes> when it could default to XY and be more easily understood. The test was added because why not.

this will break existing code where a null literal or a Nullable FlxAxes were passed in, but I'm fine disallowing that.


Geokureli:
Hmm, i see. So this fails on 3.4.7 cpp

PR #2452, opened 2021-11-23
change FlxColor.GREEN from dark green to green

Geokureli:


Geokureli:
not sure why cpp is failing and others aren't, testing locally

Geokureli:
Just Noticed LIME. wtf? should be GREEN and DARK_GREEN imo. giving up on this, though

PR #2466, opened 2021-12-13
Add the "FLX_SUPPRESS_ICON" define

MasterEric:
This resolves an issue I have been experiencing, in which my project (which utilizes HXP for project configuration rather than XML) is having its icons be overridden by the one provided by the HaxeFlixel library.

The solution may be instead to resolve the underlying issue, see haxelime/lime#1508

### DIFF
--- a/include.xml
+++ b/include.xml
@@ -2,7 +2,7 @@
 <project xmlns="http://lime.software/project/1.0.2" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
 	xsi:schemaLocation="http://lime.software/project/1.0.2 http://lime.software/xsd/project-1.0.2.xsd">
 
-	<icon path="assets/images/logo/HaxeFlixel.svg" unless="SUPPRESS_FLIXEL_ICON" />
+	<icon path="assets/images/logo/HaxeFlixel.svg" unless="FLX_SUPPRESS_ICON" />
 	
 	<section unless="unit-test">
 		<assets path="assets/sounds" rename="flixel/sounds" include="*.mp3" if="web || air" embed="true" />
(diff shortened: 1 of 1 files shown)


Gama11:
Hm, I'd prefer to avoid hacks like this. The hxp format is used so rarely that it doesn't seem justified just for that.

player-03:
I've come up with another possible solution: haxelime/lime#1509. The intended usage would look like this:

```xml
<icon path="assets/images/logo/HaxeFlixel.svg" priority="-1" />
```

Lower-priority icons would only get used if there wasn't already an icon of the needed size, even if they appear at the end of the array (as happens with HXP projects).

But the benefits wouldn't be limited to HXP projects. Even XML users have to worry about this, as demonstrated by a line that appears in [78 of your 84 demos](https://github.com/HaxeFlixel/flixel-demos/search?q=higher+priority+to+override+the+HaxeFlixel+icon):

```xml
<!--Place custom nodes like icons here (higher priority to override the HaxeFlixel icon)-->
```

With this change, there would no longer be any need to warn users about ordering, making your demos ever so slightly simpler. So the question is: would you use this implementation? Or can you think of a better one?

Gama11:
@player-03 Sure, why not. That's much more explicit that relying on ordering.

player-03:
Great! Once everything is ready on Lime's side, I'll submit a pull request here.

(Maybe I'll suggest `priority="-10"`, just so others can put numbers in between...)

Gama11:
> maybe I'll suggest priority="-10", just to leave room in between...

It's z-index all over again. :D

Geokureli:
closing for #2577

PR #2470, opened 2021-12-28
Make FlxRandom portable

### LINKED ISSUE #2469: Make FlxRandom portable
After some digging, it seems like `FlxRandom.hx` might be the best cross-platform RNG implementation in pure Haxe. I actually want to use it for projects that go beyond HaxeFlixel -- but this doesn't work currently because FlxRandom.hx references flash types (all indirectly, through FlxG, FlxSprite, etc.)

By editing the source directly in my haxelib folder, I was able to add just a few `#if flash` directives that made it possible for me to add `-lib flixel` to a non-flixel project and import FlxRandom for my purposes. Maybe this would be useful to others too? Would a PR be welcome for this?

NQNStudios:
This adds a few `#if flash` directives, allowing the FlxRandom class to be imported by projects that aren't HaxeFlixel games.

Close #2469

### DIFF
--- a/flixel/math/FlxMath.hx
+++ b/flixel/math/FlxMath.hx
@@ -1,8 +1,10 @@
 package flixel.math;
 
+#if flash
 import flash.geom.Rectangle;
 import flixel.FlxG;
 import flixel.FlxSprite;
+#end
 #if FLX_TOUCH
 import flixel.input.touch.FlxTouch;
 #end
@@ -174,6 +176,7 @@ class FlxMath
 		return false;
 	}
 
+	#if flash
 	/**
 	 * Returns true if the given x/y coordinate is within the given rectangular block
 	 *
@@ -226,6 +229,7 @@ class FlxMath
 	{
 		return pointX >= rect.x && pointX <= rect.right && pointY >= rect.y && pointY <= rect.bottom;
 	}
+	#end
 
 	/**
 	 * Adds the given amount to the value, but never lets the value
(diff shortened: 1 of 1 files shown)


Geokureli:
These aren't flash only features they're available on all platforms, so by adding these conditionals you're breaking all non flash targets.

 Note that `flash.geom.Rectangle` actually points to `openfl.geom.Rectangle` on non-flash targets:
 https://github.com/openfl/openfl/blob/develop/lib/flash-externs/src/flash/geom/Rectangle.hx#L37
 
 There's plenty of standalone random number generating haxelibs, already. I'm not convinced that it's worthwhile to make FlxRandom standalone

NQNStudios:
In going through all the other RNG haxelibs, I saw that none of them have had commits within the last year. I'm hoping to use one that is being maintained, which HaxeFlixel is. And I could copy and modify the code, but that brings up a licensing mismatch with my codebase, and would make it harder to contribute back any improvements I might make.

What about using `#if (flash || openfl)`?

Geokureli:
FlxRandom is such a tiny class and it's under MIT, you could literally just make a new haxelib, change a couple names, remove flixel refs and put any license on it. At most maybe put a blurb in the documentation about how it came from HaxeFlixel. there's plenty of classes in the flixel engine that say it was stolen from other libs.

I don't think flixel should worry about the "portability" of it's tiny features and I don't recommend using flixel just for FlxRandom

NQNStudios:
OK thank you for the licensing advice!

PR #2473, opened 2021-12-30
Add FlxAnimationController.getNamesList() to get the current list of animation names

Cheemsandfriends:
This adds a variable with all the animations currently added.
So that you can make plenty of things with it.

### DIFF
--- a/flixel/animation/FlxAnimationController.hx
+++ b/flixel/animation/FlxAnimationController.hx
@@ -792,7 +792,7 @@ class FlxAnimationController implements IFlxDestroyable
 		return arrayNames;
 	}
 
-	public function namesList():Array<String>
+	public function nameList():Array<String>
 	{
 		var anims = animationList();
 		var names:Array<String> = [];
@@ -806,13 +806,10 @@ class FlxAnimationController implements IFlxDestroyable
 
 	public function exists(name:String):Bool
 	{
-		if (namesList().contains(name))
-			return true;
-		else
-			return false;
+		return nameList().contains(name);
 	}
 	
-	public function renameName(name:String, value:String)
+	public function renameAnimation(name:String, value:String)
 	{
 		var prevAnims = animationList();
 
--- a/flixel/animation/FlxAnimationController.hx
+++ b/flixel/animation/FlxAnimationController.hx
@@ -196,9 +196,7 @@ class FlxAnimationController implements IFlxDestroyable
 	 * @param   FlipY       Whether the frames should be flipped vertically.
 	 */
 	public function add(Name:String, Frames:Array<Int>, FrameRate:Float = 30, Looped:Bool = true, FlipX:Bool = false, FlipY:Bool = false):Void
-	{
-		_list.push(Name);
-		
+	{	
 		// Check _animations frames
 		var framesToAdd:Array<Int> = Frames;
 		var numFrames:Int = framesToAdd.length - 1;
@@ -393,8 +391,6 @@ class FlxAnimationController implements IFlxDestroyable
 	public function addByIndices(Name:String, Prefix:String, Indices:Array<Int>, Postfix:String, FrameRate:Int = 30, Looped:Bool = true, FlipX:Bool = false,
 			FlipY:Bool = false):Void
 	{
-		_list.push(Name);
-		
 		if (_sprite.frames != null)
 		{
 			var frameIndices:Array<Int> = new Array<Int>();
@@ -475,8 +471,6 @@ class FlxAnimationController implements IFlxDestroyable
 	 */
 	public function addByPrefix(Name:String, Prefix:String, FrameRate:Int = 30, Looped:Bool = true, FlipX:Bool = false, FlipY:Bool = false):Void
 	{
-		_list.push(Name);
-		
 		if (_sprite.frames != null)
 		{
 			var animFrames:Array<FlxFrame> = new Array<FlxFrame>();
@@ -797,7 +791,14 @@ class FlxAnimationController implements IFlxDestroyable
 		
 	function get_list():Array<String>
 	{
-		return _list;
+		var arrayNames:Array<String> = [];
+		for (i in _animations)
+		{
+			if (!arrayNames.contains(i.name))
+				arrayNames.push(i.name);
+		}
+		
+		return _list = arrayNames;
 	}
 
 	inline function get_curAnim():FlxAnimation
--- a/flixel/animation/FlxAnimationController.hx
+++ b/flixel/animation/FlxAnimationController.hx
@@ -31,7 +31,7 @@ class FlxAnimationController implements IFlxDestroyable
 	/**
 	 * Gets or sets a list with all the added animations (warning: can be `null`).
 	 */
-	public var list(get, never):Array<String>;
+	public var list(get, set):Array<String>;
 	
 	/**
 	 * Pause or resume the current animation.
@@ -799,6 +799,29 @@ class FlxAnimationController implements IFlxDestroyable
 		}
 		
 		return _list = arrayNames;
+	}
+	
+	function set_list(value:Array<String>):Array<String>
+	{
+		if (value.length > _list.length)
+		{
+			FlxG.log.warn("value Exceeds the amount of names currently!");
+			return _list;
+		}
+		var savePrevAnims:Array<FlxAnimation> = [];
+		
+		for (i in _animations)
+		{
+			savePrevAnims.push(i);
+		}
+		_animations = new Map<String, FlxAnimation>();
+		for (i in 0...value.length)
+		{
+			savePrevAnims[i].name = value[i];
+			_animations.set(value[i], savePrevAnims[i]);
+		}
+
+		return _list = value;
 	}
 
 	inline function get_curAnim():FlxAnimation
--- a/flixel/animation/FlxAnimationController.hx
+++ b/flixel/animation/FlxAnimationController.hx
@@ -805,20 +805,30 @@ class FlxAnimationController implements IFlxDestroyable
 	{
 		if (value.length > _list.length)
 		{
-			FlxG.log.warn("value Exceeds the amount of names currently!");
+			FlxG.log.warn("value exceeds the amount of animation names!");
 			return _list;
 		}
+
 		var savePrevAnims:Array<FlxAnimation> = [];
 		
 		for (i in _animations)
 		{
 			savePrevAnims.push(i);
 		}
+
 		_animations = new Map<String, FlxAnimation>();
-		for (i in 0...value.length)
+
+		for (i in 0...savePrevAnims.length)
 		{
-			savePrevAnims[i].name = value[i];
-			_animations.set(value[i], savePrevAnims[i]);
+			if (value[i] != null)
+			{
+				savePrevAnims[i].name = value[i];
+				_animations.set(value[i], savePrevAnims[i]);
+			}
+			else
+			{
+				_animations.set(savePrevAnims[i].name, savePrevAnims[i]);
+			}
 		}
 
 		return _list = value;
--- a/flixel/animation/FlxAnimationController.hx
+++ b/flixel/animation/FlxAnimationController.hx
@@ -804,7 +804,17 @@ class FlxAnimationController implements IFlxDestroyable
 
 	public function exists(name:String):Bool
 	{
+		#if (haxe > "4.0.5")
 		return nameList().contains(name);
+		#else
+		for (i in 0...nameList().length)
+		{
+			if (nameList[i] == name)
+				return true;
+		}
+
+		return false;
+		#end
 	}
 	
 	public function renameAnimation(name:String, value:String)
--- a/flixel/animation/FlxAnimationController.hx
+++ b/flixel/animation/FlxAnimationController.hx
@@ -27,7 +27,7 @@ class FlxAnimationController implements IFlxDestroyable
 	 * Gets or sets the currently playing animation (warning: can be `null`).
 	 */
 	public var name(get, set):String;
-	
+
 	/**
 	 * Pause or resume the current animation.
 	 */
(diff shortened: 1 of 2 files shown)


Geokureli:
@Cheemsandfriends I noticed `remove()` doesn't remove names from the list, which points out how easy it can be to forget about this feature when adding new features. 

 what's the use-case you're imagining for this feature? I wonder if it's better to have a function like `getAnimationNames()` generate a new array each time it's called. Or would something like `hasAnimation(name:String)` suffice?

Cheemsandfriends:
> @Cheemsandfriends I noticed `remove()` doesn't remove names from the list, which points out how easy it can be to forget about this feature when adding new features.
> 
> what's the use-case you're imagining for this feature? I wonder if it's better to have a function like `getAnimationNames()` generate a new array each time it's called. Or would something like `hasAnimation(name:String)` suffice?

Well, I first imagined like a list of animation names that has the thing, but I'd maybe change it to `array<FlxAnimation>` better. So that instead of animation Names, You would see the list with the animation stuff.
I'd look more into it but I'll check it as a draft PR cos it's pretty dumb the idea tbh lol

Geokureli:
I think it's a good idea, but I have issues with the execution

Cheemsandfriends:
> I think it's a good idea, but I have issues with the execution

Well, maybe not exactly names, but I'll try to see like animations in general, so instead of names, the in specific

Cheemsandfriends:
@Geokureli, I added a set option so that you can change the name of the whole list and the anims in-game, kinda experimental cos if the value array isn't the same length as the animations length it will be breaking but it's a nice feature imo

Geokureli:
I don't see the point of the setter, it seems like things could easily go very wrong.

If they want to mass rename the animations `animations.list = ["idle", "walk_l", "walk_r"];` seems like a very unintuitive way to do it for several reasons:
1. I don't really see the need to mass rename animation names in one line
2. The `_animations` map is orderless, there's no guarantee that you're setting a specific  animation's name this way across all targets.
3. animations.list[0] = "idle" wouldn't change anything, since animations.list wasn't set
4. I'm not even a fan of the getter, I would much rather have a function that returns the Array, disguising it as a property doesn't really help anything
5. `list` isn't very descriptive, we may want to reserve that if we decide to expose a list of `FlxAnimation`s

I'm gonna hard-pass on `set_list` entirely. and rather than `_list` and `list` vars it should be a method like `getNameList()` or `getNames()`. I also think `has(name)` would make a great addition.

 if people want to mass rename their animations (again, why?), i would rather have a `renameAnim()` method, but even then I don't really see the point

Cheemsandfriends:
> I don't see the point of the setter, it seems like things could easily go very wrong.
> 
> If they want to mass rename the animations `animations.list = ["idle", "walk_l", "walk_r"];` seems like a very unintuitive way to do it for several reasons:
> 
>     1. I don't really see the need to mass rename animation names in one line
> 
>     2. The `_animations` map is orderless, there's no guarantee that you're setting a specific  animation's name this way across all targets.
> 
>     3. animations.list[0] = "idle" wouldn't change anything, since animations.list wasn't set
> 
>     4. I'm not even a fan of the getter, I would much rather have a function that returns the Array, disguising it as a property doesn't really help anything
> 
>     5. `list` isn't very descriptive, we may want to reserve that if we decide to expose a list of `FlxAnimation`s
> 
> 
> I'm gonna hard-pass on `set_list` entirely. and rather than `_list` and `list` vars it should be a method like `getNameList()` or `getNames()`. I also think `has(name)` would make a great addition.
> 
> if people want to mass rename their animations (again, why?), i would rather have a `renameAnim()` method, but even then I don't really see the point

yeah, the renameNames is something maybe useful to have just in case someone wants to have a animation name and wants to change it for some reason.
Like if they were making a animation maker thing idk.
set list was basically renaming the names so I should prob think how it could work lol.

Geokureli:
To be honest, the ability to rename anim names seemed like an afterthought in this PR. like you wanted an animation list, made a getter, and then said "hey, a setter could be useful, too". Do any of your current projects need anim renaming?

Cheemsandfriends:
> To be honest, the ability to rename anim names seemed like an afterthought in this PR. like you wanted an animation list, made a getter, and then said "hey, a setter could be useful, too". Do any of your current projects need anim renaming?

Yeah, It was an after thinking of the PR. I haven't need a anim renaming but some wants to rename a name because lol.
but yeah, I'll think about a way to add a function that goes well with _animations cos it isn't static

Cheemsandfriends:
@Geokureli, I made the list of animations and name animations, and also if a animation exists and the possibility to rename names (Which idk why you can't see the potential with it but ok).
and removed the variable lol

Cheemsandfriends:
Oh wait, maybe it isn't ready

Geokureli:
be sure to check out the [Files changed](https://github.com/HaxeFlixel/flixel/pull/2473/files) to keep an eye on unintended changes or things you may have missed

Cheemsandfriends:
> be sure to check out the [Files changed tab](https://github.com/HaxeFlixel/flixel/pull/2473/files) to keep an eye on unintended changes or things you may have missed

I think I removed all the _list things, I should rewatch the whole thing again

flixel/animation/FlxAnimationController.hx
```diff
@@ -777,6 +777,55 @@ class FlxAnimationController implements IFlxDestroyable
 		play(AnimName);
 		return AnimName;
 	}
+		
+	public function animationList():Array<FlxAnimation>
```
Geokureli, line 781: `animationList` is a noun, flixel functions should be either `verb` or `verbNoun`, also, `animation.animationList` is redundant. This should be `getList`.

This also needs documentation. Above the function, there should should be a description of that this does, like so: 
```hx
/**
 * blah blah
 */
```

flixel/animation/FlxAnimationController.hx
```diff
@@ -777,6 +777,55 @@ class FlxAnimationController implements IFlxDestroyable
...
+		
+		for (i in _animations)
+		{
+			arrayNames.push(i);
+		}
+
+		return arrayNames;
+	}
+
+	public function nameList():Array<String>
```
Geokureli, line 793: should be getNameList

Needs documentation

flixel/animation/FlxAnimationController.hx
```diff
@@ -777,6 +777,55 @@ class FlxAnimationController implements IFlxDestroyable
...
+		var names:Array<String> = [];
+		for (i in 0...anims.length)
+		{
+			names.push(anims[i].name);
+		}
+
+		return names;
+	}
+
+	public function exists(name:String):Bool
```
Geokureli, line 805: needs documentaion

flixel/animation/FlxAnimationController.hx
```diff
@@ -777,6 +777,55 @@ class FlxAnimationController implements IFlxDestroyable
...
+	}
+
+	public function nameList():Array<String>
+	{
+		var anims = animationList();
+		var names:Array<String> = [];
+		for (i in 0...anims.length)
+		{
+			names.push(anims[i].name);
+		}
```
Geokureli, line 800: rather than generating 2 arrays to return one, can be
```hx
		var names:Array<String> = [];
		for (name in _animations.keys())
		{
			names.push(name);
		}
```

flixel/animation/FlxAnimationController.hx
```diff
@@ -777,6 +777,55 @@ class FlxAnimationController implements IFlxDestroyable
 		play(AnimName);
 		return AnimName;
 	}
+		
+	public function animationList():Array<FlxAnimation>
+	{
+		var arrayNames:Array<FlxAnimation> = [];
```
Geokureli, line 783: `arrayNames` is an incorrect name

flixel/animation/FlxAnimationController.hx
```diff
@@ -777,6 +777,55 @@ class FlxAnimationController implements IFlxDestroyable
...
+
+		return names;
+	}
+
+	public function exists(name:String):Bool
+	{
+		return nameList().contains(name);
+	}
+	
+	public function renameAnimation(name:String, value:String)
```
Geokureli, line 820: can just be called `rename`

needs documentation

for clarity, use `oldName` and `newName`

flixel/animation/FlxAnimationController.hx
```diff
@@ -777,6 +777,55 @@ class FlxAnimationController implements IFlxDestroyable
 		play(AnimName);
 		return AnimName;
 	}
+		
+	public function animationList():Array<FlxAnimation>
+	{
+		var arrayNames:Array<FlxAnimation> = [];
+		
+		for (i in _animations)
```
Geokureli, line 785: suggestion: I try to avoid `i` except for traditional `for (i in 0...length)` loops. I recommend `anim`

Cheemsandfriends:
contains doesn't exist in arrays?????

flixel/animation/FlxAnimationController.hx
```diff
@@ -777,6 +777,55 @@ class FlxAnimationController implements IFlxDestroyable
...
+			if (prevAnims[i].name == name)
+			{
+				prevAnims[i].name = value;
+				_animations.set(value, prevAnims[i]);
+			}
+			else
+			{
+				_animations.set(prevAnims[i].name, prevAnims[i]);
+			}
+		}
```
Geokureli, line 837: 1. we don't need to create a new animation map just to rename 1 entry, maps have a remove function. we also don't need to loop through animations, we can access by name via the map
2. we should throw a warning when they try to rename a non-existent animation.
3. We should return a bool indicating whether it was successful.

putting it all together, something like:
```hx
var anim = _animations[oldName];
if (anim == null)
{
	FlxG.log.warn('No animation called "$newName"');
	return false;
}
_animations[newName] = anim;
anim.name = newName;
_animations.remove(oldName);
return true;
```

flixel/animation/FlxAnimationController.hx
```diff
@@ -777,6 +777,65 @@ class FlxAnimationController implements IFlxDestroyable
...
+		return nameList().contains(name);
+		#else
+		for (i in 0...nameList().length)
+		{
+			if (nameList[i] == name)
+				return true;
+		}
+
+		return false;
+		#end
```
Geokureli, line 817: this should just be
```hx
return _animations.exists(name);
```

Geokureli (changes requested):
mainly, it needs documentation, and we shouldn't use a new generated array list just to loop through the animations for internal use, that's what _animations is for

flixel/animation/FlxAnimationController.hx
Cheemsandfriends: I think it's now resolved

flixel/animation/FlxAnimationController.hx
Cheemsandfriends: Yeah, I changed the name of the things so it's resolved

flixel/animation/FlxAnimationController.hx
Cheemsandfriends: Documentation given and changed the function, resolved, maybe

flixel/animation/FlxAnimationController.hx
Cheemsandfriends: changed the thing so yeah, resolved

flixel/animation/FlxAnimationController.hx
Cheemsandfriends: added the documentation, at least is what I can give as my perspective lol.
Resolved

flixel/animation/FlxAnimationController.hx
Cheemsandfriends: Thanks, I'm just so stupid that I can't code properly.
Resolved

flixel/animation/FlxAnimationController.hx
Cheemsandfriends: Yeah, changed the names of the vars and the function itself, also added some docs.
Resolved I guess

flixel/animation/FlxAnimationController.hx
Cheemsandfriends: I changed the code a bit but I think it gives the same thing and it gives some nice good-looking code imo.
Resolved

flixel/animation/FlxAnimationController.hx
Cheemsandfriends: I don't think anyone will see that but I changed the name var just in case lol
Resolved

Cheemsandfriends:
I added most of the stuff you requested,
Maybe the docs aren't `professional` looking, but that can change of course, I'm not very formal tbh.
If I have missed something, please add a review again, I would like to know what I made some error and correct it lol

Geokureli:
FYI 4.0.5 checks are failing until firetongue releases [this change](https://github.com/larsiusprime/firetongue/pull/47#issuecomment-1015058759). So I'm gonna have to do some manual checks for this

Cheemsandfriends:
> FYI 4.0.5 checks are failing until firetongue releases [this change](https://github.com/larsiusprime/firetongue/pull/47#issuecomment-1015058759). So I'm gonna have to do some manual checks for this

Sorry to ask but, you sure it was for Std? I think it was the array String that didn't have `contains` back in the day, but now that it isn't contains, it should be ok now

Geokureli:
> you sure it was for Std?
I knew that your previous errors were from contains, but as of late our 4.0.5 actions have been failing from firetongue, but apparently the actions use latest git for all dependencies rather than the latest haxelib release, which I didn't know.

I still need to do one small thing locally, however, 1 sec

Geokureli (approved):

Cheemsandfriends:
Thanks for changing the docs!
I don't know a lot of how giving a proper documentation so at least you changed that lol.

if I can ask, what are the @:Test things?

Geokureli:
I went ahead and added some lesser-known conventions rather than spewing an entire wall of requests at you, notably:

- tests: tests are run by our github actions to make sure future changes don't break the features you've added
- `@since` tags: all new public fields are required to say when they were added

Other than that I cleaned up some of the wording and whitespace. this is approved, but it may take a bit for me to actually merge it. I've just received the ability to approve PRs yesterday and there's 11 other PRs so I want to plan out the merge order and the release.

Cheemsandfriends:
@Geokureli, You forgot the Rename test, I tried to make a commit adding the test but idk if it works well, can you check it out pls?

Issue #2318, opened 2021-06-10
Flixel demos and examples crash/segfaults on the HL target  


MetalMaxMX:
- **Haxe version:** 4.2.2
- **Flixel version:** 4.9.0
- **OpenFL version:** 9.1.0
- **Lime version:** 7.9.0
- **Affected targets:** Linux's HashLink, latest stable

________________________________

**Code snippet reproducing the issue**:

```package;

import flixel.FlxState;

class PlayState extends FlxState
{
	override public function create()
	{
		super.create();
		
	var text = new flixel.text.FlxText(0, 0, 0, "Hello World", 64);
    text.screenCenter();
    add(text);
		
	}

	override public function update(elapsed:Float)
	{
		super.update(elapsed);
	}
}

```

________________________________

**Observed behavior:** When doing a lime test hl or lime test haxelink, the binary segfaults, the window on "Hello World" first appears for lower than a sec and then crashes immediately, this issue does not happens on: HTML5, Windows, Linux, Neko, etc. targets.  

**Expected behavior:** That the simple hello world HL binary runs as well as those observed from that of the other targets. 

On a side-note: A few of the flixel demos like flappybalt also manage to segfault the app as well.

Metlmeta:
What did you change the values to for NONE and ANY?

> ... numbers above 256 for HL switch statements
Can you clarify what you mean by that?

MetalMaxMX:
> What did you change the values to for NONE and ANY?
> 
> > ... numbers above 256 for HL switch statements
> 
> Can you clarify what you mean by that? Do you mean that HashLink can't support more than 256 switch statements or that any number above 256 used as a switch subject causes a crash?

Sure, I'm happy to answer and collaborate! 
I put in the values of NONE as equal 9 and ANY as equal 1. And with that, I was able to get flixel demos compiling on the HL target. On a stock installation of HaxeFlixel and HL on Linux, if you don't modify the values, it just crashes. 

I'm not entirely sure about the behavior of the latter as far as switch statements behavior go actually. Of HashLink I mean. Because I was told that this is strictly a bug inside Haxe's compiler that affects HashLink Linux support. In essence, the issue would lie within Haxe's compiler itself, not entirely with HashLink. I haven't done any more testing on this as I'm quite satisfied with my workaround. 

[...]

PR #2487, opened 2022-01-24
Prevent hashlink segfaults on linux

jobf:
This is to address the issue - https://github.com/HaxeFlixel/flixel/issues/2318

It doesn't fix the underlying issue (apparently something in hashlink) but the changes here would mean flixel can run on linux unchanged.

I've been running flixel with hashlink on linux for a long time with the changes in this PR and have not seen issues. I had a look through OpenFL and Lime key codes to be sure nothing else was using 254 or 255 already.

Aside from anecdotal personal use I have made a test flixel program here to ensure the FlxKey still works with the changes - https://github.com/jobf/flxkey-smoketest

I've tested with the following and it's working as expected.

- linux
- hashlink (on linux)
- html5 (on linux)
- html5 (on windows)

Ideally if someone can help test the following, that would be great. Currently compiling on windows is not working here but if I get it working I will update.

- windows
- mac
- hashlink (on mac)
- html5 (on mac)

### DIFF
--- a/flixel/input/FlxBaseKeyList.hx
+++ b/flixel/input/FlxBaseKeyList.hx
@@ -22,7 +22,7 @@ class FlxBaseKeyList
 
 	inline function check(keyCode:Int):Bool
 	{
-		return keyManager.checkStatus(keyCode, status);
+		return keyManager.checkStatusUnsafe(keyCode, status);
 	}
 
 	function get_ANY():Bool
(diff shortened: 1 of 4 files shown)


Geokureli:
I don't even know how to checkout this branch locally, because it has a `/` in the name

Geokureli:
FYI I'm gonna have to look into FlxActions and make sure this doesn't impact previous saved bindings or anything

Geokureli:
I'm actually wary of accepting this, after reading into it. I don't know if 254-255 are actual keys a dev might want to use, but this seems like an issue that should be fixed in hl, rather than adding a workaround for what might be a temporary hl bug

Geokureli:
FYI me job and zeta chatted about this in the discord. Rather than changing the value of keys we changed the switch to a series of ifs, with a warning comment.

Also: key checks like FlxG.keys.pressed.A now  bypass ANY/ALL checks

PR #2502, opened 2022-02-16
Changing two things in specific that might be good to add

Cheemsandfriends:
There are 2 things that I think it might be good to add.
1. A function to load the saved preferences of sounds when you change the save Path.
2. Adding an if condition in case that the save data is null

### DIFF
--- a/flixel/system/ui/FlxSoundTray.hx
+++ b/flixel/system/ui/FlxSoundTray.hx
@@ -117,9 +117,12 @@ class FlxSoundTray extends Sprite
 				active = false;
 
 				// Save sound preferences
-				FlxG.save.data.mute = FlxG.sound.muted;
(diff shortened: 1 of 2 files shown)


Cheemsandfriends:
Why does this give "errors" but compiles everything and checks as completed anyway?

Geokureli:
#2396 
something like this was proposed earlier, but, imo "switching saves" is weird in general, and should be minimized to fringe cases, instead of being an expected step of every new game. and I think there's a lot of work we could do to improve this, where this PR is just a bandaid.

Closing this now, I'll try to formalize the root problem, better, in the existing PR

Also you're getting errors because [FlxG.sound does not always exists, in all games](https://github.com/HaxeFlixel/flixel/blob/dev/flixel/FlxG.hx#L318-L323)

PR #2511, opened 2022-03-04
Apply scaling to checkPixelPerfectOverlap()

EyeDaleHim:
Noticed an issue with scaling and this class, thought I might try to add a fix to it.


Geokureli:
I'd like to try this change out, can you recommend a demo I should use, or a simple repro that showcases this fix? I haven't used FlxMouseEventManager very much

Geokureli:
this doesn't seem to be working as intended, also FlxMouseEventManager was changed and this branch doesn't use the latest changes, I can tell because the Sprite arg is capitalized.
https://imgur.com/rf3jp06

you bring up a valid issue that should be fixed, though

Geokureli:
replaced by #2576

PR #2518, opened 2022-03-11
Fixed `loadRotatedGraphic` to solve `Mod by 0`

SeiferTim:
Would happen if using a single-frame graphic image.

Error occurred *only* on Windows targets that I could see - not HashLink, or HTML5.

### DIFF
--- a/flixel/FlxSprite.hx
+++ b/flixel/FlxSprite.hx
@@ -438,7 +438,7 @@ class FlxSprite extends FlxObject
 			// we assume that source graphic has one row frame animation with equal width and height
 			var brushSize:Int = brush.height;
 			var framesNum:Int = Std.int(brush.width / brushSize);
-			Frame = (framesNum > Frame) ? Frame : (Frame % framesNum);
+			Frame = (framesNum > Frame || framesNum == 0) ? Frame : (Frame % framesNum);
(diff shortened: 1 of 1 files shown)


Geokureli:
sorry, forgot to circle back on this. Wouldn't `framesNum` only be 0 if the image is smaller than the frameWidth? Should we log an an error in this case? perhaps I'm not understanding, can you give me a repro to try?

SeiferTim:
Sorry, I don't see these notifications very often - they are usually buried in about 10k that have nothing to do with me...

I'm not 100% sure of what the actual problem is, I'm using it like this:

```haxe
sprite.loadRotatedGraphic('assets/images/Graphic.png', 180, 0, false, true, '${Graphic}-rotated');
```
...and it just dies on Windows targets, not on Hashlink/HTML5.
Changing the code to what I committed fixed it.

In my case, my `Graphic.png` is an image that is 10x24px and I want to use the whole image (frame `0`) and generate rotations...

Geokureli:
I wanna double-check that this isn't the symptom of some other issue, can you give me the dimensions of Graphic.png?

Geokureli:
Sorry this took forever, I just wanted to make sure I understood the problem

PR #2520, opened 2022-03-13
Adding the alpha support when tinting

Cheemsandfriends:
When you try to tint stuff, you basically need to forget about the alpha cos you can't tint it.
This PR basically adds the alpha support + changing the color value to Transparent, which is useless cos it's the default one.

### DIFF
--- a/flixel/FlxSprite.hx
+++ b/flixel/FlxSprite.hx
@@ -879,7 +879,7 @@ class FlxSprite extends FlxObject
 		if (useColorTransform)
 		{
 			colorTransform.setMultipliers(colorTransform.redMultiplier - color.alphaFloat, colorTransform.greenMultiplier - color.alphaFloat, colorTransform.blueMultiplier - color.alphaFloat, alpha);
-			colorTransform.setOffsets(Math.round(color.red * (color.alphaFloat - 0.01)), Math.round(color.green * (color.alphaFloat - 0.01)), Math.round(color.blue * (color.alphaFloat - 0.01)), 0);
+			colorTransform.setOffsets(Math.round(color.red * color.alphaFloat), Math.round(color.green * color.alphaFloat), Math.round(color.blue * color.alphaFloat), 0);
 		}
 		else
 		{
--- a/flixel/FlxSprite.hx
+++ b/flixel/FlxSprite.hx
@@ -158,7 +158,7 @@ class FlxSprite extends FlxObject
 	 * Tints the whole sprite to a color (`0xRRGGBB` format) - similar to OpenGL vertex colors. You can use
 	 * `0xAARRGGBB` colors, but the alpha value will simply be ignored. To change the opacity use `alpha`.
 	 */
-	public var color(default, set):FlxColor = 0xffffff;
+	public var color(default, set):FlxColor = 0x00000000;
 
 	public var colorTransform(default, null):ColorTransform;
 
@@ -866,7 +866,7 @@ class FlxSprite extends FlxObject
 		colorTransform.setMultipliers(redMultiplier, greenMultiplier, blueMultiplier, alphaMultiplier);
 		colorTransform.setOffsets(redOffset, greenOffset, blueOffset, alphaOffset);
 
-		useColorTransform = alpha != 1 || color != 0xffffff || colorTransform.hasRGBOffsets();
+		useColorTransform = alpha != 1 || color != 0x00000000 || colorTransform.hasRGBOffsets();
 		dirty = true;
 	}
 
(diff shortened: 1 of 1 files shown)


RichardBray:
I didn't know this was an issue. Do you have before and after screenshots so I can get a better understanding of what this does?

Cheemsandfriends:
> I didn't know this was an issue. Do you have before and after screenshots so I can get a better understanding of what this does?
Yeah, Sure @RichardBray!

image

Basically, This checks the alpha that it's put, (which, in this case is like 0.81 of color alpha) and tries to replicate what Flash uses for Color effects.

Cheemsandfriends:
Removing the white thing was because I thought it was kinda limiting to ppl who want to tint their sprite white, so I basically made the variable set transparent, cos it's the basic setter of the color transform (the floats, aka the multipliers are 1 and the ints, aka the offsets are 0)

Cheemsandfriends:
Removed it cos when I was using this for my project, I was trying to make it perfect with my calculator, and the most close thing that was near to what I wanted in that moment was `- 0.01` but now that I use Math.round to round the decimal to an int it isn't useful to use it anymore

RichardBray:
Cool. I'm not as much of an expert on Flixel as @Geokureli or @Gama11 so it would be good to get their thoughts on this too.

Geokureli:
can you be more explicit about what this fixes? can you give code examples alongside expectations and actual results?

As far as I know sprite.color avoids alpha by design, you change that by setting the sprite's alpha. it also looks like you change the way the color is applied, as well as the default color, which could mess with existing games. None of that is listed here and it seems separate to the title of this PR

Cheemsandfriends:
> can you be more explicit about what this fixes? can you give code examples alongside expectations and actual results?

This fixes that the alpha of the color is completely ignored and when you tint the sprite, it isn't 100% accurate to an actual tint, but it just sets the multipliers of the color.

> As far as I know sprite.color avoids alpha by design, you change that by setting the sprite's alpha. That way you can pass in values like 0xFF0000 or FlxColor.RED, so people may get confused when they update flixel and find their sprites invisible (I think)

The alpha color isn't set to the alpha, the alpha color is used just so the colors are more or less visible depending on the alpha of its own.

> it also looks like you change the way the color is applied, as well as the default color, which could mess with existing games

The way the color is applied is changed because of how you tint sprites in Flash basically, which is the reference I'm using for that. And about the change of the default color: the default color was basically white, which it made that you couldn't tint your sprite white, I just made the color variable to Transparent, which is what ColorTransform initialises and what you're setting.
Messing existing games could be a huge problem, but my solution would be making a compilation target to enable that or disable that to add the legacy one.

> None of that is listed here and it seems separate to the title of this PR

Wdym with that? and about the title, I thought it was a decent title for this PR tbh

Geokureli:
> The alpha color isn't set to the alpha, the alpha color is used just so the colors are more or less visible depending on the alpha of its own.

it sounds like you're treating the alpha value as the "tint" value in flashes Tint Color Effect, is that true?
Screen Shot 2022-03-22 at 1 07 35 PM

if so this is not the purpose of the sprite's color field, you should make a tinting tool that takes a color and tint strength and calls `sprite.setColorTransform` with the correct values, rather than hacking it into the existing color field using unused bit values that were meant for alpha

Geokureli:
would this work for you?
```hx
function setTint
	( redMultiplier:Float = 1.0, greenMultiplier:Float = 1.0, blueMultiplier:Float = 1.0
	, redOffset:Int = 0, greenOffset:Int = 0, blueOffset:Int = 0
	, tint:Float = 1.0
	):Void
{
	setColorTransform(redMultiplier - tint, greenMultiplier - tint, blueMultiplier - tint, 1.0,
		Math.round(redOffset * tint), Math.round(greenOffset * tint), Math.round(blueOffset * tint), 0);
}
```

Cheemsandfriends:
> if so this is not the purpose of the sprite's color field

Wait, it isn't? I thought it was trying to look into the tinting option but didn't know how to do it so made up a basic tinting option
If the color isn't planned like that, maybe add a new variable instead that does the same thing?

Cheemsandfriends:
> would this work for you?

Yeah, it's basically that lol

Geokureli:
Closing: just make a helper in your project.

The color tool is meant to just be a rgb multiplier over the sprite, im hesitant to add a new tint property since it might confuse people and setColorTransform handles just about everything you can ever want

Cheemsandfriends:
Seems fair, I just tought that color was trying to replicate the tint effect in Flash.
I was wrong, so I guess it doesn't need to change tbh

PR #2532, opened 2022-04-15
FlxCollision.pixelPerfectCheck check for preRotated graphics

SeiferTim:
`FlxCollision.pixelPerfectCheck` does not need to use 'advanced' detection when both graphics are using prerotated graphics and/or `angle = 0`

### DIFF
--- a/flixel/util/FlxCollision.hx
+++ b/flixel/util/FlxCollision.hx
@@ -44,10 +44,12 @@ class FlxCollision
 	 * @param	Camera			If the collision is taking place in a camera other than FlxG.camera (the default/current) then pass it here
 	 * @return	Whether the sprites collide
 	 */
+	@:access(flixel.animation.FlxAnimationController._prerotated)
 	public static function pixelPerfectCheck(Contact:FlxSprite, Target:FlxSprite, AlphaTolerance:Int = 1, ?Camera:FlxCamera):Bool
 	{
 		// if either of the angles are non-zero, consider the angles of the sprites in the pixel check
-		var advanced = (Contact.angle != 0) || (Target.angle != 0)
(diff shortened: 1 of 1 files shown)


Geokureli:
Love this! I want to try this on the [pixelPerfectCollision](https://haxeflixel.com/demos/PixelPerfectCollision/) demo and see if we get even better performance, as well as visually confirm that it works as intended

Geokureli:
I'm seeing false positives and false negatives in my demo, I notice that it's still rotating the graphic [here](https://github.com/HaxeFlixel/flixel/blob/00e7d31bf1c3f78156dedf93127ffb0ac38d71e8/flixel/util/FlxCollision.hx#L91) and [here](https://github.com/HaxeFlixel/flixel/blob/00e7d31bf1c3f78156dedf93127ffb0ac38d71e8/flixel/util/FlxCollision.hx#L108)

so i tried wrapping those lines in an if, like so:
```hx
if (Target.animation._prerotated == null)
	testMatrix.rotate(Target.angle * FlxAngle.TO_RAD);
```
It seems better, but not 100%, on html5. as you can see below all colliding sprites should be red, otherwise green. I suspect framePixels may still reference the unrotated graphic, rather than the current prerotated frame.
Screen Shot 2022-04-15 at 2 25 55 PM

on flash I get errors [here](https://github.com/HaxeFlixel/flixel/blob/00e7d31bf1c3f78156dedf93127ffb0ac38d71e8/flixel/util/FlxCollision.hx#L163). unexpected end of file. I expected flash to work better, as it's a blit target

SeiferTim:
> I suspect framePixels may still reference the unrotated graphic, rather than the current prerotated frame.

I think you're right! :( I will do some more research/testing.

PR #2536, opened 2022-04-18
Align FlxText between all targets

Metlmeta:
The desktop and mobile versions of FlxText do not render in the same visual location as web, instead rendering +1 x/y from the origin. See here for discussion:  https://github.com/HaxeFlixel/flixel/discussions/2528

Separately, the code for centering text on a button has always used (x - 1, y +3), even as far back as 2013. Changes to text (possibly through OpenFL) have improved centering and no longer need the (x - 1), which currently causes it to render off-center.

### DIFF
--- a/flixel/text/FlxText.hx
+++ b/flixel/text/FlxText.hx
@@ -871,6 +871,12 @@ class FlxText extends FlxSprite
 
 			return;
 		}
+		#elseif !web
+		// Fix to render desktop and mobile text in the same visual location as web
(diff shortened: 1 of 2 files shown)


Geokureli:
care to add some images verifying the improvements?

Metlmeta:
Sure thing!

Comparison of text rendering location:

**Old** vs **New**
old_align  new_align






Comparison of button label centering:

**Old** vs **New**
old_center  new_center

PR #2548, opened 2022-05-01
Shake effect on FlxSprite

SavanDev:
In my game I needed only one sprite to shake instead of the whole screen.
So, based on what is already in FlxCamera, I adapted it so that each sprite can be shaken separately.

https://user-images.githubusercontent.com/20217646/166168135-3e27fb90-326e-4318-aa38-d9499df17e40.mp4

https://user-images.githubusercontent.com/20217646/166168137-315fc793-d9dc-469a-a428-07942c0f6ac5.mp4

### DIFF
--- a/flixel/FlxSprite.hx
+++ b/flixel/FlxSprite.hx
@@ -1,5 +1,6 @@
 package flixel;
 
+import flixel.util.FlxAxes;
 import flash.display.BitmapData;
 import flash.display.BlendMode;
 import flash.geom.ColorTransform;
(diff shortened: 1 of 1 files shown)


Geokureli:
I like the idea, but it shouldn't be added directly to FlxSprite, FlxSprite is already too bloated. I would rather there be some external sprite shaker tool, it could even be a tween type

SavanDev:
I'll try to move it to the FlxTween and see how it looks.
Something like FlxTween.shake(sprite:FlxSprite,...) could be?
I will upload it in a new pull request when I finish it.

Geokureli:
closing this for #2549

PR #2480, opened 2022-01-07
FlxPathfinder - Overhaul to pathfinding

Geokureli:
I moved all of the pathfinding logic to a new class, so that it can be extended with different algorithms
Excerpt from FlxPathfinder docs:
>  Used to find paths in a FlxBaseTilemap. extend this class and override
> `getNeighbors` and `getDistance` to create you're own! For top-down maps, it may
> be wiser to extend FlxDiagonalPathfinder, as it already has a lot of basic helpers.
> 
> To use, either call `myPathfinder.findPath(myMap, start, end)` or
> `myMap.findPathCustom(myPathfinder, start, end)`

I plan to add a platformer pathfinder, too, but that doesn't need to be included here.

No breaking changes, just more ways to execute pathfinding. Im open to ideas for unit tests, I'll try to think of some.

My goal was to have `FlxPathfinder<Tilemap:FlxBaseTilemap<FlxObject>>` where FlxTilemap would accept `FlxPathfinder<FlxBaseTilemap<FlxObject>>`, `FlxPathfinder<FlxTilemap>` and `FlxPathfinder<FlxBaseTilemap<FlxTile>>` but not `FlxPathfinder<FlxTilemapExt>`, but I could not figure it out :(

PR #2549, opened 2022-05-02
Add `FlxTween.shake` for sprite shaking.

SavanDev:
Implement the same function that exists in `FlxCamera` to be used on an individual `FlxSprite` (and not on the whole screen).
Same as #2548 but now from FlxTween:
```haxe
FlxTween.shake(Sprite, Intensity, Duration, Axes, Options);
```
(I did it based on the tweens that are already made to work the same way)

Here two examples:

https://user-images.githubusercontent.com/20217646/166291048-42ae6f51-4823-4699-adad-52a1f218363d.mp4

https://user-images.githubusercontent.com/20217646/166291051-d01fa635-d48d-4c48-a704-6effac181f83.mp4

### DIFF
--- a/flixel/tweens/misc/ShakeTween.hx
+++ b/flixel/tweens/misc/ShakeTween.hx
@@ -24,9 +24,9 @@ class ShakeTween extends FlxTween
 	var sprite:FlxSprite;
 
 	/**
-	 * Defines the initial position of the sprite at the beginning of the shake effect.
+	 * Defines the initial offset of the sprite at the beginning of the shake effect.
 	 */
-	var initialXY:FlxPoint;
+	var initialOffset:FlxPoint;
 
 	/**
 	 * A simple shake effect for FlxSprite.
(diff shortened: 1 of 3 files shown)


TahirArch:
yes

flixel/tweens/misc/ShakeTween.hx
```diff
@@ -0,0 +1,73 @@
...
+		sprite = null;
+	}
+
+	override function update(elapsed:Float)
+	{
+		super.update(elapsed);
+		if (axes != Y)
+			sprite.x = initialXY.x + FlxG.random.float(-intensity * sprite.width, intensity * sprite.width);
+		if (axes != X)
+			sprite.y = initialXY.y + FlxG.random.float(-intensity * sprite.height, intensity * sprite.height);
```
Geokureli, line 66: rather than storing the initial x and y, I'm wondering if it should store the sprite's offset, and use the offset to shake it. this would allow it to shake objects in motion.

Also `initialXY` doesn't fit the name scheme, it should be initialPosition or (if changed) `initialOffset`

Geokureli:
If this tween is cancelled will it be reverted back to its original position, or the position it was shaken to? I'd prefer the former

Geokureli (changes requested):
Unless you have a good argument against it, this should use sprite.offset rather than the position directly. this will also ensure that the shake effect doesn't interfere with collision bounds and is a purely graphical effect.

WIth that, renaming initialXY, and a unit test that ensures offset is properly reset on complete and cancel, this should be good to go

SavanDev:
Now as soon as I have some time (probably at the weekend) I will start to make the changes in the code and the tests to check that it works well.
Sorry for not replying sooner.

SavanDev:
I have tried to run the unit test but I am not very familiar with it yet.
If you see that I'm missing something or it's done wrong let me know and I'll correct it (what I've tried passes the test).

Other than that, I've been testing it on a new project to see if it actually works correctly when finalizing and/or canceling the interpolation and I haven't seen any problems with either.
I uploaded it to GitHub so you can see it: https://github.com/SavanDev/ShakeTest

Geokureli (approved):

Geokureli:
For some reason I'm not seeing utnit test outcomes in github actions, so I checked you're locally and it looks good.

Thanks for this!

P.S.: I'm making [a change to FlxAxes](https://github.com/HaxeFlixel/flixel/pull/2480/commits/cd379a046944cc5bf2ae3705a634ff0c2aef3ba0) in another branch, so you'll see a similar change to ShakeTween, from me, pretty soon

hayfii:
this got added right? i cant use it

Geokureli:
Its not released yet, youll need to use `haxelib dev` or `haxelib git`

PR #2557, opened 2022-05-15
rename FlxPoint->FlxBasePoint, FlxVector->FlxPoint

Geokureli:
## The One True FlxPoint
Renamed FlxPoint to FlxBasePoint, which will be hidden from docs and completion, FlxVector was renamed to FlxPoint, a new typedef FlxVector was made as a simple alias to FlxPoint (with a deprecation warning) for backwards compatibility.

## Math Operators
 - `A += B` adds the value of `B` to `A`
 - `A -= B` subtracts the value of `B` from `A`
 - `A *= k` scales `A` by float `k` in both x and y components
 - `A + B` returns a new point that is sum of `A` and `B`
 - `A - B` returns a new point that is difference of `A` and `B`
 - `A * k` returns a new point that is `A` scaled with coefficient `k`
 
 ## Pooling
Operators grab points from the pool but don't put any (non-weak) points back, this could be a serious caveat. It makes me want to remove point-pooling altogether and inline the point constructor, however because set_x, set_y and set are not inlined there would be almost no benefit to this. these setters can't be inlined because they are overriden in `FlxCallbackPoint`, which is needed for the monstrosity known as `FlxSpriteGroup`

I'm assuming the impact to GC and pooling is small, but I added FLX_NO_POINT_POOL to allow side by side benchmark testing on existing games. In some cases, pool management is worse than letting GC take care of it


Geokureli:
> Isn't this gonna affect old repos that used FlxVector?

FlxVector still exists, but they'll get deprecation warnings

Cheemsandfriends:
Oh
Oh wait you're right, I haven't seen that commit lmao srry

Cheemsandfriends:
wait, what happens with the old FlxPoint?

Geokureli:
> wait, what happens with the old FlxPoint?

wdym? references to `flixel.math.FlxPoint` will still work

Cheemsandfriends:
> wdym? references to `flixel.math.FlxPoint` will still work

Yeah but it'll be using the new one right? What happens to the old one?

Geokureli:
can you give an example of existing code that would cause problems? How can code specifically target an "old" implementation of a class?

Cheemsandfriends:
Oh wait
Now that you mention it, FlxVector has the same functions as FlxPoint.
I thought they were completely different shit

ShaharMS:
might be a good time to also add some nice features to FlxBasePoint, like math functions with FlxBasePoints (additions, subtractions...)

PR #2560, opened 2022-05-21
add `FlxInputText` to the flixel.text package

ShaharMS:
This is a patched version of FlxInputText that doesnt crash on native when 2 spacebars are pressed/multilines on JS.

This can be pulled right now (works as intended), and i can polish it later.
Whats offerred now:

basic input
correct word wrapping
cross platform multiline

what isnt offered now:

multi-language input
RTL on non-JS platforms

### DIFF
--- a/flixel/text/FlxInputText.hx
+++ b/flixel/text/FlxInputText.hx
@@ -54,6 +54,23 @@ class FlxInputText extends FlxText
 
 	/**
 	 * A function called whenever the value changes from user input, or enter is pressed
+	 * 
+	 * It is structured like this:
+	 *  
+	 *  - The first argument is the `FlxInputText`s `text` property.
+	 *  - The second argument is the currently pressed key. It might be easier for you to check equality to the `FlxInputText..._ACTION` string.
+	 * 
+	 * Usage example:
+	 * 
+	 * ```haxe
+	 * inputText.callback = function(text:String, key:String) {
+	 * 		if (key == FlxInputText.ENTER_ACTION) {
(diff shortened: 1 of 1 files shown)


ShaharMS:
i still need to add documentation and remove some boilerplate

ShaharMS:
i just realized i haven't handled input from mobile anywhere in my code...

i did on a **heavily** modified version of FlxInputText, but that one wasn't as stable, so i decided to start development on an older, but stable mod. stuff i need to change/handle:

 - change input listener from openfl to lime (multilingual input)
 - focus on mobile
 - soft keyboard support
 - remove `_charBoundaries` as its not needed
 - add some more convenience functions and fields (`getCaretIndexUnderMouse()`, more listener fields, make key callback intuitive)

Geokureli:
This shouldn't go in vanilla flixel, this should replace [the current FlxInputText in flixel-ui](https://github.com/HaxeFlixel/flixel-ui/blob/master/flixel/addons/ui/FlxInputText.hx). Also, you should make a demo for this in [Flixel-Demos](https://github.com/HaxeFlixel/flixel-demos). Something that showcases all of the features, not only to help newcomers, but also so we can test them all on all targets with Github actions and manual testing

Geokureli (changes requested):
changes listed in previous comment

Geokureli:
Changes requested was what I set this PR status to. this should definitely go in Flixel UI and replace the existing one there. make a PR on that repo: https://github.com/HaxeFlixel/flixel-ui

PR #2564, opened 2022-05-26
Adding pitch support

Cheemsandfriends:
I know there's already a PR that's been merged in lime [right here](https://github.com/openfl/lime/pull/1510)
but that build doesn't seem to be public yet, sooo I decided to make this support as local as possible (also that it may work on more than one version while using lime idk)


flixel/system/FlxSound.hx
```diff
@@ -737,19 +732,55 @@ class FlxSound extends FlxBasic
...
+			@:privateAccess
+			if (_channel != null)
+				_channel.__source.pitch = v;
+			#else
+			FlxG.log.error("Pitch is supported only in sys systems while using openfl legacy!! switch the build into sys or upgrade your openfl!");
+			return 1;
+			#end
+		#else
+		#if flash
+		FlxG.log.error("Pitch is not supported on Flash!!!");
```
Geokureli, line 754: Rather than throwing a runtime error on flash we should hide the feature entirely and throw a compile time error when they try to use it

Geokureli, line 754: Rather than throwing a runtime error on flash we should hide the feature entirely and throw a compile time error when they try to use it

flixel/system/FlxSound.hx
```diff
@@ -94,12 +94,10 @@ class FlxSound extends FlxBasic
 	 */
 	public var volume(get, set):Float;
 
-	#if (sys && openfl_legacy)
```
Geokureli, line 97: We may to add a new FlxDefine for FLX_PITCH, defined as `#if ((sys && openfl_legacy) || (lime > 3.10.0 && !flash))` or whatever version of lime is released next

Geokureli, line 97: We may to add a new FlxDefine for FLX_PITCH, defined as `#if ((sys && openfl_legacy) || (lime > 3.10.0 && !flash))` or whatever version of lime is released next

flixel/system/FlxSound.hx
```diff
@@ -737,19 +732,55 @@ class FlxSound extends FlxBasic
...
+				backend.timer.stop();
+			}
+
+			var timeRemaining = Std.int((length - time) / v);
+			if (timeRemaining > 0)
+			{
+				backend.timer = new haxe.Timer(timeRemaining);
+				backend.timer.run = backend.timer_onRun;
+			}
+		}
```
Geokureli, line 774: I'm not sure what's being done here, can you add comments explaining?

Geokureli, line 774: I'm not sure what's being done here, can you add comments explaining?

Cheemsandfriends:
I'll work on this, it's just that it works when you play and change the pitch, but when you play it with the pitch already given, it doesn't for some reason?

Geokureli (changes requested):
failing on cpp

flixel/system/FlxSound.hx
Cheemsandfriends: Basically, in a song (at least how lime treats it on cpp) there's a timer that if it finishes, it calls the finished stuff.
if you changed the pitch without changing the timer of the song, you would have the pitch, but the song would "end" too late or too early depending of the pitch cos the time would be on pitch 1.
basically what it does is calling internal lime stuff so that it would be available not only on lime 3.1.0 (or idk what version is gonna be) but other versions too.

flixel/system/FlxSound.hx
Cheemsandfriends: That could be good, but I think that it should be able to change between targets without any compile errors?

flixel/system/FlxSound.hx
Cheemsandfriends: We could but all the code that's doing isn't trying to grab anything from the dev lime, just grabbing internal shit that the latest released lime version should work with

Cheemsandfriends:
I think the failing on cpp is a false alarm, cos I have everything ok (testing on desktop at least), and if I check on the cpp builds, there seem to have an error somewhere in `Run HaxeFlixel/setup-flixel@master` which honestly I can't find the error.

Cheemsandfriends:
I also addressed the concerns that you have about the code I made, when you can tell me if you have any questions about what I replied or adding some suggestion of my code

flixel/system/FlxSound.hx
Geokureli: what I mean is, we should use #if directives to hide the sound.pitch field entirely on targets that do not support it rather than exposing it when it doesn't work.

for instance, if someone is on flash `mySound.pitch` should give an error.

In this case I recommend creating the compiler directive FLX_SOUND_PITCH which behave similarly to `#if ((sys && openfl_legacy) || (lime > 3.10.0 && !flash))`, this way if we make features that use pitch, or if devs want to conditionally use the pitch feature they have an easy way to do so.

I'm not sure if openfl_legacy or lime 3.10 are still in use, so it may just be `#if !flash`, but in general it's good practice to use a specific compiler directive just in case this feature breaks on a certain target, or gets added to flash, or whatever

Geokureli:
> I think the failing on cpp is a false alarm, cos I have everything ok (testing on desktop at least), and if I check on the cpp builds, there seem to have an error somewhere in `Run HaxeFlixel/setup-flixel@master` which honestly I can't find the error.

I don't think it's a false alarm, I'm seeing the unit tests failing at:
```
Class: flixel.system.FlxSoundGroupTest .
Segmentation fault (core dumped)
```

flixel/system/FlxSound.hx
ninjamuffin99: I think this might be a bit better off being more integrated with OpenFL, and then flixel simply calls to set the new pitch, a bit similarly how it'd just simply set to change the pitch with the old pitch logic, although of course this time, should just need something like george suggests, `#if !flash` or something like that :)

Cheemsandfriends: > I think this might be a bit better off being more integrated with OpenFL, and then flixel simply calls to set the new pitch, a bit similarly how it'd just simply set to change the pitch with the old pitch logic, although of course this time, should just need something like george suggests, `#if !flash` or something like that :)

this would be less code and would make this 100% better, but the thing is, idk when lime is gonna release, and if I just add that up, this would a feature that would only work for a git version of lime, which normally is broken cos it's on dev.
if the next lime version is releasing, I would totally do that but if you do that rn, it wouldn't be a feature unless lime releases the next version and it could be days/weeks/months till it releases.
unless George doesn't care too much about it oc

Geokureli: keep in mind the next flixel release isn't happening soon (5.0.0) so might as well wait for lime, right?

Geokureli: looks like lime is planning to release, soon
https://community.openfl.org/t/lime-and-openfl-update/13539

MemeHoovy (approved):

Cheemsandfriends:
I hope you don't mind re-re-requesting another review about this @Geokureli lmao

flixel/system/FlxSound.hx
```diff
@@ -736,17 +735,24 @@ class FlxSound extends FlxBasic
...
 
 	function set_pitch(v:Float):Float
 	{
 		if (_channel != null)
+			#if openfl_legacy
 			_channel.pitch = v;
+			#else
+			@:privateAccess
+			if (_channel.__source != null)
+				_channel.__source.pitch = v;
```
Geokureli, line 753: getting errors here:
> .../FlxSound.hx:752: characters 17-25 : flash.media.SoundChannel has no field __source
.../FlxSound.hx:753: characters 14-22 : flash.media.SoundChannel has no field __source

There must be a problem with the FLX_PITCH definition

flixel/system/FlxSound.hx
Cheemsandfriends: I believe this is resolved in some way

flixel/system/FlxSound.hx
Cheemsandfriends: I this is resolved 100% uncertain

flixel/system/FlxSound.hx
Cheemsandfriends: resolved

flixel/system/macros/FlxDefines.hx
```diff
@@ -129,6 +131,9 @@ class FlxDefines
 		defineInversion(FLX_NO_FOCUS_LOST_SCREEN, FLX_FOCUS_LOST_SCREEN);
 		defineInversion(FLX_NO_DEBUG, FLX_DEBUG);
 		defineInversion(FLX_NO_POINT_POOL, FLX_POINT_POOL);
+		#if (openfl_legacy && sys || lime >= "8.0.0" && !flash)
+		defineInversion(FLX_NO_PITCH, FLX_PITCH);
+		#end
```
Geokureli, line 136: the #if shouldn't be here, `FLX_PITCH` and `FLX_NO_PITCH` will always be inversions, you need to define FLX_PITCH [here](https://github.com/HaxeFlixel/flixel/blob/dev/flixel/system/macros/FlxDefines.hx#L134):

Geokureli (changes requested):
still failing on flash

Geokureli:
also do you have a little tiny project that uses this feature you could share?

flixel/system/macros/FlxDefines.hx
Cheemsandfriends: why it shouldn't be there?
afaik, defineInversion is for stuff that has 2 sides, the yes and the no version, which, what [`defineInversions()`](https://github.com/HaxeFlixel/flixel/blob/0f3ae2daa75b962956db1ce3b3bac6a299a5755a/flixel/system/macros/FlxDefines.hx#L122) does afaik.
[defineHelperDefines()](https://github.com/HaxeFlixel/flixel/blob/0f3ae2daa75b962956db1ce3b3bac6a299a5755a/flixel/system/macros/FlxDefines.hx#L134) only adds defines depending if it can help the first thing
unless I'm overlooking things

Cheemsandfriends:
> also do you have a little tiny project that uses this feature you could share?

I have one little project that I used for testing back in the lime pr.
do you want me to publish that thing?

Geokureli:
> do you want me to publish that thing?
Either that, or send me a relevant snippet, or zip it on discord

Cheemsandfriends:
> Either that, or send me a relevant snippet, or zip it on discord

Sure! [here it is!](https://github.com/Cheemsandfriends/FlxPitch-sample)

flixel/system/macros/FlxDefines.hx
Geokureli: None of the other inversion are wrapped in #ifs, take a look at [what 
defineInversion() actually does](https://github.com/HaxeFlixel/flixel/blob/dev/flixel/system/macros/FlxDefines.hx#L173). I believe this is the reason flash is failing

Cheemsandfriends:
and I believe this should wrap it up @Geokureli?

Cheemsandfriends:
> something is odd about this, I feel I was wrong about the order here. FLX_NO_PITCH needs to be set to true on flash and openfl legacy before defineInversions is called.
> 
> Right now it's still failing on flash

I believe it was still failing on flash cos I forgot a semicolon when setting the FLX_NO_PITCH thing.
let's wait for the checker to check everything, but even then, I do believe that my setting of stuff wasn't bad but whatev lmao

Geokureli:
oh no, it was failing on flash because I was testing with your demo, which should fail on flash, Im dumb

Cheemsandfriends:
> oh no, it was failing on flash because I was testing with your demo, which should fail on flash, Im dumb

Oh LMAO

ninjamuffin99:
I think it would be nice / accessible to have a demo, or at the very least a snippet that nicely demos the new pitch support, just suggestion tho !!!

Cheemsandfriends:
> I think it would be nice / accessible to have a demo, or at the very least a snippet that nicely demos the new pitch support, just suggestion tho !!!

I don't mind if geokureli wants to use [this](https://github.com/Cheemsandfriends/FlxPitch-sample) as a pitch sample.
I doubt it but if he wants to I won't mind lmao

Geokureli:
thanks cheems, I'll probably use that as a starting point, if not, as is

PR #2565, opened 2022-05-27
Changing the way FlxStates are called

### LINKED ISSUE #2541: Take ()->FlxState in FlxG.switchState
In it's current form: `FlxG.switchState(new PlayState())` the state's constructor is called immediately, and the stateSwitch happens later, hence why we have `create()`. I'm proposing this should take a function that returns a newly created FlxState instance. This would allow devs to put all of their init logic in the constructor, where as of now this pattern would cause issues, since the constructor is called while the old state is still active.

This will also allow us to call `FlxG.resetState` with states that have constructor args

Similarly, the initialState passed in the FlxGame constructor should also be changed to a function.

## Easy migration guide
```hx
// old code                                  -->    new code
FlxG.switchState(new PlayState());        /* --> */ FlxG.switchState(PlayState.new);
FlxG.switchState(new PlayState(levelID)); /* --> */ FlxG.switchState(PlayState.new.bind(levelID));
```

~~This is literally gonna break every demo~~ Edit: IT DIDN'T!
This is still a breaking change, becuase it requires the removal of `FlxState.switchTo`

Recommended by @Gama11

Cheemsandfriends:
I was looking into the issues tab from flixel when I saw #2541 and I felt like it looked pretty damn cool so I thought, why not try to make a pr about it?

Fixes #2541

### DIFF
--- a/flixel/FlxG.hx
+++ b/flixel/FlxG.hx
@@ -367,10 +367,10 @@ class FlxG
 	 * Attempts to switch from the current game state to `nextState`.
 	 * The state switch is successful if `switchTo()` of the current `state` returns `true`.
 	 */
-	public static inline function switchState(nextState:FlxState):Void
+	public static inline function switchState(nextState:() -> FlxState):Void
 	{
-		if (state.switchTo(nextState))
-			game._requestedState = nextState;
+		if (state.switchTo(nextState()))
+			game._requestedState = nextState();
 	}
 
 	/**
@@ -379,7 +379,7 @@ class FlxG
 	 */
 	public static inline function resetState():Void
 	{
-		switchState(Type.createInstance(Type.getClass(state), []));
+		switchState(() -> Type.createInstance(Type.getClass(state), []));
 	}
 
 	/**
--- a/flixel/FlxG.hx
+++ b/flixel/FlxG.hx
@@ -369,7 +369,7 @@ class FlxG
 	public static inline function switchState(nextState:() -> FlxState):Void
 	{
 		if (state.switchTo(nextState()))
-			game._requestedState = nextState();
+			game._requestedState = nextState;
 	}
 
 	/**
@@ -378,7 +378,7 @@ class FlxG
 	 */
 	public static inline function resetState():Void
 	{
-		switchState(() -> Type.createInstance(Type.getClass(state), []));
+		switchState(game._requestedState);
 	}
 
 	/**
--- a/flixel/FlxState.hx
+++ b/flixel/FlxState.hx
@@ -80,6 +80,8 @@ class FlxState extends FlxGroup
 	@:noCompletion
 	var _subStateClosed:FlxTypedSignal<FlxSubState->Void>;
     
+	public function new() {super();}
+	
 	/**
 	 * This function is called after the game engine successfully switches states.
 	 * Override this function, NOT the constructor, to initialize or set up your game state.
--- a/flixel/system/FlxSplash.hx
+++ b/flixel/system/FlxSplash.hx
@@ -15,7 +15,7 @@ import flixel.util.FlxTimer;
 
 class FlxSplash extends FlxState
 {
-	public static var nextState:Class<FlxState>;
+	public static var nextState:()->FlxState;
 
 	/**
 	 * @since 4.8.0
@@ -199,7 +199,7 @@ class FlxSplash extends FlxState
 		#end
 		FlxG.stage.removeChild(_sprite);
 		FlxG.stage.removeChild(_text);
-		FlxG.switchState(() -> Type.createInstance(nextState, []));
+		FlxG.switchState(nextState);
 		FlxG.game._gameJustStarted = true;
 	}
 }
--- a/flixel/system/FlxSplash.hx
+++ b/flixel/system/FlxSplash.hx
@@ -199,7 +199,7 @@ class FlxSplash extends FlxState
 		#end
 		FlxG.stage.removeChild(_sprite);
 		FlxG.stage.removeChild(_text);
-		FlxG.switchState(Type.createInstance(nextState, []));
+		FlxG.switchState(() -> Type.createInstance(nextState, []));
 		FlxG.game._gameJustStarted = true;
 	}
 }
--- a/flixel/FlxGame.hx
+++ b/flixel/FlxGame.hx
@@ -92,7 +92,7 @@ class FlxGame extends Sprite
 	/**
 	 * Class type of the initial/first game state for the game, usually `MenuState` or something like that.
 	 */
-	var _initialState:Class<FlxState>;
+	var _initialState:()->FlxState;
 
 	/**
 	 * Current game state.
@@ -201,13 +201,18 @@ class FlxGame extends Sprite
 	/**
 	 * If a state change was requested, the new state object is stored here until we switch to it.
 	 */
-	var _requestedState:FlxState;
+	var _requestedState(default, set):()->FlxState;
 
+	/**
+	 * A flag for checking if you can change states or not.
+	 */
+	var _requestState:Bool = true;
+	
 	/**
 	 * A flag for keeping track of whether a game reset was requested or not.
 	 */
 	var _resetGame:Bool = false;
-
+	
 	#if FLX_RECORD
 	/**
 	 * Container for a game replay object.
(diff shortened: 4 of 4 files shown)


Geokureli:
this is a good start, but there's more to it than just that. The whole point is to call the state constructor after the old state is destroyed so that we don't need to create() method anymore

flixel/FlxG.hx
```diff
@@ -367,10 +367,10 @@ class FlxG
 	 * Attempts to switch from the current game state to `nextState`.
 	 * The state switch is successful if `switchTo()` of the current `state` returns `true`.
 	 */
-	public static inline function switchState(nextState:FlxState):Void
+	public static inline function switchState(nextState:() -> FlxState):Void
 	{
-		if (state.switchTo(nextState))
-			game._requestedState = nextState;
+		if (state.switchTo(nextState()))
```
Geokureli, line 372: state.switchTo() should take a `()->FlxState` rather than a FlxState instance, it should store the function in `_requestedState` and  wait until the new stuff is all set up to call it, creating the state

flixel/FlxG.hx
```diff
@@ -379,7 +379,7 @@ class FlxG
 	 */
 	public static inline function resetState():Void
 	{
-		switchState(Type.createInstance(Type.getClass(state), []));
+		switchState(() -> Type.createInstance(Type.getClass(state), []));
```
Geokureli, line 382: once we start storing the state creator function we can just call that to create a new one, rather than needing to use Type.createInstance.

with Type.createInstance FlxStates can't have constructor args, by passing in functions we can do things like
```hx
FlxG.switchState(()->{ return new MyState(arg1, arg2); });
```
and we'll know how to switch to that state AND reset it

flixel/FlxState.hx
```diff
@@ -80,6 +80,8 @@ class FlxState extends FlxGroup
 	@:noCompletion
 	var _subStateClosed:FlxTypedSignal<FlxSubState->Void>;
     
+	public function new() {super();}
```
Geokureli, line 83: shouldn't be needed, by default if a constructor isn't given to an extending class it just calls super automatically

flixel/system/FlxSplash.hx
```diff
@@ -199,7 +199,7 @@ class FlxSplash extends FlxState
 		#end
 		FlxG.stage.removeChild(_sprite);
 		FlxG.stage.removeChild(_text);
-		FlxG.switchState(Type.createInstance(nextState, []));
+		FlxG.switchState(() -> Type.createInstance(nextState, []));
```
Geokureli, line 202: nextState should also be a `()->FlxState`, and the FlxGame constructor should take a function instead of a Class Ref

Geokureli (changes requested):
This is going to be a pretty massive PR. I don't recommend making all of the changes directly in github on a web browser. it should be made locally, and we'll need to change Flixel-ui, flixel-addons, AAAAAALLLL of the demos, some snippets, the tutorial, and probably more, and we'll need to test them all together before anything is merged

flixel/FlxState.hx
Cheemsandfriends: Well, I had to add that cos the FlxGroup had a public function new with args, making every state, no matter if you don't have any, use bind

flixel/system/FlxSplash.hx
Cheemsandfriends: Hm So like making the PlayState.new also in FlxGame?
I liked that you didn't need to add anything, just the name but ok

flixel/FlxG.hx
Cheemsandfriends: Oh, so like a function? I like it, I like it

Cheemsandfriends:
@Geokureli wdym with "doing it locally?" you want me to make like a small fork introducing that or some shit like that?

Geokureli:
Yeah, typically I make changes to flixel using my forked repos. I set haxelib to use my local git versions via `haxelib dev flixel path/to/my/repo/`., 

I do this for addons, ui, demos, snippets, templates and tools. when you make a change to flixel, Github Actions test your new version of flixel against the dev branch of all of these additional repos, and in this case it's definitely going to fail, so we test them all locally and make branches for each repo that will merged at the same time, once all our local tests work well together.

NGL, this is a big task, and you'll likely have to learn a lot to get it done, but time is on your side, and I'm feeling pretty patient right now. especially because Austin East is gonna try to rewrite the flixel rendering system for flixel 5.0.0

flixel/system/FlxSplash.hx
Geokureli: where before you had
```hx
addChild(new FlxGame(400, 300, MenuState);
```
now you can just do
```hx
addChild(new FlxGame(400, 300, MenuState.new);
```

the reason for this is to allow constuctor args, like
```hx
addChild(new FlxGame(400, 300, MenuState.new.bind(arg);
```

Cheemsandfriends:
Hm...
so how can I let you know about the changes? I post them here or something like that?

Geokureli:
you can check out [Cheemsandfriends:patch-5](https://github.com/Cheemsandfriends/flixel/tree/patch-5) in VSCode via its github ui:
Screen Shot 2022-05-28 at 1 22 10 PM
just make sure your local repo points to your personal fork, you may have to pull once to load all the new remote branches

once that is set up changes you commit and push will be added to the patch-5 branch and will show up here, triggering a new round of github actions

you can also ditch this branch and PR and make your own locally, publich it and then you'll be able to start a new PR using that branch when you go to https://github.com/Cheemsandfriends/flixel

Cheemsandfriends:
I've been trying to add this for some days but idk why flixel-addons seems to work weird with my implementation?
even tho I changed the variables to `()->FlxState`

Cheemsandfriends:
I've got a question @Geokureli, what is this code supposed to do exactly?
image
I don't quite understand it

Geokureli:
> I've got a question @Geokureli, what is this code supposed to do exactly? [FlxGame.hx#L527](https://github.com/HaxeFlixel/flixel/blob/dev/flixel/FlxGame.hx#L527) I don't quite understand it

For full context, here's the full code:
```hx
if (FlxG.vcr.paused)
{
	if (FlxG.vcr.stepRequested)
	{
		FlxG.vcr.stepRequested = false;
	}
	else if (_state == _requestedState)
	{
		return;
	}
}
```
you can pause the game using the debug menu's VCR, if you do, `return` is called preventing the game from updating or drawing, however if `_state == _requestedState` is false, than `return` should be bypassed, allowing the game to update and draw normally. a different `requestedState` means that `FlxG.switchState` was called, and at the end of the game's update loop, the actual stateSwitch is done, [here](https://github.com/HaxeFlixel/flixel/blob/dev/flixel/FlxGame.hx#L728).

Geokureli:
hey @Cheemsandfriends! I'm planning on merging this for 6.0.0, I just want it to be the last thing merged before releasing 6.0.0 because it's such a breaking change, like, it breaks everything!

Also I never mentioned this but I actually made this change in [my own branch](https://github.com/Geokureli/flixel/tree/flxstate_new) in June, before you finalized this branch. I also made branches in flixel-addons and all the others, that will coincide. I'll use this branch, so you get the contribution stats, but I'll cross-check both our branches to see if either one of us missed something. but when I last looked it all seemed good

T1mL3arn:
Seen this too late to make a proposal, but anyway - why not use [Either](https://code.haxe.org/category/other/passing-different-types-to-a-function-parameter.html) ?  I believe with this pattern it wouldn't be a breaking change.

Geokureli:
@EliteMasterEric came up with a way to give a deprecation warning on the old type
```hx
function main()
{
	output(new FlxState("a")); // Warning : use ()->FlxState)
	output(FlxState.new.bind("b"));
}

function output(state:NextState)
{
	if (state is FlxState)
		trace("received state instance");
	else
		trace("received state maker");
}

abstract NextState(Dynamic) to FlxState to ()->FlxState
{
	@:from
	@:deprecated("use ()->FlxState)")
	static public function fromState(state:FlxState):NextState
	{
		return cast state;
	}
	
	@:from
	static public function fromMaker(func:()->FlxState):NextState
	{
		return cast func;
	}
}

class FlxState
{
	public var foo:String;
	
	public function new (foo:String) this.foo = foo;
}
```

Geokureli:
> it lets you reset with constructor arguments by using bind()

that's one of the big features, here. State constructor arguments are extremely powerful, but no one uses them

> Would this change essentially obsolete the create() function as well?

We would keep create for backwards compatibility, but the main reason for this change is to prevent issues from creating Flixel objects in the state constructor, (since, in current flixel, the next state's constructor is called before the previous state is destroyed) however we will likely always keep the create() function in all future versions because FlxG.state will be always be null in your state's constructor. so if you make objects in your state constructor that reference FlxG.state, or if global utils reference FlxG.state, you'll have a bad time

EliteMasterEric:
Ah, so `new FlxState()` is pre-switch and `create()` is post-switch? Good to know.

Once the OneOfTwo change is implemented (could we make a generic `DeprecatedInFavorOf<FlxState, Void->FlxState>` instead for later reuse?), what other obstacles are there to merging this change?

Geokureli:
looks like its doable, https://github.com/HaxeFlixel/flixel/pull/2733 assuming this passes CI. feel free to try out the branch, just know that this isn't a real pr

EliteMasterEric:
Checking back in, the new `@:op(a())` might be useful here to make the code more clean, there's definitely a way to avoid a type check here. `if (state is FlxState)` is expensive on several platforms IIRC.

Geokureli:
> Checking back in, the new `@:op(a())` might be useful here to make the code more clean, there's definitely a way to avoid a type check here. `if (state is FlxState)` is expensive on several platforms IIRC.

Is the "cost" of it actually perceivable in any of it's cases? When it comes to optimizing code its easy to say code A is more efficient than code B, but what that really means is "there will be noticeable slowdown if done 1,000 times a second". Calling `is` a couple times a frame should have no real perceivable impact when it's a couple times a frame, and it seems easier to read and maintain

EliteMasterEric:
I guess I'm just brainstorming use cases for newly available language features.

I do think that creating an abstract which convert a class instance to a supplier at compile time is better for code readability, and is easier to phase out later on.

Here's what my proposed code would look like for context:

```haxe
function main()
{
	output(new FlxState("a")); // Warning : use ()->FlxState)
	output(FlxState.new.bind("b"));
	output(()->new FlxState("c"));
}

function output(buildState:StateSupplier)
{
        var state:FlxState = buildState();
}

abstract StateSupplier(()->FlxState) to FlxState to 
{
        @:op(a())
        public function build():FlxState
        {
                return this();
        }

	@:from
	@:deprecated("use ()->FlxState)")
	static public function fromState(state:FlxState):StateSupplier
	{
		return () -> state;
	}
	
	@:from
	static public function fromMaker(func:()->FlxState):StateSupplier
	{
		return cast func;
	}
}

class FlxState
{
	public var foo:String;
	
	public function new (foo:String) this.foo = foo;
}
```

The major benefit here is that `output(StateSupplier)` doesn't even need to KNOW if the StateSupplier was created from an FlxState or from a function, and when you're ready to fully remove the deprecated functionality you can simply replace the method signature with `output(buildState:()->FlxState)` without modifying the method body.

Geokureli:
> I guess I'm just brainstorming use cases for newly available language features.

How new is this feature? I'm hesitant to raise HaxeFlixel's minimum and for this case the () op has very little benefit over having a create() method, especially since this is only used on the backend. The noteworthy differences of your approach are:

1. It converts the FlxState to a function immediately, rather than on create.
2. You convert the state instance to a function that returns that instance.

Your method is more simple but unfortunately doesn't work with `FlxG.resetState`, which switches to a new instance of the current state by calling the supplied state-creating function. In your case reset state would just switch to the current state instance. and we can't just switch to `Type.createInstance(Type.getClass(currentState), []);` is `resetState` because that ignores constructor args in cases such as: `FlxG.switchState(()->new PlayState(foo, bar))`. This is why [I have a](https://github.com/HaxeFlixel/flixel/pull/2733/files#diff-6ec53deec9419b3e6c475f82f26915a2b180f1d2155370f90eee0dbba5f29025R25-R44) `create()` and a `getConstructor`, each using  `isTypeOf()` each of which is only called once per state switch

Geokureli:
added https://github.com/HaxeFlixel/flixel/pull/2768, once it's merged I'll update FlxTransitionState and everything that uses `switchTo` for an outro to use `startOutro`. After that, I'll test https://github.com/HaxeFlixel/flixel/pull/2733 with `switchTo` removed and it should pass, but it will be a breaking change

Geokureli:
good news, it worked! Gonna close this PR in light of #2733 which is a more frictionless way of introducing this.

I appreciate everyone's help on this!

PR #2566, opened 2022-05-28
Save overhaul

### LINKED ISSUE #2539: FlxSave overhaul
> I think we could eliminate the most needs to rebind, or ever call bind in general by setting web's default binding to bind(appTitle, urlDomainName). It should also check bind("flixel", null) as well for backwards compatibility.
> 
> Right now FlxSave's intended use isn't very inutitive for web. I always call bind with custom args because of this and I assume it's the norm among web devs (at least the ones aware of this issue)

quote from #2396

## Breaking changes
Should be none, existing games that called `bind` with specific args will still work, and games that didn't should still check the old save path before copying over to the new default path

Geokureli:
Fixes #2539

related to #2202 #2502  #2396

## The Main Problem
Flixel creates FlxG.save to store internal things, but people tend to bind their own save, usually with a specified path, for a number of reasons:
- To avoid losing saves when things are moved, or when updating their game on web portals like Newgrounds
- To have a unique save id
- To cross-reference saves across applications
- The tutorials and demos told them to bind and close their own save rather than using FlxG.save.

This causes people to call `FlxG.save.bind("MyGameName", "MyName")` after the initial setup from the default save. meaning changes to sound and window preferences are saved to a new save rather than the one loaded at the start of the game. This has led to many people requesting a way to set the initial save path rather than creating a new one.

## Additional Problems
- FlxSoundTray, ConsoleHistory and Window crash when there is no valid FlxG.save.data available
- FlxSave has a lot of global vars used for a callback system that doesn't even need to exist, as no part of FlxSave is asynchronous (any more).

## Proposed changes
Rather than binding to "flixel" automatically, bind to the game's file name and author data, via Project.xml metadata.  This will prevent url paths from changing with updates and devs will have a simple, known id to cross-reference saves in other games.

For backwards compatibility on games that used the old save id, it will automatically search for the old save id and migrate the data into the new default save location, using the new `FlxG.save.migrateDataFrom("flixel")`.

Alternatively, devs can bind FlxG.save to their own desired name and path, FlxGame will no longer attempt to bind to the default save id if the save is already set up. Existing games that already have custom save bindings can do this for backwards compatibility.

To fix save related crashes from internal utils, they will all first check whether the save is bound, using the new `FlxG.save.isBound` getter.  All callback related args and instance vars have been removed.

## Breaking Changes
- Games expecting the old save id will need to be updated
- Calls to flush with the onComplete specified will now
- Use of FlxSave's old private callback-related vars will cause an error

## What now
- More thorough testing on all targets, might want to check out flixel for Switch
- Consider better names for `isBound` and `FlxSaveStatus.BOUND`
- Find a way to unit test setting a save before instantiating FlxGame

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@
 - `WatchFrontEnd`: added `FlxG.watch.addFunction` ([#2500](https://github.com/HaxeFlixel/flixel/pull/2500))
 - `FlxPoint`: added binary operators `+`, `-`, `+=`, `-=`, `*`, and `*=` ([#2557](https://github.com/HaxeFlixel/flixel/pull/2557))
 - `FlxColor`: added `rgb` getter and setter ([#2555](https://github.com/HaxeFlixel/flixel/pull/2555))
-- `FlxSave`: added `migrateDataFrom` ([#2566](https://github.com/HaxeFlixel/flixel/pull/2566))
+- `FlxSave`: added `migrateDataFrom`, `status` and `isBound` ([#2566](https://github.com/HaxeFlixel/flixel/pull/2566))
 
 #### Bugfixes:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 - `WatchFrontEnd`: added `FlxG.watch.addFunction` ([#2500](https://github.com/HaxeFlixel/flixel/pull/2500))
 - `FlxPoint`: added binary operators `+`, `-`, `+=`, `-=`, `*`, and `*=` ([#2557](https://github.com/HaxeFlixel/flixel/pull/2557))
 - `FlxColor`: added `rgb` getter and setter ([#2555](https://github.com/HaxeFlixel/flixel/pull/2555))
+- `FlxSave`: added `migrateDataFrom` ([#2566](https://github.com/HaxeFlixel/flixel/pull/2566))
 
 #### Bugfixes:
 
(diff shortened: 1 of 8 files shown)


Cheemsandfriends:
Wait, wouldn't the flixel save would be a one time migrating?

Geokureli:
yes it should, we still have to look for it everytime, because we don't know whether this is the first time

Cheemsandfriends:
> @Cheemsandfriends any concerns? I'm thinking about merging this today

Oops, sorry for not respondin, I had no Internet yesterday lmao

What I'm kinda concerned is that flixel will make a migrating once and it will disappear, cos in the function had a `erase();` iirc

Geokureli:
if the data is moved to the new save, why would we ever need the old save?

Cheemsandfriends:
> if the data is moved to the new save, why would we ever need the old save?

Because it's making a one time thing
Imagine that flixel save has multiple game saves and you open a new game without anything.
All those saves will be in that new game, completely useless to it and it loses the data of everything else
Im not saying the function is useless, but the flixel migrating thing should be changed?

Geokureli:
each game had it's own save location. when you call `bind("flixel")` it actually binds to `bind("flixel", "http://full/path/to/where/the/game/is/played.com")

Cheemsandfriends:
Oh
I was thinking that in the desktop target perspective, not the html one

Geokureli:
This is the case with all targets, desktop uses the full file path to the game as well to distinguish, we are almost never using the same save file for multiple games, if we did there would be a lot of problems

PR #2568, opened 2022-05-29
FlxGraphicsShader fixed for android and web platforms

MAJigsaw77:


Geokureli:
Can you give me some small snippet or demo so I can see the difference this PR makes? Maybe some kind of scenario that failed before and works now. I'm not super familiar with FlxGraphicsShader, so if you could at least talk about the problem you had and how this solves it, that would be helpful

jigsaw-4277821:
> Can you give me some small snippet or demo so I can see the difference this PR makes? Maybe some kind of scenario that failed before and works now. I'm not super familiar with FlxGraphicsShader, so if you could at least talk about the problem you had and how this solves it, that would be helpful

Sometimes the shaders are bad coded so this will replace some of the code then is there with the new one (you can see that in code)

Geokureli:
looks like flash is failing from this change. also can you be more specific? a code example highlighting the difference would really help

jigsaw-4277821:
> looks like flash is failing from this change. also can you be more specific? a code example highlighting the difference would really help

You can just to see the code, I can't explain myself that much

jigsaw-4277821:
And I really don't know why flash isn't working, really now

Geokureli:
Ive never had a problem with the old way, I can't merge it if i don't know what it fixes

McJonnycat:
So basically it can run uniforms with values inside the shader code without it crashing it for macos,linux,android,web. 
it also runs the bad shader code without replacing anything and it runs fine.

McJonnycat:
but the only problem is when ever im hiting a note o friday night funkin it lowers the fps every note and it lags

Geokureli:
Why was this closed? sorry I haven't had time to look into this recently, but I was planning to learn more about it. For flash could we just wrap it in some #if flash conditionals?

ItsyourboyJonnycat:
you can fix it by doing #if !flash

ItsyourboyJonnycat:
please close this i have found a fix and its in openfl and i made a pull request

PR #2571, opened 2022-05-31
FlxDrawQuadsItem Crash fix

jigsaw-4277821:
Sometimes the shader is null so it can cause a crash but with this it fixes it!


Geokureli:
the shader being null is a symptom of a separate issue. every time I've ever come across this, it's because a sprite was destroyed and is still set to be rendered by a camera. the leading cause of this is from creating FlxSprites in a FlxState's new() rather than in create().

I plan to make this error more understood, and I plan to [change when FlxState constructors are called](https://github.com/HaxeFlixel/flixel/issues/2541), which should prevent it from happening in a lot of cases. this change will not fix the underlying dev error that caused the problem it will only make the problem happen silently

TheRealJake12:
@MaJigsaw I got a patch of Flixel that fixes that issue.

PR #2575, opened 2022-06-03
Change AssetPaths args

### LINKED ISSUE #2547: AssetPaths changes
In my opinion, this util in its current state is more problems than it's worth, it seems to be a barrier for newcomers, and throws a lot of errors for stupid reasons. My default response to people asking for help with AssetPaths issues is "just delete it and use string paths", especially since VSCode's file tree has a simple "copy relative path" tool.

The simplest solution is just to remove it from the template and change tutorials to use string paths.

Of course, we could try fixing it. The main problem now is that it throws errors whenever two files have the same name, even when they are in different folders, and there's no way exclude certain subfolders as a way to alleviate this. We recently added a [custom renaming arg](https://github.com/HaxeFlixel/flixel/pull/2227) but the errors seem to be thrown before renaming.

Features starting with the highest priority:
 - downgrade duplicate AssetPaths fields to warnings, and assign the value of whichever was processed first
 - allow specific paths or subfolders to be excluded
 - check for duplicates among the final result of the renaming, rather than the filename
 - create a default renaming method that allows

The current rename func is passed the filename and extension, without the directory. I want to change this to take the entire file path. 

This would allow:
 - return null to exclude files based of subfolder or any other arbitrary culling method
 - directory tiebreakers for identically named files in different directories

Ideally, we could change the current rename func from:
```hx
(String)->String
```
to
```hx
(ReadOnlyArray<String>)->Array<Null<String>>
```
Meaning we give them a list of file paths and they return a list of valid haxe field names.

Geokureli:
fixes #2547

AssetPaths has always been a clunky system, whenever people come to me with issues regarding it, I tell them just to delete it and use String paths.

## Problems
- Duplicate filenames break down the entire system. even when they are in different directories, or if one of them is excluded via Project.xml
- Custom `rename` arg only takes the file name, and ignores the directory

## Solutions
- Allow duplicate field names, warn the user and omit the duplicated field
- Change filterExtensions to an include/exclude regular expression (or wildcard strings), allowing exclusion for any context.
- Change custom renaming to take the entire filepath, not just name and extension

## Excerpt from the 5.0.0 migration guide

AssetPaths (or any class built using `FlxAssets.buildFileReferences`) will no longer throw an error when two files have the same file name. Now it will give a warning and ignore whichever file is nested more deeply in the folder, or whichever file was found later.

The `filterExtensions` arg was removed and replaced with `include` and `exclude` args. These new args can either be an `EReg`, or a wildcard string, similar to openfl's project.xml args. For example to exclude everything in a folder called "test", as well as any .psd files, you would either use `"*/test/*|*.ase"` or `~/\/test\/|\.ase/`.

Lastly, the `rename` arg was changed, significantly. Custom renaming is already a relatively new and unknown feature of AssetPaths, but now it may be the most powerful! The `rename` arg is a function that will take a filepath (a relative filepath from the project.xml) and returns a field name used to access that path. The previous versions there would be no way to include two files with the same name, for instance, "assets/images/mario/walk.png" and "assets/images/luigi/walk.png" wouldn't be allowed, now you can call them `mario_walk.png` and `luigi_walk.png`. Additionally, you can return `null` to exclude a specific file.


Geokureli:
@player-03 
> Plus your exclude example should be `"/exclude/|\.ogg$"`.

"\.ogg$" is correct, I'll change
"/exclude" is actually preferred since it will skip the directory, where "/exclude/" will recursively call the directory and skip all files within it. I'm still thinking about how to explain that. ("/exclude$" would be best but I thought it would be confusing)

> The old version, though not very flexible, was easy for new programmers to understand. Jumping straight from there to regular expressions may scare them away.

IMO the old system was practically unusable, it was rare that this tool was actually worthwhile. There's still some stuff I want to try and do to improve this. but even with these changes my gut says to just remove AssetPaths from the templates/demos and reinforce the usage of string paths, which is even easier for new programmers to understand, imo. especially since VSC has "right click->copy relative path"

Geokureli:
that said, input is absolutely welcome as I'm still not 100% sure what the best solution is. So far this is my best guess

player-03:
I'd say the best solution is to allow both wildcard strings and regexes. To distinguish, just check if the user wrote `"*.gif"` or `~/\.gif$/`. Yes it's possible; see my pull request.

Geokureli:
Another thing I was wondering, since include/exclude will always have to match the Project.xml include/excludes, is it possible to auto-apply those filters and only check the files that will actually be exported?

Even if we could, I wonder if there are drawbacks to this.
Would it bog down compile time?
Is it worth forcing everyone to wrap `AssetPaths.some_file__ogg` refs in `#if desktop` blocks?

PR #2577, opened 2022-06-06
Specify a negative icon priority.

player-03:
This will make it easy for other libraries to override the Flixel's default icon, even with an inconvenient include order. Same with end users; any icons they set will take precedence, even if they set them before `<haxelib name="flixel" />`.

At the time of this writing, this is a development-only feature. Lime versions 7.9.0 and under will simply ignore the `priority` attribute.

Supersedes #2466.

### DIFF
--- a/include.xml
+++ b/include.xml
@@ -2,7 +2,7 @@
 <project xmlns="http://lime.software/project/1.0.2" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
 	xsi:schemaLocation="http://lime.software/project/1.0.2 http://lime.software/xsd/project-1.0.2.xsd">
 
-	<icon path="assets/images/logo/HaxeFlixel.svg" />
+	<icon path="assets/images/logo/HaxeFlixel.svg" priority="-10" />
 	
 	<section unless="unit-test">
 		<assets path="assets/sounds" rename="flixel/sounds" include="*.mp3" if="web || air" embed="true" />


Geokureli:
is there lime or openfl change that needs to be referenced here? You mention a needed change in 2466

player-03:
You're right, it takes several clicks to find the change. I thought it was accessible right from 2466, but nope.

This PR rely on openfl/lime#1509.

Geokureli:
we need to wait until lime releases that change to haxelib, before approving this PR

Geokureli:
"higher priority to override the HaxeFlixel icon" doesn't suggest "declare later" to me, it seems to suggest that some priority feature exists. or perhaps it's just worded badly? either way, we won't need it there, soon so I have no complaints. I just find the wording odd, there

player-03:
It was definitely poor wording. Eric's original problem back in December was that Lime always used the last-declared icon, with no way to override that. We looked for a priority feature and found none.

Geokureli:
in that case, let's just merge the change to demos, since they only serve to confuse. I still think we should wait to merge this until lime is released

Geokureli:
> Probably not. It's the same situation as the PR itself. There's no harm in doing it early, and odds are no one will even notice. At least, I can't imagine anyone's in a rush to move their <icon /> tags to the top of the file.

hmm I didn't realize the need was that specific, I never mess with icons. fuck it, lets just merge this

PR #2589, opened 2022-06-25
Fixing `FlxBitmapText.cutLines()`

SeiferTim:
In `FlxBitmapText` when `autoSize` is `false` it will use `wordWrap` by default.
If `wordWrap` is false, it will use `cutLines` to cut (truncate) lines that are too long.
However, if the string in the field is smaller than the maximum width of the field, `cutLines` ends up truncating the entire string, which causes a crash when the `textHeight` or `textWidth` is computed in order to display the `FlxBitmapText`
This should fix that issue.

### DIFF
--- a/flixel/text/FlxBitmapText.hx
+++ b/flixel/text/FlxBitmapText.hx
@@ -756,6 +756,11 @@ class FlxBitmapText extends FlxSprite
 					subLine = subLine.addChar(charCode);
 					subLineWidth += charWidth;
 				}
+				
+				if (subLineWidth > 0)
+				{
+					newLines.push(subLine.toString());
+				}
 
 				c++;
 			}


Geokureli:
I'm not familiar with this feature, can you give me a small situation that showcases this change?

SeiferTim:
This is roughly how I discovered the problem:

Create a `FlxBitmapText` and use `FlxBitmapFont.fromAngelCode` as the font.
Set your `FlxBitmapText` to have `autoSize` and `multiLine` and `wordWrap` be `false`.
Set it's `text` to a string.
Set the `fieldWidth` to be something longer than length of the string in pixels.
For example, if your letters are all roughly 16px wide, and you set your `text = "A";` set `fieldWidth = 100;` should work.

Try `updateHitbox()` on the `FlxBitmapText` or even `trace(bmptext.textHeight);` and it will crash because the `height` and `width` of the `FlxBitmapText` object will be `0`.

Geokureli:
I tried the following, using the [FlxBitmapText demo](https://github.com/HaxeFlixel/flixel-demos/tree/dev/UserInterface/FlxBitmapText)
```hx
var fontAngelCode = FlxBitmapFont.fromAngelCode("assets/NavTitle.png", "assets/NavTitle.fnt");
var field = new FlxBitmapText(fontAngelCode);
field.y = 100;
field.text = "Hello World!";
field.fieldWidth = FlxG.width;
field.autoSize = false;
field.multiLine = false;
field.wordWrap = false;
field.screenCenter(X);
add(field);
```

with the old code it crashes on `field.screenCenter(X);`  which uses the `width` getter which calls `cutLines`. With the new code it only displays the letter "H". That's a lot better than crashing, but this change may have only put a bandaid on the real issue. lets take a look at `cutLines` more throughly and make sure it looks correct in as many cases as possible

PR #2590, opened 2022-06-27
Fix `FlxBitmapText.cutLines()`

Geokureli:
an alternative to #2589

@SeiferTim This seems to work for all the cases I tried, does this work in your case?


SeiferTim:
This code still fixes my issue and I don't see any new issues.

PR #2600, opened 2022-07-15
Add completeFade to stopFX

slameron:
This allows you to call a new camera.fade() while one is running.


Geokureli:
I think we should just set `_fxFadeCompleted` to true, rather than calling the callback because we are not completing the fade we are cancelling it. None of the other effects call their callback, either.

another approach could be to remove the `_fxFadeCompleted` bool altogether, and have fade() behave exactly like shake(), where it would simply check `_fxFadeDuration`: 
https://github.com/HaxeFlixel/flixel/blob/dev/flixel/FlxCamera.hx#L1469

Geokureli:
Closing this, gonna go with #2652, instead which doesn't call the complete callback

PR #2620, opened 2022-08-10
Fixed the lag  issue

ItsyourboyJonnycat:
Don't merge it yet untill https://github.com/openfl/openfl/pull/2577 is merged

### DIFF
--- a/flixel/graphics/tile/FlxGraphicsShader.hx
+++ b/flixel/graphics/tile/FlxGraphicsShader.hx
@@ -4,7 +4,9 @@ package flixel.graphics.tile;
 import openfl.display.GraphicsShader;
 
 class FlxGraphicsShader extends GraphicsShader
-{
+{ 
+	
+	@:glVersion(120)
 	@:glVertexSource("
 		#pragma header
 		


ItsyourboyJonnycat:
the reason i say this is because when that is pull request from openfl  merged then you can merge this one too

Geokureli:
This doesn't seem like a "lag" issue. It sounds like performance will be improved by the openfl PR, regardless of this flixel PR. Intead, setting version 120 will prevent shaders from breaking entirely on certain targets.

That seems fine, but I just want to make sure we know what we're fixing, and whether this fixes that.

@MasterEric can you shed some light on this PR? Am I correct?

MasterEric:
I had a conversation with Jonny earlier. This PR somewhat confuses me since glVersion acts like the other metadata in that it is INHERITED.

He says he gets improved performance on Mac with this change but the glVersion should be 120 already (since the parent class sets it to 120).

Having the glVersion = 120 should do what you described in terms of compatibility though.

Geokureli:
Thanks Eric

@ItsyourboyJonnycat sounds like this PR doesn't do anything, if you can prove that this line does anything (maybe by looking at the resulting cpp build) then glVersion must not be inherited, like previously thought. if that is true, we can either reopen this PR or figure out why FlxGraphicShader isn't inheriting the gl version of the parent class

ItsyourboyJonnycat:
oh ok

PR #2621, opened 2022-08-15
Added custom loop point to FlxAnimation

MasterEric:
This simple change adds the ability to set a custom loop point on an FlxAnimation.

This allows you to conveniently perform behaviors such as having a "lead-in" to a looping animation.

For example, if you have a 5-frame animation, and you set `loopPoint` to 3, the frames you see will be:

`1-2-3-4-5~3-4-5~3-4-5~3-4-5~3-4-5`

### DIFF
--- a/flixel/animation/FlxAnimation.hx
+++ b/flixel/animation/FlxAnimation.hx
@@ -44,6 +44,12 @@ class FlxAnimation extends FlxBaseAnimation
 	public var looped(default, null):Bool = true;
 
 	/**
+	 * The custom loop point for this animation.
+	 * This allows you to skip the first few frames of an animation when looping.
+	 */
+	public var loopPoint:Int = 0;
+
+	/**
 	 * Whether or not this animation is being played backwards.
 	 */
 	public var reversed(default, null):Bool = false;
@@ -191,7 +197,7 @@ class FlxAnimation extends FlxBaseAnimation
 			_frameTimer -= delay;
 			if (reversed)
 			{
-				if (looped && curFrame == 0)
+				if (looped && curFrame == loopPoint)
 					curFrame = numFrames - 1;
 				else
 					curFrame--;
@@ -199,7 +205,7 @@ class FlxAnimation extends FlxBaseAnimation
 			else
 			{
 				if (looped && curFrame == numFrames - 1)
-					curFrame = 0;
+					curFrame = loopPoint;
 				else
 					curFrame++;
 			}


EyeDaleHim (approved):
Very convenient. Would love to see this used.

Geokureli:
I'm confused by the reverse behavior. It seems like a reverse animation similar to the one you describe would be:`5,4,3,5,4,3,5,4,3...` when I would expect `5,4,3,2,1,3,2,1,3,2,1...`

Am I correct? do others agree with my my expectation?

MemeHoovy (approved):
nice

Cheemsandfriends:
> I'm confused by the reverse behavior. It seems like a reverse animation similar to the one you describe would be:`5,4,3/5,4,3/5,4,3...` when I would expect `5,4,3,2,1/3,2,1/3,2,1...`
> 
> Am I correct? do others agree with my preference?

I think loopPoint should be where does the animation start, like if the animation is reversed, the loopPoint should be at the end of the animation or something that would be equivalent
for example, if you had an animation that would have the loopPoint of 0, and you reverse it would equal to the length of the animation and if the loopPoint is 1, then it's the length - 1.

Geokureli:
> > I'm confused by the reverse behavior. It seems like a reverse animation similar to the one you describe would be:`5,4,3/5,4,3/5,4,3...` when I would expect `5,4,3,2,1/3,2,1/3,2,1...`
> > Am I correct? do others agree with my preference?
> 
> for example, if you had an animation that would have the loopPoint of 0, and you reverse it would equal to the length of the animation and if the loopPoint is 1, then it's the length - 1.

That doesn't seem intuitive to me, since the animation's frames aren't decided that way. also if you add frames to an animation it would change the effective loopPoint. I think frame numbers should always refer to the absolute frame number

Geokureli:
> I'm confused by the reverse behavior. It seems like a reverse animation similar to the one you describe would be:`5,4,3/5,4,3/5,4,3...` when I would expect `5,4,3,2,1/3,2,1/3,2,1...`
> 
> Am I correct? do others agree with my preference?

I was mistaken, here.  For some reason I forgot `reverse` was an arg in `play()` and not `add()` in terms of playing my example of a looped anim in reverse, `5,4,3/5,4,3...` makes the most sense.  TBH, no solution is perfect here, since the reverse of `1,2,3,4,5/3,4,5/3,4,5/...` is an infinite number of `5,4,3` followed by a final `5,4,3,2,1` which isn't feasible.

Geokureli:
thanks!
[Updated the changelog](https://github.com/HaxeFlixel/flixel/commit/0f3ae2daa75b962956db1ce3b3bac6a299a5755a)

MasterEric:
Thank you for the merge!

To clarify, this is not an argument of `.add` you have to set the value on the FlxAnimation after adding it.

I did it that way because I noticed the `.add` function was starting to get bloated with arguments and didn't want to make it worse.

Geokureli:
yeah this is too niche of a feature to add to such an already large function

MasterEric:
> Maybe in 5.0.0 we can change loop, and flipX/Y to be enums. or add a addAdvanced() to prevent breaking games

Yeah, in the future an addAdvanced function could take a parameter object to allow for much finer tuning of configuration.

PR #2638, opened 2022-09-18
Add some missing key codes to FlxKey

Starmapo:
I found out that some key codes from my keyboard weren't defined in FlxKey, so I added them. This adds the following keys:
* SCROLL_LOCK
* NUMLOCK
* START (aka Windows key)
* MENU
* BREAK (aka Pause key)
* NUMPADSLASH

Tested on HTML5, HashLink, and Windows; might need more testing on Mac and Linux

### DIFF
--- a/flixel/input/keyboard/FlxKey.hx
+++ b/flixel/input/keyboard/FlxKey.hx
@@ -63,6 +63,8 @@ abstract FlxKey(Int) from Int to Int
 	var RBRACKET = 221;
 	var BACKSLASH = 220;
 	var CAPSLOCK = 20;
+	var SCROLL_LOCK = 145;
+	var NUMLOCK = 144;
 	var SEMICOLON = 186;
 	var QUOTE = 222;
 	var ENTER = 13;
@@ -79,7 +81,10 @@ abstract FlxKey(Int) from Int to Int
 	var LEFT = 37;
 	var RIGHT = 39;
 	var TAB = 9;
+	var START = 15;
+	var MENU = 302;
 	var PRINTSCREEN = 301;
+	var BREAK = 19;
 	var F1 = 112;
 	var F2 = 113;
 	var F3 = 114;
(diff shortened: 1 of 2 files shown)


Raltyro (approved):

flixel/input/keyboard/FlxKey.hx
```diff
@@ -63,6 +63,8 @@ abstract FlxKey(Int) from Int to Int
 	var RBRACKET = 221;
 	var BACKSLASH = 220;
 	var CAPSLOCK = 20;
+	var SCROLL_LOCK = 145;
```
Geokureli, line 66: I tried pressing scroll lock on my On-Screen Keyboard, and `FlxG.keys.pressed.SCROLL_LOCK` never returned true. have you tried this on your machine?

flixel/input/keyboard/FlxKey.hx
```diff
@@ -79,7 +81,10 @@ abstract FlxKey(Int) from Int to Int
 	var LEFT = 37;
 	var RIGHT = 39;
 	var TAB = 9;
+	var START = 15;
```
Geokureli, line 84: This should just be called WINDOWS, I can't find any other places referring to this as the  "start" key

Geokureli (changes requested):
1. Verify that `FlxG.keys.pressed.SCROLL_LOCK` will return true on keyboards that have a scroll lock key. If not, we should remove this to avoid confusion.

2. rename START to WINDOWS

flixel/input/keyboard/FlxKeyList.hx
```diff
@@ -338,11 +348,26 @@ class FlxKeyList extends FlxBaseKeyList
 	inline function get_TAB()
 		return check(FlxKey.TAB);
 
+	public var START(get, never):Bool;
+
+	inline function get_START()
+		return check(FlxKey.START);
```
Geokureli, line 354:
```suggestion
	public var WINDOWS(get, never):Bool;

	inline function get_WINDOWS()
		return check(FlxKey.WINDOWS);
```

flixel/input/keyboard/FlxKey.hx
Geokureli, line 84:
```suggestion
	var WINDOWS = 15;
```

Geokureli (approved):

PR #2643, opened 2022-10-01
Clip rect scale

### LINKED ISSUE #2301: Oddities when using clipRect + scale.
So ive noticed some strange things happening when using clip rects and scales, here is the test app (source code below) working:

flixel-clip-scale-3

And here is the same test when the `_scale` var is set to 3:

flixel-clip-scale-4

Whats even more strange is that its possible to "fix" the app by uncommenting the lines in the test app below, which, honestly, doesnt make any sense to me as you have to have _both_ of the lines, and i would have thought that one would have been enough...

Cheers,
Ian

- **Haxe version:** 4.2.0-rc.1+e34463538
- **Flixel version:**  git@2c451b5b019805b4ac22701494204102387262e7
- **OpenFL version:** git@36d184817973d80746340ef92b0dd7550a87af82
- **Lime version:** git@c743fe10c404fb73003e24100f319a472cb4203f
- **Affected targets:** all tested

```haxe
    private var _s:FlxSpriteGroup = null;
    
    private var _scale:Int = 3;
    
        var s:FlxSprite = new FlxSprite();
        
        s.loadGraphic(AssetPaths.haxeui_invert__png);
        s.origin.set(0, 0);
        s.scale.set(_scale, _scale);
        
        _s = new FlxSpriteGroup();
        _s.clipRect = new FlxRect(0, 0, 50, 50);
        _s.add(s);
        
        var g = new FlxSpriteGroup();
        var t = new FlxText();
        t.origin.set(0, 0);
        t.scale.set(_scale, _scale);
        t.text = "Foo!";
        t.color = 0xFFFF0000;
        g.add(t);
        _s.add(g);
        
        add(_s);
        
        super.create();
    }
    
	override public function update(elapsed:Float):Void {
		super.update(elapsed);
        
        // uncommenting out _both_ these lines "fixes" the issue, but it needs both, which doesnt make alot of sense to me
        //_s.origin.set(_s.frameWidth * 0.5, _s.frameHeight * 0.5);
        //_s.origin.set(0, 0);
        
        if (FlxG.keys.pressed.RIGHT) {
            _s.x += 1;
        } else if (FlxG.keys.pressed.LEFT) {
            _s.x -= 1;
        }
        
        if (FlxG.keys.pressed.DOWN) {
            _s.y += 1;
        } else if (FlxG.keys.pressed.UP) {
            _s.y -= 1;
        }
        
        if (FlxG.keys.pressed.D) {
            _s.clipRect.x += 1;
            _s.clipRect = _s.clipRect;
        } else if (FlxG.keys.pressed.A) {
            _s.clipRect.x -= 1;
            _s.clipRect = _s.clipRect;
        }
        
        if (FlxG.keys.pressed.S) {
            _s.clipRect.y += 1;
            _s.clipRect = _s.clipRect;
        } else if (FlxG.keys.pressed.W) {
            _s.clipRect.y -= 1;
            _s.clipRect = _s.clipRect;
        }
    }
}
```

**Observed behavior:**
When increasing the scale the clipping seems to behave strangely.

**Expected behavior:**
Clipping to be consistent regardless of scale.

Geokureli:
Closes #2301
Example of old/new behavior:

https://github.com/user-attachments/assets/bee6f157-6597-41f7-8d5a-087a6358a0ce

can toggle sprites to use the old behavior with `legacyClipRectMode`

### DIFF
--- a/flixel/FlxSprite.hx
+++ b/flixel/FlxSprite.hx
@@ -1063,18 +1063,43 @@ class FlxSprite extends FlxObject
 	 * Converts the point from world coordinates to this sprite's pixel coordinates where (0,0)
 	 * is the top left of the graphic.
 	 * 
-	 * @param   point   The world coordinates
-	 * @param   camera  The camera, used to compare screen coordinates
+	 * @param   point   The world coordinates.
+	 * @param   camera  The camera, used for `scrollFactor`. If `null`, `FlxG.camera` is used.
 	 * @param   result  Optional arg for the returning point
 	 */
-	function transformWorldToPixels(point:FlxPoint, ?camera:FlxCamera, result:FlxPoint):FlxPoint
+	public function transformWorldToPixels(worldPoint:FlxPoint, ?camera:FlxCamera, ?result:FlxPoint):FlxPoint
 	{
 		if (camera == null)
 			camera = FlxG.camera;
-
-		var screenPoint = FlxPoint.weak(point.x - camera.scroll.x, point.y - camera.scroll.y);
-		point.putWeak();
+		
+		var screenPoint = FlxPoint.weak(worldPoint.x - camera.scroll.x, worldPoint.y - camera.scroll.y);
+		worldPoint.putWeak();
 		return transformScreenToPixels(screenPoint, camera, result);
+	}
+	
+	/**
+	 * Converts the point from world coordinates to this sprite's pixel coordinates where (0,0)
+	 * is the top left of the graphic. Same as `worldToPixels` but never uses a camera,
+	 * therefore `scrollFactor` is ignored
+	 * 
+	 * @param   point   The world coordinates.
+	 * @param   result  Optional arg for the returning point
+	 */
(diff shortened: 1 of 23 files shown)


flixel/FlxSprite.hx
```diff
@@ -275,7 +275,14 @@ class FlxSprite extends FlxObject
...
 	 */
 	public var clipRect(default, set):FlxRect;
-
+	
+	/**
+	 * If true, clipRect will behave as it did in flixel 4 and earlier.
+	 * Where the clipRect will scale up proportionally with the sprite.
+	 * @since 5.0.0
+	 */
+	public var clipRectIgnoreScale(default, set):Bool = false;
```
Geokureli, line 284: TODO:
- keep legacy behavior, make this an opt-in feature
- come up with a better name maybe `scaleClipRect` or something

Geokureli:
Closing for https://github.com/HaxeFlixel/flixel/pull/3419

PR #2648, opened 2022-10-05
Min max

Geokureli:
better docs


Cheemsandfriends:
I've got a question, why are you doing a flixelType instead of Std.isOfType?
Is there something wrong with Std.is that should be avoided that I don't know?

Cheemsandfriends:
even tho it sounds like a pain in the ass to do that all the time, maybe do like a flixel util or something like that?

Geokureli:
right now there's only 5 files that use Std.isOfType, I see no reason make big changes for this any time soon

Issue #2403, opened 2021-09-11
Toggling FlxSpriteGroup visibility resets visibility of all sprites in it


arm32x:
- **Haxe version:** 4.2.3
- **Flixel version:** 4.9.0
- **OpenFL version:** 9.1.0
- **Lime version:** 7.9.0
- **Affected targets:** Tested on Linux and HTML5, probably affects others

________________________________

**Code snippet reproducing the issue**:

```haxe
package;

import flixel.FlxSprite;
import flixel.group.FlxSpriteGroup;
import flixel.FlxState;

class PlayState extends FlxState
{
	override public function create()
	{
		super.create();
		
		// Create an FlxSpriteGroup
		var group = new FlxSpriteGroup(50, 50);
		add(group);
		
		// Add sprites to the group
		var red = new FlxSprite(0, 0).makeGraphic(100, 100, 0xFFFF0000);
		group.add(red);
		var green = new FlxSprite(50, 50).makeGraphic(100, 100, 0xFF00FF00);
		group.add(green);
		
		// Make one invisible
		green.visible = false;
		
		// Toggle the visiblilty of the group
		group.visible = false;
		group.visible = true;
	}
}
```

________________________________

**Observed behavior:**
Both sprites are visible at the end of `create()`.

**Expected behavior:**
The green sprite stays invisible, even when the visibility of the `FlxSpriteGroup` is toggled off and on.

TheHENOOB:
I think the "bug" makes sense, when you are toggling the group visibility to false it means all objects of the group will not be visible without changing the visibility to false, what it was true it will be visible, what it was false it will remain invisible, **in my theory**

To fix this you could use the forEach method from the FlxGroup

```
group.forEach(function(object:FlxSprite) {
  object.visible = true;
}
```

arm32x:
I think you've mixed up the observed and expected behaviours in my report.

Currently, when you do `group.visible = true`, it acts like your `forEach` call and sets all sprites' `visibility` to `true`. I would expect it to work as you described, with the sprite remaining invisible if its `visibility` was `false` before changing the group's `visibility`.

TheHENOOB:
> I think you've mixed up the observed and expected behaviours in my report.
> 
> Currently, when you do `group.visible = true`, it acts like your `forEach` call and sets all sprites' `visibility` to `true`. I would expect it to work as you described, with the sprite remaining invisible if its `visibility` was `false` before changing the group's `visibility`.

Ish, my bad for that one

[...]

PR #2654, opened 2022-10-16
A system for visibility #2403

Wither362:
The function of this pull request is to do what #2403 wants to do: a system for not toggling the visibility of the entire FlxGroup.


MemeHoovy (approved):

Geokureli:
if you want different, specific behavior on FlxSpriteGroups then you should extend the class and override the set_visible function with your desired behavior.

PR #2655, opened 2022-10-16
Added `excludes` param to `FlxSoundGroup`

Wither362:
For simplifying code.

(Sorry, I don’t know how to make better descriptions)

### DIFF
--- a/flixel/system/FlxSoundGroup.hx
+++ b/flixel/system/FlxSoundGroup.hx
@@ -57,22 +57,28 @@ class FlxSoundGroup
 
 	/**
 	 * Call this function to pause all sounds in this group.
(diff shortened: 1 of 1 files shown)


MemeHoovy (approved):

Geokureli:
I don't get this, maybe put sounds in a different sound group if you want them to behave differently

Wither362:
> I don't get this, maybe put sounds in a different sound group if you want them to behave differently

yeah, youre right, but for only pausing some sounds, and other let then play, for simplified code...

Geokureli:
I don't want to add complexity to the engine just for the sake of slightly reducing a one-off game's own complexity, this isn't really a common thing, and it's contrary to what groups are designed to do

PR #2656, opened 2022-10-16
Fix Garbage Memory in FlxText

NeeEoo:
There has been a bug in FlxText where updating the text takes up more memory due to it having to regenerate the bitmap, this was actually due to a bad condition check against an Int and a Float.

This has been tested and has no noticeable visual changes.

### DIFF
--- a/flixel/text/FlxText.hx
+++ b/flixel/text/FlxText.hx
@@ -795,7 +795,7 @@ class FlxText extends FlxSprite
 			newHeight = oldHeight;
 		}
 
-		if (oldWidth != newWidth || oldHeight != newHeight)
+		if (oldWidth != Std.int(newWidth) || oldHeight != Std.int(newHeight))
 		{
 			// Need to generate a new buffer to store the text graphic
 			height = newHeight;
--- a/flixel/text/FlxText.hx
+++ b/flixel/text/FlxText.hx
@@ -785,9 +785,9 @@ class FlxText extends FlxSprite
 			oldHeight = graphic.height;
 		}
 
-		var newWidth:Float = textField.width;
+		var newWidth:Int = Math.ceil(textField.width);
 		// Account for gutter
-		var newHeight:Float = textField.textHeight + VERTICAL_GUTTER;
+		var newHeight:Int = Math.ceil(textField.textHeight) + VERTICAL_GUTTER;
 
 		// prevent text height from shrinking on flash if text == ""
 		if (textField.textHeight == 0)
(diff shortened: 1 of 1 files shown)


MemeHoovy (approved):

Geokureli:
@NeeEoo can you give me a small test case that shows the specific issue, that is fixed by this change?

@MemeHoovy why do you keep approving changes? You're not a maintainer. If a specific change is favorable or useful to you, then just say that

MemeHoovy:
> @NeeEoo can you give me a small test case that shows the specific issue, that is fixed by this change?
> 
> @MemeHoovy why do you keep approving changes? You're not a maintainer. If a specific change is favorable or useful to you, then just say that

okay? Idc if I'm not a maintainer or not, I'm still gonna approve changes

Geokureli:
> okay? Idc if I'm not a maintainer or not, I'm still gonna approve changes

why? it's just confusing to people making PRs. I'd much rather you say what you like about the changes or why you think they should be merged. otherwise I'm just going to ignore it

MemeHoovy:
then just ignore them lol

Geokureli:
> then just ignore them lol

I will, but I'm still asking why and telling you how to be more helpful

MemeHoovy:
> > then just ignore them lol
> 
> I will, but I'm still asking why and telling you how to be more helpful

okay

NeeEoo:
@Geokureli here is the bug it's fixing, It only happens on windows, since the font code differs between windows, mac and html5

https://user-images.githubusercontent.com/23155359/196196843-21971f73-de84-495d-b0b2-b1dd53d3aa07.mp4

And here if you want to try it
https://github.com/NeeEoo/FlxText-Fix

Hold A to turn off the fixed text updating
Hold B to turn off the old text updating

The numbers on each side at the bottom of the screen is the total of bitmaps it generated, the one in the center is the ram the current process takes, which includes the non garbage collected bitmaps

NeeEoo:
There also seems to be a memory leak when updating text, after like 30 minutes it took 650mb ram, i suspect this is in the harfbuzz internal code.

Raltyro:
> There also seems to be a memory leak when updating text, after like 30 minutes it took 650mb ram, i suspect this is in the harfbuzz internal code or in https://github.com/openfl/openfl/blob/develop/src/openfl/text/_internal/TextLayout.hx.

but why that seems to be the case?
that sounds like its hxcpp gc issues instead of the internal code

NeeEoo:
Some of the ram wasnt counted by the hxcpp memory counter, meaning it was c++ memory, though there also seems to be a small memory leak with the hxcpp memory too when updating text

Geokureli:
I tried on both my machines and I saw no difference in memory from the old way to the new way. that said, this change is pretty simple, but I feel like if we're going to change this we might as well make `newWidth` and `newHeight` Ints

Raltyro:
What if you just lock the BitmapData on updating the bitmap for text changes and then unlock them
wouldnt that be better?

Raltyro:
also maybe for this pr, make it so there a property for flxtext to destroy previous bitmapdatas instead of keeping them in the cache
just saying

Geokureli:
> What if you just lock the BitmapData on updating the bitmap for text changes and then unlock them after updating it wouldnt that be better?

Locking only really helps when you're performing multiple changes to a bitmap. locking the bitmap to draw a textfield, shouldn't impact performance.

> also maybe for this pr, make it so there a property for flxtext to destroy previous bitmapdatas instead of keeping them in the cache just saying

This seems like a separate issue to this pr. if we did want to improve bitmap caching it should be done in `FlxGraphic` or `BitmapFrontEnd` and there's already systems is place that handle clearing bitmap memory based on usage.

Geokureli:
I went ahead and changed those vars to ints and went with `Math.ceil`over `Std.int` to avoid cutting the edge off the textfield.

Thanks for bringing this issue to our attention!

PR #2665, opened 2022-11-10
callCenterOrigin bool in FlxSprite.updateHitbox()

CyndaquilDAC:
A project I've been working on has been having issues with updateHitbox causing animated sprites to be heavily displaced from their intended positions, and I've figured out that this is why. Simply adds a boolean variable in updateHitbox() that allows you to disable it calling centerOrigin() (Enabled by default for backwards compatibility). If there are any problems with this, let me know.

### DIFF
--- a/flixel/FlxSprite.hx
+++ b/flixel/FlxSprite.hx
@@ -711,14 +711,17 @@ class FlxSprite extends FlxObject
 
 	/**
 	 * Updates the sprite's hitbox (`width`, `height`, `offset`) according to the current `scale`.
-	 * Also calls `centerOrigin()`.
+	 * Also calls `centerOrigin()` when callCenterOrigin is true.
 	 */
-	public function updateHitbox():Void
+	public function updateHitbox(callCenterOrigin:Bool = true):Void
 	{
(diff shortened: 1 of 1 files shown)


Geokureli:
This feels odd to me, the point of updateHitbox is to make the hitbox match the graphic. If you wanted to do this with a non-centered origin, you would need to calculate a very different offset. For instance, if the origin is 0,0 the offset would need to also be 0,0, so I don't really understand why it's useful to ignore the origin but not the offset, here.

If it were me and I needed this functionality I would just make a helper util that does what you need. I don't see the use for this specific change. I do think updateHitbox has issues, and there are more useful option that could be made, but this isn't it, in my opinion

CyndaquilDAC:
Offset is automatically overwritten by our code so that isn't an issue, our issue is centerOrigin making it so all our intended positions are heavily off when updateHitbox is used (which is needed for certain things).

Geokureli:
I recommend using a custom helper method in your project instead of changing flixel. for instance if your game uses a FlxSprite extended base class you can just override it with new behavior, otherwise you can use static extension or some global utility class.

Unfortunately, I see very little desire for this change outside of your specific situation.

PR #2670, opened 2022-11-16
Added `toTitleCase` & `slugify` functions to `FlxStringUtil`

SeiferTim:
Added a couple of simple helper functions to `FlxStringUtil`:
`toTitleCase` will turn something like `"a tale of two cities, pt ii"` into: `"A Tale of Two Cities, Part II"`
and `slugify` which will turn a string into a 'slug': Example: `"A Tale of Two Cities, Part II"` becomes `"a_tale_of_two_cities__part_ii"`

### DIFF
--- a/flixel/util/FlxStringUtil.hx
+++ b/flixel/util/FlxStringUtil.hx
@@ -649,10 +649,10 @@ class FlxStringUtil
 	}
 	
 	/**
-	 * Returns a slugified string
+	 * Returns an Underscored, or "slugified" string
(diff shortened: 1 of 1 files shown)


Geokureli (changes requested):
instead of `slugify` can you call it `toUnderscoreCase`. This is the first time I've ever personally heard the term "slugify", I've heard of snake_case, but underscore case is pretty straightforward

Geokureli (approved):

Issue #2581, opened 2022-06-10
disbale FlxTilemap's useScaleHack by default


Geokureli:
it's not a good solution and in most cases with pixel art it makes most things worse.

lets add a static property `FlxTilemap.defaultTileFrameBuffer = 2;` and if it's > 0 graphics passed into tilemap will automatically have each tile frame's edges extruded by that amount

Geokureli:
fixed with https://github.com/HaxeFlixel/flixel/commit/c9eb3f9100ca5efd4adeb020669a9bee87d499f1
Accidentally forgot to use a branch

PR #2673, opened 2022-11-18
prevent frame padding before tilegraphic is loaded

Geokureli:
This fixes a bug introduced by #2581. Where frame padding was applied before the bitmap was loaded. this is a pretty hacky fix, but FlxGraphic.fromClass is almost never used so w/e


Geokureli:
not going to add this to changelog because this bug was introduced and fixed without a release in between

PR #2679, opened 2022-11-22
Fix borders on rotating camera

NeeEoo:
This fixes the black borders that appear when you rotate a FlxCamera, since it rotates the openfl Sprite which the camera renders to.

Reason it's a draft is because i haven't been able to test it on blitting render, and might need some testing since isOnScreen probably doesn't account for the camera angle.

### DIFF
--- a/flixel/FlxCamera.hx
+++ b/flixel/FlxCamera.hx
@@ -12,6 +12,7 @@ import flixel.graphics.FlxGraphic;
 import flixel.graphics.frames.FlxFrame;
 import flixel.graphics.tile.FlxDrawBaseItem;
 import flixel.graphics.tile.FlxDrawTrianglesItem;
+import flixel.math.FlxAngle;
 import flixel.math.FlxMath;
 import flixel.math.FlxMatrix;
 import flixel.math.FlxPoint;
@@ -52,7 +53,7 @@ class FlxCamera extends FlxBasic
 	/**
 	 * Used behind-the-scenes during the draw phase so that members use the same default
 	 * cameras as their parent.
-	 * 
+	 *
 	 * Prior to 4.9.0 it was useful to change this value, but that feature is deprecated.
 	 * Instead use the `defaultDrawTarget` argument in `FlxG.cameras.add `.
 	 * or`FlxG.cameras.setDefaultDrawTarget` .
@@ -60,11 +61,11 @@ class FlxCamera extends FlxBasic
 	 */
 	@:deprecated("`FlxCamera.defaultCameras` is deprecated, use `FlxG.cameras.setDefaultDrawTarget` instead")
 	public static var defaultCameras(get, set):Array<FlxCamera>;
-	
+
 	/**
 	 * Used behind-the-scenes during the draw phase so that members use the same default
 	 * cameras as their parent.
-	 * 
+	 *
 	 * This is the non-deprecated list that the public `defaultCameras` proxies. Allows flixel classes
 	 * to use it without warning.
 	 */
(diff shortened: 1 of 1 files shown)


Raltyro:
maybe make it so your still able to change the flashsprite rotation, maybe FlxCamera.rotation?

Raltyro:
and also make the sprites able to render inside the bound that has angle applied, cause it seems like any sprites outside the original bound won't render even if it's in the camera view

Geokureli:
@NeeEoo, Do you have a test case for this? I don't see how I'm supposed to change `_sinAngle` and `_cosAngle`

NeeEoo:
It just changes how the angle property looks like visually by making it not rotate the openfl sprite that the camera uses

Geokureli:
So I made this test case:
https://github.com/Geokureli/flixel-tests/blob/main/Source/states/CameraAngleTestState.hx

I think I misunderstood the purpose for this. I assumed the goal was to actually rotate the camera's view space, but after trying this out, it seems it was literally made to "prevent the black borders when rotating the camera", not to show FlxSprites in that region

If that's all you really wanted to do, all you need to do is:
```hx
final bgColor = 0xFF404040;
FlxG.cameras.bgColor = bgColor;
FlxG.stage.color = bgColor;
```

NeeEoo:
It was meant to show FlxSprites in that rotated space, but i haven't figured out the code in the isOnScreen to show the stuff outside of the view

Starmapo:
I'm using this in my fork of HaxeFlixel and I made these changes:

```
// Modified to take angle into account
function transformRect(rect:FlxRect):FlxRect
{
  if (FlxG.renderBlit)
  {
	  rect.offset(-viewMarginX, -viewMarginY);
  
	  if (_useBlitMatrix)
	  {
		  rect.x *= zoom;
		  rect.y *= zoom;
		  rect.width *= zoom;
		  rect.height *= zoom;
  
		  if (angle != 0)
		  {
			  var origin = FlxPoint.weak(FlxMath.lerp(viewMarginLeft, viewMarginRight, 0.5), FlxMath.lerp(viewMarginTop, viewMarginBottom, 0.5));
			  origin.x -= FlxMath.lerp(rect.left, rect.right, 0.5);
			  origin.y -= FlxMath.lerp(rect.top, rect.bottom, 0.5);
			  rect.getRotatedBounds(angle, origin, rect);
		  }
	  }
  }
  
  return rect;
}

// Modified to take angle into account
public function getViewMarginRect(?rect:FlxRect)
{
  if (rect == null)
	  rect = FlxRect.get();
  
  rect.set(viewMarginLeft, viewMarginTop, viewWidth, viewHeight);
  if (angle != 0)
  {
	  rect.getRotatedBounds(angle, FlxPoint.weak(FlxMath.lerp(rect.left, rect.right, 0.5), FlxMath.lerp(rect.top, rect.bottom, 0.5)), rect);
  }
  return rect;
}

public inline function containsPoint(point:FlxPoint, width:Float = 0, height:Float = 0):Bool
{
  var viewRect = getViewMarginRect();
  var contained = (point.x + width > viewRect.left) && (point.x < viewRect.right) && (point.y + height > viewRect.top) && (point.y < viewRect.bottom);
  point.putWeak();
  viewRect.put();
  return contained;
}

public inline function containsRect(rect:FlxRect):Bool
{
  var viewRect = getViewMarginRect();
  var contained = (rect.right > viewRect.left) && (rect.x < viewRect.right) && (rect.bottom > viewRect.top) && (rect.y < viewRect.bottom);
  rect.putWeak();
  viewRect.put();
  return contained;
}
```

and it works perectly in tile render mode. I haven't tested it in blit render mode however.
I don't know how to request these changes to the pull request, so I'm just putting them here.

gamerbross:
any updates on this??

NeeEoo:
> any updates on this??

Nah not really been kinda unmotivated as of recently. The only thing thats basically needed is fixing the isOnScreen

PR #2681, opened 2022-11-23
Thread safe pool

MasterEric:


Geokureli:
Honestly, thread safety is not a concern of flixel. And it feels like a can of worms to make flixel more thread safe, if the code is going to look this ugly. Flixel is meant to be accessible, something new devs could use and contribute to. And things like this just make the code more fragile and hard to maintain

I would just use `FLX_NO_POINT_POOL`

MasterEric:
I did not intend to PR this code to the main repository, it was a misclick.

It is worth mentioning that `FLX_NO_POINT_POOL` is not useful, as several classes such as FlxRect ignore this option.

Geokureli:
it's not that they ignore point-pooling directives, they just aren't points. we can add more directives or a global pool directive, but honestly lets remove rect pooling and inline the FlxRect constructor so pooling isn't even needed

PR #2683, opened 2022-11-24
Fix version mishap

ACrazyTown:
HaxeFlixel 5.0.1 reports as HaxeFlixel 5.0.0 in the debugger, this fixes it


Geokureli:
can't believe I forgot this. likely gonna have to make it 5.0.2 to fix this in release

PR #2685, opened 2022-11-25
Added a swap function to FlxArrayUtil

UncertainProd:
This pr adds a simple swap function to FlxArrayUtil, which should be useful for re-ordering sprites like this:

https://user-images.githubusercontent.com/83609901/204037478-fb613af5-a993-4e0d-b972-f37689a16bc3.mp4

The code:
```haxe
package;

import flixel.FlxG;
import flixel.FlxSprite;
import flixel.FlxState;
import flixel.text.FlxText;
import flixel.util.FlxArrayUtil;
import flixel.util.FlxColor;

{
	var card1:FlxSprite;
	var card2:FlxSprite;
	var helpText:FlxText;

	{
		super.create();
		card1 = new FlxSprite().makeGraphic(100, 100, FlxColor.RED);
		card1.screenCenter();
		card1.x -= 20;

		card2 = new FlxSprite().makeGraphic(100, 100, FlxColor.YELLOW);
		card2.screenCenter();
		card2.x += 20;
		card2.y += 20;

		helpText = new FlxText(0, FlxG.height - 20, 0, 'Press ENTER to toggle which card is at the front');
		helpText.setFormat(null, 12, FlxColor.BLACK);
		helpText.screenCenter(X);
		add(helpText);

		FlxG.camera.bgColor = FlxColor.WHITE;
		add(card1);
		add(card2);
	}

	{
		super.update(elapsed);
		if (FlxG.keys.justPressed.ENTER)
		{
			FlxArrayUtil.swap(members, members.indexOf(card1), members.indexOf(card2));
		}
	}
}
```

### DIFF
--- a/flixel/util/FlxArrayUtil.hx
+++ b/flixel/util/FlxArrayUtil.hx
@@ -83,6 +83,25 @@ class FlxArrayUtil
 	}
 
 	/**
+	 * Swaps two elements of an array, which are located at indices `index1` and `index2`
+	 * @param array The array whose elements need to be swapped
+	 * @param index1 The index of one of the elements to be swapped
+	 * @param index2 The index of the other element to be swapped
+	 * @return The array
+	 */
+	@:generic
+	public static inline function swap<T>(array:Array<T>, index1:Int, index2:Int):Array<T>
+	{
+		if(index1 >= 0 && index1 < array.length && index2 >= 0 && index2 < array.length)
+		{
+			var temp = array[index1];
+			array[index1] = array[index2];
+			array[index2] = temp;
+		}
+		return array;
+	}
+
+	/**
 	 * Clears an array structure, but leaves the object data untouched
 	 * Useful for cleaning up temporary references to data you want to preserve.
 	 * WARNING: Can lead to memory leaks.


Geokureli:
just in case you're unaware, `FlxArrayUtil` is meant to be used via static extension, or  by adding `using flixel.util.FlxArrayUtil` to your imports. by adding a using statement to your example, you could replace
```hx
FlxArrayUtil.swap(members, members.indexOf(card1), members.indexOf(card2));
```
with a simpler line of:
```hx
members.swap(members.indexOf(card1), members.indexOf(card2));
```

flixel/util/FlxArrayUtil.hx
```diff
@@ -82,6 +82,25 @@ class FlxArrayUtil
...
 	}
 
+	/**
+	 * Swaps two elements of an array, which are located at indices `index1` and `index2`
+	 * @param array The array whose elements need to be swapped
+	 * @param index1 The index of one of the elements to be swapped
+	 * @param index2 The index of the other element to be swapped
+	 * @return The array
+	 */
+	@:generic
```
Geokureli, line 92: No need to use `@:generic`

flixel/util/FlxArrayUtil.hx
```diff
@@ -82,6 +82,25 @@ class FlxArrayUtil
...
+	public static inline function swap<T>(array:Array<T>, index1:Int, index2:Int):Array<T>
+	{
+		if(index1 >= 0 && index1 < array.length && index2 >= 0 && index2 < array.length)
```
Geokureli, line 95: This shouldn't check whether the indices provided are valid, if people provide invalid indices it should just throw whatever error it normally would, we could add `safeSwap` that *does* check. this way we aren't adding needless overhead unless desired by the dev.

Geokureli (changes requested):
1. I think your example illustrates that there should also be a way to swap elements of an array based on references to those elements rather than their indices. I would call that `swap`, and this should be `swapIndices`. example:
```hx
public static function swap<T>(array:Array<T>, item1:T, item2:T):Array<T>
{
	return swapIndices(array, array.indexOf(item1), array.indexOf(item2));
}

public static function swapIndices<T>(array:Array<T>, index1:Int, index2:Int):Array<T>
{
	// ...
}
```

2. All new features in this class should have unit tests, in [FlxArrayUtilTest.hx](https://github.com/HaxeFlixel/flixel/blob/dev/tests/unit/src/flixel/util/FlxArrayUtilTest.hx). Normally I add these, but this should be pretty easy, I'm the busiest I'll ever be, and you seem to know what your doing. 

Check out the unit test [readme.md](https://github.com/HaxeFlixel/flixel/blob/dev/tests/unit/README.md). if you need help testing, let me know.

note that you can do things like:
```hx
Assert.isTrue([0, 1, 2, 3, 4, 5].swap(2, 4).equals([0, 1, 5, 3, 4, 2]));
```

UncertainProd:
> I think your example illustrates that there should also be a way to swap elements of an array based on references to those elements rather than their indices. I would call that swap, and this should be `swapIndices` or `swapByIndex`

`swapByIndex` sounds good, I've added functions for swapping based on index and based on the element itself, and added a safe version of them as well which will validate the input elements/indices

> All new features in this class should have unit tests, in [FlxArrayUtilTest.hx](https://github.com/HaxeFlixel/flixel/blob/dev/tests/unit/src/flixel/util/FlxArrayUtilTest.hx).

I've added in some tests as well. Using `haxe test-cpp.hxml` on my machine works for me but for some reason `test-web.hxml` shows an error saying 'ERROR: Local results server appeared to hang so test reporting was cancelled.', maybe it's because I don't have flash installed?

Geokureli (approved):

Geokureli:
> What do you mean adding to your imports? Do you mean adding "using bla.bla.Bla" to the playstate class??

yeah I mean adding a using statement at the top of the file (where the imports go), like so:
https://github.com/HaxeFlixel/flixel/blob/dev/tests/unit/src/flixel/util/FlxArrayUtilTest.hx#L5

allowing you to use these methods as static extensions rather than static methods, like so:
https://github.com/HaxeFlixel/flixel/blob/dev/tests/unit/src/flixel/util/FlxArrayUtilTest.hx#L88

PR #2693, opened 2022-12-05
Fix chaining for screenCenter()

TheWorldMachinima:
This PR fixes chaining for `screenCenter` function in `FlxObject`.
Currently you cannot really chain FlxSprites or whatever with `screenCenter` since it returns `FlxObject`, not the extended class.
Example:
```haxe  
var sprite:FlxSprite = new FlxSprite().screenCenter(); //is not allowed
var sprite:FlxSprite = cast new FlxSprite().screenCenter(); //is allowed
```

### DIFF
--- a/flixel/FlxObject.hx
+++ b/flixel/FlxObject.hx
@@ -1211,7 +1211,7 @@ class FlxObject extends FlxBasic
 	 * @param   axes   On what axes to center the object (e.g. `X`, `Y`, `XY`) - default is both. 
 	 * @return  This FlxObject for chaining
 	 */
-	public inline function screenCenter(axes:FlxAxes = XY):FlxObject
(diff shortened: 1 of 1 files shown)


Geokureli:
this isn't a good solution, I recommend either using the cast, avoiding chains or create a helper that centers and returns FlxSprite.

I'd rather wait for haxe to have [polymorphic This types](https://community.haxe.org/t/feature-proposal-polymorphic-this-types/174).

Geokureli:
also note,  you can override function to return an extension of the original class:
```hx
class Test {
    static function main() {
        var instance:B = new B ().foo();
    }
}

class A {
    public function new() {}
    public function foo():A return this;
}

class B extends A {
    override function foo():B return this;
}
```

PR #2695, opened 2022-12-10
Possible fix for drawing tiled sprites with FlxStrip

### LINKED ISSUE #2694: Changes to FlxStrip break FlxTiledSprite
With the addition of changes made by @UncertainProd in #2671, FlxTiledSprite no longer renders properly. The tiled sprite is now stretched out instead of repeated over the sprite's width.

I considered posting this issue to `flixel-addons`, but no changes were made to FlxTiledSprite; rather, the code change to FlxStrip has caused an issue which potentially affects anything using FlxStrip.

- **Haxe version:** 4.2.5
- **Flixel version:** git commit [e1fbdc4](https://github.com/HaxeFlixel/flixel/commit/e1fbdc40df8c9bfc9a73b0bdfbca4cc96da5485a)
- **OpenFL version:** 9.2.0
- **Lime version:** 8.0.0
- **Affected targets:** Windows and Linux native were both tested

https://github.com/MasterEric/FlxTiledSprite-TestBench

**Observed behavior:**

incorrect

**Expected behavior:**

correct

UncertainProd:
This adds a field called 'repeat' into `FlxDrawTrianglesItem`, which will allow for tiling the bitmap used when rendering using drawTriangles. This should fix #2694 as whenever `flxstripobject.repeat` is set to true, it will use `beginBitmapFill` to perform rendering, as opposed to `beginShaderFill`. However this will also disable shaders if the repeat field is set to true and only allow shaders when it is set to false.

### DIFF
--- a/flixel/FlxCamera.hx
+++ b/flixel/FlxCamera.hx
@@ -45,9 +45,10 @@ private typedef FlxDrawItem = #if FLX_DRAW_QUADS flixel.graphics.tile.FlxDrawQua
 class FlxCamera extends FlxBasic
 {
 	/**
-	 * Any `FlxCamera` with a zoom of 0 (the default value) will have this zoom value.
+	 * While you can alter the zoom of each camera after the fact,
+	 * this variable determines what value the camera will start at when created.
 	 */
-	public static var defaultZoom:Float = 1.0;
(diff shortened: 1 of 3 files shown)


UncertainProd:
Moved the fix over to #2696

PR #2699, opened 2022-12-14
Added new compile defines: FLX_NO_ERROR_SOUND and FLX_NO_ERROR_CONSOLE

MasterEric:
Added two new compile defines:

* `FLX_NO_ERROR_SOUND` disables any sounds which play when a warning or error is logged.
* `FLX_NO_ERROR_CONSOLE` disables the behavior where the console automatically opens when a warning or error is logged.

These are useful in the even that a developer wants to compile a debug build, but does not want to be annoyed by regular beeping of the console. This allows them to temporarily disable the beeps by adding compile defines to their project file.

### DIFF
--- a/flixel/system/debug/log/LogStyle.hx
+++ b/flixel/system/debug/log/LogStyle.hx
@@ -60,8 +60,8 @@ class LogStyle
 		bold = Bold;
 		italic = Italic;
 		underlined = Underlined;
-		errorSound = ErrorSound;
-		openConsole = OpenConsole;
+		errorSound = #if FLX_NO_ERROR_SOUND null #else ErrorSound #end;
+		openConsole = #if FLX_NO_ERROR_CONSOLE false #else OpenConsole #end;
 		callbackFunction = CallbackFunction;
 	}
 }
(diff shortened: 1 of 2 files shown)


MemeHoovy:
honestly, I appreciate these changes since they can be annoying at times, especially when a million errors occur, and tons of beeps play

Geokureli:
Not sure I want these to be compile flags, this seems like an esoteric solution that no one would ever find.
I'm pretty sure you can always change the error sound, and perhaps we should replace the current default to something less annoying. I would really like to avoid a stupidly long list of frivolous FlxDefines. especially in this case when it's not really needed at compile time. Seems like a runtime setting would not only suffice but would allow you to toggle it multiple times in-game

Also the idea that people hear this sound every time they play their game and want to mute it rather than fixing the issue seems crazy to me

Geokureli:
@MasterEric @MemeHoovy 
looks like there already is a runtime feature for this: https://github.com/HaxeFlixel/flixel/blob/dev/flixel/system/debug/log/LogStyle.hx

to use, just say
```hx
LogStyle.ERROR.openConsole = false;
LogStyle.ERROR.errorSound = null;
```

to set a new errorsound you can do:
```hx
LogStyle.ERROR.errorSound = "assets/sounds/errorBeep";
```
however, errorSound assumed .ogg, so I made this PR to allow for mp3 sounds: https://github.com/HaxeFlixel/flixel/pull/2702

MemeHoovy:
good to know 👍🏻

PR #2644, opened 2022-10-02
Added diffX and diffY properties to FlxMouse

MasterEric:
This function makes it easy to compare how the mouse has moved since the last `update()`. This makes the internal _prevX and _prevY values, previously only used by the `justMoved` property, more useful to the user. 

Behaving somewhat similarly to `wheel`, these properties provide the following:

- `diffX`: The distance the mouse has moved horizontally since the previous update.
- `diffY`: The distance the mouse has moved vertically since the previous update.
- `diffScreenX`: The distance the mouse has moved in screen space horizontally since the previous update.
- `diffScreenY`: The distance the mouse has moved in screen space vertically since the last update.

PR #2709, opened 2022-12-30
Added deltaX and deltaY properties to FlxMouse.

MasterEric:
This function makes it easy to compare how the mouse has moved since the last update(). This makes use of the internal `_prevX` and `_prevY` values, previously private and inaccessible to the user.

Behaving somewhat similarly to `wheel`, these properties provide the following:

- `diffX`: The distance the mouse has moved horizontally since the previous update.
- `diffY`: The distance the mouse has moved vertically since the previous update.
- `diffScreenX`: The distance the mouse has moved in screen space horizontally since the previous update.
- `diffScreenY`: The distance the mouse has moved in screen space vertically since the last update.

This pull request is a rewrite of #2644, which was abandoned due to a faulty Git rebase.

### DIFF
--- a/flixel/input/mouse/FlxMouse.hx
+++ b/flixel/input/mouse/FlxMouse.hx
@@ -70,6 +70,26 @@ class FlxMouse extends FlxPointer implements IFlxInputManager
 	public var justMoved(get, never):Bool;
 
 	/**
+	 * Distance in pixels the mouse has moved since the last frame in the X direction.
+	 */
+	public var deltaX(get, never):Int;
+
+	/**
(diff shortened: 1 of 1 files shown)


Geokureli:
x, y, screenX and screenY are defined in FlxPointer which, FlxMouse extends, might as well add these properties to FlxPointer instead of FlxMouse

Geokureli:
on second thought, I see that _prevX was already defined in FlxMouse rather than FlxPointer, so those would need to be moved to the super class.

this makes me wonder if adding 4 new floats to every pointer, by default is too much overhead, so for now let's just keep it in FlxMouse and revisit this after I finish a tool I'm working on that may help optimize this

PR #2712, opened 2023-01-03
Convert FlxTextBorderStyle to an abstract enum (string)

gamerbross:


Geokureli:
Why?

gamerbross:
> Why?

Sometimes it's more convenient to use strings instead of enums, like in hscripts. Also, the text alignment enum works like a string.

mayo78:
FYI if hscript has the type library shared you can get any enum like this:
```haxe
var OUTLINE = Type.resolveEnum('flixel.text.FlxTextBorderStyle').OUTLINE;
```

PR #2713, opened 2023-01-14
Fix findMinY() and findMaxY() misconception

EyeDaleHim:
Both findMinY() and findMaxY() return X when both variables declare it'll return a Y value when the sprite's length is 0?


Geokureli:
Good find, thanks!

Geokureli:
@EyeDaleHim this pull reuest was made to master, in the future all PRs should be made to the dev branch, please

PR #2719, opened 2023-01-23
Add reverse to FlxArrayUtil and FlxStringUtil

TheWorldMachinima:
Simple enough. Both functions reverses the provided element.

Example:
```haxe
var array:Array<Int> = [2, 3, 5, 7, 9];
trace(flixel.util.FlxArrayUtil.reverse(array)); //[9,7,5,3,2]

var string = "12345";
trace(flixel.util.FlxStringUtil.reverse(string)); //54321
```

### DIFF
--- a/flixel/util/FlxArrayUtil.hx
+++ b/flixel/util/FlxArrayUtil.hx
@@ -106,7 +106,7 @@ class FlxArrayUtil
 	 */
 	public static inline function safeSwapByIndex<T>(array:Array<T>, index1:Int, index2:Int):Array<T>
 	{
-		if(index1 >= 0 && index1 < array.length && index2 >= 0 && index2 < array.length)
+		if (index1 >= 0 && index1 < array.length && index2 >= 0 && index2 < array.length)
 		{
 			swapByIndex(array, index1, index2);
 		}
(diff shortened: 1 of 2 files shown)


Geokureli:
Reminder that FlxArrayUtil and FlxStringUtil was designed to be used via static extension through `using` statements , not as static methods, like this:
```hx
using flixel.util.FlxArrayUtil;

class TestState extends flixel.FlxState
{
    {
        trace([1, 2, 3, 4, 5, 6].swap(2, 5)) // outputs: [1, 5, 3, 4, 2, 6];
    }
}
```

The `using` statement adds FlxArrayUtil.swap as a method to Arrays. I bring this up because [Arrays already have a reverse method](https://api.haxe.org/Array.html#reverse).

TheWorldMachinima:
> Reminder that `FlxArrayUtil` and `FlxStringUtil` was designed to be used via static extension through `using` statements, not as static methods, for example:
> 
> ```hx
> using flixel.util.FlxArrayUtil;
> 
> class TestState extends flixel.FlxState
> {
>     override public function create()
>     {
>         trace([1, 2, 3, 4, 5, 6].swap(2, 5)) // outputs: [1, 5, 3, 4, 2, 6];
>     }
> }
> ```
> 
> The `using` statement adds `FlxArrayUtil.swap` as a method to Arrays. I bring this up because [Arrays already have a reverse method](https://api.haxe.org/Array.html#reverse).
> 
> Futhermore, if we're going to create a string reverse method, which seems like such a rare need, we're only really making a helper for `s.split("").reverse().join("");`, which seems frivolous since it doesn't save a lot of typing, and there's [much much faster ways to accomplish this](https://stackoverflow.com/questions/13317329/reversing-a-string-in-haxe)
> 
> Edit: I double-checked these benchmarks on hashlink, and they're 96% faster https://try.haxe.org/#Ce700749

I didn't know arrays had reverse method. This PR became pretty much useless now. Closing this PR.

PR #2723, opened 2023-01-31
Add functions for keys that are inputed at the same time.

Cheemsandfriends:
It adds a function whether all the keys that you want to press are pressed are the same time, or released.

so instead of having to make a conditional like `if (FlxG.keys.justPressed.CONTROL && FlxG.keys.justPressed.A)` to set an example, you would only need to do `if (FlxG.keys.allJustPressed([CONTROL, A]))`, making it easier to type if you have several keys you want to input at the same time.

### DIFF
--- a/flixel/input/FlxKeyManager.hx
+++ b/flixel/input/FlxKeyManager.hx
@@ -293,6 +293,64 @@ class FlxKeyManager<Key:Int, KeyList:FlxBaseKeyList> implements IFlxInputManager
 
 		return false;
 	}
+	
+	/**
+	 * Check if every key of an array if keys has been just pressed.
+	 * @param KeyArray an Array of keys.
+	 */
+	public function allJustPressed(KeyArray:Array<Key>)
+	{
+		return checkConsecutiveKeyArrayState(KeyArray, JUST_PRESSED);
+	}
+	/**
+	 * Check if every key of an array if keys are being pressed.
+	 * @param KeyArray an Array of keys.
+	 */
+	public function allPressed(KeyArray:Array<Key>)
+	{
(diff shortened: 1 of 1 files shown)


TheRealJake12 (approved):
very nice. Makes things alot easier.

Geokureli:
I'm confused by this. When you save a file you don't press Ctrl and S keys at the same time, you hold ctrl down and then press the S key. so the usual code would be something like:
```hx
if (FlxG.keys.pressed.CONTROL && FlxG.keys.justPressed.S)
```

I don't think I would ever make a feature in a game or app that required the user to ""justPress" 2 keys at the same time, I would detect both keypresses and see if they both happened within some margin of time

Geokureli:
@Cheemsandfriends did this feature come about because you needed it for for some project, or did you think it just might be useful? I'm having trouble imagining a reasonable use-case for this.

Cheemsandfriends:
Oh sorry, I forgot completely to answer
I was playing to Manhunt some days ago before that day and I remembered that there was this move where you had to press 2 buttons at the same time to grab the enemy to make an attack combo, and because I thought that if you had several keys that you needed to be pressed at the same time with the same press state, then this helper function would be ideal for that case

Geokureli:
My guess is that game you're talking about didn't check whether you pressed 2 buttons in the same frame, but rather gave a few frames (or some fraction of a second) of leeway to try and press them both, which would require an entirely different system, FlxActions might already have this ability.

Let's just wait until someone with a project that actually needs allJustPressed or allJustReleased functionality comes forward before adding, because I really feel the need for this is purely theoretical

PR #2725, opened 2023-02-06
prevent crash when hiding debug windows

Geokureli:
While it's still not recommended to call FlxG.save.bind, this prevents a crash when you hide a debug window while using a custom save path on FlxG.save


chosencharacters:
<3

PR #2728, opened 2023-02-12
Fix `firstPressed` functions in FlxGamepad

Starmapo:
* `firstPressedRawID` now returns the first button that's pressed instead of the first button that's released
* `firstPressedID`, `firstJustPressedID` and `firstJustReleasedID` now properly return `NONE` on XInput and Logitech controllers (before, if no buttons were detected, it would return `GUIDE` as opposed to `NONE`)


Geokureli:
thanks!

Issue #2667, opened 2022-11-12
add sprite.graphicWidth and graphicHeight getters, stop calling updateHitbox on setGraphicSize()


Geokureli:
no reason to change the hitbox in setGraphicSize other than to have a handy getter for the new graphic width. maybe add new helpers that scale the graphic and hitbox, too
https://github.com/HaxeFlixel/flixel-docs/issues/258

PR #2730, opened 2023-02-14
Small implementations for HaxeFlixel

Just-Feeshy:
The minor fix is to make the width for FlxText change the field width, too. When you decrease the regular width, the text won't be offset to the right.

### DIFF
--- a/flixel/math/FlxMatrix.hx
+++ b/flixel/math/FlxMatrix.hx
@@ -18,17 +18,19 @@ class FlxMatrix extends Matrix
 	 */
 	public inline function rotateWithTrig(cos:Float, sin:Float):FlxMatrix
 	{
-		var a1:Float = a * cos - b * sin;
+		var temp1:Float;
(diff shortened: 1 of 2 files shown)


Geokureli:
Currently haxeflixel has a problem where there is a confusing lack of distinction between a sprite's "graphic" size and it's "collision" size, and we're aiming to make them more distinct: https://github.com/HaxeFlixel/flixel/issues/2667

I fear this is a step in the wrong direction, a text's collision width has different uses than it's graphical width. Also, it might be helpful for a majority of use-cases but I don't think we should make it more difficult to have a different fieldWidth and collision width.

also there's a lot of proposed changes that may make this change unnecessary very soon.

Just-Feeshy:
Thank you, I appreciate you bringing up the potential issues with the proposed change, and I will carefully consider the impacts before making further comments. Thank you for your feedback and understanding.

PR #2734, opened 2023-02-18
add FlxTypedTilemap

Geokureli:
testing some stuff

Github actions are failing on cpp with weird issues, likely gonna move this off of 5.4.0 unless someone knows what's going on

Here's the errors
[error.txt](https://github.com/HaxeFlixel/flixel/files/11424334/error.txt)


Geokureli:
the errors are the result of making FlxBaseTilemap an abstract class, it has nothing to do with giving FlxTilemap a type param. gonna keep this on 5.4.0, but move the abstract change to another release, likely when haxe fixes the cpp error

PR #2739, opened 2023-02-22
check exists in findMin/Max helpers

Geokureli:
WIP, need to prevent values of INFINITY when a group only has non-existent or null members


Geokureli:
gonna revisit this, need to add a `checkExistsInBounds` flag for backwards compat

PR #2741, opened 2023-02-24
Implement RGB/RGBA detection to `FlxColor.fromString`

gamerbross:
Convert strings like `rgb(184, 76, 217)` to a FlxColor using fromString. (case-insensitive & ignores white spaces)

Examples:
- `rgb(184, 76, 12)`
- ` rgb  ( 123, 21,0 )`
- `rgb ( 32 , 56 , 63 , 127 )` 
- `rgba(32,56,63,0)` 

```haxe
~/^rgb\s*\((\s*\d{1,3}\s*,?){3,4}\)/i
```

### DIFF
--- a/flixel/util/FlxColor.hx
+++ b/flixel/util/FlxColor.hx
@@ -392,6 +392,17 @@ abstract FlxColor(Int) from Int from UInt to Int to UInt
 	}
 
 	/**
+	 * Return a String representation of the color in the format rgba(R, G, B, A)
+	 *
+	 * @param Alpha Whether to include the alpha value in the RGB string
+	 * @return	A string in the format rgba(R, G, B, A)
+	 */
+	public inline function toRGBString(Alpha:Bool = false):String
+	{
+		return (Alpha ? "rgba" : "rgb") + '($red, $green, $blue' + (Alpha ? ', $alpha' : "") + ")";
+	}
+
+	/**
 	 * Return a String representation of the color in the format #RRGGBB
 	 *
 	 * @return	A string of length 7 in the format #RRGGBB
--- a/flixel/util/FlxColor.hx
+++ b/flixel/util/FlxColor.hx
@@ -381,7 +381,7 @@ abstract FlxColor(Int) from Int from UInt to Int to UInt
 	/**
 	 * Return a String representation of the color in the format 0xAARRGGBB
 	 *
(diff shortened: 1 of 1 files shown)


Geokureli:
why use this instead of `FlxColor.fromRGB(184, 76, 12)`?

gamerbross:
> why use this instead of `FlxColor.fromRGB(184, 76, 12)`?

To automate things.
For example, when we want a color from a JSON file, and we use the `fromString` function.

Geokureli:
What tool is creating color strings like this? it seems like a very weird format when tools typically convert 4 ints to a hex int trivially via  `(a << 24 | r << 16 | g << 8 | b)`

I feel like color strings formats specific to your game should be converted in your game's code rather than being automatically done in flixel. otherwise we're adding bloat to a common flixel util for an uncommon use case

Geokureli:
apparently even github recognizes this format and automatically puts color dots in code, so perhaps its more common than I thought
image

Although, I think this should have it's own method rather than using toString(), lets call it `fromRGBString()`, and add `toRGBString()` as well

Geokureli (changes requested):
create new `fromRGBString` and `toRGBString` methods

Geokureli:
also worth noting that `rgba(32,56,63,0)` doesn't work, despite being listed as one of your examples
https://try.haxe.org/#8FBec88e

gamerbross:
> create new `fromRGBString` and `toRGBString` methods (or suggest alternative names)
> 
> also you should add unit tests here: https://github.com/HaxeFlixel/flixel/blob/dev/tests/unit/src/flixel/util/FlxColorTest.hx to test various inputs/outputs

Is the function `fromRGBString` really necessary?
The point of the `fromString` function is to be able to parse a color in any format.

I'm also working on more changes and being able to parse CMYK, HSB/HSV and HSL color strings and convert a FlxColor to a string with those formats

Geokureli:
> Is the function `fromRGBString` really necessary?

Yes

> I'm also working on more changes and being able to parse CMYK, HSB/HSV and HSL

This is exactly why we should have specific "fromRGBString" converters, every new method added add more overhead for every call even though 99.9% of people know exactly what conversion method they need

Furthermore, I still don't know of any game tools that spit out string values like this, do you know of any?

PR #2742, opened 2023-02-25
Cache FlxGradient bitmaps

Starmapo:
This makes bitmaps outputted from `createGradientBitmapData` get added to the cache, so it can be reused instead of creating a new bitmap object when a gradient is needed with the same parameters.

### DIFF
--- a/flixel/util/FlxGradient.hx
+++ b/flixel/util/FlxGradient.hx
@@ -111,6 +111,10 @@ class FlxGradient
 		{
 			height = 1;
 		}
+		
+		var cacheName = 'gradient:$width:$height:$colors:$chunkSize:$rotation:$interpolate';
+		if (FlxG.bitmap.checkCache(cacheName))
+			return FlxG.bitmap.get(cacheName).bitmap;
 
 		var gradient:GradientMatrix = createGradientMatrix(width, height, colors, chunkSize, rotation);
 		var shape = new Shape();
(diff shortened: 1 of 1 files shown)


Starmapo:
If the user keeps `destroyOnNoUse` to true and `persist` to false, like they are by default, then the graphic will be destroyed once it's not used, so I don't believe there will be a memory issue. Besides this is different from FlxUI9SliceSprite as that has to generate a new graphic every time it's resized, whereas with the gradient graphic you can just use `scale` from FlxSprite to scale it after it's been generated.

When you use `loadGraphic` with a bitmap on a sprite, it automatically looks for a graphic in the cache that uses the same bitmap, so in this case it will mark the graphic as being used. I decided to keep it outputting the bitmap to preserve the original return type, though changing it to return a graphic is definitely possible.

I'll test it some more and make some changes.

Starmapo:
Well turns out this will need changes on other pieces of code. FlxBar and FlxButtonPlus use `createGradientBitmapData` and cache the bitmap manually with their own keys, so I would need to add an argument to be able to use another key as well as changing the code on FlxBar and FlxButtonPlus to use the arguments.

I'll close this since it needs some more work.

Geokureli:
@Starmapo I think it would make more sense to make a new function that takes a bitmap and draws to it rather than creating a bitmap all. Then we can change those other tools to use that, and the func that creates a bitmap can call it with it's own caching logic

PR #2745, opened 2023-02-28
Allow access to FlxSignal's destroy method

### LINKED ISSUE #2743: Cannot access a FlxSignal's destroy method
I was tracing out the destroy function for a FlxSignal and noticed `destroy()` is not natively called when a FlxSignal is destroyed as part of a FlxState. I then wanted to manually call `destroy()`, but the interface is not set for it.

I can create a PR to allow calling `destroy()` on a FlxSignal called directly. Thoughts on this?

DigiEggz:
Closes #2743

### DIFF
--- a/flixel/util/FlxSignal.hx
+++ b/flixel/util/FlxSignal.hx
@@ -39,6 +39,11 @@ abstract FlxTypedSignal<T>(IFlxSignal<T>)
 		this.removeAll();
 	}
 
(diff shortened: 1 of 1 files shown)


Geokureli:
CI is failing but clearly it has nothing to do with your changes, I'm just gonna try running the failed jobs again for my own curiosity before I merge

Geokureli:
thanks!

PR #2746, opened 2023-03-03
add "addFrameOffset" and "setFrameOffset" to FlxFramesCollection

BeastlyGabi:
for a good while now, I've noticed that the Animation Controller class lacks a proper way to set offsets to individual animations rather than relying on the `_sprite`'s animation, I've made a simple map that mitigates that issue

### DIFF
--- a/flixel/animation/FlxAnimationController.hx
+++ b/flixel/animation/FlxAnimationController.hx
@@ -524,8 +524,8 @@ class FlxAnimationController implements IFlxDestroyable
 			return;
 		}
 
-		var AnimOffsets:Array<Float> = [OffsetX, OffsetY];
-		offsets.set(AnimName, AnimOffsets);
+		var animOffsets:Array<Float> = [OffsetX, OffsetY];
+		offsets.set(AnimName, animOffsets);
 	}
 
 	/**
@@ -569,8 +569,8 @@ class FlxAnimationController implements IFlxDestroyable
 
 		if (offsets.exists(AnimName))
 		{
-			var _curOffset:Array<Float> = offsets.get(AnimName);
-			_sprite.offset.set(_curOffset[0], _curOffset[1]);
+			var currentOffset:Array<Float> = offsets.get(AnimName);
+			_sprite.offset.set(currentOffset[0], currentOffset[1]);
 		}
 
 		if (oldFlipX != _curAnim.flipX || oldFlipY != _curAnim.flipY)
--- a/flixel/animation/FlxAnimationController.hx
+++ b/flixel/animation/FlxAnimationController.hx
@@ -3,6 +3,7 @@ package flixel.animation;
 import flixel.FlxG;
 import flixel.FlxSprite;
 import flixel.graphics.frames.FlxFrame;
+import flixel.math.FlxPoint;
 import flixel.util.FlxDestroyUtil.IFlxDestroyable;
 
 class FlxAnimationController implements IFlxDestroyable
@@ -59,7 +60,7 @@ class FlxAnimationController implements IFlxDestroyable
 	/**
 	 * Stores all the animation offsets that were added to this sprite.
 	 */
-	public var offsets:Map<String, Array<Float>>;
+	public var offsets = new Map<String, FlxPoint>();
 
 	/**
 	 * Internal, reference to owner sprite.
@@ -83,7 +84,6 @@ class FlxAnimationController implements IFlxDestroyable
 	{
 		_sprite = Sprite;
 		_animations = new Map<String, FlxAnimation>();
-		offsets = new Map<String, Array<Float>>();
 	}
 
 	public function update(elapsed:Float):Void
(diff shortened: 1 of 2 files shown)


Geokureli:
can you share a test project that uses this, I'm not quite getting it

ACrazyTown:
> can you share a test project that uses this, I'm not quite seeing the usability

I think FNF uses this, to keep the characters in place instead of appearing like they're "gliding"

Geokureli:
> I think FNF uses this, to keep the characters in place instead of appearing like they're "gliding"

I believe there's some issues with atlases generated from Animate. not sure if it's a flixel issue or an Animate issue, but rather than putting in a bandaid we should see what's causing them. Flixel's atlases are far from perfect, I'm sure there's some bugs

Can you offer an example this is fixed by this feature.

BeastlyGabi:
it's likely an animate issue given how this doesn't happen just with flixel spritesheets

https://user-images.githubusercontent.com/93128334/222934993-e31a5d0a-180d-4e03-a57f-d3a5f2699f80.mp4


this PR aims to "fix" this (or rather, provide a way of doing so), while I believe there's another better way, it 
is also an issue with the way frame offsets are exported on adobe animate


https://user-images.githubusercontent.com/93128334/222935122-00c36109-35ff-4afa-af34-7f0eb1419e55.mp4

this is how it looks with the fix applied and with the offsets attached to the sprite's animations

Geokureli:
I'm having trouble spotting the difference in the 2 vids, can you just have a new project with the same animation, one with the fixed offset and one without?, and share the code that is used to fix the one

Geokureli:
any chance you could share the json and png? i just wanna test some things

BeastlyGabi:
sure

[Boyfriend.zip](https://github.com/HaxeFlixel/flixel/files/10902702/Boyfriend.zip)

Geokureli:
@BeastlyGabi FYI please use code tags to share code rather than screenshots, like this:
\`\`\`hx
// code here
/`/`/`

shows up as formatted and colored haxe code that I can copy and paste rather than retyping it by hand

Geokureli:
I made some changes, but I'm still thinking this isn't the right move. What I don't like about this is that it sets the offset rather than adjusting the position on top of the sprite's existing offset. FlxFrames has an [offset field](https://github.com/HaxeFlixel/flixel/blob/dev/flixel/graphics/frames/FlxFrame.hx#L83), that doesn't change the FlxSprite's offset. I think texture packer flies use this.

We should instead create a way to adjust frame offsets to FlxAtlasFrames, and add a helper to set the offset of all frames with a given prefix (for your specific use-case). I'm also fairly certain that a per-animation offset won't fix everyone's issues, I think it just-so-happens to fix yours because of how it was animated

Geokureli:
Rather than applying offsets to animations via the animation controller, I've added the ability to offset frames in the atlas. this will:
1. allow people to use this offset even on sprites with custom offsets (since offset is often used for collision boxes)
2. will easily apply offsets to all sprites using the atlases, where before they all needed to have the animation offsets applied

I made a [small test project](https://github.com/Geokureli/flixel-tests/blob/main/Source/states/AtlasOffsetTestState2746.hx) that uses the new method. [Watch an example vid, here](https://imgur.com/a/W67hV2I).

Note that sprite.offset and frame.offset seem to be applied in opposite ways, so I had to add the negative values to the new system.

Instead of 
```hx
sprite.animation.addFramesOffsetByPrefix("idle"     ,  5 ,  0 );
sprite.animation.addFramesOffsetByPrefix("hey"      ,  7 ,  5 );
sprite.animation.addFramesOffsetByPrefix("singLEFT" ,  4 , -7 );
sprite.animation.addFramesOffsetByPrefix("singDOWN" , -22, -51);
sprite.animation.addFramesOffsetByPrefix("singUP"   , -47,  28);
sprite.animation.addFramesOffsetByPrefix("singRIGHT", -48, -5 );
```

it's 
```hx
atlas.addFramesOffsetByPrefix("BF idle dance" , -5 , -0 );
atlas.addFramesOffsetByPrefix("BF HEY!!"      , -7 , -5 );
atlas.addFramesOffsetByPrefix("BF NOTE LEFT0" , -4 ,  7 );
atlas.addFramesOffsetByPrefix("BF NOTE DOWN0" ,  22,  51);
atlas.addFramesOffsetByPrefix("BF NOTE UP0"   ,  47, -28);
atlas.addFramesOffsetByPrefix("BF NOTE RIGHT0",  48,  5 );
```

Geokureli:
@BeastlyGabi planning to merge this sometime after CI finishes. Until then, any comments/concerns you have with the new approach?

BeastlyGabi:
the new approach seems much better in comparison to mine, it solves my issue while being better in the long run, so it's perfectly fine for me

Geokureli:
forgot to add doc comments, added them here: https://github.com/HaxeFlixel/flixel/commit/e988404ac491510099ee0d4369c8c85071e9d7fb

Thanks BeastlyGabi for bringing the issue to my attention and giving examples

PR #2747, opened 2023-03-03
Move doc from AssetPaths to FlxAssets

T1mL3arn:
As it was suggested in https://github.com/HaxeFlixel/flixel-templates/pull/42#discussion_r1122083274

### DIFF
--- a/flixel/system/FlxAssets.hx
+++ b/flixel/system/FlxAssets.hx
@@ -56,54 +56,54 @@ class FlxAssets
 	
 	#if (macro || doc_gen)
 	/**
-		Reads files from a directory relative to this project and generates `public static inline`
-		variables containing the string paths to the files in it.
-
-		**Example usage:**
-
-		```haxe
-		@:build(flixel.system.FlxAssets.buildFileReferences("assets/images/"))
-		class Images {}
-		```
-
-		**Renaming Duplicates**
-
-		If you have files with the same names, whichever file is nested deeper or found later 
-		will be ignored. You can provide `rename` function to deal with this case. The function takes a filepath 
-		(a relative filepath from the `Project.xml`) and returns a field name used to access that path. 
-		Returning `null` means "ignore the file".
-
-		```haxe
-		// assets structure:
-		// assets/music/hero.ogg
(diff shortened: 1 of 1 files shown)


Geokureli:
@T1mL3arn how does this look?

flixel/system/FlxAssets.hx
```diff
@@ -104,7 +104,6 @@ class FlxAssets
 	 *                          will exclude .ogg files and everything in the exclude folder
 	 * @param   rename          A function that takes the file path and returns a valid haxe field name.
 	 *
-	 * @see [AssetPaths](https://api.haxeflixel.com/AssetPaths.html) for more examples
```
T1mL3arn: I think `AssetPaths` can be added to api.flixel.com, it has now good doc and examples, thus the link should be here.

Geokureli: figure out how to add it, first. api.flixel pages are generated from from files in [this directory](https://github.com/HaxeFlixel/flixel/tree/dev/flixel), as I said before, AssetPaths doesn't exists in this repo, its created when the dev creates a project.

I also don't think we should add it to the api, because it's not actually part of the api. Different games can have different implementations on AssetPath. this is why we added docs to the template

T1mL3arn: >it's not actually part of the api.

`spinehaxe` and `nape` too, yet [they have their doc](https://api.haxeflixel.com/spinehaxe/). They are explicitly added to the doc generation process in [project.xml](https://github.com/HaxeFlixel/flixel-docs/blob/c2a3205a94558d9d39eee831727b34e57f3794c8/api/xml/project.xml#L13-L16).

>figure out how to add it

There are two options:

1. The shortest: update docgen script to download AssetPaths.hx and save it next to [`project.xml`](https://github.com/HaxeFlixel/flixel-docs/blob/master/api/xml/project.xml). Include AssetPaths.hx as haxeflag. I did it, and it worked:
image
2. Longer, but probably better: alter `flixel-templates` library so it provides AssetPaths as a real haxe module, then setup docgen `project.xml` to use it.

PR #2754, opened 2023-03-12
Added warnings for when 'animation.addBy*' doesn't find any frames

UncertainProd:
When adding animation frames using `animation.addByPrefix` or other such `animation.addBy____` functions, flixel usually doesn't warn the user if no frames were found until the user calls `animation.play()` where it warns that the "animation doesn't exist". This pr should warn the user right when they're trying to add frames that don't exist.


Geokureli:
I was actually thinking about re-imagining this and just having a getFramesByPrefix/forEachByPrefix method

PR #2756, opened 2023-03-16
Revert FlxButton Status Change

MAJigsaw77:
I revert that change because it made the input to be really weird.

### DIFF
--- a/flixel/ui/FlxButton.hx
+++ b/flixel/ui/FlxButton.hx
@@ -514,7 +514,7 @@ class FlxTypedButton<T:FlxSprite> extends FlxSprite implements IFlxInput
 	 */
 	function onUpHandler():Void
 	{
-		status = FlxButton.HIGHLIGHT;
+		status = FlxButton.NORMAL;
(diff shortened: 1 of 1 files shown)


Geokureli:
Give a code sample that shows the issue your having, that is fixed by this PR. after that we will come up with a solution that resolves that issue as well as the issue in PR #2657

MAJigsaw77:
> Give a code sample or scenario that shows the issue you're having, that is fixed by this PR. after that we will come up with a solution that resolves that issue as well as the issue in PR #2657

I'm using this fix on a custom project, but the reason is, when I keep my finger on the button then I tap with another one it will call on out function but by revering this it gets fixed

Geokureli:
Tbh the issue solved by the old pr is more common than the one you describe. FlxButtons arent really good with multitouch, at the moment. And in many cases devs might even want an onOut to trigger when a new touch is started elsewhere.

Either we fix both of the issues, or you extend FlxButton in your project and make the change there

PR #2771, opened 2023-04-11
Allow `FlxCamera#followLerp` to be used with `NO_DEAD_ZONE`

DalekCraft2:


flixel/FlxCamera.hx
```diff
@@ -1178,7 +1178,8 @@ class FlxCamera extends FlxBasic
 		{
 			target.getMidpoint(_point);
 			_point.addPoint(targetOffset);
-			focusOn(_point);
+			_scrollTarget.set(_point.x - width * 0.5, _point.y - height * 0.5);
+			_point.putWeak();
```
Geokureli, line 1182: _point is an instance var, it does not get put and will never be weak

Geokureli (changes requested):
CI is failing for what seems like network issues, remove the putWeak and I'll re-run CI as needed

Geokureli (approved):

DalekCraft2:
I was having connection issues when running the `haxelib` command in command prompt around the same time as you were doing the CIs, but it's running fine now, so maybe the CIs will work if they are re-run now.

Geokureli:
thanks!

PR #2772, opened 2023-04-13
Multi atlas

Geokureli:
## Multi-Atlases
Allows FlxAtlasFrames to be combined or merged, for use in a single sprite. 
- `addAtlas`: To add one atlas's frames to another.
- `concatenate` To form a new atlas containing all the frames of 2 atlases.

## Revert small part of https://github.com/HaxeFlixel/flixel/pull/2752
I've removed the newly added `setFramesDurationByPrefix` from `FlxFramesCollection` out of fear that it may cause confusion. You should never need to set the fps of an entire animation this way, instead frame speed should be controlled via `FlxAnimationController`.

## To Do Before Merging
- [x] Test out different scenarios with combining different atlases with different `persist`, `destroyOnNoUse` while switching states
- [x] ~~Look into `FlxAtlas`, how is it different from `FlxAtlasFrames`? Do I need to upkeep that, or make similar changes?~~ do later
- [x] Fix checkstyle failing on `static final` vars, or convert them to vars


Geokureli:
I tested scenarios where atlases with different values of `persist` and `destroyOnNoUse`. As I expected, certain graphics were destroyed causing errors when attempting to reuse the main FlxGraphic. This seems like a pretty rare case, but nonetheless, the errors it caused were incredibly confusing. I'm now looking for destroyed graphics when creating animations, so the dev understands what they did wrong.

Vortex2Oblivion:
No way
I need

PR #2778, opened 2023-04-21
update FlxMouseEventManager class doc

T1mL3arn:


flixel/input/mouse/FlxMouseEventManager.hx
```diff
@@ -38,6 +32,17 @@ import flixel.util.FlxDestroyUtil;
  * function onMouseOver(object:FlxObject) {}
  * function onMouseOut(object:FlxObject) {}
  * ```
+ *
```
T1mL3arn, line 34: Doc below this comment probably should be moved to `FlxMouseEvent`, but let it be here for now.

Geokureli: Why not do it now?

flixel/input/mouse/FlxMouseEventManager.hx
```diff
@@ -6,28 +6,21 @@ import flixel.FlxBasic;
...
 import flixel.input.mouse.FlxMouseButton.FlxMouseButtonID;
 import flixel.math.FlxPoint;
 import flixel.util.FlxDestroyUtil;
 
 /**
  * Provides mouse event detection for `FlxObject` and `FlxSprite` (pixel-perfect for those).
- * To use it, initialize the manager and register objects / sprites.
+ * The manager get initialized by `Flixel`. To use it, get the manager with 
+ * [`FlxG.plugins.get()`](https://api.haxeflixel.com/flixel/system/frontEnds/PluginFrontEnd.html#get) 
+ * and register objects / sprites:
```
Geokureli, line 18: the static `FlxMouseEvent` class is meant to be the access point for the global manager, not via `FlxG.plugins`. It's Similar to `FlxTween` or `FlxTimer`. All of this doc should probably be moved there

Geokureli (changes requested):
Move all of this to `FlxMouseEvent`.  FlxMouseEventManager should really just tell them that this global managing functionality was moved there, and that a new instance can be made for private management (but isn't recommended)

Geokureli:
Thanks!

PR #2781, opened 2023-04-28
Fix issue with createCameraWall

slameron:
When making a cameraWall using placeOutside, the left and right walls are pushed down, causing a gap between the top wall and the side walls.


Geokureli:
can you provide an example that highlights the case that this fixes?

slameron:
Before the change, the math with `placeOutside` caused there to be a gap of `thickness` on the top and bottom.

image

After the change there are no gaps

image

PR #2787, opened 2023-05-03
Fix for html5 when OS display scale is >100%

aeveis:
See https://github.com/aeveis/haxeflixel-display-scale-fix example project for details.  This fixes the display scale cropping I see and also should not affect how something renders regardless of set offset.


aeveis:
Testing some more cases the offset I'm setting is still not enough to prevent the camera cropping so I will do some more investigation.

aeveis:
Again testing it seems things are still off so I will do some more investigation on my end.

aeveis:
This fix for this is setting this in the project.xml:
	`<window if="html5" resizable="true" allow-high-dpi="false" />`

PR #2789, opened 2023-05-07
Add `fieldHeight` prop for `FlxText` to control textfield height

### LINKED ISSUE #2783: `FlxText` always shrinks its "graphic object" height
No matter how I try I cannot set desired height for `FlxText`, its height shrinks to the real text height . After some time digging I found the culprit ([see FlxText.hx#710](https://github.com/HaxeFlixel/flixel/blob/3418173e0de6f44e30206383df0b2abb9741242a/flixel/text/FlxText.hx#L790)):

```haxe
var newHeight:Int = Math.ceil(textField.textHeight) + VERTICAL_GUTTER;
...
// which later becomes field's and render frame's height
height = newHeight;
frameHeight = newHeight;
```

```haxe
/**
  The height of the text in pixels.
**/
public var textHeight(get, never):Float;
```

I believe instead of `textField.textHeight` there should be `textField.height`, though maybe there was a reason to not use `height` prop.

T1mL3arn:
closes #2783 

A fix with backward compatibility. Though I would add `fieldHeight` to the constructor. Also I made sort of a demo with tests, if you want it I will clean up and post here.


Geokureli:
Sorry for taking a while to get to this, hopefully soon

Geokureli:
@T1mL3arn you said you had a demo, can you share it?

T1mL3arn:
https://github.com/T1mL3arn/Flixel-text-height-demo

flixel/text/FlxText.hx
```diff
@@ -35,11 +35,46 @@ import openfl.utils.AssetType;
...
+ * ```
  */
 class FlxText extends FlxSprite
 {
 	/**
 	 * 2px gutter on both top and bottom
+	 * 
+	 * NOTE: Right now it is needed when new height is calculated from
+	 * `textHeight` to get a nice padding from bottom line of text
+	 * to bottom edge of text field.
```
Geokureli, line 77: this is verbose and confusing, lets just do
```hx
/**
 * 2px gutter on both top and bottom when the text has dynamic height
 */

File: flixel/text/FlxText.hx [hunk 2]
```diff
@@ -35,11 +35,46 @@ import openfl.utils.AssetType;
...
+ * var text = new FlxText(0, 0, -1, "Hello");
+ * // or
+ * text.autoSize = true;
+ * text.wordWrap = false;
+ * // or
+ * text.fieldWidth = text.fieldHeight = -1;
+ * 
+ * // 4. auto width but fixed height
+ * // has no sense and works as default autosize
+ * ```
```
Geokureli, line 68: I think this would be much clearer without the examples and they should just be in the doc for autoSize, fieldWidth and fieldHeight, maybe here it can have an autosizing section that refers the dev to those fields. i.e:

```
 * ## Autosizing
 * 
 * By default `FlxText` is autosized to fit it's text. To set a fixed size,
 * use the `fieldWidth`, `fieldHeight` and `autoSize` fields.
```

flixel/text/FlxText.hx
```diff
@@ -122,14 +157,50 @@ class FlxText extends FlxSprite
...
+		}
+		else
+		{
+			_autoHeight = false;
+
+			textField.height = value;
+		}
+		_regen = true;
+		return value;
+	}
```
Geokureli, line 194: setters and getters are defined below

flixel/text/FlxText.hx
```diff
@@ -122,14 +157,50 @@ class FlxText extends FlxSprite
...
+
+	function get_fieldHeight():Float
+	{
+		return (textField != null) ? textField.height : 0;
+	}
+
+	function set_fieldHeight(value:Float):Float
+	{
+		if (textField == null)
+			return value;
```
Geokureli, line 181: put an empty line after if-bodies when there is code immediately after it

Geokureli (changes requested):
mostly needs doc and organizational changes, otherwise seems good

flixel/text/FlxText.hx
T1mL3arn: Also. "2px on both top and bottom" is not correct, it is should be "4px at the bottom". You can clearly see it on flash target.

flixel/text/FlxText.hx
T1mL3arn: Maybe add a snippet or update a text demo or even both? What do you think?

T1mL3arn:
What do you think about adding `fieldHeight` to the constructor?

flixel/text/FlxText.hx
Geokureli: some snippets for text would be awesome, but that would happen after this gets merged

Geokureli:
> What do you think about adding `fieldHeight` to the constructor?

That would be a breaking change, unfortunately, also it's not very commonly used, tbh

Geokureli (approved):

Geokureli:
Thanks!

PR #2796, opened 2023-05-09
Custom bouncing (FlxEase.hx) +  changed variables to defaults

Wither362:
Todo?
- [x] Clean code
- [ ] Test it.

### DIFF
--- a/flixel/tweens/FlxEase.hx
+++ b/flixel/tweens/FlxEase.hx
@@ -161,11 +161,17 @@ class FlxEase
 	}
 
 	/**
-	 * Returns A NEW EASE FUNCTION. So it should be used a like different than often:
-	 * // Normal method:
-	 * `ease: FlxEase.bounceIn`
-	 * // New method:
-	 * `ease: FlxEase.bounceInCustom(FlxEase.B1 * 0.5, FlxEase.B2 * 0.5, FlxEase.B3 * 0.5, FlxEase.B4 * 0.5, FlxEase.B5 * 0.5, FlxEase.B6)`
+	 * Creates a new ease function with the specified constants
+	 * 
+	 * Note: This method cannot be used as an ease param, you must pass the result into an ease param.
+	 * For example: 
+	 * ```haxe
+	 * final myEase = FlxEase.createBounceInOut(1 / 2.75, 2 / 2.75, 1.5 / 2.75, 2.5 / 2.75, 2.25 / 2.75, 2.625 / 2.75);
+	 * FlxTween.tween(sprite, { x: 600, y: 800 }, 2.0, { ease: myEase });
+	 * ```
+	 * 
+	 * @return A function that can be passed into a tween's ease property
+	 * @since 5.4.0
 	 */
 	public static function createBounceIn(b1:Float, b2:Float, b3:Float, b4:Float, b5:Float, b6:Float):EaseFunction
 	{
@@ -173,11 +179,17 @@ class FlxEase
 	}
 
 	/**
-	 * Returns A NEW EASE FUNCTION. So it should be used a like different than often:
-	 * // Normal method:
-	 * `ease: FlxEase.bounceOut`
-	 * // New method:
-	 * `ease: FlxEase.bounceOutCustom(FlxEase.B1 * 0.5, FlxEase.B2 * 0.5, FlxEase.B3 * 0.5, FlxEase.B4 * 0.5, FlxEase.B5 * 0.5, FlxEase.B6)`
+	 * Creates a new ease function with the specified constants
+	 * 
+	 * Note: This method cannot be used as an ease param, you must pass the result into an ease param.
+	 * For example: 
+	 * ```haxe
+	 * final myEase = FlxEase.createBounceIn(1 / 2.75, 2 / 2.75, 1.5 / 2.75, 2.5 / 2.75, 2.25 / 2.75, 2.625 / 2.75);
+	 * FlxTween.tween(sprite, { x: 600, y: 800 }, 2.0, { ease: myEase });
+	 * ```
+	 * 
+	 * @return A function that can be passed into a tween's ease property
+	 * @since 5.4.0
 	 */
 	public static function createBounceOut(b1:Float, b2:Float, b3:Float, b4:Float, b5:Float, b6:Float):EaseFunction
 	{
@@ -185,11 +197,17 @@ class FlxEase
 	}
 
 	/**
-	 * Returns A NEW EASE FUNCTION. So it should be used a like different than often:
-	 * // Normal method:
-	 * `ease: FlxEase.bounceInOut`
-	 * // New method:
-	 * `ease: FlxEase.bounceInOutCustom(FlxEase.B1 * 0.5, FlxEase.B2 * 0.5, FlxEase.B3 * 0.5, FlxEase.B4 * 0.5, FlxEase.B5 * 0.5, FlxEase.B6)`
+	 * Creates a new ease function with the specified constants
+	 * 
+	 * Note: This method cannot be used as an ease param, you must pass the result into an ease param.
+	 * For example: 
+	 * ```haxe
+	 * final myEase = FlxEase.createBounceInOut(1 / 2.75, 2 / 2.75, 1.5 / 2.75, 2.5 / 2.75, 2.25 / 2.75, 2.625 / 2.75);
+	 * FlxTween.tween(sprite, { x: 600, y: 800 }, 2.0, { ease: myEase });
+	 * ```
+	 * 
+	 * @return A function that can be passed into a tween's ease property
+	 * @since 5.4.0
 	 */
 	public static function createBounceInOut(b1:Float, b2:Float, b3:Float, b4:Float, b5:Float, b6:Float):EaseFunction
 	{
--- a/flixel/tweens/FlxEase.hx
+++ b/flixel/tweens/FlxEase.hx
@@ -39,6 +39,16 @@ class FlxEase
 		return t;
 	}
 
+	public static inline function inOutHelper(t:Float, easeIn:EaseFunction, easeOut:EaseFunction)
+	{
+		return t <= .5 ? easeIn(t * 2) / 2 : (1 + easeOut(t * 2 - 1)) / 2;
+	}
+
+	public static inline function invertEase(t:Float, ease:EaseFunction)
+	{
+		return 1 - ease(1 - t);
+	}
+
 	public static inline function quadIn(t:Float):Float
 	{
 		return t * t;
@@ -46,12 +56,12 @@ class FlxEase
 
 	public static inline function quadOut(t:Float):Float
 	{
-		return -t * (t - 2);
+		return invertEase(t, quadIn);
 	}
 
 	public static inline function quadInOut(t:Float):Float
 	{
-		return t <= .5 ? quadIn(t) * 2 : 1 - (--t) * t * 2;
+		return inOutHelper(t, quadIn, quadOut);
 	}
 
 	public static inline function cubeIn(t:Float):Float
@@ -61,12 +71,12 @@ class FlxEase
 
 	public static inline function cubeOut(t:Float):Float
 	{
-		return 1 + (--t) * t * t;
+		return invertEase(t, cubeIn);
 	}
 
 	public static inline function cubeInOut(t:Float):Float
 	{
-		return t <= .5 ? cubeIn(t) * 4 : -t * (t - 2) * 4;
+		return inOutHelper(t, cubeIn, cubeOut);
 	}
 
 	public static inline function quartIn(t:Float):Float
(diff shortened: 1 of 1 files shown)


Wither362:
Once it’s pushed (if it gets to get pushed), I’ll implement more custom functions (or maybe I can code it in this patch…)

Wither362:
The compile didn’t work because it didn’t download Haxe with total support, but it works.

Geokureli:
gonna take a bit before I get to this, there's some things I wanna test and potentially change and I won't have the time for a couple days. Sorry for the wait, and thanks for doing this

flixel/tweens/FlxEase.hx
```diff
@@ -151,25 +155,60 @@ class FlxEase
...
+			if (t < Bounce1)
+				return 7.5625 * t * t;
+			if (t < Bounce2)
+				return 7.5625 * (t - Bounce3) * (t - Bounce3) + .75;
+			if (t < Bounce4)
+				return 7.5625 * (t - Bounce5) * (t - Bounce5) + .9375;
+			return 7.5625 * (t - Bounce6) * (t - Bounce6) + .984375;
+		};
+		return func;
+	}
```
Geokureli, line 190: 1. remove commented args
2. use lowercase args
3. use default arg values  from consts instead of null floats, this will work if B1-B6 are changed to inline
4. define an inlined private helper rather than using anonymous functions
```suggestion
		return 1 - bounceOutHelper(t);
	}
	
	public static function bounceInCustom(b1 = B1, b2 = B2, b3 = B3, b4 = B4, b5 = B5, b6 = B6):Float
	{
		return (t)->1 - bounceOutHelper(t, b1, b2, b3, b4, b5, b6);
	}
	
	public static function bounceOutCustom(b1 = B1, b2 = B2, b3 = B3, b4 = B4, b5 = B5, b6 = B6):Float
	{
		return (t)->bounceOutHelper(t, b1, b2, b3, b4, b5, b6);
	}
	
	static inline function bounceOutHelper(t:Float, b1 = B1, b2 = B2, b3 = B3, b4 = B4, b5 = B5, b6 = B6):Float
	{
			if (t < b1)
				return 7.5625 * t * t;
			if (t < b2)
				return 7.5625 * (t - b3) * (t - b3) + .75;
			if (t < b4)
				return 7.5625 * (t - b5) * (t - b5) + .9375;
			return 7.5625 * (t - b6) * (t - b6) + .984375;
	}
```

flixel/tweens/FlxEase.hx
```diff
@@ -151,25 +155,60 @@ class FlxEase
...
+	public static function bounceInOutCustom(/*bounces:Int = 6, */?Bounce1:Float, ?Bounce2:Float, ?Bounce3:Float, ?Bounce4:Float, ?Bounce5:Float, ?Bounce6:Float):EaseFunction
+	{
+		var outFunc:EaseFunction = bounceOutCustom(/*bounces, */Bounce1, Bounce2, Bounce3, Bounce4, Bounce5, Bounce6);
+		var func:EaseFunction = function(t:Float)
+		{
+			return t < 0.5
+				? (1 - outFunc(1 - 2 * t)) / 2
+				: (1 + outFunc(2 * t - 1)) / 2;
+		};
+		return func;
```
Geokureli, line 201: same changes as above and use the bounceOutHelper as well

flixel/tweens/FlxEase.hx
```diff
@@ -151,25 +155,60 @@ class FlxEase
 
 	public static function bounceIn(t:Float):Float
 	{
-		return 1 - bounceOut(1 - t);
+		return BOUNCE_IN(t);
+	}
+
+	public static function bounceInCustom(/*bounces:Int = 6, */?Bounce1:Float, ?Bounce2:Float, ?Bounce3:Float, ?Bounce4:Float, ?Bounce5:Float, ?Bounce6:Float):EaseFunction
```
Geokureli, line 161: these need doc comments.

how are the args determined? right now they seem quite arbitrary, and only determined via long trial and error. what do these values mean to the dev using this function?

flixel/tweens/FlxEase.hx
```diff
@@ -151,25 +155,60 @@ class FlxEase
...
 	public static function bounceOut(t:Float):Float
 	{
-		if (t < B1)
-			return 7.5625 * t * t;
-		if (t < B2)
-			return 7.5625 * (t - B3) * (t - B3) + .75;
-		if (t < B4)
-			return 7.5625 * (t - B5) * (t - B5) + .9375;
-		return 7.5625 * (t - B6) * (t - B6) + .984375;
+		return BOUNCE_OUT(t);
```
Geokureli, line 206: use the above mentioned bounceOutHelper

flixel/tweens/FlxEase.hx
```diff
@@ -151,25 +155,60 @@ class FlxEase
...
 	}
 
 	public static function bounceInOut(t:Float):Float
 	{
-		return t < 0.5
-			? (1 - bounceOut(1 - 2 * t)) / 2
-			: (1 + bounceOut(2 * t - 1)) / 2;
+		return BOUNCE_IN_OUT(t);
```
Geokureli, line 211: use the above mentioned bounceOutHelper

Geokureli (changes requested):
I still need to do some tests in a couple days, but this could be much more optimized and less verbose

flixel/tweens/FlxEase.hx
```diff
@@ -50,7 +54,7 @@ class FlxEase
 
 	public static inline function quadInOut(t:Float):Float
 	{
-		return t <= .5 ? t * t * 2 : 1 - (--t) * t * 2;
+		return t <= .5 ? quadIn(t) * 2 : 1 - (--t) * t * 2;
```
Geokureli, line 57: is `1 - (--t) * t * 2` just a fancy way of saying `quadOut((t - 0.5) * 2)`? changing would be more readible and it's all inlined to reduced math expressions by the compiler anyway

Geokureli: also I believe this should be `quadIn(t * 2)` not `quadIn(t) * 2`. so please visually verify each function that was changed

Geokureli: Here's an example of how I personally, visually verified the changes of your previous PR
https://github.com/Geokureli/flixel-tests/blob/main/Source/states/BounceEaseTestState2788.hx

if the equations were not changed blue and dark blue should always be touching, regardless of the mouse position, same with red and green

flixel/tweens/FlxEase.hx
Geokureli: furthermore, can you provide for me an example of a good custom bounce function that I may try

Wither362:
Ok, see your point, I’ll try to fix it (cpp gave a strange error transferring them to inlines…)

By the way, is there any difference between using `0.5` and using `.5`?

flixel/tweens/FlxEase.hx
```diff
@@ -65,7 +69,7 @@ class FlxEase
 
 	public static inline function cubeInOut(t:Float):Float
 	{
-		return t <= .5 ? t * t * t * 4 : 1 + (--t) * t * t * 4;
+		return t <= .5 ? cubeIn(t) * 4 : quadOut(t) * 4;
```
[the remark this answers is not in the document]
Wither362: > quadOut should be cubeOut

Yeah. I personally don’t know what were I thinking of

flixel/tweens/FlxEase.hx
```diff
@@ -20,17 +20,21 @@ package flixel.tweens;
...
+	static var DEFAULT_B2(default, null):Float = 2 / 2.75;
+	static var DEFAULT_B3(default, null):Float = 1.5 / 2.75;
+	static var DEFAULT_B4(default, null):Float = 2.5 / 2.75;
+	static var DEFAULT_B5(default, null):Float = 2.25 / 2.75;
+	static var DEFAULT_B6(default, null):Float = 2.625 / 2.75;
+	static var DEFAULT_ELASTIC_AMPLITUDE(default, null):Float = 1;
+	static var DEFAULT_ELASTIC_PERIOD(default, null):Float = 0.4;
+	static var BOUNCE_OUT:Float->Float = bounceOutCustom();
+	static var BOUNCE_IN:Float->Float = bounceInCustom();
+	static var BOUNCE_IN_OUT:Float->Float = bounceInOutCustom();
```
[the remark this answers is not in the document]
Wither362: > * The prefix DEFAULT_ is verbose an unneeded
> * these should be inlined (unless they use PI, since PI isn't inlined)
> * capitalized field names should only be used for inline constants
> * Why even define a var to `bounceOutCustom` why not just reference `bounceOutCustom` directly? this just seems like needless obfuscation

* I defined it to `DEFAULT_` because it kinda looks better to me, since it (or will be) a non-changeable variable, I tried to make it clear.
* Before I tried to do it (`static inline var`), and cpp gave an error, so I thought the problem was in using inlines.
* Ok, I thought… well, I can’t make up excuses. I’ll change it.
* I get it, and I can’t tell any excuse right.

Wither362:
Forget it.

I have to rewrite it all again.


Stupid head of mine…

Geokureli:
No difference between 0.5 and .5 but it's easier to notice the period in 0.5

flixel/tweens/FlxEase.hx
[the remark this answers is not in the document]
Geokureli: Reversed? You used quadOut in cubeInOut, so it should be cubeOut

Wither362:
Ok, I’ve been reviewing your code, and I give you THIS answer:

bounceInCustom should return an `EaseFunction`.

Wither362:
Wait I have to-

Wither362:
I also gave elastic a chance (in another fork), so, if this one gets pushed, I’ll also request the custom elastic one.

Geokureli:
Still gonna take a couple of days before i can test everything.

I also asked tor an examples of custom eases that you think people might use

Geokureli (changes requested):
Hold off on making more changes, I have some changes I'm about to commit

Lots of issues, none of the bounce functions work correctly. You flipped boundIn and bounceOut and bounceInOut bounces in the middle of the tween rather than the ends. You need to actually test out the code and verify that it works.

I made a graph of the old functions vs your branch.

- darker lines represent the previous ease functions
- lighter lines are your ease functions
- blue is bounceIn
- green is bounceOut
- red is bounceInOut

Notice that light green matches dark blue and light blue matches dark green, and light red is inverted

Screenshot 2023-05-11 at 2 26 58 PM


I also have a personal rule, no new features unless someone has a real need for it in a project. That's why I'm asking for an example use-case of the custom ease function. If this is a feature you think someone might hypothetically want someday, I'm not all that interested in it. If you or someone you know has a project that would use this custom bounce feature (or if another framework you've used has a good implementation of it that you have used), then I totally welcome it. Without an ACTUAL use-case, features made with "hypothetical intent" rarely end up being practical. And I fear that's the case here given that B1 through B6 just seem like magic numbers

Geokureli:
I made the changes I described earlier, but I'm still not sure if this custom feature is worthwhile

your example in the doc comment is confusing, 1. because FlxEase.B1 and the like are private 2. because you multiple b1-5 by the same value but not b6, was this intentional? lastly because the result looks like this.
Screenshot 2023-05-11 at 4 33 55 PM

multiplying B6 by 0.5 results in this, which isn't much better
Screenshot 2023-05-11 at 4 33 55 PM


I also simplified the in/out/inOut bodies of various eases. everything will be inlined and optimized into reduced math expressions when the dev uses analyser optimizations

Geokureli:
there seems to be an issue with github actions

Wither362:
With your changes, what does the program prints?

Wither362:
> I made the changes I described earlier, but I'm still not sure if this custom feature is worthwhile, or made in a way that anyone will find useful
> 
> your example in the doc comment is confusing, 1. because FlxEase.B1 and the like are private 2. because you multiple b1-5 by the same value but not b6, was this intentional? lastly because the result looks like this. Screenshot 2023-05-11 at 4 30 40 PM
> 
> multiplying B6 by 0.5 results in this, which isn't much better Screenshot 2023-05-11 at 4 33 55 PM
> 
> I also simplified the in/out/inOut bodies of various eases. everything will be inlined and optimized into reduced math expressions when the dev uses analyser optimizations

It will take time, but I’ll implement options for how many jumps, strength of each one, and etc

Geokureli:
Once again, is this a feature that you're making because you need it in a project? Or because you imagine people may one day need it?

Geokureli:
After thinking about it a bit, I think this isn't a great idea. there's a million valid ways to implement a "custom bounce ease" that I don't think we need a specific one in flixel.

If you do make one I would put it in a separate, personal repo, since it's pretty easy to copy and paste ease code and add it to a project. when that happens, I'll add it to the [External Tools](https://github.com/HaxeFlixel/flixel/wiki/External-Tools) page in the wiki

Wither362:
> After thinking about it a bit, I think this isn't a great idea. there's a million different valid ways to implement a "custom bounce ease" that I don't think we need a specific one in flixel.
> 
> If you do make one I would put it in a separate, personal repo, since it's pretty easy to copy and paste ease code and add it to a project. when that happens, I'll add it to the [External Tools](https://github.com/HaxeFlixel/flixel/wiki/External-Tools) page in the wiki

Ok, anyways, thanks for helping me

PR #2802, opened 2023-05-15
Replace self with abstract in FlxAxes

EliteMasterEric:
This new syntax was made available in Haxe 4.3.0.

Made via web so let's see if this builds.

### DIFF
--- a/flixel/util/FlxAxes.hx
+++ b/flixel/util/FlxAxes.hx
@@ -17,29 +17,19 @@ enum abstract FlxAxes(Int)
 	 */
 	public var y(get, never):Bool;
 	
-	/**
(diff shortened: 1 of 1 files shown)


Geokureli:
the minimum Haxe version is 4.2.5 and we don't have plans to go to 4.3 any time soon. It's only month old and this particular feature doesn't benefit anyone. Closing this issue but we'll reopen it when we raise the minimum version

I'll wait until CI finishes, I'm not sure if closing cancels CI and I want to see the results

Geokureli:
@EliteMasterEric FYI, we're raising the minimum haxe version after flixel 6.2

PR #2806, opened 2023-05-18
large FlxGroup changes

Geokureli:
- Changed a lot of while loops to for loops, this was a relic from the flash days, js haxe will turn them back into for loops but this is more readable
- Move function scoped vars to for loop scoped vars when they are only used in the loop, same reason as above
- Add `killMembers` and `reviveMembers` and `keyValueIterator`
- Add new searching utils `getFirst`, `getLast`, `getFirstIndex`, `getLastIndex`, `any` and `every`
- Use inlined `getFirst` for all other, similar search utils
- Replace `getFirstNull`'s loop with a simple indexOf check
- add unit tests for new and changed FlxGroup features

### DIFF
--- a/flixel/FlxObject.hx
+++ b/flixel/FlxObject.hx
@@ -934,7 +934,7 @@ class FlxObject extends FlxBasic
 		var group = FlxTypedGroup.resolveGroup(objectOrGroup);
 		if (group != null) // if it is a group
 		{
-			return FlxTypedGroup.overlaps(overlapsCallback.bind(_, 0, 0, inScreenSpace, camera), group);
+			return group.any(overlapsCallback.bind(_, 0, 0, inScreenSpace, camera));
 		}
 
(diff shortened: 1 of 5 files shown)


NeeEoo:
I always thought the old group loop method were to reduce memory allocation or cpu cycles

Geokureli:
I'm pretty sure its a remnant from the flash days, declaring loop vars outside of the loop was very common in flash. either way, if there were benefits, they're likely insignificant and not worth the lack of readibility, but I'll see if i can back up that claim

NeeEoo:
I could be wrong, but iirc i also saw the old method being done in javascript projects.

Geokureli:
for loops are converted to while loops for javascript, anyway. https://try.haxe.org/#8F85C046
Ignore the `haxe_iterators_ArrayIterator`, thats an DCE issue and you can see that a while loop is used

PR #2811, opened 2023-05-21
Fix reversed animations not finishing

Starmapo:
I noticed reversed animations didn't call the animation finish callback. This fixes that by checking `tempFrame` instead of `frame`.


Geokureli:
Thanks!

Tested this via https://github.com/Geokureli/flixel-tests/blob/main/Source/states/ReverseAnimTestState2811.hx

PR #2812, opened 2023-05-23
Fixed bug causing focus to be lost when pressing TAB.

EliteMasterEric:
The associated bug can be reproduced in this way:

- Build any project with FLX_DEBUG enabled.
- While in-game, press `TAB`.

Expected behavior is that nothing should happen in-game (unless something is bound to TAB).

Actual behavior is that the window loses focus and the cursor becomes detached from the window.

The root cause is that the console textbox is being focused by OpenFL even when the debugger is not open.

Tested and with this fix the textbox still focuses when the debugger is OPEN (as expected).

### DIFF
--- a/flixel/system/debug/console/Console.hx
+++ b/flixel/system/debug/console/Console.hx
@@ -117,8 +117,10 @@ class Console extends Window
 		// openfl/openfl#1856
 		FlxG.stage.addEventListener(KeyboardEvent.KEY_DOWN, function(e:KeyboardEvent)
 		{
-			if (FlxG.debugger.visible && FlxG.game.debugger.console.visible && e.keyCode == Keyboard.TAB)
+			if (FlxG.debugger.visible && FlxG.game.debugger.console.visible && e.keyCode == Keyboard.TAB) {
 				FlxG.stage.focus = input;
+			}
+			e.preventDefault();
 		});
 		#end
 		#end
--- a/flixel/system/debug/console/Console.hx
+++ b/flixel/system/debug/console/Console.hx
@@ -117,10 +117,11 @@ class Console extends Window
 		// openfl/openfl#1856
 		FlxG.stage.addEventListener(KeyboardEvent.KEY_DOWN, function(e:KeyboardEvent)
 		{
-			if (FlxG.debugger.visible && FlxG.game.debugger.console.visible && e.keyCode == Keyboard.TAB) {
+			if (FlxG.debugger.visible && FlxG.game.debugger.console.visible && e.keyCode == Keyboard.TAB)
 				FlxG.stage.focus = input;
-			}
-			e.preventDefault();
+			#if windows
+			if (e.keyCode == Keyboard.TAB) e.preventDefault();
+			#end
 		});
 		#end
 		#end


MemeHovy:
I feel like I remember something like that happening before, and that can honestly be a bit of a pet peeve to people, so this is a good change ngl (I don't understand why it wasn't added sooner)

Geokureli:
nice I'll try this out today

flixel/system/debug/console/Console.hx
```diff
@@ -117,8 +117,10 @@ class Console extends Window
 		// openfl/openfl#1856
 		FlxG.stage.addEventListener(KeyboardEvent.KEY_DOWN, function(e:KeyboardEvent)
 		{
-			if (FlxG.debugger.visible && FlxG.game.debugger.console.visible && e.keyCode == Keyboard.TAB)
+			if (FlxG.debugger.visible && FlxG.game.debugger.console.visible && e.keyCode == Keyboard.TAB) {
```
Geokureli, line 120: don't cuddle brackets
```suggestion
			if (FlxG.debugger.visible && FlxG.game.debugger.console.visible && e.keyCode == Keyboard.TAB)
			{
```

flixel/system/debug/console/Console.hx
```diff
@@ -117,8 +117,10 @@ class Console extends Window
...
 				FlxG.stage.focus = input;
+			}
+			e.preventDefault();
```
Geokureli, line 123: do we want to preventDefault on all keys, or just on tab? should we prevent it on all targets? does this issue happen on all targets?

I'm thinking "no" for all 3

Geokureli (changes requested):

flixel/system/debug/console/Console.hx
Geokureli: also why add the bracket here in the first place, at first I assumed it was to include `e.preventDefault();`

Geokureli (approved):

Geokureli:
Thanks, I still need to test this on my windows pc. Also when i press tab on mac it just crashes (tried on dev, btw) Was gonna see why, and see if this fixes that

Geokureli:
linking this as it seems related:
https://github.com/openfl/openfl/issues/1856

flixel/system/debug/console/Console.hx
```diff
@@ -119,6 +119,9 @@ class Console extends Window
 		{
 			if (FlxG.debugger.visible && FlxG.game.debugger.console.visible && e.keyCode == Keyboard.TAB)
 				FlxG.stage.focus = input;
+			#if windows
```
Geokureli, line 122: nitpicky, I know but please put an empty line above here. I always try to put one empty after an if's block
```suggestion

			#if windows
```

Geokureli (changes requested):

Geokureli:
Also I tried these steps on my windows pc (without your changes):
> Build any project with FLX_DEBUG enabled.
> While in-game, press TAB.

the "actual behavior" you describe doesn't happen for me, not sure if it's becuase it's windows 11, I can try my win10 pc. But the system mouse appears when i press "~" revealing the debug menu, not when i press tab. My game doesn't lose focus in any way, i'm still able to control it.

Geokureli:
I wasn't seeing the issue because I didn't have hscript installed. with hscript installed I now see that pressing tab is a way to select the console input, which seems to pause the game. Unfortunately this change doesn't prevent that, and even clicking the console field pauses the game

EliteMasterEric:
> Also I tried these steps on my windows pc (without your changes):
> > Build any project with FLX_DEBUG enabled.
> > While in-game, press TAB.
> 
> the "actual behavior" you describe doesn't happen for me, not sure if it's because it's windows 11, I can try my win10 pc. But the system mouse appears when i press "~" revealing the debug menu, not when i press tab. My game doesn't lose focus in any way, i'm still able to control it.

To clarify on the symptoms 

If the debug menu is open, pressing TAB switches focus to the text field for the console.

If the debug menu is CLOSED, pressing TAB seems to switch focus to the inactive element causing focus to detach from Flixel entirely.

Geokureli:
> If the debug menu is CLOSED, pressing TAB seems to switch focus to the inactive element causing focus to detach from Flixel entirely.
I tried this as well, and saw no issue on any of my machines. maybe it's because I didn't create any textFields?

NeeEoo:
I thought #2760 fixed this issue

Geokureli:
@EliteMasterEric can you try and confirm that the issue is still happening to you on the latest release? not sure if you made this branch based on experiences happening prior to upgrading to flixel 5.3

EliteMasterEric:
> @EliteMasterEric can you try and confirm that the issue is still happening to you on the latest release? not sure if you made this branch based on your experiences prior to upgrading to flixel 5.3. especially since I cannot reproduce the error

It looks like I was on an old branch and hadn't fast-forwarded. The issue appears to be resolved now.

PR #2818, opened 2023-05-25
Add hxformat.json to published Flixel release

T1mL3arn:
Sometimes I need to do some tests with Flixel code, and if my project has different formatter rules it affects Flixel formatting. This pr fixes this (Lime, for example, publishes its formatter rules too).


Geokureli:
sure, why not?

PR #2819, opened 2023-05-25
Allow force-complete for `FlxSplash`.

T1mL3arn:
Without this fix `preGameStart` signal is not dispatched in some cases.

```haxe
package;

class TestState extends flixel.FlxState {
  override function create() {
    trace('TestState create()');
  }
}
class Main extends Sprite {

  public function new() {
    super();
    
    trace(#if debug 'debug' #else 'release' #end);
    flixel.FlxG.signals.preGameStart.addOnce(() -> trace('DISPATCHED!'));
    addChild(new flixel.FlxGame());
    Flixel.switchState(new TestState());
  }
}
```

Outputs in release and debug differ (due to `FlxSplash`, it is skipped in debug), but expected to be the same.

```
Output in release:

source/Main.hx:28: release
source/Main.hx:11: TestState create()
```

```
Output in debug:

source/Main.hx:28: debug
source/Main.hx:30: DISPATCHED!
source/Main.hx:11: TestState create()
```

So it is possible to trigger state switching before splash sets `_gameJustStarted` and without this flag `preGameStart` is not called.

### DIFF
--- a/flixel/system/FlxSplash.hx
+++ b/flixel/system/FlxSplash.hx
@@ -1,8 +1,8 @@
 package flixel.system;
 
+import flash.Lib;
 import flash.display.Graphics;
 import flash.display.Sprite;
-import flash.Lib;
 import flash.text.TextField;
 import flash.text.TextFormat;
 import flash.text.TextFormatAlign;
@@ -33,6 +33,7 @@ class FlxSplash extends FlxState
 	var _cachedBgColor:FlxColor;
 	var _cachedTimestep:Bool;
 	var _cachedAutoPause:Bool;
+	var complete:Bool = false;
 
 	override public function create():Void
 	{
@@ -83,6 +84,19 @@ class FlxSplash extends FlxState
 			FlxG.sound.load(FlxAssets.getSound("flixel/sounds/flixel")).play();
 		}
 		#end
+
+		FlxG.signals.preStateSwitch.addOnce(onStateSwitch);
+	}
+
+	function onStateSwitch()
+	{
+		/**
+			Normally FlxSplash switches to a new state on its completion,
+			but in case when the new state was requested before that moment
+			FlxSplash must be force-completed without using FlxSplash.nextState.
+		**/
+		if (!complete)
+			onComplete(null);
 	}
 
 	override public function destroy():Void
@@ -93,6 +107,7 @@ class FlxSplash extends FlxState
 		_times = null;
 		_colors = null;
 		_functions = null;
+
 		super.destroy();
 	}
 
@@ -191,6 +206,11 @@ class FlxSplash extends FlxState
 
 	function onComplete(Tween:FlxTween):Void
 	{
+		// If `Tween` is null then method was called not by FlxTween
+		// but by us which means splash must be force-completed
+		// due ongiong state switch
+		var forceComplete = Tween == null;
+
 		FlxG.cameras.bgColor = _cachedBgColor;
 		FlxG.fixedTimestep = _cachedTimestep;
 		FlxG.autoPause = _cachedAutoPause;
@@ -199,7 +219,11 @@ class FlxSplash extends FlxState
 		#end
 		FlxG.stage.removeChild(_sprite);
 		FlxG.stage.removeChild(_text);
-		FlxG.switchState(Type.createInstance(nextState, []));
+
+		if (!forceComplete)
+			FlxG.switchState(Type.createInstance(nextState, []));
+
 		FlxG.game._gameJustStarted = true;
+		complete = true;
 	}
 }


T1mL3arn:
The fix does not interfere with arg in `FlxGame` constructor. If user intentionally skips the splash with the arg, it behaves the same way as without the fix - no splash no problem.

T1mL3arn:
If user does not use `skipSplash` arg in `FlxGame.new` then this means "show the splash" which leads to a bug. If user sets this arg to `true` then splash is gone and that's fine - no splash no bug.

T1mL3arn:
Btw this also fixes stale "HaxeFlixel" text.

```haxe
addChild(new flixel.FlxGame(0, 0, null, false));

// dalay is used to show more stalled objects
// with no delay it is just "HaxeFlixel" text
haxe.Timer.delay(() -> {
  Flixel.switchState(new flixel.FlxState());
}, 250);

// NOTE: must be compiled with release flag !
```
<details>
    <summary>Click to see screenshot</summary>
    
image

</details>

In the case when state switch happened before splash completes, the splash lefts its flash display objects on the stage.

Geokureli:
I'm confused by your "stale text" issue becuase I don't consider your example to be reasonably expected code from a haxeflixel dev, is this the only type this bug happens, or have you seen this issue in a more practical project?

Geokureli:
So it looks like I misread your original post, I missed the line  `Flixel.switchState(new TestState());`. assuming "Flixel" is supposed to be "FlxG", I'm curious why anyone would ever do this? is this an attempt to skip the splash screen early? If so, there are easier ways to accomplish that

Geokureli:
I don't think we should allow people to skip the splash screen by calling switchState during, they should just make a new custom splash screen that allows for skipping, we can make this easier, by making FlxSplash more modular

Geokureli:
In our upcoming flixel 6 [feature branch](https://github.com/HaxeFlixel/flixel/pull/2733) I've made FlxSplash more modular. I in response to this PR, I added [this](https://github.com/HaxeFlixel/flixel/pull/2733/commits/bbf509e53ac20c358626df7681d8345ce4d37c76) which allows you skip in the way you like, but the [prior commit](https://github.com/HaxeFlixel/flixel/pull/2733/commits/324acd4d930f58785d8e2ffe74647446ab3bc71b) allows you to extend FlxSplash and make a skippable version, only difference is that `preGameStart` and `postGameStart` are dispatched with the splash state rather than after, but this seems pretty easy to deal with

Example:
```hx
class Main extends openfl.display.Sprite
{
    public function new()
    {
        super();
        addChild(new flixel.FlxGame(()->new SkippableSplash(flixel.FlxState.new), true));
    }
}

class SkippableSplash extends flixel.system.FlxSplash
{
	public function new (nextState)
	{
		super(nextState);
	}
	
	{
		super.create();
		
		#if FLX_KEYBOARD
		FlxG.keys.enabled = true;
		#end
	}
	
	{
		super.update(elapsed);
		
		final keysPressed = #if FLX_KEYBOARD FlxG.keys.justPressed.ANY #else false #end;
		final mousePressed = #if FLX_MOUSE FlxG.mouse.justPressed #else false #end;
		final touchPressed = #if FLX_TOUCH FlxG.touches.justStarted().length > 0 #else false #end;
		final gamepadPressed = #if FLX_GAMEPAD FlxG.gamepads.anyPressed(ANY) #else false #end;
		if (keysPressed || mousePressed || touchPressed || gamepadPressed)
			complete();
	}
}
```

T1mL3arn:
>I'm confused by your "stale text" issue becuase I don't consider your example to be reasonably expected code from a haxeflixel dev, is this the only time this bug happens, or have you seen this issue in a more practical project?

I agreed that in a more real project probably no one encounters this problem. The bug was discovered when I started to use some state switching for quick tests:

```haxe
// normal startup code
addChild(new FlxGame(0, 0, InitialState));

// immediate state switch to test things
FlxG.switchState(new Foo());
// FlxG.switchState(new Bar());
```

And during such test I got null-pointer exception from third-party lib. The exception was thrown only in a release build, the reason for the exception is that the lib didn't initialize some of its data due to `preGameStart` was not fired. So looking at how these flixel things work I found `FlxSplash` do not cover the case when it is switched out before its completion (and that abnormal state switching is the reason why `preGameStart` is not fired and for openfl objects leftovers). And I fixed that with this pr.

>Edit: it also seems like this fixes the stale text issue inadvertently, we could fix it in a more robust way by removing the openfl display objects in destroy().

Yep, that code from `onComplete()` could be partially moved to `destroy()`

>So it looks like I misread your original post, I missed the line Flixel.switchState(new TestState());

Sorry my bad, forgot to rename it back, I personally use such renaming because it has less camel-case and it is faster to type. Btw, could this be add as an official typedef?

T1mL3arn:
Tested `flxstate_new` branch and found no errors with abnormal switching from splash. Thanks!

Geokureli:
> Sorry my bad, forgot to rename it back, I personally use such renaming because it has less camel-case and it is faster to type. Btw, could this be add as an official typedef?

No worries, also, I feel this. 90% of the time I type "FLxG" with a capital L, I suck at typing so much its a wonder i went into software dev in the first place

PR #2820, opened 2023-05-27
Fix pixelPerfectPointCheck when using offset

lcapps-luke:
"pixelPerfectPointCheck" was not accounting for graphic offset relative to sprite hitbox. This meant that checks on sprites with hitboxes not exactly aligned with their graphic would be inaccurate.

### DIFF
--- a/flixel/util/FlxCollision.hx
+++ b/flixel/util/FlxCollision.hx
@@ -202,7 +202,7 @@ class FlxCollision
 		// How deep is pointX/Y within the rect?
 		var test:BitmapData = Target.framePixels;
 
-		var pixelAlpha = FlxColor.fromInt(test.getPixel32(Math.floor(PointX - Target.x), Math.floor(PointY - Target.y))).alpha;
+		var pixelAlpha = FlxColor.fromInt(test.getPixel32(Math.floor(PointX - Target.x + Target.offset.x), Math.floor(PointY - Target.y + Target.offset.y))).alpha;
 
 		if (FlxG.renderTile)
 		{


Geokureli:
FYI, it's easier for me to test out and make changes to PR branches if you make a new branch for the change, rather than using lcapps-luke:dev. If you're new to github, check out [the contribution wiki](https://github.com/HaxeFlixel/flixel/wiki/Contribution-Guide#use-feature-branches).

No need to make a new branch, tho just remember for next time

Geokureli:
Looks like this doesn't fully fix the issue, I also don't know if this would work with scaled, rotated sprites and/or sprites with scrollFactors other than 1.0.
https://github.com/HaxeFlixel/flixel/assets/2609513/34634a13-15c3-4eb5-b6be-c1b793ce8dac

you could use [sprite.pixelsOverlapPoint](https://github.com/HaxeFlixel/flixel/blob/dev/flixel/FlxSprite.hx#LL1016C34-L1016C34), which gets the pixel of a sprite at the desired world coordinates. however I just noticed [an error in that function](https://github.com/HaxeFlixel/flixel/pull/2824/files), so thanks for leading me to find that!

Geokureli (changes requested):
after fixing this we should probably just deprecate FlxCollision.pixelPerfectPointCheck in favor of sprite.pixelOverlapsPoint, unless you see a benefit to this

Geokureli (approved):

lcapps-luke:
@Geokureli no, looks good. Thanks

Issue #2821, opened 2023-05-28
flixel.animation.FlxAnimationController has no field getFrameDuration


JonnycatMeow:
- **Haxe version:** 4.3.1
- **Flixel version:** 5.3.1
- **OpenFL version:** 9.2.1
- **Lime version:** 8.0.1
- **Affected targets:**
Mac 
________________________________

**Code snippet reproducing the issue**:

/usr/local/opt/haxe/haxelib/flixel/5,3,1/flixel/animation/FlxAnimation.hx:231: characters 35-51 : flixel.animation.FlxAnimationController has no field getFrameDuration
![Screen Shot 2023-05-27 at 6 00 50 PM](https://github.com/HaxeFlixel/flixel/assets/107599365/41e8c8a5-c947-4c94-9290-1cceb7420daf)

MemeHovy:
it's a flixel 5.3 feature, add this to FlxAnimationController
```haxe
	#if (flixel >= "5.3.0")
	@:allow(flixel.animation.FlxAnimation)
	function getFrameDuration(index:Int)
	{
		return _sprite.frames.frames[index].duration;
	}
	#end
```

JonnycatMeow:
thank you

JonnycatMeow:
can you make a pull request regarding this issue

MemeHovy:
> thank you

you're welcome, and sure thing

JonnycatMeow:
acutally i got another error

JonnycatMeow:
![Screen Shot 2023-05-27 at 7 10 25 PM](https://github.com/HaxeFlixel/flixel/assets/107599365/0f183db7-de50-454d-ad76-820f66835046)

JonnycatMeow:
nvm i fixed my issue

JonnycatMeow:
but thank you

Geokureli:
@JonnycatMeow please share what the issue was and what fixed it, in case other people are having the same issue

JonnycatMeow:
so there was this fnf mod call tails.exe and i relized they had put the flixel.animation.Flxanimationcontroller in the source code

JonnycatMeow:
and so i replaced it with the origninal Flxanimationcontroller with the moddified one

JonnycatMeow:
and i also had to modified it as well because it needed certain variables

JonnycatMeow:
and it worked

JonnycatMeow:
[...]

PR #2822, opened 2023-05-28
Compile fix for flixel 5.3

MemeHovy:
Fixes an issue with FlxAnimationController preventing people compiling
based on #2821


Geokureli:
why would flixel need to check it's own version?

MemeHovy:
> why would flixel need to check it's own version?, anyone who is using this code will be on flixel 5.3.1 or later, so this does nothing
> 
> Edit: looks like this didn't fix the issue in #2821 anyway

it was something someone requested

PR #2825, opened 2023-05-29
X & Y parameters for alphaMaskFlxSprite

Cherif107:


Geokureli:
Im not very familiar with this feature can you provide details about why this is needed and a practical use case?

PR #2829, opened 2023-06-02
All Turkish translate

neondev27:
Here is my take on main MDs in Turkish language for Turkish developers. If requested, I will do more translate in this repo


Geokureli:
I've outlined the steps to include Turkish translations in #2830, so I'm closing this as a duplicate

PR #2832, opened 2023-06-07
Changing all the references from `Flash` to `OpenFL`.

MAJigsaw77:
The title tells everything this commit do.


Geokureli:
ive been putting this off because it will likely cause merge conflicts with almost every other active PR, but let's try it. gimme some time, first

flixel/input/mouse/FlxMouse.hx
```diff
@@ -17,11 +17,11 @@ import flixel.system.FlxAssets;
...
-import flash.Vector;
-import flash.geom.Matrix;
-import flash.geom.Point;
-import flash.ui.MouseCursor;
-import flash.ui.MouseCursorData;
+import openfl.Vector;
+import openfl.geom.Matrix;
+import openfl.geom.Point;
+import openfl.ui.MouseCursor;
+import openfl.ui.MouseCursorData;
```
Geokureli, line 24: looks like flash is failing here because there is no `openfl.ui.MouseCursorData`

Geokureli: https://github.com/HaxeFlixel/flixel/blob/dev/flixel/system/macros/FlxDefines.hx#L144
I think FLX_NATIVE_CURSOR is only used with flash, so lets just use `flash.ui.MouseCursorData`

MAJigsaw77: ok?

MAJigsaw77: Nah

Geokureli:
Thanks, hopefully this fixes most macro issues

PR #2846, opened 2023-06-21
Trigger re-draw when FlxText.text is set to empty string

### LINKED ISSUE #2845: FlxText & Classes including Text show previously existing/"ghost" characters when text is empty and fieldWidth is set
- **Haxe version:** 4.3.1
- **Flixel version:** 5.2.2
- **OpenFL version:** 9.2.1
- **Lime version:** 8,0,1
- **Affected targets:** 
HTML5

```haxe
package;

import flixel.FlxG;
import flixel.FlxState;
import flixel.text.FlxText;

class TState extends FlxState
{
	var text:FlxText;

	{
		super.create();
		text = new FlxText(20, 20, 400, "Something");
		add(text);
	}

	{
		if (FlxG.keys.justPressed.H)
			text.text = "";
		super.update(elapsed);
	}
}

```

**Observed behavior:** On the press of H, The text "Something" Stays, even though the text should is empty when referenced

**Expected behavior:** On the press of H, Text should dissapear.

UncertainProd:
Fixes #2845
When FlxText is initialized to a non-empty string and then flxtext.text is set to "" (empty string), it does not clear the text and the old text is still visible (which only seems to happen if `fieldWidth` is set to a non-zero value and only happens on html5 targets somehow). This pr should cause `drawTextFieldTo` to be called even when text is empty, which will update and clear the text correctly.

### DIFF
--- a/flixel/text/FlxText.hx
+++ b/flixel/text/FlxText.hx
@@ -869,7 +869,7 @@ class FlxText extends FlxSprite
 			}
 		}
 
-		if (textField != null && textField.text != null && textField.text.length >= 0)
+		if (textField != null && textField.text != null)
 		{
 			// Now that we've cleared a buffer, we need to actually render the text to it
 			copyTextFormat(_defaultFormat, _formatAdjusted);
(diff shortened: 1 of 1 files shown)


Geokureli:
thanks!

PR #2860, opened 2023-07-14
add a ton of aseprite atlas classes

Geokureli:
~~WIP~~Done
- Added `loadAseAtlas`, `addAseAtlasTagsByIndex`, and `addAseAtlasTagsByPrefix`
- Added `TexturePackerAtlas` and `AseAtlas` types
- Added Aseprite metadata types: `AseAtlasMeta`, `AseAtlasTag`, `AseAtlasSlice` and `AseAtlasLayer`
- Deprecated `TexturePackerObject`, `TexturePackerFrameData` and `TexturePackerFrameRect`
for `TexturePackerAtlas`, `TexturePackerAtlasFrame` and `AtlasRect`, respectively;


flixel/graphics/frames/FlxAtlasFrames.hx
```diff
@@ -414,24 +412,158 @@ class FlxAtlasFrames extends FlxFramesCollection
...
+	
+	/**
+	 * Ranges from 0 to 255
+	 */
+	@:optional var opacity:Int;
+	
+	/**
+	 * 
+	 */
+	@:optional var blendMode:AseBlendMode;
```
MondayHopscotch, line 493: It looks like `AseAtlasLayer` can also have user data attached to it. Set in-app by right clicking a cel -> `Cel Properties` -> Clicking the notepad icon

The data type appears to be of this format:
```
@:optional var cels:Array<AseCelData>;
```

```
typedef AseCelData = {
    frame:Int,
    data:String,
}
```

Geokureli: thanks! I''ll add this, btw, I've noticed that color is not included if set to #000000

flixel/graphics/frames/FlxAtlasFrames.hx
```diff
@@ -414,24 +412,158 @@ class FlxAtlasFrames extends FlxFramesCollection
...
+	 * The size of the slice at this frame
+	 */
+	var bounds:AtlasRect;
 };
 
-typedef TexturePackerFrameData =
+/**
+ * Size struct use for atlas json parsing, { w:Float, h:Float }
+ */
+typedef AtlasSize = { w:Float, h:Float };
```
MondayHopscotch, line 539: Aren't these likely to be just type `Int`? I'm currently just casting them to ints, but it got me wondering if floats even make sense here since we are referring to pixel images -- what would at atlas with size `10.3x20.25` be, for example.

Or is this just a parsing safety thing since calling it a `Float` will happily parse it either way? Just mulling this over as I'm implementing stuff in my project.

Geokureli: since were dealing in pixels, int is likely the better option. I think I tend to write float by habit

flixel/graphics/FlxAsepriteUtil.hx
```diff
@@ -0,0 +1,194 @@
...
+
+typedef AseAtlasTag = { name: String, from:Int, to:Int, direction:String };
+
+typedef AseAtlasSlice =
+{
+	var name:String;
+	var color:String;
+	
+	/**
+	 * Info of at what frames the slice changes size
```
MondayHopscotch, line 74: Minor nit:

`Info of at what frames the slice changes size/position`

Since you can move them around and resize them per frame

flixel/graphics/FlxAsepriteUtil.hx
```diff
@@ -0,0 +1,194 @@
...
+	/**
+	 * The effect added to this layer to change how the colors blend with colors on lower layers
+	 * @see AseBlendMode
+	 */
+	@:optional var blendMode:AseBlendMode;
+	
+	/**
+	 * Any notes that were left on cels of this layer
+	 */
+	@:optional var cels:Array<{ frame:Int, data:String, ?color:String }>;
```
MondayHopscotch, line 102: I failed to update my Aseprite to the latest 1.2x and am now on the release candidate for 1.3. It looks like they've added a z-index to the cel data. Also per your update to make `color` optional, it appears that `data` is also not included if it is empty in aseprite.

So I believe this is more accurate:

```
/**
 * Any properties on cels of this layer
 */
@:optional var cels:Array<{ frame:Int, ?zIndex:Int, ?data:String, ?color:String }>;
```

This is pulled from the export json with 3 frames, each having only one of the 3 values set:

```
"cels": [
    { "frame": 1, "zIndex": 4 },
    { "frame": 2, "color": "#8f8db7d8" },
    { "frame": 3, "data": "user data" }
]
```

https://www.aseprite.org/api/cel#celzindex

flixel/graphics/FlxAsepriteUtil.hx
```diff
@@ -0,0 +1,194 @@
...
+{
+	/**
+	 * The frame that the slice changes size
+	 */
+	var frame:Int;
+	/**
+	 * The size of the slice at this frame
+	 */
+	var bounds:AtlasRect;
+};
```
MondayHopscotch, line 143: Same minor nits for the descriptions here:

```
	/**
	 * The frame that the slice changes
	 */
	var frame:Int;

	/**
	 * The size and position of the slice at this frame
	 */
	var bounds:AtlasRect;
```

flixel/graphics/FlxAsepriteUtil.hx
```diff
@@ -0,0 +1,194 @@
...
+
+typedef AtlasFrame =
+{
+	/**
+	 * The name of this frame, ommitted if using a hashed atlas
+	 */
+	@:optional var filename:String;
+	
+	/**
+	 * Should alwyas be false
```
MondayHopscotch, line 168: typo in `always`

flixel/graphics/FlxAsepriteUtil.hx
Geokureli: Let me grab latest aseprite, thanks for all the feedback, so far!

Geokureli:
Fwiw i tend to do a full pass on doc once the broad structure is finalized, which i think is finally done. But i do appreciate the feedback. Much of the existing structs did not have doc comments, so i need to add that. But imo aseprite's field names are kinda weird, namely because they adopted texture packers format. Hence "filename" for frame names. 

Long story short theres a lot of doc work needed here

MondayHopscotch:
Definitely not trying to be annoying with my comments. I just happen to be implementing some stuff in my own project and am heavily using all of this, so I figure I'd point stuff out as I see/find it to take some of the burden off of you. Their field names are definitely a little strange at times.

Is there a better way you'd like me help as I run across things?

Geokureli:
it wasn't annoying, it's very welcome. Usually I'm alone, asking for more input from flixel users on my changes, and no one says anything

Geokureli:
Still need to do unit tests. apparently almost all data objects have an optional `data` and `color` field, so i made a bse object for that

MondayHopscotch:
Did you find an actual schema description you're working from? Or are you just tinkering and seeing what comes out of the export?

Geokureli:
> Did you find an actual schema description you're working from? Or are you just tinkering and seeing what comes out of the export?

just trying out a bunch of stuff in aseprite and checking the json

flixel/graphics/FlxAsepriteUtil.hx
```diff
@@ -0,0 +1,335 @@
...
+	 * @param   tagSuffix  The delimeter on each frame name that follows the animation name and precedes the frame number
+	 * @return  This `FlxSprite` instance (nice for chaining stuff together, if you're into that).
+	 * @see flixel.graphics.FlxAsepriteUtil.AseAtlasMeta
+	 * @since 5.4.0
+	 */
+	public static function addAseAtlasTags(sprite:FlxSprite, data:FlxAsepriteJsonAsset, tagSuffix:String = ":")
+	{
+		final aseData = data.getData();
+		for (frameTag in aseData.meta.frameTags)
+			sprite.animation.addByPrefix(frameTag.name, frameTag.name + tagSuffix);
```
MondayHopscotch, line 61: A question here as this was part of the work I had to do to get this working properly. `addByPrefix` relies on the existing loaded frame names. The `frameTag.name` is the name given to the animation tag in Aseprite (such as `run_left`, we'll say for this example).

The frame names (at least with the fairly default export options I'm using), are loaded based on the file name that they came from. My example is that I have a `player.ase` file, and the export json has the frames named like `player 0.ase`, `player 1.ase`, etc. So if you do `addByPrefix("run_left", "run_left:")`, there are no frames that match.

In my case, I had to build the animations based on the frame index data on the tag (`from` and `to`): https://github.com/bitDecayGames/haxeflixel-template/blob/master/source/loaders/Aseprite.hx#L54-L57

I haven't handled the case where you export with the default `--json-hash` option explicitly. There may be some strangeness with the order frames are added depending on how the hash is initially iterated when adding the frames.

Am I off-base with this, or does it make sense that this function won't properly build out animations as it sits?

Geokureli: Screenshot 2023-07-18 at 1 09 04 PM

I'm not familiar with the batch exporter, but in the editor you can specify a format for the frame names, this is the format to use with the default delimiter. I was gonna add this to the doc

> In my case, I had to build the animations based on the frame index data on the tag (from and to):

that's not a bad way to go about it either, assuming the frames are always ordered, predictably.

MondayHopscotch: Ah, I definitely missed that as I'm only using the CLI at the moment. They have `--tag-format` documented, but don't explicitly call out that it follows the same formatting rules that `--file-format` uses. However, if you click the `(?)` in the UI, it takes you here: https://www.aseprite.org/docs/cli/#filename-format

One thing I noticed with your export settings (which I think actually makes the frames way more readable than the default of `{title} {frame}.{extension}`) is that when you have overlapping tags, you lose information, for example:

I have two animations `a` and `b`, and I am using one frame of animation `a` as my knockback frame when my character takes damage
image

This exports as (shorthanded):
```
  "frames": {
   "a:0": {...},
   "b:0": {...},
   "a:2": {...},
   .
   .
   .
   },
 },
 ```
So you would lose that frame from your `a` animation if you simply load them by prefix.

I think using the `from` and `to` are the proper way to do this. They are easy to handle if you do `Array` export (via the drop-down in the UI, or via the `--json-array` flag on CLI.

The tricky part is if you select `Hash` in the UI (which is also the default for the CLI. It looks like you (or whoever originally implemented it if you pulled it from somewhere else) has the `abstract Hash<T>(Dynamic)` type that gets a key value iterator via `Reflect.fields(this).iterator();`. I'm still looking if that has any guarantees of field ordering (seems a little dangerous, either way due to potential platform differences). I'm curious why Hash is the default since there is an inherently linear nature to animation frames. 🤔 Maybe because you can have empty frames that don't have to be exported, so you can avoid sparsely populated arrays?

Comment (MondayHopscotch) [reply flixel/graphics/FlxAsepriteUtil.hx:61]: Well I'm finding more weird edge-cases. If you add an empty frame to your animation (frame 5 on this timeline):
image

Then export the sprite sheet and check the `Layout` -> `Ignore Empty` box

Your frames are exported as:
```
  "frames": {
   "a:0": {...},
   "a:1": {...},
   "a:2": {...},
   "a:3": {...},
   "a:5": {...},
   "left:0": {...},
   "left:1": {...},
   "left:2": {...},
   "left:3": {...}
 },
```

And your `frameTag` for `a` looks like:
```
{ "name": "a", "from": 0, "to": 5, ...}
```

So there is an implicit frame index referenced in the tag that isn't actually exported in the json. The two tags `a` and `left` reference 10 total frames across their `from` and `to` values, but there are only 9 elements in the `frames` Hash/Array.

Maybe it's best to just document what export settings are supported and draw some reasonable bounds to keep the scope reasonable.

Comment (Geokureli) [reply flixel/graphics/FlxAsepriteUtil.hx:61]: > To be honest, this seems like an Aseprite bug to me.

Screenshot 2023-07-18 at 2 13 17 PM

Screenshot 2023-07-18 at 2 13 54 PM

There is an option to ignore empty frames

Comment (MondayHopscotch) [reply flixel/graphics/FlxAsepriteUtil.hx:65]: > 

I was calling out specifically when there is an empty frame in the middle of a tag, it will _not include_ the frame (as expected if you check that box), but it will give your frame tag counts as though that frame still exists. I think it's well beyond the scope of what you're supporting with this PR. Here's a simplified example just to exhibit the weirdness without any additional stuff:

Simple animation with 3 frames. Middle frame is empty:
image

Export with `Ignore Empty` checked:
image

Shorthanded JSON:
```
{
  "frames": [
   {
    "filename": "a:0",...},
   {
    "filename": "a:2",...}
 ],
 "meta": {
  "frameTags": [
   { "name": "a", "from": 0, "to": 2, ...}
  ],
 }
}
```

Tag `a` says it is from frame `0` to frame `2` (these are frame indexes rather than names), which is 3 frames total. However, there are only 2 frames in the frame array. So our tag references a frame that aseprite took out during export, but didn't adjust the tag metadata, which _should_ be `from: 0, to: 1`.

But I'm way too deep in edge cases at the moment. I'm bubbling back up to just focus on the high level best way to parse tag frames out as animations.

At the moment, your way requires that you set the `Item Filename` to something other than the default for parsing to work (your `{tag}:{tagframe}` format). This will miss frames if the user has overlapping tags in their .ase file.

The other option is to go off of the `from` and `to` in the `frameTag` data, but that has the _potential_ issue of how fields are parsed when `Hash` is used.

Do you have a preference or general thoughts on those two directions?

Comment (Geokureli) [reply flixel/graphics/FlxAsepriteUtil.hx:65]: also, there's "{outertag}:{frame}" which always uses the longest tag which solves your issue above with `"a:0", "b:0" "a:2"` outertag would result in `"a:0", "a:1" "a:2"`

Comment (MondayHopscotch) [reply flixel/graphics/FlxAsepriteUtil.hx:65]: Yeah, it seems like either way, you lose information needed to parse this easily. Using `{tag}`, you lose the ability to parseByPrefix to build tag `a` properly. Using `{outertag}` loses that tag `b` even exists. 🤔 This is way more subtle than I was expecting.

And yes, the quick summary of my above rant is that ignoring empty frames makes it so the frame indexes in the tag are unreliable. Which seems really silly - it should be reasonably simple to account for having removed empty frames so those indices are correct.

**Comment by Geokureli** [comment]
@MondayHopscotch any notes or changes on the new methods?

File: flixel/graphics/FlxAsepriteUtil.hx [hunk 8]
```diff
@@ -0,0 +1,442 @@
...
+	 * Uses the tag's `to` and `from` fields to determine.
+	 * 
+	 * Note: It's recommended to disable the "Ignore Empty" option if you have empty frames (that 
+	 * is, empty on every layer) in tags, while it will reduce the size of your json, this method
+	 * will not be able to add those empty frames to an animation.
+	 * 
+	 * @param   sprite     The sprite to add the animations
+	 * @param   data       Can be an `AseAtlas` struct, a JSON string matching the `AseAtlas` or a
+	 *                     string asset path to a json
+	 * @return  This `FlxSprite` instance (nice for chaining stuff together, if you're into that).
```
MondayHopscotch, line 122: can these all just be "The `FlxSprite` (nice for [...]" to avoid potential confusion with the keyword `this`?

MondayHopscotch (commented):
Overall looks good. I forgot there is just the option to provide both methods of loading them 😄.

flixel/graphics/FlxAsepriteUtil.hx
Geokureli: this was copied from a bunch of FlxSprite methods and the wording is all over, and these methods are meant to be used via static extension  i.e. `using flixel.graphics.FlxAsepriteUtil;` and `sprite.loadAseAtlasAndTags(graphic, data, " ");`

That said, i see no issue in changing "This" to "The"

flixel/graphics/FlxAsepriteUtil.hx
```diff
@@ -0,0 +1,442 @@
...
+		return sprite;
+	}
+	
+	/**
+	 * Helper for parsing Aseprite atlas json files. Reads frame data via `loadAseAtlas`,
+	 * then, adds animations for any tags listed, via `addAseAtlasTagsByIndex`.
+	 * 
+	 * Note: It's recommended to disable the "Ignore Empty" option if you have empty frames (that 
+	 * is, empty on every layer) in tags, while it will reduce the size of your json, this method
+	 * will not be able to add those empty frames to an animation.
```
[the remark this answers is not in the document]
Geokureli: is `--ignore-empty` the default, or is it disabled by default?

flixel/graphics/FlxAsepriteUtil.hx
```diff
@@ -0,0 +1,442 @@
...
+		return sprite;
+	}
+	
+	/**
+	 * Helper for parsing Aseprite atlas json files. Reads frame data via `loadAseAtlas`,
+	 * then, adds animations for any tags listed, via `addAseAtlasTagsByPrefix`.
+	 * 
+	 * Notes: Assumes that the frame names are prefixed by tag name and assumes that the `tagSuffix`
+	 * does not appear anywhere in the tag name. You can use the filename format `{outertag}:{frame}`
+	 * for most cases.
```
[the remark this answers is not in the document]
Geokureli: added "and assumes that no tags overlap"

PR #2863, opened 2023-07-21
update transform after setting pan

Geokureli:
should fix sound.pan = #
still needs testing


Geokureli:
https://github.com/openfl/lime/blob/develop/src/lime/_internal/backend/html5/HTML5AudioSource.hx
https://community.openfl.org/t/panning-sound-while-playing/6961/10

PR #2869, opened 2023-07-27
add setTint and setBrightness

### LINKED ISSUE #2864: Adobe Animate's color effect formulas
Tint:
(except alpha)
```haxe
colorTransform.multipliers = 1 - multiplier
colorTransform.offsets = Math.round(tint.rgb * multiplier)
```
the offset has to be multiplied with the Int side of the colors.

(extracted from [FlxTint.hx](https://github.com/Dot-Stuff/flxanimate/blob/dev/flxanimate/effects/FlxTint.hx), my class)

Brightness:
(except alpha)
```haxe
colorTransform.multipliers =  1 - Math.abs(brightness)
colorTransform.offsets = brightness * 255;
```
(extracted from [FlxBrightness.hx](https://github.com/Dot-Stuff/flxanimate/blob/dev/flxanimate/effects/FlxBrightness.hx), my class)

I know those are classes that I made, but these were made reverse engineering Animate's color effects.

Edit: Made some corrections, now it should be fine

Geokureli:
Fixes #2864

Still need to test, may move these functions to FlxSpriteUtil

@Cheemsandfriends any comments or concerns?

### DIFF
--- a/flixel/FlxSprite.hx
+++ b/flixel/FlxSprite.hx
@@ -989,41 +989,6 @@ class FlxSprite extends FlxObject
 		dirty = true;
 	}
 	
-	/**
-	 * Change's this sprite's color transform to apply a tint effect.
-	 * Mimics Adobe Animate's "Tint" color effect
-	 * 
-	 * @param   color     The 0xRRGGBB color to tint the sprite
-	 * @param   strength  The amount of tint to apply, from 0.0 to 1.0
-	 * 
-	 * @since 5.4.0
-	 */
-	public function setTint(color:FlxColor, strength:Float)
-	{
-		inline function scaleInt(i:Int):Int
-		{
-			return Math.round(i * strength);
-		}
-		
-		final mult = 1 - strength;
-		setColorTransform(mult, mult, mult, 1.0, scaleInt(color.red), scaleInt(color.green), scaleInt(color.blue));
-	}
-	
-	/**
-	 * Change's this sprite's color transform to brighten or darken it.
-	 * Mimics Adobe Animate's "Brightness" color effect
-	 * 
-	 * @param   brightness  Use 1.0 to fully brighten, -1.0 to fully darken, or anything inbetween
-	 * 
-	 * @since 5.4.0
-	 */
-	public function setBrightness(brightness:Float)
-	{
-		final mult = 1.0 - Math.abs(brightness);
-		final offset = Math.round(Math.max(0, 0xFF * brightness));
-		setColorTransform(mult, mult, mult, 1.0, offset, offset, offset);
-	}
-	
 	function updateColorTransform():Void
 	{
 		if (colorTransform == null)
--- a/flixel/FlxSprite.hx
+++ b/flixel/FlxSprite.hx
@@ -1000,8 +1000,13 @@ class FlxSprite extends FlxObject
 	 */
 	public function setTint(color:FlxColor, strength:Float)
 	{
+		inline function scaleInt(i:Int):Int
+		{
+			return Math.round(i * strength);
+		}
+		
 		final mult = 1 - strength;
-		setColorTransform(mult, mult, mult, 1.0, Math.round(color.red * strength), Math.round(color.green * strength), Math.round(color.blue * strength));
+		setColorTransform(mult, mult, mult, 1.0, scaleInt(color.red), scaleInt(color.green), scaleInt(color.blue));
 	}
 	
 	/**
--- a/flixel/FlxSprite.hx
+++ b/flixel/FlxSprite.hx
@@ -988,7 +988,37 @@ class FlxSprite extends FlxObject
 		useColorTransform = alpha != 1 || color != 0xffffff || colorTransform.hasRGBOffsets();
 		dirty = true;
 	}
-
+	
+	/**
+	 * Change's this sprite's color transform to apply a tint effect.
+	 * Mimics Adobe Animate's "Tint" color effect
+	 * 
+	 * @param   color     The 0xRRGGBB color to tint the sprite
+	 * @param   strength  The amount of tint to apply, from 0.0 to 1.0
+	 * 
+	 * @since 5.4.0
+	 */
+	public function setTint(color:FlxColor, strength:Float)
+	{
+		final mult = 1 - strength;
+		setColorTransform(mult, mult, mult, 1.0, Math.round(color.red * strength), Math.round(color.green * strength), Math.round(color.blue * strength));
+	}
+	
+	/**
+	 * Change's this sprite's color transform to brighten or darken it.
+	 * Mimics Adobe Animate's "Brightness" color effect
+	 * 
+	 * @param   brightness  Use 1.0 to fully brighten, -1.0 to fully darken, or anything inbetween
+	 * 
+	 * @since 5.4.0
+	 */
+	public function setBrightness(brightness:Float)
+	{
+		final mult = 1.0 - Math.abs(brightness);
+		final offset = Math.round(Math.max(0, 0xFF * brightness));
+		setColorTransform(mult, mult, mult, 1.0, offset, offset, offset);
+	}
+	
 	function updateColorTransform():Void
 	{
 		if (colorTransform == null)
--- a/flixel/util/FlxSpriteUtil.hx
+++ b/flixel/util/FlxSpriteUtil.hx
@@ -20,6 +20,7 @@ import flixel.math.FlxMath;
 import flixel.math.FlxPoint;
 import flixel.math.FlxRect;
 import flixel.system.FlxAssets;
+import flixel.tweens.FlxEase;
 import flixel.tweens.FlxTween;
 
 // TODO: pad(): Pad the sprite out with empty pixels left/right/above/below it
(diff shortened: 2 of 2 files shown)


Cheemsandfriends:
This is a start, but I think the tint should be a variable, not functions exactly, and then with the `updateColorTransform()` you can set the tint + "concat" It with `color`.
I also think the strength of the color should be the alpha of the tint itself

Geokureli:
> This is a start, but I think the tint should be a variable, not functions exactly

if people want to control this with a variable they can but 99% of basic usage doesn't require that overhead. Right now you can easily tween tint and brightness via `FlxTween.num(0.0, 1.0 3.0, (n)->setTint(color, n));`

> and then with the `updateColorTransform()` you can set the tint + "concat" It with `color`.
I'm not following, if you mean the ability to set tint, color, brightness and other effects individually and have them stack?

> I also think the strength of the color should be the alpha of the tint itself
can you clarify exactly what you mean, here

Geokureli:
> I also think the strength of the color should be the alpha of the tint itself

Regarding your [FlxTint](https://github.com/Dot-Stuff/flxanimate/blob/dev/flxanimate/effects/FlxTint.hx) class, I think it's okay to represent a tint variable as an 0xAARRGGBB Int **under the hood** (since there are 4 components), but I think what you're suggesting is actually a bad practice. It's a quick, esoteric short-hand that can come in handy, but is not really intuitive to newcomers.

Kinda like [FlxAngle.getPolarCoords](https://github.com/Geokureli/flixel/blob/dev/flixel/math/FlxAngle.hx#L379) which returns a FlxPoint, where `p.x` represents the magnitude of a vector and `p.y` represents the degrees of that vector. I deprecated this and added fields for `length`, ` radians` and `degrees` because it's confusing to represent data as some other form of data.

Especially since haxe has many tools that allow you to represent some data as another type under the hood really easily. here's an example of a tint class represented by a single Int: https://try.haxe.org/#17100eBB

You still deal with values `red/redFloat`, `green/greenFloat`, `blue/blueFloat` and `tint.strength = 1.0;` as expected, where your class makes then do `tint.tint.alphaFloat = 1.0;`. I do, however, see value in allowing devs to quickly define a Tint via `0xF0FF008F` which I added to the example.

lastly, it's worth mentioning that that using an int's alpha byte to store strength is less precise than using a float, which causes weird issues from time to time. this is one reason why I wanted to avoid having a tint "object" on FlxSprites with getters/setters for various fields

Also, worth mentioning if someone did want to create a system that allowed people to set tint, color and brightness objects, I have concerns:
- Animate, itself doesn't have this ability, you choose between one of them, and fields like "tint strength" or "brightness" are not available in code, all you have is `sprite.transform.colorTransform`
- how do you determine the order?

Cheemsandfriends:
> if people want to control this with a variable they can but 99% of basic usage doesn't require that overhead. Right now you can easily tween tint and brightness via FlxTween.num(0.0, 1.0 3.0, (n)->setTint(color, n));

I'd prefer making it like `FlxTween.tween(sprite, {tint: 0x7DFFFF00 brightness: 0.2}, 2.0);` cuz I believe it's more intuitive to do it instead of doing like a function to tween to.
> I'm not following, do you mean the ability to set tint, color, brightness and other effects individually and have them stack?

I think it's better to make the `color` variable to make it a "base color setting". so for example, if `color` is default, then it applies the tint normally. if not, then it's gonna be that tint plus that tinted value. so like it's tinted for everytime even if you use `tint` or `brightness`.

> can you clarify exactly what you mean, here

I meant so that instead of having two different variables, making it so it's set as "how much alpha does this color has, the amount of tint is gonna be applied onto it", altho to tween the alpha it's gonna be complicated to do that.

Cheemsandfriends:
> Regarding your [FlxTint](https://github.com/Dot-Stuff/flxanimate/blob/dev/flxanimate/effects/FlxTint.hx) class, I think it's okay to represent a tint variable as an 0xAARRGGBB Int under the hood (since there are 4 components), but I think what you're suggesting is actually a bad practice. It's a quick, esoteric short-hand that can come in handy, but is not really intuitive to newcomers.

I don't think it's unintuitive to users, since you can think of it as if you were "painting the whole sprite to that color", and the amount of alpha is how much of it you want it to be put. Animate's strength's whole purpose is to act like an "imaginary" alpha channel for the tint.

> Kinda like [FlxAngle.getPolarCoords](https://github.com/Geokureli/flixel/blob/dev/flixel/math/FlxAngle.hx#L379) which returns a FlxPoint, where p.x represents the magnitude of a vector and p.y represents the degrees of that vector. I deprecated this and added fields for length,  radians and degrees because it's confusing to represent data as some other, unrelated form of data just because they look the same in memory.

I don't think it's the same thing, since you would need to know what the point's axis were like, here's a way to apply color onto it, like if you were "stamping the color onto the sprite".

> You still deal with values red/redFloat, green/greenFloat, blue/blueFloat and tint.strength = 1.0; as expected, where your class makes then do tint.tint.alphaFloat = 1.0;.

Can you elaborate with this one? I didn't understand exactly this.

> lastly, it's worth mentioning that that using an int's alpha byte to store strength is less precise than using a float, which causes weird issues from time to time. this is one reason why I wanted to avoid having a tint "object" on FlxSprites with getters/setters for various fields

Yeah... That's something I agree it should be avoid at all costs. My guess would be that maybe do another value that should be `tintStrength`. Altho I don't like it, this way you can tween the amount of tint with a single variable and change it without changing the whole value, and maybe when setting `tint`, if it's possible, maybe check if the value is _AARRGGBB_ and with the alpha value set the strength to that value?

> Animate, itself doesn't have this ability, you choose between one of them, and fields like "tint strength" or "brightness" are not available in code, all you have is sprite.transform.colorTransform

Yeah, but that's mainly cuz when Macromedia programmed the Color effects, they had in mind to only make it one by one.

> how do you determine the order?

I don't think there's a specific order tbh. If you do it like adding it and subtract it, like [this](https://try.haxe.org/#682a624c)

Geokureli:
> I'd prefer making it like `FlxTween.tween(sprite, {tint: 0x7DFFFF00 brightness: 0.2}, 2.0);` cuz I believe it's more intuitive to do it instead of doing like a function to tween to.

The fact that order makes a difference means there can't just be a tint field on FlxSprite. If you want stacking color effects it should just be an array of effects that result in a single colorTransform applied to a sprite. this allows for multiple effects of the same type, which may or may not be useful (say if you want an enemy to flash red when attacked, which takes priority over a slight green tint used to indicate poisoned status effect)

I think what you're approaching is something like this (psuedo code):
```hx
    final list = new ColorEffectList();
    list.push(new Tint(0xFF00FF, .95));
    list.push(new Brightness(0.5));
    trace("tint then brightness");
    trace(list.flatten().toString());
    
    final list = new ColorEffectList();
    list.push(new Brightness(0.5));
    list.push(new Tint(0xFF00FF, .95));
    trace("brightness then tint");
    trace(list.flatten().toString());
    
    list.getFirst(Brightness).amount = 1.0;
    trace("changing values");
    trace(list.flatten().toString());
```

Here's the [full example](https://try.haxe.org/#664ac1f6) of how I see this happening in flixel, because I don't fully understand how your example would be implemented into flixel.

> Yeah... That's something I agree it should be avoid at all costs. My guess would be that maybe do another value that should be tintStrength

my example solves this too with:
```hx
sprite.colorEffects.getFirst(Tint).strength = 0.0;
```

But really, I still wouldn't implement this into this info flixel, it's just not necessary for nearly any practical use cases. Maybe if in the future flixel becomes more component driven I'd make this as a non-default color component option. And I wouldn't implement the proposed sprite.tint field because it's also overkill, and because (imo) it's limited due to the lack or ordering.

I may make a tiny repo with this system fully fleshed out, where it just sets the player's colorTransform

Geokureli:
Ok i caved i think ill use AARRGGBB instead

Cheemsandfriends:
> Here's the [full example](https://try.haxe.org/#664ac1f6) of how I see this happening in flixel, because I don't fully understand how your example would be implemented into flixel.

My example is laying here on [`updateColorTransform()`](https://github.com/HaxeFlixel/flixel/blob/d60a10cd5995e6a8b2a141ff0a45ad72457dd209/flixel/FlxSprite.hx#L992). when the thing's using colorTransform, the `useColorTransform` is set depending on the `alpha` and the `color` isn't default, then the multiplier is set. (line [997-1001)

Cheemsandfriends:
Wait, actually, this dismisses the example I made in haxe, Fuck.

Cheemsandfriends:
I can try to make a repo with an extended FlxSprite and a test sample to showcase the example correctly

Geokureli:
> aside from that, it's then making another check for brightness and that's it.

So the order is forced to tint then brightness? did you see the example I linked in my previous message? I would recommend making that ColorEffectList component that controls a sprite's colorTransform over extending FlxSprite to add color effect fields.

Cheemsandfriends:
> I would recommend making that ColorEffectList component that controls a sprite's colorTransform over extending FlxSprite to add color effect fields.

Actually, this would be a better idea, but something like an `Array`, or like an `FlxGroup` of effects, since you can add your own effects and all that, but I somewhat think this could be overkill if it's only having two effects

Geokureli:
> Actually, this would be a better idea, but something like an `Array`, or like an `FlxGroup` of effects, since you can add your own effects and all that, but I somewhat think this could be overkill if it's only having two effects

I feel like something was missed, you're describing [the example I gave before](https://try.haxe.org/#664ac1f6), and the part where i say it's overkill. It's too much to be included in all FlxSprites, but fine as a standalone system that people can add to their project

And even though there's only 3 types, for now (tint, brightness and just a normal colorTransform) I described a scenario where it's useful:
> say if you want an enemy to flash red when attacked, which takes priority over a slight green tint used to indicate poisoned status effect

Cheemsandfriends:
> It's too much to be included in all FlxSprites, but fine as a standalone system that people can add to their project.

Wait, what if the system's added like a util? Something like `FlxSpriteColoringUtil` or something like that?
I think that's what you wanted to add lol

Geokureli:
@Cheemsandfriends I went and made the system I was talking about
Here's an example usage: https://github.com/Geokureli/flixel-tests/blob/main/Source/states/ColorEffectListTestState2869.hx
And here's the class: https://github.com/Geokureli/flixel-tests/blob/main/Source/tools/FlxColorEffectList.hx

PR #2870, opened 2023-07-29
FlxObject: motion optimisation

Sword352:
An optimisation to FlxObject's `updateMotion` function, which basically avoids doing expensive math operations for nothing. Improves performance a lot.

(Ignore the typo commits as they were already merged to this repository)


MemeHovy:
yooo

Geokureli:
> Improves performance a lot.

Do you have any metrics on this? Is there a test project that showcases a significant fps difference?

I also think I see a bug, here will test out later

flixel/FlxObject.hx
```diff
@@ -886,22 +886,31 @@ class FlxObject extends FlxBasic
...
+			final delta = velocity.x * elapsed;
+			velocity.x += velocityDelta;
+		    x += delta;
+		}
+
+		if (velocity.y != 0) {
+			final velocityDelta = 0.5 * (FlxVelocity.computeVelocity(velocity.y, acceleration.y, drag.y, maxVelocity.y, elapsed) - velocity.y);
+		    velocity.y += velocityDelta;
+		    final delta = velocity.y * elapsed;
+		    velocity.y += velocityDelta;
```
Geokureli, line 911: these indents use both spaces and tabs, please only use tab intents

flixel/FlxObject.hx
```diff
@@ -886,22 +886,31 @@ class FlxObject extends FlxBasic
...
-
-		velocityDelta = 0.5 * (FlxVelocity.computeVelocity(velocity.y, acceleration.y, drag.y, maxVelocity.y, elapsed) - velocity.y);
-		velocity.y += velocityDelta;
-		delta = velocity.y * elapsed;
-		velocity.y += velocityDelta;
-		y += delta;
+		if (velocity.x == 0 && velocity.y == 0 && angularVelocity == 0)
+			return;
+
+		if (angularVelocity != 0) {
```
Geokureli, line 892: no bracket cuddling

Geokureli (changes requested):
As I suspected, this has a bug. I'm curious how you were able to test this change and measure any performance gains without seeing that it breaks most of the flixel demos that use velocity.

I tried this change on [EZPlatformer](https://haxeflixel.com/demos/EZPlatformer/) and the platformer never moved.

Geokureli:
Even if the bugs are fixed, I think you can get nearly all of the benefits by setting moves = false

Sword352:
Fair enough

PR #2872, opened 2023-07-29
Adding `gain` variable to `FlxSound`

MAJigsaw77:


flixel/sound/FlxSound.hx
```diff
@@ -93,18 +93,27 @@ class FlxSound extends FlxBasic
...
 	public var pitch(get, set):Float;
 	#end
+
 	/**
 	 * The position in runtime of the music playback in milliseconds.
 	 * If set while paused, changes only come into effect after a `resume()` call.
 	 */
 	public var time(get, set):Float;
 
+	#if (native && !openfl_legacy)
```
Geokureli, line 110: Rather than using `#if (blah blah blah)` we need to define a FLX_SOUND_GAIN for devs to use if they are doing multitargeting

Follow [what FLX_PITCH did](https://github.com/HaxeFlixel/flixel/blob/dev/flixel/system/macros/FlxDefines.hx#L151-L152)

Geokureli (changes requested):

flixel/sound/FlxSound.hx
MAJigsaw77: Ok

MAJigsaw77:
The gain is actually used as volume in openfl

PR #2881, opened 2023-08-02
`FlxSprite.setGraphicSize` allows floats

### LINKED ISSUE #2878: `FlxSprite.setGraphicSize` only takes integers
Is there a reason why `FlxSprite.setGraphicSize` only takes integers? The `width` and `height` properties for `FlxObject` are floats, I don't see why the function is limited to integers. For example, it's a little annoying having to do:
```haxe
sprite.setGraphicSize(Std.int(otherSprite.width), Std.int(otherSprite.height));
```
rather than just:
```haxe
sprite.setGraphicSize(otherSprite.width, otherSprite.height);
```
Even if the width and height aren't decimal numbers, it prevents having to do `Std.int` twice.

Starmapo:
Allows `setGraphicSize` to use floats as inputs, instead of being limited to integers.

Also removed references to it in the documentation saying that it automatically adjusts the hitbox (it does not).

fixes #2878

### DIFF
--- a/flixel/FlxSprite.hx
+++ b/flixel/FlxSprite.hx
@@ -101,7 +101,7 @@ using flixel.util.FlxColorTransformUtil;
  * ### Scale
  * [Snippets - Scale](https://snippets.haxeflixel.com/sprites/scale/)
  * (FlxPoint) Change the size of your sprite's graphic. NOTE: The hitbox is not automatically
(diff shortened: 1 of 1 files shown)


Geokureli:
> Also removed references to it in the documentation saying that it automatically adjusts the hitbox (it does not).

weird I thought it did, lol

Geokureli:
thanks

PR #2888, opened 2023-08-10
add 'getFilters()' function

Vortex2Oblivion:
add a function to return the camera's shaders by doing
```haxe
FlxG.camera.getFilters();
```
instead of
```haxe
@:privateAccess
FlxG.camera._filters
```
could be useful for applying the filters of one camera to another

### DIFF
--- a/flixel/FlxCamera.hx
+++ b/flixel/FlxCamera.hx
@@ -1600,6 +1600,14 @@ class FlxCamera extends FlxBasic
 	}
 
 	/**
+	 * Gets the filter array applied to the camera.
+	 */
+	public function getFilters():Void
+	{
+		return _filters;
+	}
+
+	/**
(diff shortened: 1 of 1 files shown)


Geokureli:
if we're gonna add this, maybe we should reconsider making it private in the first place?

Typically things are designed this way because directly modifying the array can have weird consequences, but since you're providing the exact array in `setArray` this layer doesn't prevent that.

Can you apply the same filter instance to multiple cameras (for all existing filter types)? You mentioned needing to use `@:privateAccess` before, can you provide a concrete example of what you did and why?

Vortex2Oblivion:
https://github.com/Vortex2Oblivion/haxeflixel-getFilters-testing
i made a little example project
hopefully this should help

Geokureli:
fuck it

PR #2894, opened 2023-08-18
Isolate atlas typedefs for better macro compatibility

MondayHopscotch:
Moves atlas typedefs into their own files to avoid imports that cause issues when used with macros.


Geokureli:
Thanks!

PR #2897, opened 2023-08-21
Destroy `_bgSprite` in `FlxSubState`

Starmapo:
This fixes `_bgSprite` not being destroyed when the substate is, which causes its graphic to keep being marked as used and thus not being cleared automatically.


Geokureli:
thanks!

PR #2906, opened 2023-08-28
Fix for FlxStrip when drawDebug is `true` #2904

### LINKED ISSUE #2904: FlxStrips appear full black when debugDraw is enabled
Affects html5 but not Mac builds. no other targets tested. I've seen similar issues when testing on mac targets, though

Example:
```hx
package states;

import flixel.FlxG;
import flixel.FlxStrip;
import flixel.graphics.tile.FlxDrawTrianglesItem;

class BlackDebugSliceSpriteTestState extends flixel.FlxState
{
    override function create()
    {
        super.create();
        
        FlxG.camera.bgColor = 0xFF808080;
        
        final strip = new FlxStrip(50, 50, "assets/images/haxe.png");
        strip.vertices = DrawData.ofArray([0.0, 0.0, 80, 0.0, 40, 40]);
        strip.indices = DrawData.ofArray([0, 1, 2]);
        strip.uvtData = DrawData.ofArray([0, 0, 0, 1, 1, 1.0]);
        add(strip);
        
        FlxG.debugger.drawDebug = true;
    }
}
```
Using this image 
haxe

Here's how it looks when debug draw is false:
Screenshot 2023-08-25 at 5 45 58 PM
true:
Screenshot 2023-08-25 at 5 48 38 PM

UncertainProd:
Fixes #2904, I tested it on html5 and windows

### DIFF
--- a/flixel/graphics/tile/FlxDrawTrianglesItem.hx
+++ b/flixel/graphics/tile/FlxDrawTrianglesItem.hx
@@ -93,7 +93,7 @@ class FlxDrawTrianglesItem extends FlxDrawBaseItem<FlxDrawTrianglesItem>
 		{
 			var gfx:Graphics = camera.debugLayer.graphics;
 			gfx.lineStyle(1, FlxColor.BLUE, 0.5);
-			gfx.drawTriangles(vertices, indices);
+			gfx.drawTriangles(vertices, indices, uvtData);
 		}
 		#end
 


Geokureli:
Sweet! I'll try this out later today

Geokureli:
Thanks! this probably saves me days of debugging

Geokureli:
@UncertainProd do you know what this line actually does? commenting it out seems to change nothing, and I still get bad performance when drawDebug is enabled and FlxStrips are displayed, even in ignoreDrawDebug is enabled for them

UncertainProd:
Honestly, even I'm not too sure. It looks like it's meant to draw some extra stuff when drawDebug is enabled but I don't see anything different being drawn either way

UncertainProd:
Update: I found that if you add the following 2 lines in the `update()` method of PlayState in the FlxClothSprite demo, then it shows you the vertices used in the cloth sprite (which is an FlxStrip)
Code to add:
```haxe
{
    FlxG.debugger.drawDebug = !FlxG.debugger.drawDebug;
}
```

This results in:
debugflxstrip

Turns out FlxClothSprite overrides it's drawDebug method to achieve this. Maybe something like this was the intended effect in FlxStrip but drawTriangles doesn't seem to work with lineStyles like the other openfl drawing commands.

Geokureli:
I may have discovered the source of my performance issues, I talk about it [here](https://github.com/HaxeFlixel/flixel-addons/issues/404)

> Turns out FlxClothSprite overrides it's drawDebugOnCamera method to achieve this. Maybe something like this was the intended effect in FlxStrip but drawTriangles doesn't seem to work with lineStyles like the other openfl drawing commands

yeah maybe they wanted this kind of wireframe drawing on all triangles, rather than having each FlxStrip implementing their own version, but the one in FlxDrawTrianglesItem broke, or only worked in flash and no one noticed. Then, when FlxClothSprite was introduced they unknowingly reinvented the broken wheel?

UncertainProd:
Yeah, I think so

Geokureli:
> Update: I found that if you add the following 2 lines in the `update()` method of PlayState in the FlxClothSprite demo, then it shows you the vertices used in the cloth sprite (which also uses drawTriangles):

does this wireframe show up even if you comment out these lines: https://github.com/UncertainProd/flixel/blob/dev/flixel/graphics/tile/FlxDrawTrianglesItem.hx#L91-L98

UncertainProd:
> does this wireframe show up even if you comment out these lines: https://github.com/UncertainProd/flixel/blob/dev/flixel/graphics/tile/FlxDrawTrianglesItem.hx#L91-L98

Yeah the wireframe shows up even after commenting it out

UncertainProd:
Also, when I replace those lines with the following FlxDrawTrianglesItem, it *does* outline the triangles being drawn:
```haxe
if (FlxG.debugger.drawDebug)
{
			var gfx:Graphics = camera.debugLayer.graphics;
			gfx.lineStyle(1, FlxColor.BLUE, 0.5);
                        // draw a triangle path for each triangle in the drawitem
			var i = 0;
			while (i < indices.length)
			{
				var indexToPoint = (idx:Int) ->
				{
					return [vertices[idx * 2], vertices[idx * 2 + 1]];
				};
				var pt1 = indexToPoint(indices[i]);
				var pt2 = indexToPoint(indices[i + 1]);
				var pt3 = indexToPoint(indices[i + 2]);
				gfx.drawPath(Vector.ofArray([MOVE_TO, LINE_TO, LINE_TO, LINE_TO]), Vector.ofArray([
					pt1[0], pt1[1],
					pt2[0], pt2[1],
					pt3[0], pt3[1],
					pt1[0], pt1[1],
				]));
				i += 3;
			}
}
```

Geokureli:
> Also, when I replace those lines with the following in FlxDrawTrianglesItem, it _does_ outline the triangles being drawn:

Nice! I didn't know global vertex drawing was that easy, however I feel like this feature should be tied to some flag on the FlxStrip, something like `strip.debugDrawWireFrame` and/or `ignoreDrawDebug` rather than relying solely on `FlxG.drawDebug`

UncertainProd:
> I didn't know global vertex drawing was that easy

Yeah although it did take me a bit of searching to understand how that (and drawTriangles in general) worked 😅 

> I feel like this feature should be tied to some flag on the FlxStrip, something like `strip.debugDrawWireFrame` and/or `ignoreDrawDebug` rather than relying solely on `FlxG.drawDebug`

True

Geokureli:
> ... it did take me a bit of searching to understand how that (and drawTriangles in general) worked 😅

I'm in the same boat, thanks for all you've done!

UncertainProd:
No problem!

PR #2910, opened 2023-08-31
FlxGame: make the filters array public

Sword352:
I remember doing something similar to FlxCamera, that would be nice to do the change to FlxGame as well

### DIFF
--- a/flixel/FlxGame.hx
+++ b/flixel/FlxGame.hx
@@ -908,12 +908,12 @@ class FlxGame extends Sprite
 		return Lib.getTimer();
 	}
 
-	#if (!flash || haxe >= "4.2.5") override #end function set_filters(v:Array<BitmapFilter>):Array<BitmapFilter>
+	#if (!flash || haxe > "4.2.5") override #end function set_filters(v:Array<BitmapFilter>):Array<BitmapFilter>
 	{
 		return _filters = v;
 	}
 
-	#if (!flash || haxe >= "4.2.5") override #end function get_filters():Array<BitmapFilter>
+	#if (!flash || haxe > "4.2.5") override #end function get_filters():Array<BitmapFilter>
 	{
 		return _filters;
 	}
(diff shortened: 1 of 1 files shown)


Geokureli:
> home/runner/work/flixel/flixel/flixel/FlxGame.hx:148: characters 13-20 : Redefinition of variable filters in subclass is not allowed. Previously declared at openfl.display.DisplayObject

flixel/FlxGame.hx
```diff
@@ -762,7 +766,7 @@ class FlxGame extends Sprite
 		FlxArrayUtil.clearArray(FlxG.swipes);
 		#end
 
-		filters = filtersEnabled ? _filters : null;
+		filters = filtersEnabled ? filters : null;
```
Geokureli, line 769: This doesn't seem right

flixel/FlxGame.hx
```diff
@@ -142,10 +142,13 @@ class FlxGame extends Sprite
 	 */
 	var _lostFocus:Bool = false;
 
-	/**
-	 * The filters array to be applied to the game.
-	 */
-	var _filters:Array<BitmapFilter>;
+        /**
+         * The filters array to be applied to the game.
+         */
```
Geokureli, line 147: indented with 8 spaces, should use 1 tab

Geokureli (changes requested):
FlxGame extends DisplayObject with has `public var filters(get, set):Array<BitmapFilter>;`
https://github.com/openfl/openfl/blob/develop/src/openfl/display/DisplayObject.hx#L462
the only solution I see is that we should keep _filters, and make the getter set _filters and instead of 
```hx
filters = filtersEnabled ? _filters : null;
```
it should do 
```hx
super.set_filters(filtersEnabled ? _filters : null);
```

Sword352:
Did a few changes, this should be good now

Geokureli:
still failing on flash with haxe 4.2.5

Sword352:
Alright so there is a weird error with flash saying that flash.display.Sprite does not have a filters field, but when adding a conditional block for only compiling the override keyword to non-flash targets, it stil says it needs to be overrided on flash.
How can we solve that issue?

Sword352:
Closing this, will rework on that later after i fix the flash issue

PR #2911, opened 2023-09-01
Remove functions that should be unavailable to macros

### LINKED ISSUE #2893: FlxColor and FlxMath does not work with macros
FlxColor imports FlxMath which imports FlxG which imports a lot of things, perhaps we should make these more self contained in a macro context, since macros may use these tools, or things that use these tools

Cheemsandfriends:
when using `FlxMath` in macros, as specified by #2893, they might not work due to functions that are involved with `FlxG`, in other terms, functions that should be used in a non macro way but as helpers to the day to day coding basis.
This fixes it + adds a compiler conditional around `FlxG` due to it being used only for `FLX_MOUSE`.

Fixes #2893

### DIFF
--- a/flixel/math/FlxMath.hx
+++ b/flixel/math/FlxMath.hx
@@ -1,10 +1,14 @@
 package flixel.math;
 
 import openfl.geom.Rectangle;
+#if !macro
+#if FLX_MOUSE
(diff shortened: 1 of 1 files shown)


Geokureli:
yolo lol

PR #2912, opened 2023-09-04
Add COMMAND (MacOS) to FlxKey

EliteMasterEric:
The Command key (typically represented by the `⌘` symbol) currently has no value in the FlxKey enumeration. This PR adds it.

### DIFF
--- a/flixel/input/keyboard/FlxKey.hx
+++ b/flixel/input/keyboard/FlxKey.hx
@@ -74,6 +74,7 @@ enum abstract FlxKey(Int) from Int to Int
 	var GRAVEACCENT = 192;
 	var CONTROL = 17;
 	var ALT = 18;
+	var COMMAND = 224;
 	var SPACE = 32;
 	var UP = 38;
 	var DOWN = 40;


Cheemsandfriends:
Maybe it should be conditioned with [this key](https://github.com/HaxeFlixel/flixel/blob/5e563adb94b0f70491593e92d455d5995029b4d0/flixel/input/keyboard/FlxKey.hx#L83) instead of making a new one?
That way it's "compatible" both with a Windows keyboard and a Mac keyboard and both have the same key function afaik

EliteMasterEric:
> Maybe it should be conditioned with [this key](https://github.com/HaxeFlixel/flixel/blob/5e563adb94b0f70491593e92d455d5995029b4d0/flixel/input/keyboard/FlxKey.hx#L83) instead of making a new one?
> The name's not of my liking, but that way it's "compatible" both with a Windows keyboard and a Mac keyboard and both have the same key function afaik

The name is COMMAND because that's the name of the key. Same with WINDOWS 

I don't believe there's any conflict between these keys on Mac vs Windows, they're simply excluded from the respective layouts

Cheemsandfriends:
Eh, That's fair
It's just that it can make it convoluted making it separated for several ppl, taking in mind the `WINDOWS` is basically the Windows version of `COMMAND` like the [`SUPER`](https://en.wikipedia.org/wiki/Super_key_(keyboard_button)) key in Linux keyboards (Which technically is the oldest definition of those keys).
since those keys are set for specific keyboards, it would make sense to actually make a whole `SUPER` key and make compiler conditionals to it instead of separating everything, but idk how it would impact this when the `WINDOWS` key is set as a separate key.

Geokureli (changes requested):
You'll need to add COMMAND to [FlxKeyList](https://github.com/HaxeFlixel/flixel/blob/dev/flixel/input/keyboard/FlxKeyList.hx) So that we have `FlxG.keys.pressed.COMMAND`

Geokureli:
also, just tried this and COMMAND is 15 for me, not 224

Geokureli:
closing this as it seems it's an error, if it's not, reopen and give me some example code to try

PR #2913, opened 2023-09-09
Adding a `timeScale` variable to FlxAnimationController.

MAJigsaw77:
This can be helpful for the users who want to base thier animation's speed to a specific value.

### DIFF
--- a/flixel/animation/FlxAnimationController.hx
+++ b/flixel/animation/FlxAnimationController.hx
@@ -10,9 +10,9 @@ using StringTools;
 class FlxAnimationController implements IFlxDestroyable
 {
 	/**
-	 * The speed across every animation that is played.
+	 * How fast or slow should be every animation that is played.
 	 */
-	public static var animationSpeed:Float = 1.0;
+	public static var timeScale:Float = 1.0;
 
 	/**
 	 * Property access for currently playing animation (warning: can be `null`).
@@ -97,7 +97,7 @@ class FlxAnimationController implements IFlxDestroyable
 	{
 		if (_curAnim != null)
 		{
-			_curAnim.update(elapsed * animationSpeed);
+			_curAnim.update(elapsed * timeScale);
 		}
 		else if (_prerotated != null)
 		{
--- a/flixel/animation/FlxAnimationController.hx
+++ b/flixel/animation/FlxAnimationController.hx
@@ -10,6 +10,11 @@ using StringTools;
 class FlxAnimationController implements IFlxDestroyable
 {
 	/**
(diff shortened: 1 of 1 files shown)


flixel/animation/FlxAnimationController.hx
```diff
@@ -9,6 +9,11 @@ using StringTools;
 
 class FlxAnimationController implements IFlxDestroyable
 {
+	/**
+	 * The speed across every animation that is played.
+	 */
+	public static var animationSpeed:Float = 1.0;
```
Geokureli, line 15: `sprite.animation.animationSpeed` is redundant and, imo not very intuitive. I say use the name `timeScale` since this performs a similar function to [FlxG.timeScale](https://api.haxeflixel.com/flixel/FlxG.html#timeScale)

Geokureli: also just realizing this is static, I assumed this would be an instance var, so you can change the rate of a specific sprite's animation, I see very little use for a global animation speed

MAJigsaw77: Ok

Geokureli (changes requested):
```hx
	 /**
	 *  How fast or slow time should pass for this animation controller
	 */
	public var timeScale:Float = 1.0;
```
also move it to the bottom of the current public vars

MAJigsaw77:
> ```hx
> 	 /**
> 	 *  How fast or slow time should pass for this animation controller
> 	 */
> 	public var timeScale:Float = 1.0;
> ```
> 
> also move it to the bottom of the current public vars

Ok

flixel/animation/FlxAnimationController.hx
Geokureli: > also just realizing this is static, I assumed this would be an instance var, so you can change the rate of a specific sprite's animation, I see very little use for a global animation speed

^

Geokureli (approved):

Geokureli:
I'm hesitant to add a global animation timeScale like your original PR had, I'm not a fan of using statics in classes like FlxAnimationController since they are hidden and esoteric, I'm already having regrets for FlxSpite.defaultAntialiasing

Geokureli:
also i'm hopping on a plane in a little bit, this may not get merged until tomorrow, but - assuming CI passes - it's good

MAJigsaw77:
> also i'm hopping on a plane in a little bit, this may not get merged until tomorrow, but - assuming CI passes - it's good

Ok

Geokureli:
Thanks!

PR #2914, opened 2023-09-16
FlxSound: updateTransform crash fix

Sword352:
I've got many crashes because of `_transform` being null, then realized that destroyed `FlxSound`s does an `updateTransform` call. This pull request just make the function return nothing if `_transform` is null.

### DIFF
--- a/flixel/sound/FlxSound.hx
+++ b/flixel/sound/FlxSound.hx
@@ -588,6 +588,9 @@ class FlxSound extends FlxBasic
 	@:allow(flixel.sound.FlxSoundGroup)
 	function updateTransform():Void
 	{
+		if (_transform == null)
+			return;
+		
 		_transform.volume = #if FLX_SOUND_SYSTEM (FlxG.sound.muted ? 0 : 1) * FlxG.sound.volume * #end
 			(group != null ? group.volume : 1) * _volume * _volumeAdjust;
 


MemeHovy:
yea I literally did this for anytime someone had an issue with it, cuz I felt like the transform being null was the issue

Geokureli:
Im weary about null checks like this, they seems simple and harmless, but adding them adds uncertainty to future work and we start having frivolous null checks everywhere by habit.

Im not saying this will be rejected, but i need to look into what is causing this error, can you provide a snippet of code that causes this error. Also im on vacation so i may not get to this right away

Sword352:
I don't really have a code snippet for this, though I can confirm this happens randomly, usually when switching states. The null access might be from [here](https://github.com/HaxeFlixel/flixel/blob/master/flixel/sound/FlxSoundGroup.hx#83), I will do some tests and let you know.

Geokureli:
Usually if you debug your program in vscode it will pause on the erroring line and you can see the call stack that led to it.

It's possible you may have code that changes volume when exiting a state. it's doubtful I'll merge something like this without understanding why it's happening

Sword352:
I don't have any code changing the volume when exiting a state, and this does happens on a lot of projects of mine, though I am doing more investigations at the time of writing this.

Geokureli:
You mentioned using sound groups, can you try using those in a test project and seeing if that reproduces the error?

Sword352:
> You mentioned using sound groups, can you try using those in a test project and seeing if that reproduces the error?

Will do

Sword352:
Gosh I got busy these weeks, will work on this as soon as I can

PR #2919, opened 2023-09-26
Update FlxG.collide() and FlxG.overlap() to take callbacks with FlxObject parameters

marinight:
At the moment, `FlxG.collide` and `FlxG.overlap`'s callbacks (`NotifyCallback` & `ProcessCallback`) accept `Dynamic` arguments, rather than `FlxObject`s and anything that extends it, making the function prone to errors if a user inputs an object that isn't a `FlxObject` or derives from it due to `FlxQuadTree.load` taking in callbacks that have `FlxObject` parameters

This commit fixes that issue by changing the callbacks' parameters to accept `FlxObject`s instead of `Dynamic` objects.

I also assume that this will very slightly optimize things.

### DIFF
--- a/flixel/FlxG.hx
+++ b/flixel/FlxG.hx
@@ -417,8 +417,8 @@ class FlxG
 	 *                            will only be called if `ProcessCallback` returns true for those objects!
 	 * @return  Whether any overlaps were detected.
 	 */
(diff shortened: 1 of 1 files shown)


Geokureli:
I can't remember exactly what, but I tried this recently and noticed issues.

Let's see what CI says

Geokureli:
CI fails as this change disallows functions that take things that extend FlxObject. let's revisit this later, for now I don't think taking dynamic is causing anyone too many issues

PR #2920, opened 2023-09-27
fix camera lerping speeds on lower fps

Rudyrue:
fixes camera lerping being slower than usual on lower framerates

for example if you set the cam lerp speed to like 0.05 it'll be way slower on 15-30 fps than on 60 fps

this makes it consistent between all framerates so it always has the same speed

### DIFF
--- a/flixel/FlxCamera.hx
+++ b/flixel/FlxCamera.hx
@@ -1278,8 +1278,8 @@ class FlxCamera extends FlxBasic
 		}
 		else
 		{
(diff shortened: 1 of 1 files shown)


BlueColorsin (commented):
Peak

Rudyrue:
i don't think this fixes lag spikes slowing down the speed too but i can try and fix that later

Geokureli:
> i don't think this fixes lag spikes slowing down the speed too but i can try and fix that later

I was thinking about this, too, and yeah we don't need to deal with that now. also note: you would only account for lag when `FlxG.fixedTimeStep` is false

PR #2922, opened 2023-09-27
Better framerate independant camera lerping

### LINKED ISSUE #2921: Adjust lerp based on current frame rate
@Rudyrue just merged #2920 fixing a bug in lerp where changing the framerate would inversely affect camera lerp values. We discussed that this only accounts for games with fixed timesteps, so I made [this commit](https://github.com/Geokureli/flixel/commit/d61c01b61a0ba119b09260c4366d47ce97194091) in a branch where it will use `60 * elapsed` rather than the `60 / updateFramerate`, meaning it will work the same with fixed timestep (where elapsed is `1 / updateFrameRate`).

But I have a feeling this is a bad idea. I think the current implementation of adjusting lerp is wrong, and made worse when accounting for lag. Let's imagine our update rate is 60fps and our lerpFocus is 0.5, each frame moves the camera halfway to the target, so in 2 frames (1/30th of a second) it will move 75% towards the target. When the update frame rate is 30fps, the same lerp of 0.5 is adjusted to be 1.0 (because `0.5 * (60/30) = 1.0`). My fear with adjusting lerp based on lag is that it will cause sudden camera snaps on lag spikes, making those spikes even more apparent, 17ms of lag in enough to make the camera snap to the target with most lerp values.

## Options

1. Don't make lerp lag dependent: Just avoid bad behavior by not implementing it for non-fixedTimeStep
2. Don't adjust lerp based on framerate at all: kinda odd that we do this at all, honestly. I assumed lerpFocus wasn't even adjusted, since that how it works in most engines
3. fix the formula so it behaves more predictably with different frame rates: I believe this would need some kinda compound interest formula to develop the "effective followLerp/frame" given that the followLerp value is meant to be at 60fps

Geokureli:
Closes #2921 
this is option 3 outlined in the above issue


NexIsDumb:
So I tested this branch using this [Psych's Engine pr changes](https://github.com/ShadowMario/FNF-PsychEngine/pull/13591), also including [this changes](https://github.com/ShadowMario/FNF-PsychEngine/pull/13433) and everything seemed to work fine
I even tried many FPS options, from 60 to 240, and after even trying to induce some lag (loading large sprites while in the middle of the gameplay) everything was still fine working just as intended

Geokureli:
Awesome!

> seemed to work fine I even tried many FPS options, from 60 to 240

I assume you

Geokureli:
Thanks for checking!

PR #2923, opened 2023-09-29
Untie FlxSoundTray animation speed from framerate, and make it much smoother

CyndaquilDAC:
I noticed that higher framerates make the FlxSoundTray go-away animation go way too fast, and that it's very janky in general. This really bothered me, so I decided to make a quick fix. This also updates some very old comments to fit the standard more.

### DIFF
--- a/flixel/system/ui/FlxSoundTray.hx
+++ b/flixel/system/ui/FlxSoundTray.hx
@@ -110,16 +110,16 @@ class FlxSoundTray extends Sprite
 	/**
 	 * This function just updates the soundtray object.
 	 */
-	public function update(MS:Float):Void
+	public function update(elapsed:Float, MS:Float):Void
 	{
 		// Animate stupid sound tray thing
 		if (_timer > 0)
 		{
-			_timer -= MS / 1000;
+			_timer -= (MS / 1000) * elapsed;
 		}
 		else if (y > -height)
 		{
-			y -= (MS / 1000) * FlxG.height * 2;
+			y -= ((MS / 1000) * FlxG.height * 2) * elapsed;
 
 			if (y <= -height)
 			{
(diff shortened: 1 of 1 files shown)


flixel/system/ui/FlxSoundTray.hx
```diff
@@ -108,18 +108,18 @@ class FlxSoundTray extends Sprite
...
+		// Animate sound tray thing
 		if (_timer > 0)
 		{
-			_timer -= MS / 1000;
+			_timer -= (MS / 1000);
 		}
 		else if (y > -height)
 		{
-			y -= (MS / 1000) * FlxG.height * 2;
+			y -= FlxG.height * FlxG.elapsed;
```
Geokureli, line 122: This function is called directly from an openfl onEnterframe event listener, not Flixel's update loop, meaning it should use MS/1000 rather than FlxG.elapsed. this change does not "untie" the sound tray's animation from frameRate, as the title says, instead it is now using the (potentially) fixed update elapsed value when its actually being called from a variable, or non-fixed elapse value

Geokureli (changes requested):
the comment changes are good, removing * 2 does make it more smooth, these changes are fine but revert FlxG.elapsed to (MS / 1000)

If you believe that I'm wrong and this does make the sound tray less frame rate dependent, then I'll need some evidence

flixel/system/ui/FlxSoundTray.hx
Geokureli: side note: I think using `FlxG.height` here is a typo. I believe it was supposed to be the following, which is much, much smoother
```suggestion
			y -= (MS / 1000) * height * 2;
```

CyndaquilDAC:
Done.

Geokureli (approved):

Geokureli:
Thanks!

PR #2932, opened 2023-10-03
Small improvement to FlxPointer.overlaps

CaptainBaldi:
If the `Camera` argument is null when calling, it simply refers to `ObjectOrGroup`'s first camera.


Geokureli:
With the current setup, this will eventually call [FlxObject.overlapsPoint](https://github.com/HaxeFlixel/flixel/blob/dev/flixel/FlxObject.hx#L1042) if there are any objects in the group. overlapsPoint will check for null and use FlxG.camera, instead. this is typical for nearly all functions that take an optional camera arg. The comment for `FlxPointer.overlaps` even states this.

PR #2937, opened 2023-10-13
Implement `FlxG.animationTimeScale`

### LINKED ISSUE #2936: Making `FlxAnimationController.timeScale` static / adding a static variable for global animation speed.
I've noticed that `timeScale` is a newer addition to the FlxAnimationController class. It is rather useful, however, I would personally see more use to it if it were static.

I've noticed that when created, #2913 initially had initially defined this field as static, but was changed due to not being useful in enough scenario's.

While I agree it is certainly not something that would be extremely common to use, I personally in a flixel project of mine have made an alias for `FlxAnimationController` that makes this field static, I did this to make a mode in the project that involves slowing down and speeding it back up, including animations.

Another example is Psych Engine, which also has this sort of implementation.

If making this field static isn't viable still, what about making a new static variable, something like `globalTimeScale`, and assign it a default value of say -1?

If you were to change this value to say something like 0.5, it would then affect all `FlxAnimationController`'s.

CaptainBaldi:
A more convenient way of setting global animation speeds.

Also adjusted `FlxG.timeScale`'s initialization value to match the comment.

Fixes #2936

### DIFF
--- a/flixel/FlxG.hx
+++ b/flixel/FlxG.hx
@@ -80,7 +80,12 @@ class FlxG
 	/**
 	 * How fast or slow time should pass in the game; default is `1.0`.
 	 */
-	public static var timeScale:Float = 1;
+	public static var timeScale:Float = 1.0;
+	
+	/**
+	 * How fast or slow animations should pass in the game; default is `1.0`.
+	 */
+	public static var animationTimeScale:Float = 1.0;
 
 	/**
 	 * How many times the quad tree should divide the world on each axis.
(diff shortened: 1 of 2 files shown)


flixel/FlxG.hx
```diff
@@ -80,7 +80,12 @@ class FlxG
 	/**
 	 * How fast or slow time should pass in the game; default is `1.0`.
 	 */
-	public static var timeScale:Float = 1;
+	public static var timeScale:Float = 1.0;
```
Geokureli, line 83: FYI 1.0 and 1 are functionally equivalent when typed to Float

CaptainBaldi: I figured so, however it was a change done mostly for comment accuracy, and also due to `FlxG.reset` setting these values to `1.0`.

Either way, should I revert this change?

Geokureli: no need to revert, just saying that this change doesn't make it "more accurate" in any meaningful way

Geokureli:
Thanks!

PR #2938, opened 2023-10-14
Various FlxBar fixes

Starmapo:
- Fixed `frontFrames` graphic not changing its `useCount` (for example, if you added an FlxBar and then called `FlxG.bitmap.clearUnused()`, the front graphic would be destroyed and the game would crash when trying to draw it)
- Render-method-specific variables are now destroyed correctly. It used to try to destroy variables that are only used in the opposite render method
- `_frontFrame` is now destroyed when being set to `null`
- Fixed `_frontFrame` documentation referring to it as an FlxSprite
- Removed redundant `positionOffset = null` code (the variable is already set to `null` earlier in the function)

### DIFF
--- a/flixel/ui/FlxBar.hx
+++ b/flixel/ui/FlxBar.hx
@@ -128,7 +128,7 @@ class FlxBar extends FlxSprite
 	var _fillHorizontal:Bool;
 
 	/**
-	 * FlxSprite which is used for rendering front graphics of bar (showing value) in tile render mode.
+	 * FlxFrame which is used for rendering front graphics of bar (showing value) in tile render mode.
 	 */
 	var _frontFrame:FlxFrame;
 
@@ -198,9 +198,9 @@ class FlxBar extends FlxSprite
 	{
 		positionOffset = FlxDestroyUtil.put(positionOffset);
 
-		if (FlxG.renderBlit)
+		if (FlxG.renderTile)
 		{
-			_frontFrame = null;
+			frontFrames = null;
 			_filledFlxRect = FlxDestroyUtil.put(_filledFlxRect);
 		}
 		else
@@ -214,7 +214,6 @@ class FlxBar extends FlxSprite
 		_filledBarPoint = null;
 
 		parent = null;
-		positionOffset = null;
 		emptyCallback = null;
 		filledCallback = null;
 
@@ -984,8 +983,14 @@ class FlxBar extends FlxSprite
 	{
 		if (FlxG.renderTile)
 		{
+			if (value != null)
+				value.parent.incrementUseCount();
+				
+			if (frontFrames != null)
+				frontFrames.parent.decrementUseCount();
+
 			frontFrames = value;
-			_frontFrame = (value != null) ? value.frame.copyTo(_frontFrame) : null;
+			_frontFrame = (value != null) ? value.frame.copyTo(_frontFrame) : FlxDestroyUtil.destroy(_frontFrame);
 		}
 		else
 		{


Geokureli:
Can you create an example state that illustrates the issues in the old code that this PR fixes

Starmapo:
> Can you create an example state that illustrates the issues in the old code that this PR fixes

Sure:

```haxe
package;

import flixel.FlxG;
import flixel.FlxState;
import flixel.ui.FlxBar;

{
	var bar:FlxBar;

	{
		bar = new FlxBar();
		bar.value = 50;
		bar.screenCenter();
		add(bar);

		super.create();
	}

	@:access(flixel.ui.FlxBar)
	{
		if (bar != null)
		{
			if (FlxG.keys.justReleased.ONE)
			{
				// With the old code, the game will now crash when trying to draw the FlxBar,
				// as the front/filled graphic has been incorrectly cleared
				FlxG.bitmap.clearUnused();

				FlxG.log.add('Cleared unused graphics.');
			}
			else if (FlxG.keys.justReleased.TWO)
			{
				FlxG.log.add('Destroying the FlxBar...');

				final _frontFrame = bar._frontFrame;

				bar.destroy();

				// With the old code, these variables will not have been set to `null`
				if (FlxG.renderTile)
				{
					FlxG.log.add('frontFrames is null: ${bar.frontFrames == null}');
				}
				else
				{
					FlxG.log.add('_emptyBar is null: ${bar._emptyBar == null}');
				}

				// With the old code, _frontFrame will not have been destroyed
				FlxG.log.add('_frontFrame was destroyed: ${_frontFrame.sourceSize == null}');

				remove(bar, true);
				bar = null;
			}
		}

		super.update(elapsed);
	}
}
```

The last change (removed redundant `positionOffset = null` code) doesn't fix any bugs, just removes a redundant `null` set that already happens prior. [(L199)](https://github.com/HaxeFlixel/flixel/blob/6a723096b130f76109c7f0411abfbf63ca634a49/flixel/ui/FlxBar.hx#L199)

Geokureli:
nice!

flixel/ui/FlxBar.hx
```diff
@@ -198,9 +198,9 @@ class FlxBar extends FlxSprite
 	{
 		positionOffset = FlxDestroyUtil.put(positionOffset);
 
-		if (FlxG.renderBlit)
+		if (FlxG.renderTile)
 		{
-			_frontFrame = null;
+			frontFrames = null;
```
Geokureli, line 203: Can you shed some light on this change? was `renderBlit` a mistake, or was the comment above wrong? why frontFrames instead of _frontFrames?

flixel/ui/FlxBar.hx
```diff
@@ -984,8 +983,14 @@ class FlxBar extends FlxSprite
...
 		{
+			if (value != null)
+				value.parent.incrementUseCount();
+				
+			if (frontFrames != null)
+				frontFrames.parent.decrementUseCount();
+
 			frontFrames = value;
-			_frontFrame = (value != null) ? value.frame.copyTo(_frontFrame) : null;
+			_frontFrame = (value != null) ? value.frame.copyTo(_frontFrame) : FlxDestroyUtil.destroy(_frontFrame);
```
Geokureli, line 993: can you elaborate on why this was changed to the underscored var

flixel/ui/FlxBar.hx
Starmapo: `_frontFrame` and `_filledFlxRect` are only used in tile render mode, but the current code mistakenly tries to destroy these when the game is using blit render mode, and vice versa for the variables only used in blit mode. Simply changing line 201 to `renderTile` fixes it.

I changed line 203 to `frontFrames` as that value is left without being set to null in the current code, and `set_frontFrames` will also null out `_frontFrame`.

Geokureli:
Thanks!

PR #2942, opened 2023-10-18
Regen graphic in `FlxText.getScreenBounds()`

T1mL3arn:
It is possible to get wrong screen bounds if the bounds are read just after dimensions were changed

```haxe
var text = new FlxText(10, 10, 0, 'hello', 16);
text.fieldWidth = 100;
text.fieldHeight = 100;
add(text);
trace(text.getScreenBounds());
```

actual
```
(x: 10 | y: 10 | w: 52 | h: 24)
```

expected
```
(x: 10 | y: 10 | w: 100 | h: 100)
```

### Note

This pr changes `FlxText`, but maybe `FlxSprite` should be changed? Other similar methods (`getRotatedBounds`, `inWorldBounds`) use `width`/`height` (getters that trigger actual recalculation of size) while `getScreenBounds` uses `frameWidth`:

https://github.com/HaxeFlixel/flixel/blob/a33b47d51691b972b16b0566c7e0c2c13dd70f92/flixel/FlxSprite.hx#L1302

### DIFF
--- a/flixel/text/FlxText.hx
+++ b/flixel/text/FlxText.hx
@@ -1,5 +1,6 @@
 package flixel.text;
 
+import flixel.math.FlxRect;
 import openfl.display.BitmapData;
 import openfl.geom.ColorTransform;
 import openfl.text.TextField;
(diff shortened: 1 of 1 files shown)


Geokureli:
couple changes to address some pet-peeves, no reason to ever include public on override, I know other functions already do this, but we'll address those later, I should change the codeclimate for this, too

Geokureli:
Thanks!
Note: nightly builds are failing right now, no cause for concern

PR #2946, opened 2023-10-19
Additional glyphs

SeiferTim:
Adding `appendGlyphs` function to `FlxBitmapFont`.

See [https://github.com/AxolStudio/HFAdditionalGlyphsDemo](https://github.com/AxolStudio/HFAdditionalGlyphsDemo) for an example and details.

### DIFF
--- a/flixel/graphics/frames/FlxBitmapFont.hx
+++ b/flixel/graphics/frames/FlxBitmapFont.hx
@@ -80,6 +80,17 @@ class FlxBitmapFont extends FlxFramesCollection
 	var frame:FlxFrame;
 
 	/**
+	 * Lookup tables for additional glyphs characters.
+	 */
+	public var lookupTable:Map<String, String> = [];
+	public var revLookupTable:Map<String, String> = [];
+
+	/**
+	 * Special characters start index.
+	 */
+	public var specialStart:Int = 500;
+
+	/**
 	 * Creates a new bitmap font using specified bitmap data and letter input.
 	 */
 	function new(frame:FlxFrame, ?border:FlxPoint)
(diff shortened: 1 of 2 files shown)


Geokureli:
truthfully, while I want this feature of adding glyphs to bitmap text, I don't know if I want this specific glyph parsing method to be the defacto method of using it. I can't explore right now, but I suspect there are more generic ways to implement this, I'm afraid that adding {foo} parsing may mess or conflict with other text parsing systems, I also think there might be more generic ways to add glyph frames to existing fonts, and in general I want to add ways to split up fonts into multiple atlases for various reasons. once that is done, I think an adhoc system like this will be easy to implement, and will make a good external tool

flixel/graphics/frames/FlxBitmapFont.hx
```diff
@@ -635,4 +646,51 @@ class FlxBitmapFont extends FlxFramesCollection
...
+	{
+		specialStart = StartingCharacter;
+		
+		var offY:Int = parent.height;
+		
+		var tmpBmp:BitmapData = new BitmapData(Math.ceil(Math.max(parent.width, Frames.parent.width)), parent.height + Frames.parent.height, true, 0x00000000);
+		tmpBmp.copyPixels(parent.bitmap, parent.bitmap.rect, point);
+		tmpBmp.copyPixels(Frames.parent.bitmap, Frames.parent.bitmap.rect, new Point(0, parent.height));
+		
+		parent = FlxGraphic.fromBitmapData(tmpBmp, parent.key + "+" + Frames.parent.key, true);
```
Geokureli, line 666: this is a big part of it, with multiatlases we can simply take frames from one atlas and add them to another without needing to create a new bitmap containing every frame

Also line 666 breals flash CI, by skipping the `unique` arg

flixel/graphics/frames/FlxBitmapFont.hx
```diff
@@ -79,6 +79,17 @@ class FlxBitmapFont extends FlxFramesCollection
...
+	/**
+	 * Lookup tables for additional glyphs characters.
+	 */
+	public var lookupTable:Map<String, String> = [];
+	public var revLookupTable:Map<String, String> = [];
+
+	/**
+	 * Special characters start index.
+	 */
+	public var specialStart:Int = 500;
```
Geokureli, line 91: I also think this can be implemented without needing a special start, and an ability to parse tokens to special characters seems helpful on it's own wether that maps to some large unicode index, or not

PR #2959, opened 2023-11-10
Add `useGlobalTimeScale` to `FlxAnimationController`

crowplexus:
this is to allow users to ignore the global time scale in `FlxG` if needed

### DIFF
--- a/flixel/animation/FlxAnimationController.hx
+++ b/flixel/animation/FlxAnimationController.hx
@@ -71,6 +71,11 @@ class FlxAnimationController implements IFlxDestroyable
 	public var timeScale:Float = 1.0;
 
 	/**
+	 * Whether the animation is affected by `FlxG.animationTimeScale` or not.
+	**/
+	public var useGlobalTimeScale:Bool = true;
+
+	/**
 	 * Internal, reference to owner sprite.
 	 */
 	var _sprite:FlxSprite;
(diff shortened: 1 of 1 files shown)


Geokureli:
I'm hesitant to add this as it's super ad-hoc to a rare use case and there's already ways to implement this into the current flixel system in 2 noteworthy ways:
1. you can set an idividual sprite's timescale to the inverse of the global timescale, with `sprite.animation.timeScale = 1 / FlxG.animationTimeScale * myDesiredSpriteTimeScale;`
2. you can extend FlxanimationController and implement this feature for any desired sprite, example:
```hx
class CustomAnimationController extends flixel.animation.FlxAnimationController
{
    
    /**
     * Whether the animation's `timeScale` is affected by `FlxG.animationTimeScale` or not.
    **/
    public var ignoreGlobalTimeScale = false;
    
    {
        if (_curAnim != null)
        {
            var scaledElapsed = elapsed * timeScale;
            
            if (ignoreGlobalTimeScale == false)
                scaledElapsed *= FlxG.animationTimeScale;
            
            _curAnim.update(scaledElapsed);
        }
        else if (_prerotated != null)
        {
            _prerotated.angle = _sprite.angle;
        }
    }
}

//elsewhere
var spr = new FlxSprite("assets/images/gf.png);
spr.animation = new CustomAnimationController();
spr.animation.useGlobalTimeScale = false;
spr.frames = Path.loadSparrowAtlas(...);// call this after setting animation
```

crowplexus:
honestly I believe making a custom animation controller seals the deal pretty well, this implementation i made is mostly just kind of a hacky workaround if anything lol!

PR #2968, opened 2023-11-16
Implement `wasDestroyed` check on FlxGraphic

EliteMasterEric:
When attempting to render a destroyed graphic (which is wrong and the library user needs to diagnose the issue), determining which sprite was the cause of the issue is difficult, due to the error taking place long after the draw is queued. This change ensures that the error is thrown WHEN the sprite's `draw()` call is performed rather than after; this allows developers to step up to `FlxSprite.draw()` when debugging and view the affected sprite in the local variable scope.

### DIFF
--- a/flixel/FlxCamera.hx
+++ b/flixel/FlxCamera.hx
@@ -635,6 +635,11 @@ class FlxCamera extends FlxBasic
 			itemToReturn = new FlxDrawItem();
 		}
 
+		if (graphic.wasDestroyed) {
+			// Throw an exception here, because throwing in `FlxDrawQuadsItem.render()` is not helpful.
+			throw 'Attempted to queue invalid FlxDrawItem, did you dispose of a cached sprite?';
+		}
+
 		itemToReturn.graphics = graphic;
 		itemToReturn.antialiasing = smooth;
 		itemToReturn.colored = colored;
--- a/flixel/graphics/FlxGraphic.hx
+++ b/flixel/graphics/FlxGraphic.hx
@@ -344,6 +344,16 @@ class FlxGraphic implements IFlxDestroyable
 	public var atlasFrames(get, never):FlxAtlasFrames;
 
 	/**
+	 * This value is `true` if this graphic has been disposed.
+	 * Attempting to render a disposed graphic will cause an error.
+	 */
+	public var wasDestroyed(get, never):Bool;
+
+	function get_wasDestroyed():Bool {
+		return shader == null;
+	}
+
+	/**
 	 * Storage for all available frame collection of all types for this graphic object.
 	 */
 	var frameCollections:Map<FlxFrameCollectionType, Array<Dynamic>>;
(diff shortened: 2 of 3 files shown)


flixel/FlxCamera.hx
```diff
@@ -635,6 +635,11 @@ class FlxCamera extends FlxBasic
 			itemToReturn = new FlxDrawItem();
 		}
 
+		if (graphic.wasDestroyed) {
+			// Throw an exception here, because throwing in `FlxDrawQuadsItem.render()` is not helpful.
+			throw 'Attempted to queue invalid FlxDrawItem, did you dispose of a cached sprite?';
```
Geokureli, line 640: i want to avoid the word dispose, and just use the word "destroy" consistently. since that's the function they are likely calling

flixel/FlxCamera.hx
```diff
@@ -635,6 +635,11 @@ class FlxCamera extends FlxBasic
 			itemToReturn = new FlxDrawItem();
 		}
 
+		if (graphic.wasDestroyed) {
```
Geokureli, line 638: no cuddling

flixel/graphics/FlxGraphic.hx
```diff
@@ -343,6 +343,16 @@ class FlxGraphic implements IFlxDestroyable
 	 */
 	public var atlasFrames(get, never):FlxAtlasFrames;
 
+	/**
+	 * This value is `true` if this graphic has been disposed.
+	 * Attempting to render a disposed graphic will cause an error.
+	 */
+	public var wasDestroyed(get, never):Bool;
+
+	function get_wasDestroyed():Bool {
```
Geokureli, line 352: no cuddling

Geokureli: also let's inline this

flixel/graphics/tile/FlxDrawQuadsItem.hx
```diff
@@ -115,6 +115,12 @@ class FlxDrawQuadsItem extends FlxDrawBaseItem<FlxDrawQuadsItem>
 			return;
 
 		var shader = shader != null ? shader : graphics.shader;
+
+		if (shader == null) {
```
Geokureli, line 119: no cuddling

flixel/graphics/tile/FlxDrawQuadsItem.hx
```diff
@@ -115,6 +115,12 @@ class FlxDrawQuadsItem extends FlxDrawBaseItem<FlxDrawQuadsItem>
...
+			// If this gets called, the check for `graphic.wasDestroyed` in FlxCamera failed.
+			throw 'Attempted to render invalid FlxDrawItem, did you dispose of a cached sprite?';
```
Geokureli, line 121: use the word "destroy"

Geokureli (changes requested):
see comments below also:

I'm wondering if we should do something even sooner, perhaps we should check in `draw()`, check whether the sprite was destroyed or the graphic, use `FlxG.log.error`, and then set the sprite's visible to false to prevent repeating errors?

PR #2969, opened 2023-11-20
Make `FlxMouse._cursor` public

CaptainBaldi:
Does exactly what the title says

### DIFF
--- a/flixel/input/mouse/FlxMouse.hx
+++ b/flixel/input/mouse/FlxMouse.hx
@@ -197,6 +197,7 @@ class FlxMouse extends FlxPointer implements IFlxInputManager
 	var _rightButton:FlxMouseButton;
 	#end
 
+	@:deprecated("_cursor is deprecated, use the new, public cursor, instead")
     var _cursor(get, set):Bitmap;
 	var _cursorBitmapData:BitmapData;
 	var _wheelUsed:Bool = false;
--- a/flixel/input/mouse/FlxMouse.hx
+++ b/flixel/input/mouse/FlxMouse.hx
@@ -197,7 +197,7 @@ class FlxMouse extends FlxPointer implements IFlxInputManager
 	var _rightButton:FlxMouseButton;
 	#end
 
-    var _cursor(get, set):Bitmap = null;
+    var _cursor(get, set):Bitmap;
 	var _cursorBitmapData:BitmapData;
 	var _wheelUsed:Bool = false;
 	var _visibleWhenFocusLost:Bool = true;
--- a/flixel/input/mouse/FlxMouse.hx
+++ b/flixel/input/mouse/FlxMouse.hx
@@ -51,6 +51,11 @@ class FlxMouse extends FlxPointer implements IFlxInputManager
 	 * sits at the right "height". Not used on flash with the native cursor API.
 	 */
 	public var cursorContainer(default, null):Sprite;
+	
+	/**
+	 * This is just a reference to the current cursor image, if there is one.
+	 */
+	public var cursor:Bitmap;
 
 	/**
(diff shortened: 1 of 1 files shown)


Geokureli:
Please share an example of what you're trying to do, you should really never need to use the `Bitmap` class, and if you do you can easily get a bitmapData from it

CaptainBaldi:
> Please share an example of what you're trying to do, you should really never need to use the `Bitmap` class, and if you do you can easily get a bitmapData from it

I'm trying to change the bitmap's smoothing property. `_cursor` is not a public variable so it's just convenient to be able to pass in my own `Bitmap` so I can specify `smoothing` in the constructor. Something like:
```
var _customCursor:Bitmap = new Bitmap(BitmapData.fromFile('assets/images/CustomCursor.png'), null, true);
FlxG.mouse.load(_customCursor, 2);
```

Geokureli:
would it be better to add a smoothing arg, or expose the bitmap instance, instead

CaptainBaldi:
hmm, I think i'll go with making `_cursor` public, just incase you want to mess with `Bitmap.pixelSnapping` or other properties.

flixel/input/mouse/FlxMouse.hx
```diff
@@ -51,6 +51,11 @@ class FlxMouse extends FlxPointer implements IFlxInputManager
 	 * sits at the right "height". Not used on flash with the native cursor API.
 	 */
 	public var cursorContainer(default, null):Sprite;
+	
+	/**
+	 * This is just a reference to the current cursor image, if there is one.
+	 */
+	public var cursor:Bitmap;
```
Geokureli, line 58: lets make it read-only, you'll still be able to set properties. Also should mark it as nullable which is something I want to start doing
```suggestion
	public var cursor(default, null):Null<Bitmap>;
```

flixel/input/mouse/FlxMouse.hx
```diff
@@ -51,6 +51,11 @@ class FlxMouse extends FlxPointer implements IFlxInputManager
 	 * sits at the right "height". Not used on flash with the native cursor API.
 	 */
 	public var cursorContainer(default, null):Sprite;
+	
+	/**
+	 * This is just a reference to the current cursor image, if there is one.
```
Geokureli, line 56:
```suggestion
	 * The current cursor bitmap, if there is one. To set, use the `load` function
```

flixel/input/mouse/FlxMouse.hx
```diff
@@ -192,11 +197,6 @@ class FlxMouse extends FlxPointer implements IFlxInputManager
 	var _rightButton:FlxMouseButton;
 	#end
 
-	/**
-	 * This is just a reference to the current cursor image, if there is one.
-	 */
-	var _cursor:Bitmap = null;
```
Geokureli, line 198: we need to keep this for backwards compatibility, in case anyone is using `@:privateAccess`
```hx
var _cursor(get, set):Bitmap = null;

inline function get__cursor()
{
    return cursor;
}

inline function set__cursor(value:Bitmap)
{
    return cursor = value;
}
```

CaptainBaldi:
Alright, should be good to go

flixel/input/mouse/FlxMouse.hx
Geokureli: sorry,  need to remove the ` = null`

Geokureli:
there's one mistake, from the code I gave, commented above

CaptainBaldi:
I see

Geokureli:
Added a deprecation warning to _cursor, which I forgot. Nightly CI jobs are failing in every branch, so don't worry about that, will merge in a bit

PR #2970, opened 2023-11-22
Add `@:from` metadata to `FlxColor#fromString`

Sword352:
With that addition it's gonna be possible to manipulate `FlxColor`s with strings.

```haxe
// create an FlxColor from a string
var col:FlxColor = "white";
col = "black";

// changing `FlxColor` fields with strings
mySprite.color = "red";
mySubState.bgColor = "green";
```


Sword352:
- [ ]

PR #2974, opened 2023-11-27
Add graphic.isDestroyed, prevent render crashes

Geokureli:
Closes #2968

adds graphic.isDestroyed, detects when a destroyed graphic is being drawn, logs an error and displays the flixel logo, instead.

Note to @EliteMasterEric:
- I went with `isDestroyed` to match `isLoaded` and `isDumped`.
- I also use `FlxG.log.error` instead of throwing an error. I try to throw when I want the game to pause right when they did something wrong, not later, when I discover that something previously went wrong

### DIFF
--- a/flixel/FlxSprite.hx
+++ b/flixel/FlxSprite.hx
@@ -774,6 +774,16 @@ class FlxSprite extends FlxObject
 	{
 		if (_frame == null)
 			loadGraphic("flixel/images/logo/default.png");
+		else if (graphic != null && graphic.isDestroyed)
+		{
+			// switch graphic but log and preserve size
+			final width = this.width;
+			final height = this.height;
+			FlxG.log.warn('Cannot render a destroyed graphic, the placeholder image will be used instead');
+			loadGraphic("flixel/images/logo/default.png");
+			this.width = width;
+			this.height = height;
+		}
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


EliteMasterEric:
If you can successfully revert to rendering the Flixel logo instead of crashing, using Flx.error is way better than throwing, since gracefully handling errors is preferred whenever possible

Geokureli:
> using Flx.error is way better than throwing, since gracefully handling errors is preferred whenever possible

I mostly agree, there are some cases where I may want to be extremely loud about a dev mistake, especially if I can point to the exact offending line. But this issue is hard to predict, we can only detect late symptoms, so, yes

Geokureli:
let me know if this works for you

here's my test case for posterity
```hx
package states;

import flixel.FlxG;
import flixel.FlxBasic;
import flixel.FlxSprite;

class DestroyedSpriteTestState extends flixel.FlxState
{
	var sprite1 = new FlxSprite(100, 100, "assets/images/haxe.png");
	var sprite2:FlxSprite;
	
	override function create()
	{
		super.create();
		
		add(new FlxSprite(0, 100));
		add(sprite2 = new FlxSprite(100, 0, "assets/images/haxe.png"));
		add(sprite1);
	}
	
	{
			sprite2.destroy();
		
		if (FlxG.keys.justPressed.ENTER)
			FlxG.switchState(new DestroyedSpriteTestState());
	}
	
	override function draw()
	{
		super.draw();
	}
}
```

PR #2980, opened 2023-12-11
defaultMoves (formerly movesDefault) variable

VMan-2002:
Simple static variable to change the default value of `moves`

### DIFF
--- a/flixel/FlxObject.hx
+++ b/flixel/FlxObject.hx
@@ -87,6 +87,11 @@ class FlxObject extends FlxBasic
 	public static var SEPARATE_BIAS:Float = 4;
 
 	/**
+	 * The default `moves` value of all future `FlxObjects`
+	 */
+	public static var defaultMoves:Bool = true;
+
+	/**
 	 * Generic value for "left". Used by `facing`, `allowCollisions`, and `touching`.
 	 * Note: This exists for backwards compatibility, prefer using `FlxDirectionFlags.LEFT` directly.
 	 */
@@ -607,11 +612,6 @@ class FlxObject extends FlxBasic
 	 * decreases their rendering performance by a factor of ~10x when using blitting!
 	 */
 	public var angle(default, set):Float = 0;
-
-	/**
-	 * The default `moves` value of all future `FlxObjects`
-	 */
-	public static var defaultMoves:Bool = true;
 
 	/**
 	 * Set this to `false` if you want to skip the automatic motion/movement stuff (see `updateMotion()`).
--- a/flixel/FlxObject.hx
+++ b/flixel/FlxObject.hx
@@ -88,6 +88,7 @@ class FlxObject extends FlxBasic
 
 	/**
 	 * The default `moves` value of all future `FlxObjects`
+	 * @since 5.6.0
 	 */
 	public static var defaultMoves:Bool = true;
 
--- a/flixel/FlxObject.hx
+++ b/flixel/FlxObject.hx
@@ -87,7 +87,8 @@ class FlxObject extends FlxBasic
 	public static var SEPARATE_BIAS:Float = 4;
 
 	/**
-	 * The default `moves` value of all future `FlxObjects`
+	 * The default `moves` value of all future `FlxObjects` and `FlxSprites`
+	 * Note: Has no effect on `FlxTexts`, `FlxTilemaps` and `FlxTileBlocks`
 	 * @since 5.6.0
 	 */
 	public static var defaultMoves:Bool = true;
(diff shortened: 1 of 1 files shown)


MemeHovy:
this is already possible by just doing `sprite.moves = false;` what you did was a useless addition

flixel/FlxObject.hx
```diff
@@ -612,7 +612,12 @@ class FlxObject extends FlxBasic
 	 * Set this to `false` if you want to skip the automatic motion/movement stuff (see `updateMotion()`).
 	 * `FlxObject` and `FlxSprite` default to `true`. `FlxText`, `FlxTileblock` and `FlxTilemap` default to `false`.
 	 */
-	public var moves(default, set):Bool = true;
+	public var moves(default, set):Bool = movesDefault;
+
+	/**
+	 * The default value of the `moves` variable.
+	 */
+	public static var movesDefault:Bool = true;
```
Geokureli, line 620: Couple notes:
- static vars should come before instance vars. this should be moved so it is just after [defaultPixelPerfectPositioning](https://github.com/HaxeFlixel/flixel/blob/dev/flixel/FlxObject.hx#L80) which is on line 80
- the convention for these types of vars is `defaultMoves` not `movesDefault`
- lets change the comment to ```The default `moves` value of all future `FlxObjects` ```
    - no period on single sentence comments
    - avoid the word variable (i know `defaultPixelPerfectPosition` says "var" thats also bad
- I'm trying to move a lot of existing static vars to FlxG. I find these to be esoteric, hard to find and a little confusing. That said, there's little harm in adding one more and this could actually retroactively improve performance dramatically in many games

MemeHovy:
> > this is just useless, it's already got a default value, why need a variable for that? and you can just use `sprite.moves = false` to change that or extend FlxObject to do that too
> 
> because this is a static var that determines the default value of every FlxObject created, similar to [FlxSprite.defaultAntiAliasing](https://github.com/HaxeFlixel/flixel/blob/dev/flixel/FlxSprite.hx#L136)

I mean, that’s great but this probably should be in FlxG like you suggested earlier since it would make it in line with the other code, compared to defaultAntialiasing

MemeHovy:
> Having statics in FlxObject and FlxSprite is actually the current convention, in the future these might move to FlxG, and when that happens we can move this too. But its gonna take some planning before that happens so lets just make a static in FlxObject for now

sounds good

Geokureli (changes requested):
almost all good, but one point still remains

> Few notes:
> - static fields should come before instance fields. this should be moved so it is just after [defaultPixelPerfectPositioning](https://github.com/HaxeFlixel/flixel/blob/dev/flixel/FlxObject.hx#L80) which is on line 80

flixel/FlxObject.hx
```diff
@@ -86,6 +86,11 @@ class FlxObject extends FlxBasic
 	 */
 	public static var SEPARATE_BIAS:Float = 4;
 
+	/**
+	 * The default `moves` value of all future `FlxObjects`
+	 */
+	public static var defaultMoves:Bool = true;
+
```
Geokureli, line 93:
```suggestion
	/**
	 * The default `moves` value of all future `FlxObjects`
	 * @since 5.6.0
	 */
	public static var defaultMoves:Bool = true;

```

Geokureli:
I notice this comment on `moves`:
> `FlxObject` and `FlxSprite` default to `true`. `FlxText`, `FlxTileblock` and `FlxTilemap` default to `false`.

can you confirm that FlxText, FlxTilemap and FlxTileBlock default to false even when defaultMoves is true? In fact we should probably have unit tests that verify this

Geokureli:
in the future, please do not make PRs to flixel using your fork's dev branch, create a new feature branch, so it's easier for others to test and make changes to it

Geokureli:
tested this with the following code, it works fine
```hx
package states;

import flixel.FlxG;
import flixel.FlxObject;
import flixel.FlxSprite;
import flixel.text.FlxText;
import flixel.tile.FlxTilemap;
import flixel.tile.FlxTileblock;

class DefaultMovesTestState2980 extends flixel.FlxState
{
    override function create()
    {
        super.create();
        
        assertMoves(new FlxObject());
        assertMoves(new FlxSprite());
        assertNotMoves(new FlxText());
        assertNotMoves(new FlxTilemap());
        assertNotMoves(new FlxTileblock(0, 0, 1, 1));
        
        FlxObject.defaultMoves = false;
        assertNotMoves(new FlxObject());
        assertNotMoves(new FlxSprite());
        assertNotMoves(new FlxText());
        assertNotMoves(new FlxTilemap());
        assertNotMoves(new FlxTileblock(0, 0, 1, 1));
        
        FlxObject.defaultMoves = true;
        assertMoves(new FlxObject());
        assertMoves(new FlxSprite());
        assertNotMoves(new FlxText());
        assertNotMoves(new FlxTilemap());
        assertNotMoves(new FlxTileblock(0, 0, 1, 1));
    }
    
    function assertMoves(obj:FlxObject)
    {
        if (obj.moves == false)
            throw 'Expected obj: $obj, moves:FALSE, got TRUE';
    }
    
    function assertNotMoves(obj:FlxObject)
    {
        if (obj.moves)
            throw 'Expected obj: $obj, moves:FALSE, got TRUE';
    }
}
```

PR #2987, opened 2023-12-15
Fix plus keybind on Mac

NeeEoo:
This allows pressing + on Mac. Might only work for Swedish Mac keyboards since i don't have a english keyboard

### DIFF
--- a/flixel/input/keyboard/FlxKeyboard.hx
+++ b/flixel/input/keyboard/FlxKeyboard.hx
@@ -51,6 +51,7 @@ class FlxKeyboard extends FlxKeyManager<FlxKey, FlxKeyList>
 		_nativeCorrection.set("123_222", FlxKey.LBRACKET);
 		_nativeCorrection.set("125_187", FlxKey.RBRACKET);
 		_nativeCorrection.set("126_233", FlxKey.GRAVEACCENT);
+		#if mac _nativeCorrection.set("0_43", FlxKey.PLUS); #end
 
 		_nativeCorrection.set("0_80", FlxKey.F1);
 		_nativeCorrection.set("0_81", FlxKey.F2);


NeeEoo:
I don't see why it would be closed, the ascii code 43 means +

Geokureli:
> FlxKeys do not map to characters, they map to keyboard button locations which are labelled based on their location in the english keyboard layout

NeeEoo:
What mac do you have?

Geokureli:
US English Layout
image

https://keyshorts.com/blogs/blog/37615873-how-to-identify-macbook-keyboard-localization

NeeEoo:
I mean mac model, because it worked when i had a 2017 model, but not my 2023 model

Geokureli:
I do have a 2017, but then would this change break plus on 2017s

NeeEoo:
It wouldnt break, pressing plus there would return a different keycode, which would return the key 61_187 which has no nativeCorrection mapping which then makes it point to FlxKey.PLUS (187) due to the code below

```hx
var code = _nativeCorrection.get(e.charCode + "_" + e.keyCode);
return (code == null) ? e.keyCode : code;
```

this pr is only if the key returns 0_43

Geokureli:
aight let's test this, sorry for the misunderstanding

NeeEoo:
I just placed a trace here to get the value

Skärmavbild 2023-12-15 kl  18 08 44

`trace(e.charCode + "_" + e.keyCode);`

And it does work with setting the layout to US and pressing the \` key it returns the code `61_187`

Geokureli:
> And it does work with setting the layout to US and pressing the ` key it returns the code `61_187`

what if you press +?

NeeEoo:
When i said i pressed the ` button, that is the key which is equal to + when im on us keyboard layout

Geokureli:
> When i said i pressed the ` button, that is the key which is equal to + when im on us keyboard layout

It is not, refer back to the pictures I sent, plus is typically on the = key

NeeEoo:
Pressing the + (=) key on us keyboard layout still works with this branch

Geokureli:
thanks!

PR #2990, opened 2023-12-19
Removing all references of `openfl_legacy`.

MAJigsaw77:


Geokureli:
I don't think I truly understand the consequences of this change, and I can't find any resources of what openfl_legacy is or whether we should kill support of it

MAJigsaw77:
This define had been removed since 2016 from openfl itself https://github.com/openfl/openfl/commit/15be4f5bd5eecbd8f0074ddfe966564d99ee1dc0

flixel/text/FlxText.hx
```diff
@@ -1249,4 +1249,4 @@ enum abstract FlxTextAlign(String) from String
 	}
 }
 
-private typedef AlignType = #if openfl_legacy String #else TextFormatAlign #end;
+private typedef AlignType = TextFormatAlign;
```
Geokureli, line 1252: We should remove this type and use TextFormatAlign, since most functions use TextFormatAlign, already

Geokureli (changes requested):
thanks for finding that removal, I asked in the openfl forum and no one replied

I think having AlignType adds confusion to FlxText, let's take it out. Rest looks good

flixel/text/FlxText.hx
Geokureli: ...unless you see any reason for it to exist

MAJigsaw77: > We should remove this type and use TextFormatAlign, since most functions use TextFormatAlign, already

Ok

Geokureli (approved):

Geokureli:
Thanks bro

PR #2991, opened 2023-12-19
Fix crash on gamepad disconnection

DigiEggz:
When unplugging a controller (specifically a Switch Pro controller), this calls a `destroy()` on the gamepad and sets `buttons = null;`. This causes an application crash if another function calls `FlxGamepad.checkStatusRaw()` since it tries to reference an element in a null array.

In this example, it happens when a `FlxUICursor` is polling a controller:
```
Uncaught exception: Null access .length
Called from flixel.addons.ui.FlxMultiGamepad.checkJustPressed(flixel/addons/ui/FlxMultiGamepad.hx:33)
Called from flixel.addons.ui.FlxBaseMultiInput.justPressed(flixel/addons/ui/FlxBaseMultiInput.hx:65)
Called from flixel.addons.ui.FlxUICursor._checkKeys(flixel/addons/ui/FlxUICursor.hx:807)
Called from flixel.addons.ui.FlxUICursor.update(flixel/addons/ui/FlxUICursor.hx:340)
Called from flixel.group.FlxTypedGroup.update(flixel/group/FlxGroup.hx:170)
Called from flixel.addons.ui.FlxUISubState.update(flixel/addons/ui/FlxUISubState.hx:182)
```

A working fix is to check if the array is valid before attempting to access it.

### DIFF
--- a/flixel/input/gamepad/FlxGamepad.hx
+++ b/flixel/input/gamepad/FlxGamepad.hx
@@ -327,6 +327,8 @@ class FlxGamepad implements IFlxDestroyable
 	 */
 	public inline function checkStatusRaw(RawID:Int, Status:FlxInputState):Bool
 	{
+		if(buttons == null)
+			return false;
 		var button = buttons[RawID];
 		return button != null && button.hasState(Status);
 	}


Geokureli:
Can you provide a small sample case that triggers this error, it seems like the issue may be that another util is polling a destroyed controller, not that a FlxGamepad isn't able to be polled when destroyed

DigiEggz:
There's a point where FlxUICursor [checks all of its key arrays](https://github.com/HaxeFlixel/flixel-ui/blob/f5d595cbb3806c55cb8867322fa19fc4dcaf6ff9/flixel/addons/ui/FlxUICursor.hx#L800), which may contain key references to removed controllers. This calls the [checkJustPressed functions of FlxMultiGamepad ](https://github.com/HaxeFlixel/flixel-ui/blob/f5d595cbb3806c55cb8867322fa19fc4dcaf6ff9/flixel/addons/ui/FlxMultiGamepad.hx#L29C38-L29C38), which checks if the `gamepad == null` before attempting to look at the keys. Because the gamepad was destroyed, the gamepad has a null `buttons` array.

My main concern was with how FlxGamepad's [checkStatusRaw()](https://github.com/HaxeFlixel/flixel/blob/84ecd4c6774d445590c0e6e824f6e1d05c47dd3f/flixel/input/gamepad/FlxGamepad.hx#L328C31-L328C31) assumes that the `buttons` array has elements even though there are various states where it may have a null array, leading to a `Null access .length` error.

Would it be better to check if the gamepad exists and that it has non-null `buttons` from within FlxMultiGamepad instead? For example:

```
private override function checkJustPressed():Bool
{
	if (gamepad == null || gamepad.buttons == null)
		return false;
	return gamepad.checkStatus(input, JUST_PRESSED);
}
```

Geokureli:
I understand the call chain, I'm just wondering if there is a simple example that I can use to actually cause it to happen locally, I rarely use FlxUICursor so in order to approve this PR I would need to learn it, create a test and dig around.

> My main concern was with how FlxGamepad's [checkStatusRaw()](https://github.com/HaxeFlixel/flixel/blob/84ecd4c6774d445590c0e6e824f6e1d05c47dd3f/flixel/input/gamepad/FlxGamepad.hx#L328C31-L328C31) assumes that the `buttons` array has elements even though there are various states where it may have a null array, leading to a `Null access .length` error.

Is there a state other than "destroyed" where this is possible? Right now it really seems like FlxUICursor is the tool making incorrect assumptions, here. Adding null checks here and there is often a quick, simple adhoc fix but it usually only masks the real issue and causes incorrect null-uncertainties for tools that use these tools correctly

DigiEggz:
Sure, I've made attached a sample project. I'm not sure if you have any controllers you can plug in, but I was also able to trigger it with Bluetooth controllers by disconnecting and reconnecting controllers.

[Demo.zip](https://github.com/HaxeFlixel/flixel/files/13731000/Demo.zip)

To trigger:

1. Compile and start application (try Neko or HashLink)
2. Plug in a USB controller. It should auto-connect even if you already had it plugged in.
3. Click the "Substate" button. It will open a substate with a UI cursor.
4. Unplug the controller. It should cause the app to crash.

Let me know if you have any issues triggering the crash.

Geokureli:
The problem was in fact FlxUICursor, I fixed it by preventing future gamepad checks when the gamepad is disconnected, rather than maintaining a reference to a destroyed gamepad. If you have time please try the [fix branch](https://github.com/HaxeFlixel/flixel-ui/pull/265) and let me know if it solves your problem, if so, please close this PR

To be blunt, I strongly recommend against using FlxUICursor, and Flixel-ui in general. I find it terribly bloated, cumbersome, cryptic and slow. I always recommend [haxeui-flixel](https://github.com/haxeui/haxeui-flixel) instead

DigiEggz:
Thanks for working on a fix! I'll test it as soon as I can.

I know you're not a fan, but I tried using haxeui-flixel and it came with its own set of issues. It was also missing a lot of key functionality that I would (as of now) have to recreate and for no perceivable performance benefit to the end user. Refactoring things was also a nightmare for my main project.

DigiEggz:
Addressed by https://github.com/HaxeFlixel/flixel-ui/pull/265.

PR #2992, opened 2023-12-20
Correct FlxFlicker.flicker() doc

T1mL3arn:
It is said it uses ping-pong tween but actually it uses FlxTimer.


Geokureli:
Thanks!

PR #2995, opened 2023-12-25
Correct FlxPoint doc for overloaded compound assignment operators

### LINKED ISSUE #2994: Wrong doc for FlxPoint overloaded operators
This doc tells it returns NEW point, while actually it isn't - it uses method that returns left hand operator.

https://github.com/HaxeFlixel/flixel/blob/bc6d9a825f49f704d5107d5d0c0a5933df0518e4/flixel/math/FlxPoint.hx#L157-L167

This is relevant to probably all compound assignment operators. I can make PR to fix it but before it should be clear what exactly is wrong - either doc or behavior. I personally like current behavior, though current users might use them according to the doc i.e expecting new point objects.

T1mL3arn:
closes #2994

### DIFF
--- a/flixel/math/FlxPoint.hx
+++ b/flixel/math/FlxPoint.hx
@@ -156,7 +156,7 @@ import openfl.geom.Point;
 	}
 
 	/**
-	 * Operator that adds two points, returning a new point.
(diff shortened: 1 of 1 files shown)


Geokureli:
Thanks!

PR #2996, opened 2023-12-30
Save on some if statements

MaybeMaru:
Just gets the null checked alphaMultiplier before the for loop to save on some small if statements


Geokureli:
Thanks!

PR #3004, opened 2024-01-08
Fix one-time listener is not removed from FlxSignal on HashLink

T1mL3arn:
One-time listeners other than `Void->Void` is not removed from signals on HashLink (see https://github.com/HaxeFoundation/hashlink/issues/578 ).

Use updated tests but old FlxSignal to see `addOnce()` does not work as expected on HashLink

```
haxelib run munit gen -filter flixel.util.FlxSignalTest
haxelib run lime test hl

#---
Class: flixel.util.FlxSignalTest .....!..........
    FAIL: massive.munit.AssertionException: Value [3] was not equal to expected value [1] at flixel.util.FlxSignalTest#testDispatchOnce_signal1 (175)
==============================
FAILED
Tests: 16  Passed: 15  Failed: 1 Errors: 0 Ignored: 0 Time: 0.005
```

### DIFF
--- a/flixel/util/FlxSignal.hx
+++ b/flixel/util/FlxSignal.hx
@@ -133,14 +133,19 @@ private class FlxBaseSignal<T> implements IFlxSignal<T>
 			var handler = getHandler(listener);
 			if (handler != null)
 			{
-				if (processingListeners)
-					pendingRemove.push(handler);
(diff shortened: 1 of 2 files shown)


Geokureli:
Thanks!

PR #3006, opened 2024-01-11
Wrap bitmaps for debugger with FLX_DEBUG so they don't appear in release build

T1mL3arn:
There are some bitmaps which are used only when `FLX_DEBUG` implies but aren't wrapped with this define. So they are included in release build inflating project size and aren't used by any code

If such behavior is considered a bug I will wrap other similar cases.

### DIFF
--- a/flixel/system/debug/stats/Stats.hx
+++ b/flixel/system/debug/stats/Stats.hx
@@ -12,11 +12,13 @@ import flixel.system.debug.FlxDebugger.GraphicStats;
 import flixel.system.ui.FlxSystemButton;
 import flixel.util.FlxColor;
 
+#if FLX_DEBUG
 @:bitmap("assets/images/debugger/buttons/minimize.png")
 private class GraphicMinimizeButton extends BitmapData {}
 
 @:bitmap("assets/images/debugger/buttons/maximize.png")
 private class GraphicMaximizeButton extends BitmapData {}
+#end
 
 /**
  * A simple performance monitor widget, for use in the debugger overlay.
--- a/flixel/system/debug/stats/Stats.hx
+++ b/flixel/system/debug/stats/Stats.hx
@@ -12,13 +12,11 @@ import flixel.system.debug.FlxDebugger.GraphicStats;
 import flixel.system.ui.FlxSystemButton;
 import flixel.util.FlxColor;
 
-#if FLX_DEBUG
 @:bitmap("assets/images/debugger/buttons/minimize.png")
 private class GraphicMinimizeButton extends BitmapData {}
 
 @:bitmap("assets/images/debugger/buttons/maximize.png")
 private class GraphicMaximizeButton extends BitmapData {}
-#end
 
 /**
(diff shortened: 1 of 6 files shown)


flixel/system/debug/stats/Stats.hx
```diff
@@ -12,11 +12,13 @@ import flixel.system.debug.FlxDebugger.GraphicStats;
 import flixel.system.ui.FlxSystemButton;
 import flixel.util.FlxColor;
 
+#if FLX_DEBUG
```
Geokureli, line 15: so right now it's

```hx
#if FLX_DEBUG
@:bitmap("assets/images/debugger/buttons/minimize.png")
private class GraphicMinimizeButton extends BitmapData {}

@:bitmap("assets/images/debugger/buttons/maximize.png")
private class GraphicMaximizeButton extends BitmapData {}
#end

/**
 * A simple performance monitor widget, for use in the debugger overlay.
 *
 * @author Adam "Atomic" Saltsman
 * @author Anton Karlov
 */
#if FLX_DEBUG
```

let's just use one #if for all this

Geokureli (changes requested):
lets do this, but we should just move the existing #if to include this

Geokureli:
CI is failing because of the changes you made to the new classes, i dont see why you edited those classes

T1mL3arn:
> CI is failing because of the changes you made to the new classes

Can you elaborate more why checks are failing? I see in log that it failed on `HaxeFlixel/setup-flixel@v1` but it doesn't provide any details what's wrong and I also see that tests, demos, and snippets are all green. I am confused.

> i dont see why you edited those classes

It's not possible to just wrap bitmap classes as they are  - projects won't compile. So I had to figure out some minimum of changes that allows compilation in both debug and release modes. 

Other ways to consider:

- wrap debug things on the very top of `FlxG`, so none of them are available in release builds (e.g. users must then wrap `FlxG.debugger` in their code)
- wrap each `@:bitmap` but not the types themselves, so bitmaps are not included and types are erased if user does ``dce` or `analyzer`
- ask Haxe to fix `@:bitmap`

Geokureli:
Screenshot_20240117-094221
Type not found: StatsGraph

Geokureli:
this was the change I originally requested. Not sure why you changed any of the other files, as those are outside the scope and also breaking changes.

this should prevent the images from being embedded in release, please verify this

Geokureli (approved):

T1mL3arn:
Sorry for late response.

> Not sure why you changed any of the other file

To make it compile. This problem with embedded bitmaps goes beyond a module from the initial commit - there are quite enough such bitmaps.

Geokureli:
Are there any issues with these new changes? Can you verify that this prevents them from being embedded?

Geokureli:
@T1mL3arn ?
> Can you verify that this prevents them from being embedded?

Geokureli:
Fixed via https://github.com/HaxeFlixel/flixel/pull/3077
I went a simpler route

PR #3008, opened 2024-01-14
Add interface IFlxObject to FlxObject

MaybeMaru:
Only creates a interface for FlxObjects called ``IFlxObject`` as a middle ground between ``IFlxBasic`` and ``IFlxSprite``

### DIFF
--- a/flixel/FlxObject.hx
+++ b/flixel/FlxObject.hx
@@ -1538,7 +1538,8 @@ enum abstract CollisionDragType(Int)
 	var HEAVIER = 3;
 }
 
-interface IFlxObject extends IFlxBasic {
+interface IFlxObject extends IFlxBasic
+{
 	var x(default, set):Float;
 	var y(default, set):Float;
 
--- a/flixel/FlxObject.hx
+++ b/flixel/FlxObject.hx
@@ -1,5 +1,6 @@
 package flixel;
 
+import flixel.FlxBasic.IFlxBasic;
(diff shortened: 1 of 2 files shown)


Geokureli:
is there a particular reason you're adding this? is it needed for something you're working on?

MaybeMaru:
Yes, ive been working on some basic object classes and it came up to me that i was using an flxobject type interface quite often without FlxSprite values like alpha or scale. Just thought it could be useful to have it by default for some people.

47rooks:
I was just looking at the interfaces IFlxSprite and IFlxBasic and they don't appear to do anything at all - they aren't implemented by the either of their classes. So how do they impede changes ? How would a typedef would replace them ?

But I was wondering if there was an interface definition that would clearly define what a sprite or an object or a basic was.

Geokureli:
> I was just looking at the interfaces IFlxSprite and IFlxBasic and they don't appear to do anything at all - they aren't implemented by the either of their classes. How do they impede changes ?

That's interesting I never noticed that, but when I assumed they were implemented the caveat was that changing FlxSprite/basic requires synchronized changes in two places, which is kind of a hassle.

The only place I know `IFlxSprite` is used is by [FlxParticle](https://api.haxeflixel.com/flixel/effects/particles/FlxParticle.html) which implements `IFlxParticle` an extended interface of `IFlxSprite`

Issue #2958, opened 2023-11-09
FlxTween.quadPath() can make the object disappear from the screen


giuppe:
- **Haxe version:** 4.3.2
- **Flixel version:** 5.5.0
- **OpenFL version:** ?.?.?
- **Lime version:** ?.?.?
- **Affected targets:** All

There is a condition where `FlxTween.quadPath()` will make the FlxObject disappear from the screen.
________________________________

**Code snippet reproducing the issue**:

```haxe
package;

import flixel.FlxState;

class PlayState extends FlxState
{
	override public function create():Void
	{
		_img = new FlxSprite(0, 0);
		_img.makeGraphic(100, 100, FlxColor.MAGENTA);
		add(_img);
		var point1 = FlxPoint.get(100,100);
		var point2 = FlxPoint.get(100,100);
		var point3 = FlxPoint.get(0,100);
		var point4 = FlxPoint.get(100,0);
		var point5 = FlxPoint.get(200,200);
		FlxTween.quadPath(_img, [point1, point2, point3, point4, point5], 10);
	}
}
```

The position of the FlxObject is wrongly set to (NaN, NaN).

Requirements for the bug to happen:
- at least 5 points;
- two subsequent points in the same location.

Obviously, nobody would manually set the points in this way, but it can happen if they are generated by a game mechanic.

The culprit seems to be this line of QuadPath: https://github.com/HaxeFlixel/flixel/blob/c0a6f1c7fc7fc155641853968127a113a5b6b729/flixel/tweens/motion/QuadPath.hx#L221

When both `ba` and `c2` are zero, there is a division by zero that makes the function return NaN; that then propagates through all the other computations until it ends on the FlxObject position.
`ba == c2 == 0` when the first two arguments of the functions are the same point.

I guess the fastest way to avoid that discontinuity is to set `(ba + c2)` to a small quantity (for example 0.00001) instead of zero - it would make the division by zero disappear.

I can produce a PR if needed.

________________________________

**Observed behavior:** No square on the screen.

**Expected behavior:** Square on the screen moving in a curve from `point1` to `point5`.

Geokureli:
feel free to make a PR

giuppe:
PR produced: #3019

Geokureli:
Fixed via #3019

PR #3019, opened 2024-01-28
avoid division by zero in QuadPath (issue #2958)

giuppe:
Changes some variables value from zero to a really small value when used as denominator.

### DIFF
--- a/flixel/tweens/motion/QuadPath.hx
+++ b/flixel/tweens/motion/QuadPath.hx
@@ -212,8 +212,8 @@ class QuadPath extends Motion
 			
 		p1.x = start.x - 2 * safeControl.x + finish.x;
 		p1.y = start.y - 2 * safeControl.y + finish.y;
-		p2.x = 2 * control.x - 2 * start.x;
-		p2.y = 2 * control.y - 2 * start.y;
+		p2.x = 2 * safeControl.x - 2 * start.x;
+		p2.y = 2 * safeControl.y - 2 * start.y;
 		var a:Float = 4 * (p1.x * p1.x + p1.y * p1.y),
 			b:Float = 4 * (p1.x * p2.x + p1.y * p2.y),
 			c:Float = p2.x * p2.x + p2.y * p2.y,
--- a/flixel/tweens/motion/QuadPath.hx
+++ b/flixel/tweens/motion/QuadPath.hx
@@ -1,5 +1,6 @@
 package flixel.tweens.motion;
 
+import flixel.math.FlxMath;
 import flixel.math.FlxPoint;
 import flixel.tweens.FlxTween.FlxTweenManager;
 import flixel.tweens.FlxTween.TweenOptions;
@@ -199,11 +200,18 @@ class QuadPath extends Motion
 
 	function getCurveLength(start:FlxPoint, control:FlxPoint, finish:FlxPoint):Float
 	{
+		var safeControl = control.clone();
 		var p1 = FlxPoint.get();
 		var p2 = FlxPoint.get();
 
-		p1.x = start.x - 2 * control.x + finish.x;
-		p1.y = start.y - 2 * control.y + finish.y;
+		if (safeControl.equals(start))
+			safeControl.add(FlxMath.EPSILON, FlxMath.EPSILON);
+			
+		if (safeControl.equals(finish))
+			safeControl.add(FlxMath.EPSILON, FlxMath.EPSILON);
+			
+		p1.x = start.x - 2 * safeControl.x + finish.x;
+		p1.y = start.y - 2 * safeControl.y + finish.y;
 		p2.x = 2 * control.x - 2 * start.x;
 		p2.y = 2 * control.y - 2 * start.y;
 		var a:Float = 4 * (p1.x * p1.x + p1.y * p1.y),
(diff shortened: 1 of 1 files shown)


Geokureli:
so this will result an incredibly high number rather than infinity/NaN but will that also be problematic?

can you elaborate on what causes unexpected zeroe, why it seems to cause an infinite curve length in this formula and whether a different formula can avoid this issue?

giuppe:
Yes, the use case is this (as reported in #2958):

The position of the FlxObject is wrongly set to (NaN, NaN).

Requirements for the bug to happen:
- at least 5 points;
- two subsequent points in the same location.

Obviously, nobody would manually set the points in this way, but it can happen if they are generated by a game mechanic.

```hx
package;

import flixel.FlxState;

{
	{
		_img = new FlxSprite(0, 0);
		_img.makeGraphic(100, 100, FlxColor.MAGENTA);
		add(_img);
		var point1 = FlxPoint.get(100,100);
		var point2 = FlxPoint.get(100,100);
		var point3 = FlxPoint.get(0,100);
		var point4 = FlxPoint.get(100,0);
		var point5 = FlxPoint.get(200,200);
		FlxTween.quadPath(_img, [point1, point2, point3, point4, point5], 10);
	}
}
```

> can you elaborate on what causes unexpected zeroes, why it seems to cause an infinite curve length in this formula and whether a different formula can avoid this issue?

The culprit seems to be this line of QuadPath: https://github.com/HaxeFlixel/flixel/blob/c0a6f1c7fc7fc155641853968127a113a5b6b729/flixel/tweens/motion/QuadPath.hx#L221

When both `ba` and `c2` are zero, there is a division by zero that makes the function return NaN; that then propagates through all the other computations until it ends on the FlxObject position.

How can they be zero? See the full function:
```hx
function getCurveLength(start:FlxPoint, control:FlxPoint, finish:FlxPoint):Float
	{
		var p1 = FlxPoint.get();
		var p2 = FlxPoint.get();

		p1.x = start.x - 2 * control.x + finish.x;
		p1.y = start.y - 2 * control.y + finish.y;
		p2.x = 2 * control.x - 2 * start.x;
		p2.y = 2 * control.y - 2 * start.y;
		var a:Float = 4 * (p1.x * p1.x + p1.y * p1.y),
			b:Float = 4 * (p1.x * p2.x + p1.y * p2.y),
			c:Float = p2.x * p2.x + p2.y * p2.y,
			abc:Float = 2 * Math.sqrt(a + b + c),
			a2:Float = Math.sqrt(a),
			a32:Float = 2 * a * a2,
			c2:Float = 2 * Math.sqrt(c),
			ba:Float = b / a2;

		p1.put();
		p2.put();

		return (a32 * abc + a2 * b * (abc - c2) + (4 * c * a - b * b) * Math.log((2 * a2 + ba + abc) / (ba + c2))) / (4 * a32);
	}
```

When `start` and `control` represent the same point, then `p2` is `(0, 0)`, then `b` and `c` are zero, then `ba` and `c2` are zero. This causes a discontinuity in `Math.log((2 * a2 + ba + abc) / (ba + c2)))` when `ba + c2 == 0`.

There is another discontinuity when `start`, `control`, *and* `finish` are the same point. In this case, both `p1` and `p2` are null, `a`  `b` and `c` are zero, and the denominator `(4 * a32)` is zero.

The current proposed changes just affect those denominators. If those are removable discontinuities, the function should be fine if we substitute them with the limit (that in this case is a very small value). However, now that I think of it I can't be sure that won't cause any error magnifications somewhere else - in my case, I changed the denominators and it worked. 
So maybe in the general case it's safer to act on the `control` FlxPoint argument, adding a small value to it if it happens to have the same values as `start` or `finish`.

giuppe:
The core of the latest proposal is:
```hx
var safeControl = control.clone();

if (safeControl.equals(start))
	safeControl.add(FlxMath.EPSILON, FlxMath.EPSILON);

if (safeControl.equals(finish))
	safeControl.add(FlxMath.EPSILON, FlxMath.EPSILON);
	
//then use safeControl instead of control
```

This way, even if we have two ore three equal points as input, the second one will be always slightly different to prevent divisions by zero. If we have input points different from each other (the common case), no change at all.

Geokureli:
can we just check if the two points are equal and return a curve length of 0? I assume if 2 two points are equal it just has "no" length?

giuppe:
No, the curve is not zero length when two points are equal. `a` will still be non-zero so the component of final result `a32 * abc` will still be not zero. The curve will be zero length if three points are equal.

flixel/tweens/motion/QuadPath.hx
```diff
@@ -199,13 +200,20 @@ class QuadPath extends Motion
...
 
-		p1.x = start.x - 2 * control.x + finish.x;
-		p1.y = start.y - 2 * control.y + finish.y;
-		p2.x = 2 * control.x - 2 * start.x;
-		p2.y = 2 * control.y - 2 * start.y;
+		if (safeControl.equals(start))
+			safeControl.add(FlxMath.EPSILON, FlxMath.EPSILON);
+			
+		if (safeControl.equals(finish))
+			safeControl.add(FlxMath.EPSILON, FlxMath.EPSILON);
```
Geokureli, line 211: lets put a comment here explaining why these lines are necessary

Geokureli:
Thanks!

PR #3022, opened 2024-02-03
Allow to use async images that was already loaded

T1mL3arn:
When I want to use async assets
```xml
<assets path="assets" embed="false" preload="false"/>
```
```haxe
openfl.Assets.loadBitmapData(asset_id, true).onComplete(bd ->
{
	trace('image is loaded');
	FlxAssets.getBitmapData(asset_id);
});
```
I get this 
```
image is loaded
ERROR: IMAGE asset "asset.png" exists, but only asynchronously'
```
because Flixel doesn't use cache when it gets bmd from openfl assets for some reason, I believe it is a bug.

--- update

example is valid for html5

### DIFF
--- a/flixel/system/FlxAssets.hx
+++ b/flixel/system/FlxAssets.hx
@@ -276,7 +276,7 @@ class FlxAssets
 	public static inline function getBitmapData(id:String):BitmapData
 	{
 		if (Assets.exists(id))
-			return Assets.getBitmapData(id, false);
+			return Assets.getBitmapData(id, true);
 		FlxG.log.error('Could not find a BitmapData asset with ID \'$id\'.');
 		return null;
 	}


Geokureli:
This seems like a good change, I'd like to poke around the history and verify that there's no *valid* reason for this to be false, but it seems like it's a good change.

The only issue is that this is gonna conflict with this PR: https://github.com/HaxeFlixel/flixel/pull/2982, and this change may affect how I approach that other PR, so it just needs some consideration

T1mL3arn:
>there's no valid reason for this arg to be false

At least on openfl's side it is `true` by default.

Geokureli:
I have one theory as to why it's false, I fear that if it's `true` it will return the same bitmap instance on each call rather than fetching a new copy. This means that editing one bitmap will effect all other sprites using the same path.

this is just a guess, but it's something I would like checked before merging this

T1mL3arn:
>This means that editing one bitmap will effect all other sprites using the same path.

Yes, it works that way.

Geokureli:
I'm unable to reproduce this issue due to the following reasons:
the `preload` attribute does not actually affect asset tags in Project.xml
```xml
<assets path="assets" embed="false" preload="false"/>
```
it can only be used in libraries which forces you to use the lib name in asset_id like so: "libName:assets/psth/to/image.png", and then if I try `openfl.Assets.loadBitmapData(asset_id, true).onComplete(bd->{});` it returns an error 'could not find a library named "libName"'. 
but if I use libraries in my project.cml and do: `openfl.Assets.loadLibrary(asset_id).onComplete(bd->{});` then I'm able to load images in that library via FlxAssets.getBitmapData(asset_id) after it completes

T1mL3arn:
I updated the original post - example is valid for html5.

Geokureli:
it turns out the part i was missing was `embed="false"`, `preload` does in fact do nothing in asset tags and I was able to reproduce the issue with both: `<assets path="assets" embed="false" preload="false"/>` and `<assets path="assets" embed="false" preload="true"/>`

Geokureli:
> > This means that editing one bitmap will effect all other sprites using the same path.
> 
> Yes, it works that way.

So this doesn't seem to be the case, in the following example:
```hx
import flixel.FlxSprite;

class AsyncImageTestState extends flixel.FlxState
{
    inline static var PATH = "async-assets/images/haxe.png";
    
    override function create()
    {
        super.create();
        
        openfl.Assets.loadBitmapData(PATH, true).onComplete(
            function drawSprites(_)
            {
                for (i in 0...3)
                {
                    final sprite = new FlxSprite(10 + 110 * i, 10);
                    sprite.loadGraphic(PATH, false, 0, 0, true);
                    add(sprite);
                }
                // edit the first sprite, see if all change
                final bitmap = (cast members[0]:FlxSprite).graphic.bitmap;
                bitmap.fillRect(bitmap.rect, 0xFFff0000);
            }
        );
    }   
}
```
only the first sprite is modified, meaning `unique` is honored despite the use of cahcing

Issue #1370, opened 2014-11-10
FlxTouch and FlxButton pressed being missed


JoeCreates:
This seems to be most noticeable when using touch, for example, when touching FlxButtons. Button pressed are often missed, making the game feel unresponsive. How long you touch a button for makes a difference. It is only quick presses that get missed.

I have tried adding the following traces to FlxButton#onUpEventListener:

```
#if !FLX_NO_MOUSE
    private function onUpEventListener(_):Void
    {
        trace("ONUPLISTENER");
        if (visible && exists && active && status == FlxButton.PRESSED)
        {
            trace("handled");
            onUpHandler();
        }
    }
#end
```

With the above, a quick press on a button causes "ONUPLISTENER" to be traces, but not "handled". The reason is that the status is not yet PRESSED. If the button has not been updated before the touch has been released, it will likely be missed as the state is incorrect.

It looks like this could be due to the fact that Inputs can be pressed and released in a single update, and yet the input cannot simultaneously be in the states JUST_PRESSED and JUST_RELEASED. The rapid release is meaning that the JUST_PRESSED state is being missed.

Also, is it possible that an input is going to be updated before the button? If so the input could move from JUST_RELEASED to RELEASED before the button has had time to check if the input was just released, and thus the onUp callback will be skipped.
##

Gama11:
The fix for #1686 might have helped with this (2c9a545168). Not really sure how to test it.

JoeCreates:
I don't think so. That seems to be an unrelated issue. @Tw1ddle and I fixed this on a private branch of flixel, but completely changes the states inputs can have, and so it is incompatible with Flixel's input recording.

It has to be possible for a button to be both justPressed and justReleased on the same frame.

This is a major issue for mobile devices (ours and other's mobile games have been rejected from app store for having unresponsive buttons, although submitting multiple times eventually gets it through).

I would suggest that the existing system is simply broken, but it is unfortunate that the recording is dependent on it. Perhaps we could merge our fixed version of FlxInput in and make it optional to use the old system for flxrecord?

Gama11:
I'm not sure it's only recordings that depend on this..

JoeCreates:
@Gama11 Here's how it looks in our new version: https://gist.github.com/JoeCreates/aba33862ddfb75d3071b9e5d219878df

`justPressedQueueFalse` and `justReleasedQueueFalse` ensure the `justPressed` and `justReleased` states last for a frame.

[...]

PR #3023, opened 2024-02-03
Skip extra listener for non-Flash platforms

DigiEggz:
Adding an extra listener can cause callbacks to fire twice in rapid succession, causing errors. The Flash version may still require this, so the listener is retained for that platform.

### DIFF
--- a/flixel/FlxGame.hx
+++ b/flixel/FlxGame.hx
@@ -3,6 +3,7 @@ package flixel;
 import flixel.graphics.tile.FlxDrawBaseItem;
 import flixel.system.FlxSplash;
 import flixel.util.FlxArrayUtil;
+import flixel.util.FlxDestroyUtil;
 import flixel.util.typeLimit.NextState;
 import openfl.Assets;
 import openfl.Lib;
@@ -756,7 +757,13 @@ class FlxGame extends Sprite
 		#end
 
 		#if FLX_POINTER_INPUT
-		FlxArrayUtil.clearArray(FlxG.swipes);
+		var len = FlxG.swipes.length;
+		while(len-- > 0)
+		{
+			final swipe = FlxG.swipes.pop();
+			if (swipe != null)
+				swipe.destroy();
+		}
 		#end
 
 		filters = filtersEnabled ? _filters : null;
--- a/flixel/graphics/FlxGraphic.hx
+++ b/flixel/graphics/FlxGraphic.hx
@@ -473,7 +473,7 @@ class FlxGraphic implements IFlxDestroyable
 		key = null;
 		assetsKey = null;
 		assetsClass = null;
-		imageFrame = null; // no need to dispose _imageFrame since it exists in imageFrames
+		imageFrame = FlxDestroyUtil.destroy(imageFrame);
 
 		if (frameCollections == null) // no need to destroy frame collections if it's already null
 			return;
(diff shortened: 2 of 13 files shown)


DigiEggz:
Please see if this interferes with buttons. I did not experience any differences when `overlapFound` is removed and it allows the button to be interacted with using stageEvents.

DigiEggz:
Checking `input.pressed` instead of `overlapFound` allows the button to fire "onUp" events when triggered by elements other than the pointer.

DigiEggz:
> can you provide small, complete test case that highlights the benefit of this change?

Sure thing!

DigiEggz:
I made another change to make sure it functions correctly when the `allowSwiping` property is set.

[Demo.zip](https://github.com/HaxeFlixel/flixel/files/14181751/Demo.zip)

Here's a simple project that includes a `FlxButton` and a `FlxUIButton`. Clicking on any button should print text to the console. Pressing the **X** key or Enter will trigger the UI button.  Pressing the **S** key will disable swiping on both buttons.

When you swap between the current the new code, everything should properly react. The primary benefit is that an extra event listener is no longer added to every button instance. I've noted some occasions where the event listener can cause the "onUp" event to be fired twice, which can cause issues. Please let me know if you encounter any differences or have any concerns.

Because of the order of handling, this may resolve #1370.

Geokureli:
I haven't checked the demo yet, but one fear i have pertains to this:
```hx
/**
* Using an event listener is necessary for security reasons on flash -
* certain things like opening a new window are only allowed when they are user-initiated.
*/
```
I've also experienced similar restrictions on browsers

DigiEggz:
I left the onUp check for Flash since it likely still requires it, so Flash should function the same. The extra code added to `updateStatus()` was to fix a dragging issue introduced by the listener-less code.

The end goal is for buttons to function as they currently do, but without an extra event listener, freeing up memory and preventing execution order errors.

Geokureli:
> I left the onUp check for Flash since it likely still requires it, so Flash should function the same

Again, I've noticed the same requirements on browsers meaning html5 should not have this removed

DigiEggz:
> > I left the onUp check for Flash since it likely still requires it, so Flash should function the same
> 
> Again, I've noticed the same requirements on browsers meaning html5 should not have this removed

Oh, I see what you're saying! I thought you meant browsers as in Flash within a browser. I can continue to test HTML5 functionality between different browsers.

What would be a good test case for this? When it says open a new window, does it mean something like `FlxG.openURL()`?

Geokureli:
> What would be a good test case for this? When it says open a new window, does it mean something like `FlxG.openURL()`?

Yeah i've specifically seen this error when opening a url, I'd also like to know more about this issue about events firing twice before I make a change like this for all projects. perhaps this should be an opt-in setting?

DigiEggz:
So far I've had no issues on Firefox or Chrome when using openURL. It's been hard to replicate, but there are times when a button callback changes a graphic and opens a substate, the graphic is null for a second and an OpenFL null graphic crash occurs.

I noted that not adding the additional onUp callback listener stopped this from happening. Other than for Flash, it made sense to not add an extra callback if it was no longer necessary. Let me know if you have any issues with HTML5.

What would be the best way to make this opt-in?

Geokureli:
TBH the open url issue is something that has come and gone once or twice with various security updates for browsers, at different times for different browsers.

Chrome has a domain whitelist and blacklist for this setting:
Screenshot 2024-02-10 at 12 06 19 PM

PR #3027, opened 2024-02-08
Added `letterSpacing` variable to `FlxText`.

MAJigsaw77:
Adds access to openfl's letterSpacing in FlxText instances.

related issue, letterSpacing doesn't work on html5 - https://github.com/openfl/openfl/issues/1700

### DIFF
--- a/flixel/text/FlxText.hx
+++ b/flixel/text/FlxText.hx
@@ -66,7 +66,7 @@ class FlxText extends FlxSprite
 	 *
 	 * The default value is `null`, which means that 0 pixels of letter spacing is used.
 	 */
-	public var spacing(default, set):Null<Float> = null;
+	public var letterSpacing(default, set):Null<Float> = null;
 
 	/**
 	 * The font used for this text (assuming that it's using embedded font).
@@ -649,9 +649,9 @@ class FlxText extends FlxSprite
 		return Size;
 	}
 
-	function set_spacing(Spacing:Null<Float>):Null<Float>
+	function set_letterSpacing(LetterSpacing:Null<Float>):Null<Float>
 	{
-		_defaultFormat.letterSpacing = Spacing;
+		_defaultFormat.letterSpacing = LetterSpacing;
 		updateDefaultFormat();
 		return Spacing;
 	}
--- a/flixel/text/FlxText.hx
+++ b/flixel/text/FlxText.hx
@@ -63,10 +63,8 @@ class FlxText extends FlxSprite
 	 * A number representing the amount of space that is uniformly distributed
 	 * between all characters. The value specifies the number of pixels that are
 	 * added to the advance after each character.
-	 *
-	 * The default value is `null`, which means that 0 pixels of letter spacing is used.
 	 */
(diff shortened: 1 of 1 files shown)


moxie-coder:
nice

flixel/text/FlxText.hx
```diff
@@ -59,6 +59,15 @@ class FlxText extends FlxSprite
...
 	public var size(get, set):Int;
 
+	/**
+	 * A number representing the amount of space that is uniformly distributed
+	 * between all characters. The value specifies the number of pixels that are
+	 * added to the advance after each character.
+	 *
+	 * The default value is `null`, which means that 0 pixels of letter spacing is used.
+	 */
+	public var letterSpacing(default, set):Null<Float> = null;
```
Geokureli, line 69: 1. (default, set) will cause issues, since set_ letterSpacing only sets _defaultFormat.letterSpacing. meaning the following will happen:
```hx
text.letterSpacing = 1;
trace(text.letterSpacing); // null
```
We should add a getter that returns `_defaultFormat.letterSpacing`

looks like the other fields avoid using Null<T> and just use primitive types, I think Null<T> should be avoided if possible. let's use `Float` and set the default value to 0, and also set the _defaultFormats initial letterSpacing to 0

Geokureli (changes requested):
See below for requested changes and provide a small use case

flixel/text/FlxText.hx
MAJigsaw77: The letterSpace is null by default in openfl

MAJigsaw77: That value cannot be getted

MAJigsaw77: https://github.com/openfl/openfl/blob/6c106bc8463edb1ee71491d1e4f2b7e1dbf15f7f/src/openfl/text/TextFormat.hx#L156

Geokureli: > That value cannot be getted

Sure it can

> The letterSpace is null by default in openfl
I understand, that's why i said "set the _defaultFormats initial letterSpacing to 0", specifically, [here](https://github.com/HaxeFlixel/flixel/blob/dev/flixel/text/FlxText.hx#L221)

Geokureli:
also this is gonna wait until after the 5.6.1

Geokureli (changes requested):
To reiterate:
> set the _defaultFormats initial letterSpacing to 0
> specifically [here](https://github.com/HaxeFlixel/flixel/blob/dev/flixel/text/FlxText.hx#L221)

MAJigsaw77:
Usage example:
```haxe
import flixel.text.FlxText;

var text:FlxText = new FlxText(0, 0, 0, "Hello World", 64);
text.letterSpacing = 2.5;
text.screenCenter();
add(text);
```

Geokureli (approved):

Geokureli:
All good, I'll merge this after I get 5.6.1 released

Geokureli:
When i test the following code:
```hx
```
I get this result
Screenshot 2024-02-09 at 1 53 18 PM

PR #3028, opened 2024-02-09
Fix negative animation frame number

T1mL3arn:
I got in the situation when animations play backwards despite in aseprite editor they are clearly normal. The reason was that I used format for exported frames like this `bla-bla-{frame}` and prefix to add the anim like `bla-bla`. With this combination the frames are treated as negative integers, which results with reversed frame order. 

I think there is no such thing as "negative frame numbers" in spritesheet animations, so here is the fix.

### DIFF
--- a/flixel/graphics/frames/FlxFrame.hx
+++ b/flixel/graphics/frames/FlxFrame.hx
@@ -4,6 +4,7 @@ import openfl.display.BitmapData;
 import openfl.geom.Point;
 import openfl.geom.Rectangle;
 import flixel.graphics.FlxGraphic;
+import flixel.math.FlxMath;
 import flixel.math.FlxMatrix;
 import flixel.math.FlxPoint;
 import flixel.math.FlxRect;
@@ -54,7 +55,7 @@ class FlxFrame implements IFlxDestroyable
 		if (num2 == null)
 			num2 = 0;
 
-		return Math.abs(num1) - Math.abs(num2);
+		return FlxMath.absInt(num1) - FlxMath.absInt(num2);
 	}
 
 	public var name:String;
--- a/flixel/graphics/frames/FlxFrame.hx
+++ b/flixel/graphics/frames/FlxFrame.hx
@@ -54,7 +54,7 @@ class FlxFrame implements IFlxDestroyable
 		if (num2 == null)
 			num2 = 0;
 
-		return num1 - num2;
+		return Math.abs(num1) - Math.abs(num2);
 	}
 
 	public var name:String;


Geokureli:
Can you include a test case that reproduces the issue?

T1mL3arn:
> but you're saying it's "bla-bla" without the final dash

Yes, the prefix is `"bla-bla"`. That's why  `"-4"` number substring is extracted from a frame name like `"bla-bla-4"`.

T1mL3arn:
Example

```haxe
var sprite = new FlxSprite();

// plain frame data from atlas json
var framesSource = {
	"frames": [
		{
			"filename": "top-right-walk-1",
			"frame": {
				"x": 71,
				"y": 0,
				"w": 71,
				"h": 156
			},
			"rotated": false,
			"trimmed": true,
			"spriteSourceSize": {
				"x": 29,
				"y": 20,
				"w": 71,
				"h": 156
			},
			"sourceSize": {
				"w": 128,
				"h": 192
			},
			"duration": 200
		},
		{
			"filename": "top-right-walk-2",
			"frame": {
				"x": 142,
				"y": 0,
				"w": 71,
				"h": 156
			},
			"rotated": false,
			"trimmed": true,
			"spriteSourceSize": {
				"x": 29,
				"y": 20,
				"w": 71,
				"h": 156
			},
			"sourceSize": {"w": 128, "h": 192},
			"duration": 200
		},
	]
};

// just building frames, skipping loading assets steps
var atlasFrames = new FlxAtlasFrames(FlxG.bitmap.add(new BitmapData(100, 100), false));
for (frame in framesSource.frames)
{
	@:privateAccess
	FlxAtlasFrames.texturePackerHelper(frame.filename, cast frame, atlasFrames, true);
}
sprite.frames = atlasFrames;

// add two animations from the same frames, using different prefixes
sprite.animation.addByPrefix('walk-incorrect', 'top-right-walk');
sprite.animation.addByPrefix('walk-correct', 'top-right-walk-');

var incorrect = sprite.animation.getByName('walk-incorrect');
var correct = sprite.animation.getByName('walk-correct');

trace('anim frames indicies  BAD: ${incorrect.frames}');
trace('anim frames indicies GOOD: ${correct.frames}');

// anim frames indicies  BAD: [1,0]
// anim frames indicies GOOD: [0,1]
```

Geokureli:
In this case I would say the fix is to put the dash in the prefix since the prefix is "top-right-walk-" but I doubt people will try to use negative frame numbers. On the off-chance that someone does I've added warning logs. I've also fixed some things that have been bugging me, like wtf is a "postFix", lol

Geokureli:
Thanks!

PR #3033, opened 2024-02-15
Fix tween crash when object is destroyed

CobaltBar:


Geokureli:
Provide example code that reliable causes the crash

CobaltBar:
I'm trying to reproduce it.

PR #3045, opened 2024-02-24
Added null check in stop function to prevent crash issues.

TheoDevelops:
Just added a null check in stop function.

### DIFF
--- a/flixel/effects/FlxFlicker.hx
+++ b/flixel/effects/FlxFlicker.hx
@@ -149,8 +149,10 @@ class FlxFlicker implements IFlxDestroyable
 	 */
 	public function stop():Void
 	{
-		timer.cancel();
-		object.visible = true;
+		if (object != null) {
+			timer.cancel();
+			object.visible = true;
+		}
 		release();
 	}
 


Geokureli:
can you provide a reliable test case that reproduces this issue?

TheoDevelops:
> can you provide a reliable test case that reproduces this issue?

```haxe
        if (flick != null){
            flick.stop();
        }

        final name = romGrp.members[select].name;

        if (name == 'empty') {
            romText.visible = true;
            flick = FlxFlicker.flicker(romText, 0.4 * 6, 0.4, false);
            return;
        }
```

that is the only place where the flick variable is used,
it gave me null object reference because the object was null,
and it was not

Geokureli:
> it gave me null object reference because the object was null, and it was not

It really seems like the object is null and flick is being stoped after its completed and destroyed. Note that even though 'romText' isnt null or destroyed, flick.object  can still be null. If you were to trace flick.object i bet its null.

Rather than storing flick and calling stop, i recomend just calling [FlxFlicker.stopFlickering](https://github.com/HaxeFlixel/flixel/blob/797fa1395c7fba4515dd6d8766f2d78f2948269f/flixel/effects/FlxFlicker.hx#L77) or check FlxFlicker.isFlickering, or just check `if(flick.object != null)`.

Typically using or calling a method on a destroyed object is a mistake and we want devs to know when they make a mistake, since silencing the error may add confusion about whats actually happening, and can cause other issues down the road based on false assumptions

Geokureli:
Closing, use the suggested fixes, above

PR #3052, opened 2024-02-29
Permit centering a sprite at its origin on a flxpath

47rooks:
Currently when a FlxSprite follows a FlxPath it cannot center itself on the path at its center of rotation. This can lead to some odd visuals in tilemaps or when the path itself is visible. This change attempts to maintain backward compatibility with the existing mechanism for choosing the center point while providing a new way to control the center point. This would support eventual deprecation of the autoCenter property in the future.

### DIFF
--- a/flixel/path/FlxPath.hx
+++ b/flixel/path/FlxPath.hx
@@ -387,8 +387,6 @@ class FlxPath implements IFlxDestroyable
 		}
 
 		// first check if we need to be pointing at the next node yet
-		_point.x = object.x;
-		_point.y = object.y;
 		_point = computeCenter(_point);
 
 		var node:FlxPoint = _nodes[nodeIndex];
@@ -424,7 +422,7 @@ class FlxPath implements IFlxDestroyable
 		if (object != null && speed != 0)
 		{
 			// set velocity based on path mode
-			computeCenter(_point);
+			_point = computeCenter(_point);
 
 			if (!_point.equals(node))
 			{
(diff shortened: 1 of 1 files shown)


47rooks:
These failures are on the nightly and look the same as those George posted about in discord a few days ago. Let me know if I need to do anything but for now it seems there is nothing for me to do on it.

Geokureli:
1. CI is failing because openfl's asset system isn't compatible with nightly haxe, no worries there

This is a great change in the right direction, but I think we should go about it in a different way, It's a bit complicated to explain so I may just make some commits

47rooks:
I'm happy either way so long as I can use the sprite origin to align the sprite on the path.

Geokureli:
would you be able to test this using whatever you were using before, and also test the new `sprite.path.centerMode = CUSTOM(sprite.origin);` as well?

47rooks:
I just tested this and it works fine in both my programs. CUSTOM is a nice touch. I was thinking about arbitrary placement but had not lit on this idea.

Geokureli:
i still have a test I wanna do but before i do that, @47rooks , what do you think of just having a path.centerOffset point, and removing the the centerMode enum all together?

47rooks:
So rather than rescaling the offset it becomes like a hitbox update - after scaling it's up to the caller to set it themselves ?

Geokureli:
I think there was a miscommunication, due to me rambling on and thinking out loud. I wanted to entertain the idea of just having an centerOffset value rather than the the enum, the issues I describe above were not issues with the approach used in this PR, but issues with that new idea. I just thought I should bring it up in case you had input, but I think the current change is the way to go.

I'll have this merged in a little bit

PR #3057, opened 2024-03-03
Add `drawOnTop` to PluginFrontEnd

### LINKED ISSUE #3054: (Enhancement) Implement some way to draw plugins on top of everything else
It would be nice to be able to have a plugin be drawn on top of all other objects, as it's currently forced to be drawn under everything else. For example, this can be used for a plugin that displays achievement objects on the screen (which is what I was trying to implement a while ago), an FPS/memory display, etc.

I'm not sure what's the best way to go about this, as FlxG.plugins takes any FlxBasic instance, and I think it'll be confusing to add a variable that only works for plugins.

Please let me know your thoughts on this suggestion.

Starmapo:
This PR adds a variable that allows plugins to be drawn on top of everything else, instead of under.

Closes #3054

### DIFF
--- a/flixel/FlxGame.hx
+++ b/flixel/FlxGame.hx
@@ -863,9 +863,16 @@ class FlxGame extends Sprite
 
 		FlxG.cameras.lock();
 
-		FlxG.plugins.draw();
-
-		_state.draw();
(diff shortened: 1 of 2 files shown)


Geokureli:
Thanks!

Issue #3034, opened 2024-02-15
[ENHANCEMENT] Support FlxGraphic.dump() to allow for GPU-only textures


EliteMasterEric:
FlxGraphic has a function available, `dump()`, which flags the sprite as dumped and removes the contents of `BitmapData` from RAM (while still holding a reference to the texture that has been uploaded to VRAM). This GREATLY reduces memory usage at the expense of being unable to draw on the graphic, which is still fine in a wide variety of use cases. Depending on the type of graphic I think this should probably even be the default in many cases, requiring the graphic to be explicitly "undumped" to be drawn on.

However, `dump()` relies on `bitmap.dumpBits()`, which is a function only available on `lime_legacy`. Implementing this would modern Lime would require a rewrite, as well as stability testing.

DigiEggz:
> The above functionality has been experimented with somewhat in FNF mod projects (resulting in massive decreases in memory), but having it available in the main branch would be a major improvement and at this point in time it is probably the most important change I want to see come to Flixel 6.

Can you link to a project that features this change? I'd like to take a look at implementing and testing whatever they have done, as there have also been some PRs in OpenFL that claim to [improve performance](https://github.com/openfl/openfl/pull/2707) when using disposed image bitmaps.

EliteMasterEric:
It appears Psych Engine has it: https://github.com/ShadowMario/FNF-PsychEngine/blob/65494a3375fbceb5d74cfd2231ba668790842a4f/source/backend/Paths.hx#L273

DigiEggz:
@EliteMasterEric Thanks for the quick reply. Do you have any starting point suggestions for how we could begin to test disposing bitmaps? I've been looking through graphics the code and am not sure how it compares to Psyche Engine's approach.

EliteMasterEric:
Currently, Flixel loads Bitmaps into RAM automatically. This is done to enable functionality to draw onto sprites. Simply overriding the default behavior would cause any applications using this to cease functioning.

I think the ideal implementation is one in which one can set a flag (similar to DEFAULT_PERSIST) which defines whether graphics are automatically disposed once uploaded to the GPU; you could then override the behavior on a case-by-case basis.

DigiEggz:
Can you point to toward any core locations or starting points where I could test out disposing of bitmaps, despite the immediate problems it will cause? I'm hunting for a good jumping off point but there are quite a few moving parts to the drawing system.

PR #3059, opened 2024-03-05
Remove unused legacy code & defines

Starmapo:
This PR removes a lot of code that goes unused in current versions of HaxeFlixel.

- Removed the `FLX_POST_PROCESS` define & its related code. This was only available in versions of OpenFL before 4.0.0, which is not supported in the latest HaxeFlixel (minimum OpenFL version is 9.2.2).
- Removed references to the `openfl_next`, `lime_legacy` &  `next` defines. These defines were [removed in the same commit as `openfl_legacy`](https://github.com/openfl/openfl/commit/15be4f5bd5eecbd8f0074ddfe966564d99ee1dc0), which was also already removed from HaxeFlixel in #2990.
- Removed Lime & OpenFL version checks that are always true/false due to the current minimum versions (for example, `#if (lime >= 7.0.0)` is always true due to the minimum version being 8.0.2).

There are a few breaking changes:
- The `PostProcess` class was deleted, which was still available when `FLX_POST_PROCESS` wasn't defined, though it was merely a placeholder. `addPostProcess` and `removePostProcess` were also removed from `FlxG`, having the same effect.
- `dump` and `undump` have been removed from `FlxGraphic`, due to `dump` only doing anything when `lime_legacy` was defined. Due to this, all variables and functions that made use of these were removed, including `FlxG.bitmap.onAssetsReload` which refreshed the assets by calling `undump` whenever a change event was dispatched in `openfl.utils.Assets`. However, I could not find a single instance in Lime or OpenFL when this event would actually be dispatched, so it doesn't seem to have any purpose unless the user is meant to dispatch it themselves.


Geokureli:
Because of the breaking changes, I've changed this PR so it will go to the 6.0.0 release branch rather than dev.

CI Error:
> source/MenuState.hx:128: characters 15-24 : flixel.system.frontEnds.BitmapFrontEnd has no field dumpCache
> FAILURE - TexturePackerAtlas (html5)

Starmapo:
Huh, I wasn't expecting there to be demos that use these features. TexturePackerAtlas was added 11 years ago, and PostProcess isn't even on the site currently. I'll make a PR over there fixing it.

Merged the release6 branch like you said 👍

Geokureli:
@EliteMasterEric was requesting expansion of the `FlxGraphic.dump` feature, here: #3034

this PR seems to directly conflict with that

flixel/FlxGame.hx
```diff
@@ -379,8 +344,6 @@ class FlxGame extends Sprite
 
 		// make sure the cursor etc are properly scaled from the start
 		resizeGame(FlxG.stage.stageWidth, FlxG.stage.stageHeight);
-
-		Assets.addEventListener(Event.CHANGE, FlxG.bitmap.onAssetsReload);
```
Geokureli, line 383: are we sure this stuff isn't ever used?

Starmapo:
```haxe
Assets.addEventListener(Event.CHANGE, FlxG.bitmap.onAssetsReload);
```

This essentially listens for the `lime.utils.Assets.onChange` event, which is dispatched whenever a library that has been registered dispatches its own `onChange` event. However, I can't find any instance of code that would dispatch `AssetLibrary.onChange` at all, so it seems to me like this is an unused feature.

Here are the search results for `onChange` in Lime: https://github.com/search?q=repo%3Aopenfl%2Flime%20onChange&type=code
OpenFL does not dispatch it either: https://github.com/search?q=repo%3Aopenfl%2Fopenfl+onChange&type=code

Geokureli: it might be worth going the the openfl forum or discord channel and asking, just to be sure

Starmapo: Got a reply on the Discord channel from Josh Tynjala, quote:

> perhaps it is to allow someone to subclass `AssetLibrary` and dispatch `onChange` manually.

So you were right, it seems like this should actually stay. However, `FlxGraphic.onAssetsReload` made use of `undump` to refresh the graphic.

To be clear on what `undump` does, it refreshes the bitmap if it has an `assetsKey` or `assetsClass`, so it still does something even if it wasn't previously dumped. Maybe we should bring it back but rename it appropiately like `refreshBitmap`, or just keep the code inside `onAssetsReload`?

Geokureli:
what is the plan, now?

Starmapo:
We have to keep `FlxG.bitmap.onAssetsReload` which in turn means we have to keep the `undump` function which refreshes the bitmap data, regardless of if it was dumped before. However since `dump` is getting removed I asked you if it should be renamed to something more appropiate like `refreshBitmap`, or if the code should just be kept inside `onAssetsReload`.

Geokureli:
> I asked you if it should be renamed to something more appropriate like `refreshBitmap`, or if the code should just be kept inside `onAssetsReload`.

Sorry, I missed that, yeah I think thats a good idea

Starmapo:
I decided to bring back `undump()` and `canBeDumped` renamed to `refreshBitmap()` and `canBeRefreshed` respectively, as well as the old ones but marked as deprecated, since it's _possible_ that someone was using them for the refreshing function alone. Do let me know if you have any more suggestions.

Geokureli:
Totally forgot about this, gonna look into it for the upcoming release. I backmerged release6, lmk if anything looks wrong

Starmapo:
It should be good now, hoping CI can get worked out soon

flixel/system/frontEnds/BitmapFrontEnd.hx
```diff
@@ -53,64 +53,13 @@ class BitmapFrontEnd
 		for (key in _cache.keys())
 		{
 			var obj = _cache.get(key);
-			if (obj != null && obj.canBeDumped)
+			if (obj != null && obj.canBeRefreshed)
 			{
 				obj.onAssetsReload();
 			}
 		}
 	}
```
Geokureli, line 62: Are we still unsure whether [this event](https://github.com/HaxeFlixel/flixel/blob/release6/flixel/FlxGame.hx#L387) is ever dispatched? do we know what happens when an asset is refreshed?

Geokureli (changes requested):

flixel/FlxGame.hx
```diff
@@ -472,11 +426,6 @@ class FlxGame extends Sprite
 		var width:Int = FlxG.stage.stageWidth;
 		var height:Int = FlxG.stage.stageHeight;
 
-		#if !flash
-		if (FlxG.renderTile)
-			FlxG.bitmap.onContext();
-		#end
-
```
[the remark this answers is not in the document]
Geokureli: nevermind I see onContext was removed

flixel/system/frontEnds/BitmapFrontEnd.hx
Starmapo: > Are we still unsure whether [this event](https://github.com/HaxeFlixel/flixel/blob/release6/flixel/FlxGame.hx#L387) is ever dispatched?

After I asked in the discord server last year, it seems like its intended use is to be dispatched manually by extending AssetLibrary, so it is possible, albeit not with the out-of-the-box assets system.

flixel/graphics/FlxGraphic.hx
```diff
@@ -416,58 +413,31 @@ class FlxGraphic implements IFlxDestroyable
...
-			isDumped = true;
-		}
-		#end
-	}
-
-	/**
-	 * Undumps bits of the `BitmapData` - regenerates it and regenerate tilesheet data for this object
-	 */
-	public function undump():Void
+	public function refreshBitmap():Void
```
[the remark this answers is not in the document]
Starmapo: Done!

flixel/system/frontEnds/BitmapFrontEnd.hx
Geokureli: What does it do to the assets when dispatched? Also, under what circumstances should a graphic be refreshed

Starmapo: It refreshes whichever graphics have an `assetsKey` or `assetsClass` by retrieving the bitmap data from the assets again and switching to it.
I imagine its only useful if you want to hot-reload graphics that have been modified in some way (either externally or internally with libraries)

Geokureli (approved):

Geokureli:
Thanks!

PR #3060, opened 2024-03-06
deprecate dumping and post processing for #3059

Geokureli:
@Starmapo am I missing anything? You'll need to merge dev into your branch once this is merged, and you'll likely have merge conflicts


Starmapo:
You've gotten all the public ones, but there are a few more private variables: `_onFocusFiredOnce`, `postProcessLayer` & `postProcesses` in FlxGame.

rodney528:
Why is this change occurring? I don't really know what the post processing stuff is supposed to do and the dump stuff sounds like its vital but it gets called separately from destroy??

ACrazyTown:
> Why is this change occurring? I don't really know what the post processing stuff is supposed to do and the dump stuff sounds like its vital but it gets called separately from destroy??

These are legacy methods that don't really do anything nowadays

rodney528:
> > Why is this change occurring? I don't really know what the post processing stuff is supposed to do and the dump stuff sounds like its vital but it gets called separately from destroy??
> 
> These are legacy methods that don't really do anything nowadays

ah, ok

PR #3063, opened 2024-03-07
FlxCamera.stopFade, stopFlash, and stopShake

CaptainBaldi:
Allows stopping of specific camera effects.

### DIFF
--- a/flixel/FlxCamera.hx
+++ b/flixel/FlxCamera.hx
@@ -1603,29 +1603,41 @@ class FlxCamera extends FlxBasic
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


CaptainBaldi:
> I'd rather just have a stop function for each effect, than an arg

Done

Geokureli:
Thanks!

CaptainBaldi:
@Geokureli when adding the new functions I forgot to remove the addition of `StringTools` used for the previous implementation, will this be a problem?

Geokureli:
done

PR #3065, opened 2024-03-10
Remove (now redundant) health variable and function in FlxObject

moxie-coder:
I couldn't find any duplicates to this, but if you wanna correct me on this Geokureli, you can, but this just seems pretty redundant now, this was probably kept for backwards compatibility or legacy reasons, but it seems pretty useless now since you could just extend FlxObject to readd it or something similar like that.

There are many other ways of doing it though, so I feel like this really isn't needed now, especially since it was last updated around **11 years ago** and I don't think this is really used a lot, or just not at all in contrast to now

### DIFF
--- a/flixel/FlxObject.hx
+++ b/flixel/FlxObject.hx
@@ -700,11 +700,6 @@ class FlxObject extends FlxBasic
 	public var maxAngular:Float = 10000;
 
 	/**
-	 * Handy for storing health percentage or armor points or whatever.
-	 */
-	public var health:Float = 1;
-
-	/**
 	 * Bit field of flags (use with UP, DOWN, LEFT, RIGHT, etc) indicating surface contacts. Use bitwise operators to check the values
 	 * stored here, or use isTouching(), justTouched(), etc. You can even use them broadly as boolean values if you're feeling saucy!
 	 */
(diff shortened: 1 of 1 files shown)


CaptainBaldi:
if this were to go through, it would probably be best to deprecate these for now, and then remove them entirely in flixel 6.0.0
overall i agree with this

Geokureli:
yeah we can't just remove fields because it'll break everyone's project. first deprecate it, and then we'll talk about a good way to remove this without adding more friction to 6.0.0, which already worries me

moxie-coder:
> yeah we can't just remove fields because it'll break everyone's project. first deprecate it, and then we'll talk about a good way to remove this without adding more friction to 6.0.0, which already worries me
> 
> Edit: @moxie-coder how is it "redundant"?

because you could just add it via extending the class, or making your own class too, there’s plenty of ways to add it in, and this felt like a leftover from the early days

Geokureli:
> because you could just add it via extending the class, or making your own class too, there’s plenty of ways to add it in, and this felt like a leftover from the early days

Odd semantics, but not really "redundant", we are removing a feature so that it can be added back differently, if needed. I was also thrown off by "*now* redundant" as if something recently changed with this

moxie-coder:
> > because you could just add it via extending the class, or making your own class too, there’s plenty of ways to add it in, and this felt like a leftover from the early days
> 
> Odd semantics, but not really "redundant", we are removing a feature so that it can be added back differently, if needed. I was also thrown off by "_now_ redundant" as if something recently changed with this

sorry, I probably could’ve worded it better

PR #3066, opened 2024-03-10
Support for loading `FlxGraphic`'s through `FlxMouse.load`

CaptainBaldi:
I was working on a project and I noticed that `FlxMouse` doesn't natively support loading `FlxGraphic`s, I found this odd, it supports `BitmapData` and even custom classes, why not the main one for Flixel graphics?


Geokureli:
I think accepting graphics is misleading and will make people think the mouse supports animations or sub-frames, when it doesn't, I suggest just passing in graphic.bitmap rather than accepting graphics

PR #3067, opened 2024-03-11
add define to remove health fields

Geokureli:
Continuation of #3065

I wasn't happy with the deprecation message of "you don't need it anymore, because you can extend", that's not what redundant means, the field should have never been added, to begin with.

I've added the defines: FLX_NO_HEALTH to remove it, and it's inversion: FLX_HEALTH. this way people can remove it in their projects now and add their own health system. when flixel 6.0.0 is released FLX_NO_HEALTH will be true by default, and FLX_HEALTH will be it's inversion so people can easily fix old projects

### DIFF
--- a/flixel/FlxObject.hx
+++ b/flixel/FlxObject.hx
@@ -699,11 +699,13 @@ class FlxObject extends FlxBasic
 	 */
 	public var maxAngular:Float = 10000;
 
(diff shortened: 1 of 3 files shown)


moxie-coder:
that’s honestly more better then what I did, my bad for the original message on it 😅

PR #3073, opened 2024-03-11
Turn `FlxTextBorderStyle` into a abstract enumerator and document variables in `FlxTextAlign`

crowplexus:
## What

---

This change transforms the `FlxTextBorderStyle` enumerator into an abstract one, allowing users to define a `"stringValue"` in order to set the Border Style of a FlxText.

along with this change I also gave simple comments to `FLxTextAlign`'s variables for consistency sake (because I really bet this is gonna clear up confusion in a way)

## Why

---

This change was made particularly to accomodate to HScript users, who due to limitations, cannot use the enum without going as far as setting it as a variable on scripts.

It also sort of made me confused as for why `FlxTextAlign` is a enum abstract but not `FlxTextBorderStyle`

### DIFF
--- a/flixel/text/FlxText.hx
+++ b/flixel/text/FlxText.hx
@@ -1230,39 +1230,48 @@ class FlxTextFormatMarkerPair
 	}
 }
 
-enum FlxTextBorderStyle
+enum abstract FlxTextBorderStyle(String) from String
 {
-	NONE;
+	var NONE = null;
 
 	/**
 	 * A simple shadow to the lower-right.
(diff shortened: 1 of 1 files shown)


Geokureli:
I get conflicting reports on whether enums are "better" for hscript users, in the end you can use either, and changing enums to abstracts is a loss of functionality only situationally benefitial

Geokureli:
Last time someone tried to make this change:
https://github.com/HaxeFlixel/flixel/pull/2712

PR #3086, opened 2024-03-15
add FlxTween.flicker

Geokureli:
FlxTween based flicker tool. FlxFlicker isn't being deprecated but I don't recommend using it unless someone has a good reason

additions and features:
- Add `flicker`, `isFlickering` and `stopFlickering` instance methods to `FlxTweenManager` and static methods to `FlxTween`
- Allows for custom `ratio`: rather than 50-50 split of on/off, can do 80-20 or whatever
- Allows custom tweenFunction, meaning you can alter the parameters over time, and more

~~To do: add unit test~~

### DIFF
--- a/flixel/tweens/misc/FlickerTween.hx
+++ b/flixel/tweens/misc/FlickerTween.hx
@@ -101,6 +101,7 @@ class FlickerTween extends FlxTween
 		if (tweenFunction != null)
 		{
 			final visible = tweenFunction(this);
+			// do not call setter every frame
 			if (basic.visible != visible)
 				basic.visible = visible;
 		}
--- a/flixel/tweens/misc/FlickerTween.hx
+++ b/flixel/tweens/misc/FlickerTween.hx
@@ -122,7 +122,7 @@ class FlickerTween extends FlxTween
 	 * The default tween function of flicker tweens
 	 * @param   tween  The tween handling the flickering
 	 */
-	static public function defaultTweenFunction(tween:FlickerTween)
+	public static function defaultTweenFunction(tween:FlickerTween)
 	{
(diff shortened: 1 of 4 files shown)


flixel/tweens/misc/FlickerTween.hx
```diff
@@ -0,0 +1,130 @@
...
+	override function isTweenOf(object:Dynamic, ?field:String):Bool
+	{
+		return basic == object && (field == null || field == "visible" || field == "flicker");
+	}
+	
+	/**
+	 * The default tween function of flicker tweens
+	 * @param   tween  The tween handling the flickering
+	 */
+	public static function defaultTweenFunction(tween:FlickerTween)
```
MondayHopscotch, line 126: The rest of the functions here have explicit returns types on the signature. What is the flixel guideline for return types in function signatures?

flixel/tweens/misc/FlickerTween.hx
```diff
@@ -0,0 +1,130 @@
...
+	{
+		return (tween.time / tween.period) % 1 > tween.ratio;
```
MondayHopscotch, line 128: I was curious how you approached handling the ratio. This is pleasantly simple 👍

flixel/tweens/misc/FlickerTween.hx
Geokureli: there is no rule on this, I kinda prefer not having it, in more cases, especially here where `tweenFunction` is well defined

flixel/tweens/misc/FlickerTween.hx
Geokureli: at first I had `> 0.5` then i wondered what other values would look like

PR #3096, opened 2024-03-29
Add support for V1 Sparrow atlases + fix intentional blank instances

Cheemsandfriends:
In Adobe Animate, there's two versions of Sparrow, that basically work out as the same thing, but the only difference is that the width and height attributes are shortened by their initials, As can be shown here:

| Spritesheet exported with Sparrow V1: | Spritesheet exported with Sparrow V2:
| --- | ---
|  | 

(Also something to mention, Sparrow V1 doesn't seem to support a trimmed version, aka "frameX" and "frameY")

Plus it fixes a common issue that occurs with debugging, And that is when a spritesheet presents a 0x0 instance frame. This is due to a way to simbolise that the frame its using, its an empty one, but flixel doesn't seem to push those frames, causing sometimes massive errors all over the screen. This also fixes it.

### DIFF
--- a/flixel/graphics/frames/FlxAtlasFrames.hx
+++ b/flixel/graphics/frames/FlxAtlasFrames.hx
@@ -254,15 +254,16 @@ class FlxAtlasFrames extends FlxFramesCollection
 
 		for (texture in data.nodes.SubTexture)
 		{
+			var version2 = texture.has.width; 
 			var name = texture.att.name;
 			var trimmed = texture.has.frameX;
-			var rotated = (texture.has.rotated && texture.att.rotated == "true");
+			var rotated = (texture.has.rotated && texture.att.rotated == "true");-
 			var flipX = (texture.has.flipX && texture.att.flipX == "true");
 			var flipY = (texture.has.flipY && texture.att.flipY == "true");
 
-			var rect = FlxRect.get(Std.parseFloat(texture.att.x), Std.parseFloat(texture.att.y), Std.parseFloat(texture.att.width),
-				Std.parseFloat(texture.att.height));
-
+			var rect = FlxRect.get(Std.parseFloat(texture.att.x), Std.parseFloat(texture.att.y), Std.parseFloat((version2) ? texture.att.width : texture.att.w),
+				Std.parseFloat((version2) ? texture.att.height : texture.att.h));
+			
 			var size = if (trimmed)
 			{
 				new Rectangle(Std.parseInt(texture.att.frameX), Std.parseInt(texture.att.frameY), Std.parseInt(texture.att.frameWidth),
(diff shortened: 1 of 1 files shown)


Geokureli:
I was wondering about this, do you actually know anyone that use sparrow v1? is there a program out there that is only capable of producing v1? would it be just as effective and useful if you detected v1 and threw an error saying "Outdated sparrow version detected, be sure to export using sparrow v2 in your programs settings"?

Cheemsandfriends:
I don't see why not implement it, its just changing looking two values that are shortened basically, and to go through the hassle of throwing an error to detect an older version seems to much than just support it.

(also, I tried to find a texture packer that exported Sparrow V1s, but I struggled to even find packers that exported Sparrow itself aside from Adobe Animate)

Geokureli:
It's my policy to not add solutions to problems nobody is having, [this is open source 101](https://docs.godotengine.org/en/stable/contributing/development/best_practices_for_engine_contributors.html#the-problem-always-comes-first). Is allowing people to use v1 really helping anyone? or is it potentially confusing people about the capabilities of sparrow

Geokureli:
Can you give a small example I can use to reproduce the other issue?

Cheemsandfriends:
Surely!
I made this in a whim trying to test it myself. Don't get confused, Man hi 2 is in fact Sparrow v1 and Man hi is Sparrow v2

[Man hi.zip](https://github.com/HaxeFlixel/flixel/files/14951033/Man.hi.zip)

Cheemsandfriends:
> It's my policy to not add solutions to problems nobody is having, [this is open source 101](https://docs.godotengine.org/en/stable/contributing/development/best_practices_for_engine_contributors.html#the-problem-always-comes-first). Is allowing people to use v1 really helping anyone? or is it potentially confusing people about the capabilities of sparrow. If I'm wrong here, I wanna hear it from someone who would actually use this fix


`fromSparrow` inherently doesn't tell that its a specific version, any user could see the name, read a bit the description and assume "Well, I suppose it has to support both versions right?" Also, the fact that there are two versions imply that Sparrow (The game engine for iOS) had once used this format in early versions. Throwing an error to them could potentially make the user believe the function does not work at all (which, I reiterate again, why even do it in the first place since we're not saying that the changes are astronomically different).

Cheemsandfriends:
I suppose the solution would be to change the description to specify which version should be used, but I'm genuinely not keen on doing that myself, plus maybe throwing an error is a tad harsh?

Geokureli:
>  Throwing an error to them could potentially make the user believe the function does not work at all
afaik the only way to export v1 is to select it in a dropdown in adobe animate which also has v2, the error specifically says "Use sparrow V2"

but also the fact that no one has ever complained about a crash with fromSparrow with the error "could not find attribute 'width'" makes me think this is a non issue

PR #3100, opened 2024-04-07
FlxMath type parameters

VMan-2002:
Add type parameters in FlxMath functions so they can use both Int and Float


VMan-2002:
okay this would need to be figured out
```
public static function maxAdd<T:(Float & Int)>(value:T, amount:T, max:T, ?min:T = 0):T
                                                                                  ^ here
```
this error appears a few times, where a default value is unable to fit all available types. idk how to fix this or if it even is fixable

Geokureli:
please test your code before making PRs the syntax you're trying is not valid also this is a breaking change for very little benefit

PR #3102, opened 2024-04-10
Updated the URL to HAxeFlixel site in the debugger

CharcoalStyles:
the link was pointing to www.haxeflixel.com which now seems to not work.

### DIFF
--- a/flixel/system/debug/FlxDebugger.hx
+++ b/flixel/system/debug/FlxDebugger.hx
@@ -509,7 +509,7 @@ class FlxDebugger extends openfl.display.Sprite
 
 	inline function openHomepage():Void
 	{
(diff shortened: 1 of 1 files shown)


Geokureli:
Thanks!

moxie-coder:
I’m thinking some bad actors bought that domain since it redirects to probably some malware site but it’s broken

ACrazyTown:
> I’m thinking some bad actors bought that domain since it redirects to probably some malware site but it’s broken

haxeflixel.com works fine
It's just www.haxeflixel.com, probably just incorrectly setup subdomain/dns/whatever

Geokureli:
we had to make this same change in a few other places long before the domain issue popped up. looks like www just isn't a thing anymore?

PR #3105, opened 2024-04-13
Prevent a requirement for `@:privateAccess`-ing the _customSoundTray object.

saturn-volv:
...And also allowing for it to be changed at runtime.

### Access
Before, to change the sound tray, you would have to access the variable as such:
```haxe
var game:FlxGame;

//...

@:privateAccess(flixel.FlxGame)
game._customSoundTray = CustomSoundTray;
```
And your access would have to be called *before* the game was created.

Now, you can change the sound tray at *any* time, through the method of:
```haxe
FlxG.game.changeSoundTray(CustomSoundTray);
```

Now as long as your `CustomSoundTray` was a child of `flixel.system.ui.soundFlxSoundTray`, it would work.

### DIFF
--- a/flixel/FlxGame.hx
+++ b/flixel/FlxGame.hx
@@ -176,7 +176,16 @@ class FlxGame extends Sprite
 	 * Change this after calling `super()` in the `FlxGame` constructor
 	 * to use a customized sound tray based on `FlxSoundTray`.
 	 */
-	var _customSoundTray:Class<FlxSoundTray> = FlxSoundTray;
+	@:isVar var _customSoundTray(default, set):Class<FlxSoundTray>;
+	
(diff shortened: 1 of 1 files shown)


DetectiveBaldi:
i would say it would be better to deprecate `_customSoundTray`, make a new public variable (`customSoundTray`), and then make `_customSoundTray` a getter / setter for it. having a function to change one line feels a bit redundant

Geokureli:
I think the idea here is to extend FlxGame with a class that sets these custom classes in create

saturn-volv:
ahh I've got you now. yeah extending FlxGame seems like a much better idea rather than adding all this. But yeah coming back to it in the future is definitely a good idea, I'll look more into a design for it and come back to this PR then.

PR #3106, opened 2024-04-13
Calculate adjustedLerp only when necessary

richTrash21:
Since followLerp bound between 0 and 1 in updateLerp, let lerp be adjusted only when bound value < 1


Geokureli:
I don't understand the concern

PR #3119, opened 2024-04-17
Fix `FlxKey` maps not containing `FlxKey.NONE`

DetectiveBaldi:
A check in `FlxMacroUtil.buildMap` causes `FlxKey.NONE` to be excluded from these maps. This is a quick fix to solve this issue.


Geokureli:
this change causes null refs, and even if that's fixed I'm pretty sure existing uses of this will need to be changed to manually exclude NONE

DetectiveBaldi:
> this change causes null refs, and even if that's fixed I'm pretty sure existing uses of this will need to be changed to manually exclude NONE
> 
> why don't we make a new function that defaults exclude to [] and have this old func call that with NONE

Wouldn't it better to just have the `FlxKey` maps set exclude to [] instead of making a new function?
Otherwise, any suggestions for what to call the function? I'm not too good with these things 😭

Geokureli:
> Wouldn't it better to just have the `FlxKey` maps set exclude to [] instead of making a new function? 

It can also break other peoples usage of this function

> Otherwise, any suggestions for what to call the function? I'm not too good with these things 😭

The 3 hardest parts about programming are naming things and off-by-one errors

DetectiveBaldi:
I made a change to add back the null check and simply just set the exclude parameter in `FlxKey`. This works as expected and doesn't seem to cause any other issues

Geokureli:
> A check in `FlxMacroUtil.buildMap` causes `FlxKey.NONE` to be excluded from these maps. This is a quick fix to solve this issue.

Can I get you're use case for this? is it that you want `FlxG.key.anyPressed(["NONE"]);` or something? I'm having a hard time understanding a scenario like this

DetectiveBaldi:
> > A check in `FlxMacroUtil.buildMap` causes `FlxKey.NONE` to be excluded from these maps. This is a quick fix to solve this issue.
> 
> Can I get you're use case for this? is it that you want `FlxG.key.anyPressed(["NONE"]);` or something? I'm having a hard time understanding a scenario like this

I had a function which casted `FlxG.keys.firstJustPressed` to a string value, and returned it. I noticed this function would never return `NONE`, which required me to make an additional check for it.
I don't see why this would be intended behavior, and it's a simple fix

Geokureli:
Thanks!

PR #3121, opened 2024-04-21
Minor FlxDrawTrianglesItem optimization

MaybeMaru:
Uses openfl's vector length function instead of splice for a minor performance boost clearing the vectors.

### DIFF
--- a/flixel/graphics/tile/FlxDrawTrianglesItem.hx
+++ b/flixel/graphics/tile/FlxDrawTrianglesItem.hx
@@ -101,10 +101,17 @@ class FlxDrawTrianglesItem extends FlxDrawBaseItem<FlxDrawTrianglesItem>
 	override public function reset():Void
 	{
 		super.reset();
+		#if (flash || openfl >= "4.0.0")
+		vertices.length = 0;
+		indices.length = 0;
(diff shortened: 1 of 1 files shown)


flixel/graphics/tile/FlxDrawTrianglesItem.hx
```diff
@@ -101,10 +101,17 @@ class FlxDrawTrianglesItem extends FlxDrawBaseItem<FlxDrawTrianglesItem>
 	override public function reset():Void
 	{
 		super.reset();
+		#if (flash || openfl >= "4.0.0")
```
Geokureli, line 104: we don't support openfl versions older than 9.3.2, we plan to remove the other checks like this, soon. So let's not add more

MaybeMaru: should i remove the check then?

Geokureli: yes, please

MaybeMaru: there we go

Geokureli:
Thanks!

PR #3127, opened 2024-04-24
Fixed typo in FlxMouseEventManager.hx

TheGaloXx:
"Callback when mouse is this object." -> "Callback when mouse is over this object."


Geokureli:
Thanks!

PR #3142, opened 2024-05-16
Fix FlxBGSprite render on camera zoom != 1

richTrash21:
additionally skip draw call if alpha is 0

### DIFF
--- a/flixel/system/FlxBGSprite.hx
+++ b/flixel/system/FlxBGSprite.hx
@@ -20,6 +20,9 @@ class FlxBGSprite extends FlxSprite
 	@:access(flixel.FlxCamera)
 	override public function draw():Void
 	{
+		if (alpha == 0.0)
+			return;
+
 		for (camera in getCamerasLegacy())
 		{
 			if (!camera.visible || !camera.exists)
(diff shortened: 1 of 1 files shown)


Geokureli:
can I get a simple scenario that showcases this change?

flixel/system/FlxBGSprite.hx
```diff
@@ -20,6 +20,9 @@ class FlxBGSprite extends FlxSprite
 	@:access(flixel.FlxCamera)
 	override public function draw():Void
 	{
+		if (alpha == 0.0)
+			return;
+
```
Geokureli, line 25: I see no reason for this class to check this when all other FlxSprites do not. This change will only serve to confuse people who expect this to behave exactly like other FlxSprites. I also would not want this change made in FlxSprite, I recommend either setting visible to false, manually to prevent draw calls, or making an extending class in your program that has this change

```suggestion
```

Geokureli: also this wouldn't have had any effect, since the sprite's alpha is typically 1.0, even with a transparent bgColor. Changing the substate's bgColor only changes the sprite's pixel color

Geokureli:
Thanks!

Issue #3147, opened 2024-05-19
[ENHANCEMENT] Make SoundFrontEnd.volumeHandler a FlxSignal


ACrazyTown:
Turning volumeHandler into a FlxSignal would make it possible to dispatch multiple volume change callbacks at the same time. For example it would be useful in a case where a video player library wants to get volume changes without stealing the functionality of getting a volume callback from the user

I could and would make a pull request for this but I'm not sure how to handle a breaking change like this. Would this be a new field? Would volumeHandler get deprecated and eventually replaced with the new field?

PR #3148, opened 2024-05-20
Add onVolumeChange to SoundFrontEnd

ACrazyTown:
info: #3147 

This acts the same as volumeHandler except it's just a FlxSignal. Not sure what to do with volumeHandler, should it be marked as deprecated?

### DIFF
--- a/flixel/system/frontEnds/SoundFrontEnd.hx
+++ b/flixel/system/frontEnds/SoundFrontEnd.hx
@@ -1,5 +1,6 @@
 package flixel.system.frontEnds;
 
+import flixel.util.FlxSignal.FlxTypedSignal;
 #if FLX_SOUND_SYSTEM
 import flixel.FlxG;
 import flixel.group.FlxGroup;
@@ -36,6 +37,11 @@ class SoundFrontEnd
 	 * Function should take the form myVolumeHandler(volume:Float).
 	 */
 	public var volumeHandler:Float->Void;
+
+	/**
+	 * A signal that gets dispatched whenever the volume changes.
+	 */
+	public var onVolumeChange(default, null):FlxTypedSignal<Float->Void> = new FlxTypedSignal<Float->Void>();
 
 	#if FLX_KEYBOARD
 	/**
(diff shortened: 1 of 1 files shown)


Geokureli:
can you walk me through what the existing field `volumeHandler` is, and how it was intended to be used?

ACrazyTown:
> can you walk me through what the existing field `volumeHandler` is, and how it was intended to be used? This is the first time I'm coming across it

`volumeHandler` is just a function callback that gets called whenever the global volume is changed. It's meant to be replaced by your own function like this:
```haxe
FlxG.sound.volumeHandler = function(volume:Float) {
   // do something when the volume is changed
};
```

Geokureli:
Should we deprecate it?

ACrazyTown:
> Should we deprecate it?

I don't see why not

flixel/system/frontEnds/SoundFrontEnd.hx
```diff
@@ -36,6 +36,7 @@ class SoundFrontEnd
 	 * Set this hook to get a callback whenever the volume changes.
 	 * Function should take the form myVolumeHandler(volume:Float).
 	 */
+	@;deprecated("volumeHandler is deprecated, use onVolumeChange, instead")
```
ACrazyTown: there was a typo here, i just pushed a commit to fix it

Geokureli: D'oh!

Geokureli:
It's still giving deprecation warnings:
>  SoundFrontEnd.hx:343: characters 7-20 : Warning : volumeHandler is deprecated, use onVolumeChange, instead
> SoundFrontEnd.hx:345: characters 4-17 : Warning : volumeHandler is deprecated, use onVolumeChange, instead
> SoundFrontEnd.hx:466: characters 7-20 : Warning : volumeHandler is deprecated, use onVolumeChange, instead
> SoundFrontEnd.hx:468: characters 4-17 : Warning : volumeHandler is deprecated, use onVolumeChange, instead

I'll have to mess with this locally, when i have time later, unless you can figure it out, first

ACrazyTown:
I don't seem to be getting any deprecation warnings. I tried building the TurnBasedRPG demo on hl & html5, debug and release

Geokureli:
I was looking at the 4.2.5 job, which doesn't have the ability to suppress warnings in methods

PR #3155, opened 2024-05-30
Better description for `persistentUpdate & persistentDraw` in `FlxState`

Blue2359:
Tiny updated description documentation for persistentUpdate & persistentDraw in FlxState, so that they try to make more sense, and get to be understandable better.

This also fixes the "update" part(s) in persistentDraw's description, when they instead should be using the word "draw(n)".

### DIFF
--- a/flixel/FlxState.hx
+++ b/flixel/FlxState.hx
@@ -33,7 +33,7 @@ class FlxState extends FlxContainer
 	 *
 	 * By default, this is set to `true`, so the background states will continue to be "drawn" behind the current state.
 	 *
-	 * If the background states are not supposed to be `visible` when you have a different state on top,
+	 * If you do not want background states to be `visible` when you have a different state on top,
 	 * then you should set this to `false` for improved performance.
 	 */
 	public var persistentDraw:Bool = true;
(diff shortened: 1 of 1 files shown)


flixel/FlxState.hx
```diff
@@ -18,21 +18,23 @@ import flixel.util.typeLimit.NextState;
...
-	 * By default this is `true`, so background states will continue to be drawn behind the current state.
+	 * Determines whether the current state is drawn, even when it is not the active state.
+	 * For example, if you have your game state open first, and then you push a pause state on top of it,
+	 * if this is set to `true`, the game state would still continue to be drawn behind that pause state.
+	 *
+	 * By default, this is set to `true`, so the background states will continue to be "drawn" behind the current state.
 	 *
-	 * If background states are not `visible` when you have a different state on top,
-	 * you should set this to `false` for improved performance.
+	 * If the background states are not supposed to be `visible` when you have a different state on top,
```
Geokureli, line 36:
```suggestion
	 * If you do not want background states  to be `visible` when you have a different state on top,
```

Geokureli:
Thanks!

PR #3156, opened 2024-05-30
Fix path in include.xml

ACrazyTown:
lime.software has been dead for a while now, also bumps up the version.


Geokureli:
Thanks!

PR #3157, opened 2024-05-30
Fix logo link in README.md

ACrazyTown:
The HaxeFlixel website update moved some files around which caused the logo image in the README to point to a dead link. This likely needs to be done on all the other flixel repositories too


Geokureli:
Can't believe I didn't notice this

PR #3158, opened 2024-05-31
Major change to FlxTilemap/FlxTiles Improve collision, debug drawing, add various features

Geokureli:
### Tiles
- New dynamic `orient` method in FlxTile, used to position and resize the tile based on the grid location it is meant to represent in world space
- New dynamic `overlapsObject` method in FlxTile, can be extended or set to allow custom overlap detection for tiles whose hit shape is smaller than the tileGrid

### Tilemaps
- New `forEachOverlappingTile` method in FlxTilemap, to retrieve every tile that is overlapping the given object
- Using `forEachOverlappingTile` in `overlapsWithCallbacks`
- Alters the `createTile` method (also added in 5.9.0 and therefore not released) to remove the `allowCollisions` and `visible` args, it is recommended to simply reference the map's `drawIndex` and `collideIndex` in your override if you need these to your tile instances
- Added new `isOverlappingTile` method, allows you to check all tiles overlapping an object
- Added new `forEachOverlappingTile` calls a method with every tile that overlaps an object
- Added new `processOverlaps` to replace the now deprecated `overlapsWithCallbacks`
  - Eschews `flipCallbackParams` arg, allowing better typing of both callback params
  - Adds `isCollision` flag to control whether the Tiles' collision callbacks are fired and allows for processing non-solid tiles 
- Added new helpers: `getMapIndex`, `getRow`, `getColumn`, `getTileIndex`, `getTileData`, `tileExists`, `setTileIndex`, `getColumnAt`, `getRowAt`, `columnExists` and `rowExists`

### Objects
- Add internal helper `processCheckTilemap` for calling overlap checking utils between some combination of objects and tilemaps
- Add helper methods for `separate` and `updateTouchingFlags` 

### Debug Drawing
- Checks the previously unused `tile.ignoreDrawDebug` when debug drawing tiles
- Checks `tile.debugBoundingBoxColor` in debug drawing, if null, the map's debug drawing properties is used, as usual
- New `getDebugBoundingBoxColor` in FlxObject. Meant to replace `drawDebugBoundingBox`, will deprecate later

## (semi) Breaking changes
Previously, overlaps with callback would call the callback regardless of whether the tile overlapped the object. This change only effects objects placed **exactly** on the edge of a tile, and typically it was used to call separate, which would not do any separation in this case. this is technically a breaking change, but the case is so rare I'm going to add this on a minor release.

## TODO
 - ~~New unit tests~~

### DIFF
--- a/flixel/FlxObject.hx
+++ b/flixel/FlxObject.hx
@@ -191,6 +191,10 @@ class FlxObject extends FlxBasic
 	static function processCheckTilemap<T1:FlxObject, T2:FlxObject>
 		(object1:T1, object2:T2, func:(T1, T2)->Bool, isCollision = true):Bool
 	{
+		// two immovable objects cannot collide
+		if (isCollision && object1.immovable && object2.immovable)
+			return false;
+		
 		// If one of the objects is a tilemap, just pass it off.
 		if (object1.flixelType == TILEMAP)
 		{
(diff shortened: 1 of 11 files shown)


Geokureli:
this seems ready, still nervous as it's pretty huge (that's what she said)

gonna look over it tomorrow with fresh eyes

Geokureli:
This introduced a bug with jumping, details to come

PR #3160, opened 2024-06-01
Update documentation for `NextState`

Blue2359:
I noticed that the `NextState.hx` file had some tiny errors in it's documentation, so i went ahead and fixed them.

### DIFF
--- a/flixel/util/typeLimit/NextState.hx
+++ b/flixel/util/typeLimit/NextState.hx
@@ -3,12 +3,12 @@ package flixel.util.typeLimit;
 import flixel.FlxState;
 
 /**
- * A utility type that allows methods to accept multiple types, when dealing with "future" F`lxStates`.
- * Prior to haxeFlixel 6, `FlxG.switchState` and other similar methods took a `FlxState` instance
- * which meant FlxStates were instantiated before the previous state was destroyed, potentially
- * causing errors. It also meant states with args couldn't be reset via FlxG.resetState. In version
- * 5.6.0 and higher, these methods now take a function that returns a newly created instance. This
- * allows the state's instantiation to happen after the previous state is destroyed.
+ * A utility type that allows methods to accept multiple types, when dealing with "future" `FlxStates`.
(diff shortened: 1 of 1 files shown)


flixel/util/typeLimit/NextState.hx
```diff
@@ -3,12 +3,12 @@ package flixel.util.typeLimit;
...
- * which meant FlxStates were instantiated before the previous state was destroyed, potentially
- * causing errors. It also meant states with args couldn't be reset via FlxG.resetState. In version
- * 5.6.0 and higher, these methods now take a function that returns a newly created instance. This
- * allows the state's instantiation to happen after the previous state is destroyed.
+ * A utility type that allows methods to accept multiple types, when dealing with "future" `FlxStates`.
+ * Prior to HaxeFlixel 6, `FlxG.switchState()` and other similar methods took a `FlxState` instance,
+ * which meant `FlxStates` were instantiated before the previous state was destroyed, potentially
+ * causing errors. It also meant that states with args couldn't be reset via `FlxG.resetState()`. In version
+ * 5.6.0 and higher, these methods now take a function that returns a newly created state instance. This
+ * also allows the state's instantiation to happen after the previous state is destroyed.
```
Geokureli, line 11:
```suggestion
 * A utility type that allows methods to accept multiple types, when dealing with "future" `FlxStates`.
 * Prior to HaxeFlixel 6, `FlxG.switchState` and other similar methods took a `FlxState` instance,
 * which meant `FlxStates` were instantiated before the previous state was destroyed, potentially
 * causing errors. It also meant that states with args couldn't be reset via `FlxG.resetState`. In version
 * 5.6.0 and higher, these methods now take a function that returns a newly created state instance. This
 * allows the state's instantiation to happen after the previous state is destroyed.
```
1. don't use () to talk about methods unless you're giving an example usage, we already call them methods so people know they are called with ().
2. Disagree with the use of "also" since it is not "in addition to" anything, it is the effect and goal of the previously stated change

Geokureli:
Thanks, again!

PR #3163, opened 2024-06-04
Fix for brocken hitboxes, when scale < 0

PurSnake:
Makes _scaledOrigin values always positive

### DIFF
--- a/flixel/FlxSprite.hx
+++ b/flixel/FlxSprite.hx
@@ -1300,7 +1300,7 @@ class FlxSprite extends FlxObject
 		newRect.setPosition(x, y);
 		if (pixelPerfectPosition)
 			newRect.floor();
-		_scaledOrigin.set(origin.x * scale.x, origin.y * scale.y);
+		_scaledOrigin.set(origin.x * Math.abs(scale.x), origin.y * Math.abs(scale.y));
 		newRect.x += -Std.int(camera.scroll.x * scrollFactor.x) - offset.x + origin.x - _scaledOrigin.x;
 		newRect.y += -Std.int(camera.scroll.y * scrollFactor.y) - offset.y + origin.y - _scaledOrigin.y;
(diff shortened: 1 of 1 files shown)


Geokureli:
can you provide an small example of when the old code fails and the new code behaves as expected?

PurSnake:
The fastest and most obvious example that I have without a debug interface (Yes, fnf)
Before:
https://github.com/HaxeFlixel/flixel/assets/56314743/64beda2d-5255-4773-be88-048eced8b44f

After:
https://github.com/HaxeFlixel/flixel/assets/56314743/ca58045d-2bbd-4fa7-9399-0c8622ac43f8

Geokureli:
I just need a small example code that I can take and run locally

PurSnake:
boyfriendscale.x *= -1;
or

image
Without:
https://cdn.discordapp.com/attachments/1219255780172107814/1248303330577354792/1tc1eZ9.gif?ex=66632c92&is=6661db12&hm=c576c0ad323426f534b9b216c7c2edd2f3452f0a0b3d3588d7ad3b702c6b65e8&

With:
https://cdn.discordapp.com/attachments/1219255780172107814/1248303359044227123/Ahy2amJ.gif?ex=66632c99&is=6661db19&hm=2f6c057d69725abd41d8b46d792a894f38133ea3fb767217b9a3dd5510ac01b7&

Geokureli:
I understand now, in the future just use words to describe the unexpected behavior and provide simple code that reproduces it and things will move a lot faster

Geokureli:
Also FYI, you should be using sprite.flipX and flipY instead of scaling it to -1

Geokureli (changes requested):
This only fixes the issue if the origin is in the graphic's midpoint

I also think it's much simpler to use flipX and flipY

PR #3167, opened 2024-06-05
Use `FlxColorInt` instead of `Int` for RGBA to reduce variable size for CPP target

Just-Feeshy:
It is just an area for memory usage optimization for the CPP target. The main purpose is to use C's `uint8_t` instead of regular `int` to avoid wasting memory.

### DIFF
--- a/flixel/util/FlxColor.hx
+++ b/flixel/util/FlxColor.hx
@@ -102,7 +102,7 @@ abstract FlxColor(Int) from Int from UInt to Int to UInt #if cpp from cpp.UInt8 
 	 * @param Alpha	How opaque the color should be, from 0 to 255
 	 * @return The color as a FlxColor
 	 */
-	public static inline function fromRGB(Red:Int, Green:Int, Blue:Int, Alpha:Int = 255):FlxColor
+	public static inline function fromRGB(Red:FlxColorInt, Green:FlxColorInt, Blue:FlxColorInt, Alpha:FlxColorInt = 255):FlxColor
 	{
 		var color = new FlxColor();
 		return color.setRGB(Red, Green, Blue, Alpha);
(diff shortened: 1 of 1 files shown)


Geokureli (changes requested):
Looks like [this test](https://github.com/HaxeFlixel/flixel/blob/dev/tests/unit/src/flixel/util/FlxColorTest.hx) is failing on cpp with the error:
> FAIL: massive.munit.AssertionException: Value [-32768] was not equal to expected value [-65536] at flixel.util.FlxColorTest#subtractNotSame (63)

I wish these test output hex strings instead of decimal, a quick conversion reveals the expected value is `FFFF0000` 
 but the actual value is `FFFF8000`

flixel/util/FlxColor.hx
```diff
@@ -15,7 +15,7 @@ import flixel.system.macros.FlxMacroUtil;
  *
  * @author Joe Williamson (JoeCreates)
  */
-abstract FlxColor(Int) from Int from UInt to Int to UInt
+abstract FlxColor(Int) from Int from UInt to Int to UInt #if cpp from cpp.UInt8 to cpp.UInt8 #end
```
Geokureli, line 18: Shouldn't this be removed? a color's components may be constructed from or deconstructed to UInt8s but a color itself should always a 32 bit number

Geokureli:
furthermore, I'm a little skeptical of the benefits without seeing actual test results, please do some tests and ensure the following:
- is there any performance hit due to casting from UInt8 to Ints and vice versa?
- it there a measurable difference in memory usage from this

Just-Feeshy:
I'm so sorry, I forgot to take account for the rest of the methods when doing my own test.

Geokureli:
@Just-Feeshy is this not salvageable?

Just-Feeshy:
> @Just-Feeshy is this not salvageable?

It is, but I realized I made some initial mistakes on the placement of `FlxColorInt` and I'm also trying to think of a good test to demonstrate with as little time as possible.

PR #3178, opened 2024-06-12
Steal @Starmapo input text

Geokureli:
Just learned this exists, @Starmapo if you tried to show me this, before, I missed it. I saw Funkin was using it

I have not yet begun to consider the ramifications of this, so let's start with seeing if it clears coverage CI and code climate, I guess. If you could list - in detail - the changes, improvements and features of this that would be a great!

### DIFF
--- a/flixel/text/FlxBaseInputText.hx
+++ b/flixel/text/FlxBaseInputText.hx
@@ -46,7 +46,7 @@ class FlxBaseInputText extends FlxText
 	}
 
 	/**
-	 * An integer (1-based index) that indicates the bottommost line that is currently visible in the text object.
+	 * An integer (`-based index) that indicates the bottommost line that is currently visible in the text object.
 	 * 
 	 * All the text between the lines indicated by `scrollV` and `bottomScrollV` is currently visible in the text object.
 	 */
@@ -1097,7 +1097,7 @@ class FlxBaseInputText extends FlxText
 	function _onWindowKeyDown(key:KeyCode, modifier:KeyModifier):Void
 	{
 		onWindowKeyDownHandler(key, modifier);
-		#if html5
+		#if (html5 && FlX_KEYBOARD)
 		if (key == SPACE && FlxG.keys.preventDefaultKeys.contains(SPACE))
 		{
 			onWindowTextInputHandler(" ");
--- a/flixel/text/FlxBaseInputText.hx
+++ b/flixel/text/FlxBaseInputText.hx
@@ -27,7 +27,7 @@ import openfl.text.TextFormat;
  * @author Starmapo
  */
 @:access(openfl.text.TextField)
-class FlxInputBaseText extends FlxText
+class FlxBaseInputText extends FlxText
 {
 	/**
 	 * Checks whether the control key is pressed.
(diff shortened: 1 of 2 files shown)


Geokureli:
Because it extends TextField methods that don't fit flixel's checkstyle settings. looking into a way to suppress these

Starmapo:
I'll list the improvements later once I can properly check the differences. I'd just like to warn you first that, at least in its current state, it does not compile for Flash, as the TextField extension sets the `stage` variable which is required for some of the input stuff to work, and this variable isn't able to be modified on Flash.

Starmapo:
Unfortunately the Flash issue is deeper than I assumed, since this text input uses private variables and functions from OpenFL classes, which are not available when compiling to Flash. I suppose that's the consequence of relying heavily on OpenFL's text input implementation.

> We should study [texter], as well, it boasts about "RTL support", which would be nice!, and might work on flash, iunno

It seems to extend the flixel-ui text input, so if that one works on Flash, I think texter would too.

ShaharMS:
> Unfortunately the Flash issue is deeper than I assumed, since this text input uses private variables and functions from OpenFL classes, which are not available when compiling to Flash. I suppose that's the consequence of relying heavily on OpenFL's text input implementation.
> 
> > We should study [texter], as well, it boasts about "RTL support", which would be nice!, and might work on flash, iunno
> 
> It seems to extend the flixel-ui text input, so if that one works on Flash, I think texter would too.

IIRC texter's input text does work on flash, and i don't mind providing an implementation for flixel

the main question is how advanced we want the input to be (multiline, copy-paste, selection shortcuts...)
all should be somewhat easy to implement if i remember my design from back then correctly

Geokureli:
Let's start with "what is the best way to detect character input and other text-related keyboard input". I'd like to make a standalone util for this. if possible, I would like to make things that work with both FlxText and FlxBitmapText

ShaharMS:
Already done in texter, and not too complex - 3/2 event listeners:
- window.onTextInput
- window.onKeyDown
- window.onFocusLost (needed in web)

TextInput for regulat text, KwyDown for keyboard shortcuts & special characters, and the focus one is for js for when you click out of the openfl element

Geokureli:
> * window.onTextInput
> * window.onKeyDown
> * window.onFocusLost (needed in web)

Any chance you can reduce that down to barebones, I don't really understand what bidi means and the code is a little hard to follow.

what would be nice is if it reduced all events down to a single action that anything could implement, like
```hx
enum TypingAction
{
   ADD_CHAR(charCode:Int);
   MOVE_CURSOR(type:MoveCursorAction);// left/right/up/down, home, end, word_left/right
   MOVE_SELECTION(moveEnd:Bool, type:MoveCursorAction);
   COMMAND(cmd:TypingCommand);
}

enum MoveCursorAction
{
    LEFT;
    RIGHT;
    UP;
    DOWN;
    HOME;
    END;
    LINE_LEFT;
    LINE_RIGHT;
    WORD_LEFT;
    WORD_RIGHT;
}

enum TypingCommand
{
   COPY;
   CUT;
   PASTE;
   SELECT_ALL;
}
```

Starmapo:
> Let's start with "what is the best way to detect character input and other text-related keyboard input". I'd like to make a standalone util for this. if possible, I would like to make things that work with both FlxText and FlxBitmapText

I do like that idea, though does that mean the base FlxText and FlxBitmapText classes will have input support? Or will you still have to use extended classes like FlxInputText?

Starmapo:
I've opened a draft for my new text input implementation here #3191. It's still heavily work-in-progress, any feedback or suggestions are very much appreciated.

Starmapo:
@Geokureli You can close this one now btw

PR #3179, opened 2024-06-12
Implement FlxFlicker.pause() and FlxFlicker.resume()

EliteMasterEric:
During the development of Funkin', I found the need to pause an active FlxFlicker without actively destroying it and recreating it. This PR adds that functionality.

### DIFF
--- a/flixel/effects/FlxFlicker.hx
+++ b/flixel/effects/FlxFlicker.hx
@@ -145,6 +145,22 @@ class FlxFlicker implements IFlxDestroyable
 	}
 
 	/**
+	 * Temporarily pause the flickering, so it can be resumed later.
+	 */
+	public function pause():Void {
+		if (timer == null) return;
+		timer.active = false;
+	}
+
+	/**
+	 * Resume the flickering after it has been temporarily paused.
+	 */
+	public function resume():Void {
+		if (timer == null) return;
+		timer.active = true;
+	}
+
+	/**
 	 * Prematurely ends flickering.
 	 */
 	public function stop():Void
--- a/flixel/effects/FlxFlicker.hx
+++ b/flixel/effects/FlxFlicker.hx
@@ -147,16 +147,22 @@ class FlxFlicker implements IFlxDestroyable
 	/**
 	 * Temporarily pause the flickering, so it can be resumed later.
 	 */
-	public function pause():Void {
-		if (timer == null) return;
+	public function pause():Void
+	{
+		if (timer == null)
+			return;
+		
 		timer.active = false;
 	}
 
 	/**
 	 * Resume the flickering after it has been temporarily paused.
 	 */
-	public function resume():Void {
-		if (timer == null) return;
+	public function resume():Void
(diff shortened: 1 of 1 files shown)


Geokureli:
Down to merge, but note that `FlxTween.flicker` now exists which has all the same tween methods for pausing, resuming and cancelling. It also allows for custom "ease" logic, so you can use one tween to do a flicker that changes frequency rather than using two calls, like you mentioned

EliteMasterEric:
> Down to merge, but note that `FlxTween.flicker` now exists which has all the same tween methods for pausing, resuming and cancelling. It also allows for custom "ease" logic, so you can use one tween to do a flicker that changes frequency rather than using two calls, like you mentioned

Good to know, thanks for that.

flixel/effects/FlxFlicker.hx
```diff
@@ -144,6 +144,22 @@ class FlxFlicker implements IFlxDestroyable
 		timer = new FlxTimer().start(interval, flickerProgress, Std.int(duration / interval));
 	}
 
+	/**
+	 * Temporarily pause the flickering, so it can be resumed later.
+	 */
+	public function pause():Void {
+		if (timer == null) return;
```
Geokureli, line 151:
```suggestion
		if (timer == null)
			return;
```

flixel/effects/FlxFlicker.hx
```diff
@@ -144,6 +144,22 @@ class FlxFlicker implements IFlxDestroyable
...
+		timer.active = false;
+	}
+
+	/**
+	 * Resume the flickering after it has been temporarily paused.
+	 */
+	public function resume():Void {
+		if (timer == null) return;
```
Geokureli, line 159:
```suggestion
		if (timer == null)
			return;
		
```

Geokureli (changes requested):
see suggestions

flixel/effects/FlxFlicker.hx
```diff
@@ -144,6 +144,22 @@ class FlxFlicker implements IFlxDestroyable
 		timer = new FlxTimer().start(interval, flickerProgress, Std.int(duration / interval));
 	}
 
+	/**
+	 * Temporarily pause the flickering, so it can be resumed later.
+	 */
+	public function pause():Void {
```
Geokureli, line 150:
```suggestion
	public function pause():Void
	{
```

flixel/effects/FlxFlicker.hx
```diff
@@ -144,6 +144,22 @@ class FlxFlicker implements IFlxDestroyable
...
+		if (timer == null) return;
+		timer.active = false;
+	}
+
+	/**
+	 * Resume the flickering after it has been temporarily paused.
+	 */
+	public function resume():Void {
```
Geokureli, line 158:
```suggestion
	public function resume():Void
	{
```

Geokureli (approved):

Geokureli:
Thanks!

PR #3180, opened 2024-06-12
Don't attempt to retrieve frameCollections on a destroyed graphic.

EliteMasterEric:
This fixes a crash bug I encountered while developing Funkin'.

### DIFF
--- a/flixel/graphics/FlxGraphic.hx
+++ b/flixel/graphics/FlxGraphic.hx
@@ -514,8 +514,12 @@ class FlxGraphic implements IFlxDestroyable
 	 */
 	public inline function getFramesCollections(type:FlxFrameCollectionType):Array<Dynamic>
 	{
-		if (this.isDestroyed) return [];
-
+		if (this.isDestroyed)
+		{
+			FlxG.log.warn('Invalid call to getFramesCollections on a destroyed graphic');
+			return [];
+		}
+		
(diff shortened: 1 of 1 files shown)


Geokureli:
I'd like to see if previous destroy checks have covered this, do you know a solid way to reproduce this?

EliteMasterEric:
I think something was calling getFrameCollections manually. I don't recall exactly what I was working on but I think it was FlxAnimate?

flixel/graphics/FlxGraphic.hx
```diff
@@ -514,6 +514,8 @@ class FlxGraphic implements IFlxDestroyable
 	 */
 	public inline function getFramesCollections(type:FlxFrameCollectionType):Array<Dynamic>
 	{
+		if (this.isDestroyed) return [];
+
```
Geokureli, line 518:
```suggestion
		if (this.isDestroyed)
		{
			FlxG.log.warn('Invalid call to getFramesCollections on a destroyed graphic');
			return [];
		}
		
```

Geokureli (changes requested):
see suggestions

flixel/graphics/FlxGraphic.hx
EliteMasterEric: Seems good! I think I may have had a warning in a previous iteration of this.

Geokureli:
Thanks!

Issue #3081, opened 2024-03-13
Better bitmap keying for FlxText


Geokureli:
source: https://github.com/HaxeFlixel/flixel/blob/dev/flixel/text/FlxText.hx#L893
FlxText bitmaps are keyed by their text, this seems like it would generate a lot of bitmaps unnecessarily. ideally we would generate a new bitmap if the dimensions of a textfield change, without having different fields of the same size use the same bitmap. maybe we can use ID, or get some kinda mem address?

Geokureli:
> HaxeFlixel/flixel-addons#393

no that seems unrelated

superpowers04:
On an unrelated note, the current method also causes a memory leak whenever you try to edit a FlxText. I *believe* it's somehow not disposing of graphics or something

Geokureli:
> On an semi-related note, the current method also causes a memory leak whenever you try to edit a FlxText. I _believe_ it's somehow not disposing of graphics or something. I tried just using the existing key and the memory leak went away however all of the FlxText's broke for obvious reasons

Do you have a decent way to reproduce this memory leak?

Geokureli:
> In my experience, it happens whenever I change the text field of an FlxText.

how are you measuring memory?

> I don't really have the time to make a reproducible test because it takes an absurd amount of time for a new Flixel project to compile

it compiles immensely faster if you target something other than native desktop, I typically use HTML5 for testing, many use HL which is lightning fast

PR #3181, opened 2024-06-12
Modify FlxText to create a bitmap with an informative key.

EliteMasterEric:
In the past, I've had to look over the list of graphics in the BitmapFrontEnd cache to determine the cause of memory issues.

In this circumstance, having readable keys is important; for example, if one bit of text appears in the cache a lot, there may be something forcing it to re-render too often.

This PR changes auto-generated graphics created by FlxText to include the actual text in their key (`text123` will instead be `text(Hello World)123`). This shouldn't have any negative impact but should improve debugging.


Geokureli:
This seems relevant to: https://github.com/HaxeFlixel/flixel/issues/3081

I'd like to determine whether this affects memory leaks resulting from FlxText bitmap keying

PR #3182, opened 2024-06-12
Debug pointer accounts for FlxSprite opacity

EliteMasterEric:
When using the debugger cursor, if an FlxSprite had an alpha of 0, it would still select it. You'd also select a sprite if you clicked one of its transparent pixels. Both of these were annoying when trying to use the cursor to select a specific sprite, this fixes both of those issues.


Geokureli:
When making PRs please have your settings.json include:
```json
"[haxe]":
{
    "editor.formatOnSaveMode": "modifications"
}
```

Geokureli:
This doesn't work in my tests, closing for https://github.com/HaxeFlixel/flixel/pull/3184

PR #3186, opened 2024-06-14
Fnf 6

Geokureli:
@EliteMasterEric This is the only remaining change between https://github.com/FunkinCrew/flixel/tree/dev-6.0.0 and release6. I can't find out what this change does, can you clarify, what does adding a true arg to `@:glVertexSource` and `@:glFragmentSource` do?

also you can update your dev-6.0.0 branch with https://github.com/FunkinCrew/flixel/compare/dev-6.0.0...Geokureli:fnf-6?expand=1

### DIFF
--- a/flixel/graphics/tile/FlxGraphicsShader.hx
+++ b/flixel/graphics/tile/FlxGraphicsShader.hx
@@ -23,7 +23,7 @@ class FlxGraphicsShader extends GraphicsShader
 				openfl_ColorOffsetv = colorOffset / 255.0;
 				openfl_ColorMultiplierv = colorMultiplier;
 			}
-		}")
+		}", true)
 	@:glFragmentHeader("
 		uniform bool hasTransform;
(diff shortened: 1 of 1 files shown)


EliteMasterEric:
Adding a `true` argument to the vertex source and fragment source annotations does nothing on the official build of OpenFL.

However, it DOES do something on my shader rewrite branch of OpenFL (see openfl/openfl#2715).

That branch implements the `glVersion` parameter to shaders, allowing you to use other GLSL versions (such as `120` on MacOS and `300 es` on web).

However, this has a problem; the annotations on any parent classes were written for GLSL `100`, and therefore need to be rewritten to support the new GLSL version. Thankfully, this is relatively trivial, as you can simply perform a find/replace and no special syntax changes are needed, but these changes should only be done on the internal annotations and not the user-provided shader annotations. Thus, I added an optional argument to the annotation to tell the macro system when to rewrite the shader source.

PR #3191, opened 2024-06-19
New text input object

Starmapo:
After #3178 was opened and I figured that my original text input library just would not work with Flash, I've started making a new one from the ground up, which does not depend on any internal OpenFL functions/variables.

It's still a work-in-progress as of now, you're free to request any changes of my implementation.

Showcase of its current state:

https://github.com/HaxeFlixel/flixel/assets/85134252/c34655ab-8afa-4d2b-a5a8-c6c75b02b4d4

# To-do list:

- [ ] FlxInputText
  - [x] Multiline variable
  - [x] Place caret according to the mouse
  - [x] Text selection
  - [x] Text scrolling
  - [x] Action callbacks (input, delete, etc.)
  - [x] Force case variable (flixel-ui)
  - [x] Filter mode variable (flixel-ui)
  - [ ] Block other hotkeys while input text is active (e.g. volume mute)
  - [ ] Touch support
  - [ ] Caret flashing on/off
  - [x] Test how well it works with multiple text formats
  - [ ] Documentation

### DIFF
--- a/checkstyle.json
+++ b/checkstyle.json
@@ -35,7 +35,8 @@
 					"STATIC",
 					"MACRO",
 					"INLINE",
-					"DYNAMIC"
+					"DYNAMIC",
+					"FINAL"
 				]
 			}
 		},
--- a/flixel/text/FlxInputText.hx
+++ b/flixel/text/FlxInputText.hx
@@ -5,7 +5,10 @@ import flixel.math.FlxMath;
 import flixel.math.FlxPoint;
 import flixel.system.frontEnds.InputTextFrontEnd;
 import flixel.util.FlxColor;
+import flixel.util.FlxDestroyUtil;
 import lime.system.Clipboard;
+import openfl.geom.Rectangle;
+import openfl.text.TextFormat;
 import openfl.utils.QName;
 
 class FlxInputText extends FlxText implements IFlxInputText
@@ -14,7 +17,7 @@ class FlxInputText extends FlxText implements IFlxInputText
 	
 	public var caretColor(default, set):FlxColor = FlxColor.WHITE;
 	
-	public var caretIndex(default, set):Int = -1;
+	public var caretIndex(get, set):Int;
 	
 	public var caretWidth(default, set):Int = 1;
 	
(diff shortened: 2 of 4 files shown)


Geokureli:
I've taken the quickest of glances at the source and this looks promising!

Geokureli:
Also, fwiw, `FlxBitmapInputText` is not a requirement for this PR, if you only have time for FlxInputText we can always do bitmap text sometime later, after this merge

flixel/text/FlxInputText.hx
```diff
@@ -19,6 +19,8 @@ class FlxInputText extends FlxText implements IFlxInputText
 	public var hasFocus(default, set):Bool = false;
 	
 	public var maxLength(default, set):Int = 0;
+
+	public var multiline(get, set):Bool;
```
Geokureli, line 23: So I was wrong, it's not in FLxText, but it should be... or at last input text shouldn't be different from FlxText in this regard.

Starmapo:
> Also, fwiw, `FlxBitmapInputText` is not a requirement for this PR, if you only have time for FlxInputText we can always do bitmap text sometime later, after this merge

> in fact not everything in the FlxInputText TODO needs to be done in the first pass either, my minimum requirement is whatever it takes to make people stop using addons and your `flixel-text-input` lib. we can see if Shahar wants to add RTL

Alright, I'll update the to-do list accordingly.

flixel/text/FlxInputText.hx
Starmapo: Should I make a PR to add it to FlxText, or can you do it instead?

Geokureli: not sure when i'll get to it, but i can

Starmapo:
> what is "text scrolling", is that and "force case" something that should also go in FlxText?

Text scrolling is being able to scroll the text horizontally and vertically, which for FlxInputText the user can do by scrolling the mouse wheel or moving the mouse around while selecting text. I suppose it could also have some uses in FlxText independent of text input, maybe even the mouse wheel scrolling could be an optional feature for base FlxText (it seems to already be the case for OpenFL text fields).

`forceCase` is in flixel-ui's FlxInputText and it just forces all letters the user types to be either lowercase or uppercase, I don't think it has any use for non-input text.

Geokureli:
if these features are already in ui.FlxInputText then adding them to FlxText would require a PR to flixel-ui for compatibility. so just keep doing it how you're doing it and we'll migrate these to FlxText some other time. thanks for the info

checkstyle.json
```diff
@@ -35,7 +35,8 @@
 					"STATIC",
 					"MACRO",
 					"INLINE",
-					"DYNAMIC"
+					"DYNAMIC",
+					"FINAL"
```
Geokureli, line 39: holy crap this has been driving me nuts, I was told this wasn't possible

Starmapo: Yeah its not even shown in the docs, I had to go into the default config json on the repository to find it

AlexHaxe: it seems like documentation for ModifierOrder check wasn't properly updated when Haxe 4 / checkstyle 2.5.0 came out.
`FINAL` had been in checkstyle for a bit longer, but was hidden behind a `#if haxe4` conditional. I should've updated documentation when I moved to Haxe 4 releases, but it seems I've missed it. sorry.

luckily the default config is generated automatically so it always contains whatever features and settings are available and active in any given version.

I've updated docs to include `FINAL` and the new message format.

Starmapo: Sweet, thank you!

Geokureli:
Unsubscribing for the moment, if you need me for anything, ping me please

Starmapo:
There's still a handful of things left so I think I'll just close the PR for now and reopen it once it's ready.

PR #3192, opened 2024-06-21
Add FlxBuffer to hide Vector<Float> as a Array<T>

Geokureli:
details incoming

### DIFF
--- a/flixel/system/FlxBuffer.hx
+++ b/flixel/system/FlxBuffer.hx
@@ -1,0 +1,392 @@
+package flixel.system;
+
+import haxe.macro.TypeTools;
+#if macro
+import haxe.macro.Type;
+import haxe.macro.Expr;
+import haxe.macro.Context;
+
+using haxe.macro.Tools;
+
+class BufferMacro
+{
+	public static function build(includeGetters = true):ComplexType
+	{
+		final local = Context.getLocalType();
+		switch local {
+			// Extract the type parameter
+			case TInst(local, [type]):
+				return buildBuffer(getFields(type, includeGetters), type);
+			default:
+				throw "Expected TInst";
+		}
+	}
+	
+	static function getFields(type:Type, includeGetters:Bool):Array<ClassField>
+	{
+		// Follow it to get the underlying anonymous structure
+		switch type.follow()
+		{
+			case TAbstract(abType, []):
+				final fields = getFields(abType.get().type, includeGetters).copy();
+				// add abstract fields too
+				if (includeGetters && (abType.get().impl != null || abType.get().impl.get() != null))
+				{
+					final statics = abType.get().impl.get().statics.get();
+					for (field in statics)
+					{
+						if (field.kind.match(FVar(AccCall, _)))
+						{
+							fields.push(field);
+						}
+					}
+				}
+				return fields;
+			case TAnonymous(type):
+				return type.get().fields;
+			case TInst(found, _):
+				throw 'Expected type parameter to be an anonymous structure, got ${found.get().name}';
+			case TEnum(found, _):
+				throw 'Expected type parameter to be an anonymous structure, got ${found.get().name}';
+			case found:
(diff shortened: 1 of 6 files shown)


MaybeMaru:
Pretty cool to see updates on this! Just make sure to not use typedefs structures stuff for the final thing since those are pretty slow compiling to dynamic

Geokureli:
> Pretty cool to see updates on this! Just make sure to not use typedefs structures stuff for the final thing since those are pretty slow compiling to dynamic

I don't know what you mean by "pretty slow compiling to dynamic", typedefs compile crazy fast to non-static targets. the goal here is to have seemingly typed arrays/vectors without incurring any garbage collection from the actual types

I've done some initial tests and no types are actually used as everything is inlined away, see here: https://try.haxe.org/#ADa541d7

But I plan to do actual benchmark tests with these flixel classes, too

MaybeMaru:
In hxcpp (im not sure about other targets) typedefs compile has annonymous dynamic structures but yeah better to double check with benchmarks.

Geokureli:
> In hxcpp (im not sure about other targets) typedefs compile has annonymous dynamic structures but yeah better to double check with benchmarks.

Again I don't see why we're talking about compiling here, hxcpp takes me almost an hour to build, already and dynamic targets take an extra 10th of a second

Geokureli:
@CrowPlexus and @NovaTheFurryDev, you voted thumbs down, care to share your concerns?

crowplexus:
removing the DV, my biggest concern with this was perhaps it being slower than the current method, I will definitely give this a try later on a blank project to see if it makes a major difference in comparison to the old method

Geokureli:
> removing the DV

What's the DV?

crowplexus:
> > removing the DV
> 
> What's the DV?

downvote or thumbs down

flixel/system/FlxBuffer.hx
```diff
@@ -0,0 +1,478 @@
...
+{
+	public static function build(isVector:Bool, includeGetters = true):ComplexType
+	{
+		final local = Context.getLocalType();
+		switch local {
+			// Extract the type parameter
+			case TInst(local, [type]):
+				return buildBuffer(getFields(type, includeGetters), type, isVector);
+			default:
+				throw "Expected TInst";
```
UncertainProd, line 21: I think using `Context.error()` would be better than throwing since it shows the error at the location where TInst was expected

SomeGuyWhoLovesCoding:
Is this a performance improvement thing or what?

Geokureli:
> Is this a performance improvement thing or what?

The goal is better readability and less human-error when maintaining or editing, at the hopes of not losing any performance

SomeGuyWhoLovesCoding:
> > Is this a performance improvement thing or what?
> 
> The goal is better readability and less human-error when maintaining or editing, at the hopes of not losing any performance

I'd recommend switching from typedefs to classes for QuadRectRaw, QuadColorMult, and QuadColorOffset

Geokureli:
> I'd recommend switching from typedefs to classes for QuadRectRaw, QuadColorMult, and QuadColorOffset

the typedefs are never instantiated, they are inlined away at compile time
Screenshot 2024-07-29 at 7 19 57 PM

Geokureli:
> I tried this pr and the rendering was actually made slower than usual on hl

I have no plans to merge this atm, I still need to check a billion things and there's more important things to do, but if perf is worsened it means this will not ever get merged.

Also whenever you talk about performance it's always "i just ran it a bunch of times and it seems slower when I watched it". if you have an actual reproducible benchmark test, then share that, otherwise I'll assume you're just guessing

Raltyro:
Really want to use this instead of transferring Vectors into Array just for ShaderParameter, etc.
I guess it's not happening soon.

Geokureli:
> Really want to use this instead of transferring Vectors into Array just for ShaderParameter, etc. I guess it's not happening soon.

Yeah, it's cool in theory, but isn't panning out

PR #3201, opened 2024-07-04
Add getGlobalScreenPosition to FlxPointer

01010111:
At the moment, FlxPointer does not have a publicly accessible way of getting it's global screen position. This change adds that.


Geokureli:
Fixed by https://github.com/HaxeFlixel/flixel/pull/3211
Added FlxG.mouse.gameX and gameY which are functionally identical to _globalScreenX/Y

PR #3202, opened 2024-07-05
Add `multiLine` variable to FlxText

Starmapo:
This PR adds a `multiLine` variable to FlxText which dictates whether or not new lines are allowed in the text (if disabled, they'll just be removed from the text).

Since word wrapping will create new lines in the OpenFL text field regardless of the multiline value, I added a note about that in the documentation, but let me know if you'd prefer to just toggle off word wrapping automatically, or something else.

### DIFF
--- a/flixel/text/FlxText.hx
+++ b/flixel/text/FlxText.hx
@@ -98,7 +98,15 @@ class FlxText extends FlxSprite
 	public var underline(get, set):Bool;
 
 	/**
-	 * Whether to use word wrapping and multiline or not (`true` by default).
+	 * Whether new lines in the text should be allowed or not (`true` by default).
+	 * 
+	 * **NOTE:** If word wrapping is on, the text field could still have multiple
+	 * lines even if this is disabled.
+	 */
+	public var multiLine(get, set):Bool;
+	
+	/**
+	 * Whether to use word wrapping or not (`true` by default).
 	 */
 	public var wordWrap(get, set):Bool;
 
@@ -625,6 +633,11 @@ class FlxText extends FlxSprite
 
 	function set_text(Text:String):String
 	{
+		if (!multiLine && (Text.contains("\n") || Text.contains("\r")))
+		{
+			Text = Text.split("\n").join("").split("\r").join("");
+		}
+
 		text = Text;
 		if (textField != null)
 		{
@@ -758,6 +771,28 @@ class FlxText extends FlxSprite
 		{
 			_defaultFormat.underline = value;
 			updateDefaultFormat();
+		}
+		return value;
+	}
+
+	function get_multiLine():Bool
+	{
+		return (textField != null) ? textField.multiline : false;
+	}
+	
+	function set_multiLine(value:Bool):Bool
+	{
+		if (textField == null)
+			return value;
+			
+		if (textField.multiline != value)
+		{
+			textField.multiline = value;
+			if (!value)
+			{
+				// Refresh text to remove any new lines
+				text = text;
+			}
 		}
 		return value;
 	}


flixel/text/FlxText.hx
```diff
@@ -625,6 +633,11 @@ class FlxText extends FlxSprite
 
 	function set_text(Text:String):String
 	{
+		if (!multiLine && (Text.contains("\n") || Text.contains("\r")))
+		{
+			Text = Text.split("\n").join("").split("\r").join("");
+		}
+
```
Geokureli, line 640: I disagree with this implementation, this isn't how FlxInputText works and it's not how FlxBitmapText works.

plus this will cause weird behavior in the following:
```hx
text.multiline = false;
text.text = myMultilineString;
text.multiline = true;
```

flixel/text/FlxText.hx
```diff
@@ -762,6 +775,28 @@ class FlxText extends FlxSprite
...
+	}
+	
+	function set_multiLine(value:Bool):Bool
+	{
+		if (textField == null)
+			return value;
+			
+		if (textField.multiline != value)
+		{
+			textField.multiline = value;
```
Geokureli, line 790: should add `_regen = true;` here too

Geokureli (changes requested):
changes requested

flixel/text/FlxText.hx
Starmapo: Changing `multiline` in the OpenFL text field doesn't require a graphic regeneration. The only thing it changes for non-html text is whether or not new lines can be added by the user in an input text field.

`set_multiline()` here doesn't require refreshing the layout: https://github.com/openfl/openfl/blob/2e105419e01d4b34855260253f6136e4de5f182f/src/openfl/text/TextField.hx#L2835

flixel/text/FlxText.hx
Starmapo: > I disagree with this implementation, this isn't how FlxInputText works and it's not how FlxBitmapText works.

Do you mean that this should be implemented in a better way or that the text shouldn't be modified at al?

Geokureli: the latter

Starmapo: Alright, in that case though, the only change that this variable does is whether or not new lines can be added by the user in the new FlxInputText. It doesn't stop the text field from having new lines set through text.

Geokureli: This field simply sets the underlying text field's multiline field, how input text handles new lines from user input is an entirely separate issue that should be handled by those input handlers

Starmapo: Okay, I understand now 👍

Geokureli:
Looking into multiline more, now. it seems that I have misunderstood what multiline is used for. looks like it has no effect on textfields UNLESS they are input fields. for non-input fields, it looks like new line chars in text strings will always show up, and the field will wrap text based on the value of wordWrap.

Seems like there is no reason for there to be a multiline field in FlxText, and I was wrong for suggesting this. If there is a reason, please reopen this PR and inform me about it

Geokureli:
Also, I noticed that you added `multiLine` with a capital "L", despite openfl having a lowercase l, since multiline is one word, if you do add a multiline property to input text or some input text handler of some kind please match this case

Starmapo:
> Also, I noticed that you added `multiLine` with a capital "L", despite openfl having a lowercase l, since multiline is one word, if you do add a multiline property to input text or some input text handler of some kind please match this case

Got it, I saw FlxBitmapText had its multiline property spelled like that so I just thought to match it to that.

Geokureli:
> > Also, I noticed that you added `multiLine` with a capital "L", despite openfl having a lowercase l, since multiline is one word, if you do add a multiline property to input text or some input text handler of some kind please match this case
> 
> Got it, I saw FlxBitmapText had its multiline property spelled like that so I just thought to match it to that.

uh oh

Issue #3204, opened 2024-07-05
FlxAnimationController.callback should be an FlxSignal


EliteMasterEric:
https://github.com/HaxeFlixel/flixel/blob/5ee97353cfca85a25bd835588c879fc005a95871/flixel/animation/FlxAnimationController.hx#L61

Since this (and the finishCallback just below it) are Functions rather than FlxSignals, you can only assign one per sprite, and assigning a second one silently removes the first, causing unexpected issues.

`callback` and `finishCallback` should be deprecated and replaced with appropriate FlxTypedSignals.

The only related issue I found is #2816 which proposes a separate thing.

(Just writing this down before I forget, it should be extremely easy to implement in practice but I'm busy with something else at the moment).

richTrash21:
I was thinking about making this a pr lately, will try to work on it today/tomorrow

DetectiveBaldi:
mentioned this a few days ago in the Haxe discord, totally agree

Geokureli:
piggybacking related issue: https://github.com/HaxeFlixel/flixel/issues/3212

I'd like to address this now while we are adding new fields

PR #3205, opened 2024-07-06
Make FlxAnimationController.callback an FlxTypedSignal

richTrash21:
Closes: #3204

~~Maybe there's a better way to name FlxTypedSignal variables.~~
Thanks Geo!

### DIFF
--- a/checkstyle.json
+++ b/checkstyle.json
@@ -35,7 +35,8 @@
 					"STATIC",
 					"MACRO",
 					"INLINE",
-					"DYNAMIC"
+					"DYNAMIC",
+					"FINAL"
 				]
 			}
 		},
--- a/flixel/animation/FlxAnimation.hx
+++ b/flixel/animation/FlxAnimation.hx
@@ -132,6 +132,7 @@ class FlxAnimation extends FlxBaseAnimation
 	 *                     (`Frame = numFrames - 1 - Frame`), so `Frame` value will mean frame index
 	 *                     from the animation's end in this case.
 	 */
+	@:haxe.warning("-WDeprecated")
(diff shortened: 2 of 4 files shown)


DetectiveBaldi:
@Geokureli is there a reason in certain cases flixel has an private signal for certain functionality, and then a public one (which is just a getter for the internal one)
this is seen in FlxState https://github.com/HaxeFlixel/flixel/blob/dev/flixel%2FFlxState.hx#L92 and also seems to be adopted here

Geokureli:
> @Geokureli is there a reason in certain cases flixel has an private signal for certain functionality, and then a public one (which is just a getter for the internal one) this is seen in FlxState

This is likely a leftover from the flash Flixel, where this was common practice (for good reason) but it has little advantage in haxe

Geokureli:
Changes:
- Rename `finishedCallback` to `onFinish` and `frameCallback` to `onFrameChange`
- Remove private versions of both and getters/setters, this is almost never needed since you can just use `var foo(default, null)` to prevent external write-access. I also removed the instantiation in the getters, I'm strongly against that practice
- Made the signals `final` (The new standard for signals now that code climate allows it)
- Add unit tests
- Improve documentation
- convert signal types from `A->B->Void` to `(a:A, b:B)->Void` for better code completion and readability

Geokureli:
Not sure why cpp unit tests are failing and the error is not helpful, gonna look into this, may need to revert unit tests for now, iunno

Geokureli:
Found the cause of failing CI: https://github.com/HaxeFoundation/hxcpp/pull/1125

Issue #3212, opened 2024-07-11
FlxAnimationController.finishCallback does not trigger on looped animations


balajeerc:
Not sure if I'm misunderstanding how `FlxAnimationController.finishCallback` works but looks like the callback is not getting triggered.

To reproduce this, one can add the following lines after [line 83 of FlxAsperiteUtil demo](https://github.com/HaxeFlixel/flixel-demos/blob/dd0afb866ed57b0507e8679d191ce68d4a8e2407/Graphics/FlxAsepriteUtil/source/PlayState.hx#L84):

```
player.animation.finishCallback = (animationName:String) -> {
    trace("animation complete callback: " + animationName);
 };
player.animation.callback = (name:String, frameNumber:Int, frameIndex:Int) -> {
    trace("per frame animation callback: " + name + " " + frameNumber + " " + frameIndex);
};
```

On running the demo post addition of lines above, I notice that the `finishCallback` is not called, whereas the per frame triggered `callback` is indeed called.

Geokureli:
It appears that finishCallback is not called when the animation is set to loop, I suppose this is by design, since a looping animation is never "finished". That said, I agree that this is weird behavior, there is [a separate issue to create a finished signal](https://github.com/HaxeFlixel/flixel/issues/3204) to replace this callback, perhaps a change can be made there, alternatively we can have 2 signals, one when the end is reached and one when it stops playing

Geokureli:
@balajeerc, looking into this now, note that finished callback is fired when the final frame is done, rather than when it is reached. If this were extended to include loops, it would be fired when the animation goes from the last frame back to the first frame, so this might be a little confusing seeing the finished callback whenever frame 0 is "reached". It is also confusing because `animation.finished` is still false.

I think an immediate fix for your case is to use:
```hx
animation.callback = function(name, num, index)
{
    if (num == animation.curAnim.numFrames)
         trace("final frame entered");
    
    // alternatively
    
    if (num == 0)
         trace("loop complete");
}
```

In the future, I think if there is an event that includes both "finished" and "loop complete"  animations, it should avoid the word "finished". LMK if you have naming ideas

balajeerc:
Firstly, thanks for looking into this @Geokureli 
I should have indeed filed it in flixel repo itself rather than the demo repo. Thanks for moving it.

The workaround you suggested is pretty much what I ended up using yesterday.

> LMK if you have naming ideas

[...]

PR #3216, opened 2024-07-12
Add onLoop to FlxAnimationController

richTrash21:
Solves #3212 by adding onLoop signal field to the FlxAnimationController that dispatches each time the looped animation starts over.

### DIFF
--- a/flixel/animation/FlxAnimation.hx
+++ b/flixel/animation/FlxAnimation.hx
@@ -214,16 +214,26 @@ class FlxAnimation extends FlxBaseAnimation
 			if (reversed)
 			{
 				if (looped && curFrame == loopPoint)
+				{
 					curFrame = numFrames - 1;
+					parent.fireLoopCallback(name);
+				}
 				else
+				{
 					curFrame--;
+				}
 			}
 			else
 			{
 				if (looped && curFrame == numFrames - 1)
+				{
 					curFrame = loopPoint;
+					parent.fireLoopCallback(name);
+				}
 				else
+				{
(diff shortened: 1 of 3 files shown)


Geokureli:
what is the behavior when an animation has 1 frame and is set to loop? Does it fire once, never or repeatedly?

flixel/animation/FlxAnimation.hx
```diff
@@ -228,7 +228,7 @@ class FlxAnimation extends FlxBaseAnimation
 				if (looped && curFrame == numFrames - 1)
 				{
 					curFrame = loopPoint;
-					parent.fireLoopCallback(name);
+					parent.onLoopComplete.dispatch(name);
```
Geokureli, line 231: let's revert this since it's how all the others work

richTrash21: reverted

richTrash21:
> what is the behavior when an animation has 1 frame and is set to loop? Does it fire once, never or repeatedly?

Currently testing this, but it will probably just repeatedly fire the callback.

Geokureli:
> yes, it repeatedly fires the callback, multiple times at once even. Currently adjusting the behavior

That's fine, and what I expected

richTrash21:
> Lastly, thoughts on this (edited so might have missed)
> 
> > What are your thoughts on `onLoop` rather than `onLoopComplete`?

I think that's a good change

richTrash21:
> That's fine, and what I expected

Should I push fix then?

PR #3217, opened 2024-07-13
Fix FlxCamera.fill() blending

### LINKED ISSUE #3203: FlxCamera.fill() uses blend mode of the last drawn item on camera
- **Haxe version:** 4.3.4
- **Flixel version:** 5.8.0
- **OpenFL version:** 9.3.3
- **Lime version:** 8.1.2
- **Affected targets:** all non blitting targets

While messing around with blend modes I noticed the following bug: if the last drawn item on camera had any blend mode, it will also apply to the camera's fill function.
Here's an example of this bug from a little [test project](https://github.com/richTrash21/fade-bug-test) I made, that replicates an environment where I encountered the bug:

https://github.com/HaxeFlixel/flixel/assets/87835336/9ed630a9-b62b-46d8-93f9-fe9bf9204ed7

As it was seen in the video the fade effect (which uses the fill function) have an ADD blend mode applied to it which was caused by the last drawn item on camera - a sprite with an ADD blend mode.

This is how it suppose to behave:

https://github.com/HaxeFlixel/flixel/assets/87835336/90f66dba-5638-4b08-ab8b-dc6a7adeb5e4

I ran this test project on Windows, Neko, HashLink, HTML5 and Flash and only non blitting targets were affected by this bug (blend modes just doesn't work on Flash so there's nothing to break I suppose). I also found a solution for this issue but it needs further testing before I can make a PR with a proper implementation of the fix. Might also wait a bit since from what I've read in #2966 blend modes are not implemented but also partially are?

**Observed behavior:**
Last rendered item blend mode affects the fill function, applying blend to the resulting fill.

**Expected behavior:**
Last rendered item blend mode should not affect fill function at all.

richTrash21:
Fixes #3203 by overriding the blend mode of target graphics in FlxCamera.fill() to null (BlendMode.NORMAL). Does not affect blitting targets since blend modes just doesn't work there.
Additionally makes drawFX more readable

### DIFF
--- a/flixel/FlxCamera.hx
+++ b/flixel/FlxCamera.hx
@@ -1722,40 +1722,35 @@ class FlxCamera extends FlxBasic
 	@:allow(flixel.system.frontEnds.CameraFrontEnd)
 	function drawFX():Void
 	{
-		var fxColor:FlxColor;
-		var alphaComponent:Float;
-
 		// Draw the "flash" special effect onto the buffer
 		if (_fxFlashAlpha > 0.0)
 		{
-			fxColor = _fxFlashColor;
-			alphaComponent = fxColor.alphaFloat * _fxFlashAlpha;
-
 			if (FlxG.renderBlit)
 			{
-				fxColor.alphaFloat = alphaComponent;
-				fill(fxColor);
+				var color = _fxFlashColor;
+				color.alphaFloat *= _fxFlashAlpha;
+				fill(color);
 			}
 			else
 			{
-				fill(fxColor.rgb, true, alphaComponent, canvas.graphics);
+				final alpha = color.alphaFloat * _fxFlashAlpha;
+				fill(_fxFlashColor.rgb, true, alpha, canvas.graphics);
 			}
 		}
-
+		
 		// Draw the "fade" special effect onto the buffer
 		if (_fxFadeAlpha > 0.0)
 		{
-			fxColor = _fxFadeColor;
-			alphaComponent = fxColor.alphaFloat * _fxFadeAlpha;
-
 			if (FlxG.renderBlit)
 			{
-				fxColor.alphaFloat = alphaComponent;
-				fill(fxColor);
+				var color = _fxFadeColor;
+				color.alphaFloat *= _fxFadeAlpha;
+				fill(color);
 			}
 			else
 			{
-				fill(fxColor.rgb, true, alphaComponent, canvas.graphics);
+				final alpha = _fxFadeColor.alphaFloat * _fxFadeAlpha;
+				fill(_fxFadeColor.rgb, true, alpha, canvas.graphics);
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


richTrash21:
Alternatively this can be turned into a feature by passing blend to FlxCamera.fill()
Like this:
```haxe
final someColor = new FlxColor(0x99ff00ff);
final someBlendMode = BlendMode.ADD;
FlxG.camera.fill(someColor.rgb, true, someColor.alphaFloat, FlxG.camera.canvas.graphics, someBlendMode);
```

Note: just piggybanking an idea, will probably use it for personal stuff, tho let me know if it's good or meh.

flixel/FlxCamera.hx
```diff
@@ -1703,8 +1707,11 @@ class FlxCamera extends FlxBasic
 			if (FxAlpha == 0)
 				return;
 
-			var targetGraphics:Graphics = (graphics == null) ? canvas.graphics : graphics;
+			final targetGraphics = (graphics == null) ? canvas.graphics : graphics;
 
+			#if (openfl > "8.7.0")
+			targetGraphics.overrideBlendMode(null);
+			#end
```
Geokureli, line 1714: At a glance this seems like the only change related to the titular bug, and the rest seem like unrelated style changes, is that the case?

richTrash21: Yes.

The style change for drawFX() was made to make it more readable, but if it was too unnecessary - I don't see a problem in reverting it and leaving only the bugfix.

Geokureli:
Reverted both parts like this, i don't want unrelated functional changes amended to PRs and this may technically be a "breaking change"
```hx
if (_fxFlashColor.alpha == 0)
	_fxFlashColor.alpha = 0xff;
```

also cleaned up the style changes a little more, not a fan of having vars crated outside of code blocks, only to be used inside of code blocks

Geokureli:
Thanks!

richTrash21:
> Reverted both parts like this, i don't want unrelated functional changes amended to PRs and this may technically be a "breaking change"
> 
> ```hx
> if (_fxFlashColor.alpha == 0)
> 	_fxFlashColor.alpha = 0xff;
> ```

The reason why I added these parts was because of how drawFX() used alpha component of the effect color before this pr's changes, where if color had no alpha it would just set it to 255, so technically it wasn't a "breaking change".
```hx
// Draw the "fade" special effect onto the buffer
if (_fxFadeAlpha > 0.0)
{
	alphaComponent = _fxFadeColor.alpha;

	if (FlxG.renderBlit)
	{
		fill((Std.int(((alphaComponent <= 0) ? 0xff : alphaComponent) * _fxFadeAlpha) << 24) + (_fxFadeColor & 0x00ffffff));
	}
	else
	{
		fill((_fxFadeColor & 0x00ffffff), true, ((alphaComponent <= 0) ? 0xff : alphaComponent) * _fxFadeAlpha / 255, canvas.graphics);
	}
}
```

Note: Just giving a reasoning to why such changes were made, not to imply that they were necessary, I was just following the functionality of drawFX().

PR #3219, opened 2024-07-14
Improved FlxInputText

Starmapo:
This is the completed version of #3191 that adds an improved version of FlxInputText from flixel-ui to the base library.

## Improvements over the original FlxInputText

- Multiple characters can now be selected at once, either through holding and dragging the mouse or holding the shift key while moving the selection cursor.
  - Double pressing on the text field will select the closest word at that position.
  - You can select all of the text with Ctrl+A.
  - The currently selected text can be copied or cut with Ctrl+C and Ctrl+X respectively.
  - `selectionBeginIndex` and `selectionEndIndex` variables added so you can get the span of the current selection.
  - `selectionColor` and `selectedTextColor` variables can be changed to set the selection background and the selected text's color respectively. The custom format for the selected text can be disabled with `useSelectedTextFormat`.
  - You can set the selection directly with `setSelection()` and replace the currently selected text with `replaceSelectedText()`.
- Multiline is now properly supported, with an added `multiline` variable to dictate whether new lines can be created by the user.
- The text field can now be scrolled by using the mouse wheel, and is automatically scrolled whenever moving the selection cursor to a character that is out of view.
  - The scroll can be modified with the `scrollH` and `scrollV` variables. `bottomScrollV`, `maxScrollH` and `maxScrollV` have also been added as helper read-only variables.
  - `mouseWheelEnabled` dictates whether or not the text field can be scrolled with the mouse wheel, by default set to true.
- The selection cursor can now be moved to the start/end of the current line (Home/End or Ctrl+Up/Down) or to the previous/next word (Ctrl+Left/Right).
- The `background` variable can now be changed to toggle the background on/off.
- `selectable` and `editable` variables to dictate whether the text field can be selected/edited.
- Enums have been added for `forceCase`, `filterMode` and the callback's action types.
  - The callback is now also dispatched for when the text field is scrolled in some way (SCROLL_ACTION).
- Other Flixel keybinds will be paused while an input text is active (e.g. volume keys and debugger toggle key).
- The window's text input rect is now set accordingly so the text field isn't blocked by any keyboard overlays.
- Touches are now supported.
- `InputTextFrontEnd` added (accessible via `FlxG.inputText`) to allow for easier implementations of other types of input texts.

## Things to consider

- The only variable from the original that wasn't added was `lines`. From what I could see, this didn't actually dictate the amount of lines that could be added, and was basically just a setter for if the text was multiline or not (which on top of that, multiline wasn't properly implemented in the original input text).
- Demos should be added to showcase the new input text, but I'm not good at coming up with ideas for those.

### DIFF
--- a/flixel/FlxG.hx
+++ b/flixel/FlxG.hx
@@ -1,10 +1,5 @@
 package flixel;
 
-import openfl.Lib;
-import openfl.display.DisplayObject;
-import openfl.display.Stage;
-import openfl.display.StageDisplayState;
-import openfl.net.URLRequest;
 import flixel.effects.postprocess.PostProcess;
 import flixel.math.FlxMath;
 import flixel.math.FlxRandom;
@@ -17,6 +12,7 @@ import flixel.system.frontEnds.CameraFrontEnd;
 import flixel.system.frontEnds.ConsoleFrontEnd;
 import flixel.system.frontEnds.DebuggerFrontEnd;
 import flixel.system.frontEnds.InputFrontEnd;
+import flixel.system.frontEnds.InputTextFrontEnd;
 import flixel.system.frontEnds.LogFrontEnd;
 import flixel.system.frontEnds.PluginFrontEnd;
 import flixel.system.frontEnds.SignalFrontEnd;
@@ -28,6 +24,11 @@ import flixel.system.scaleModes.RatioScaleMode;
 import flixel.util.FlxCollision;
 import flixel.util.FlxSave;
 import flixel.util.typeLimit.NextState;
+import openfl.Lib;
+import openfl.display.DisplayObject;
+import openfl.display.Stage;
+import openfl.display.StageDisplayState;
+import openfl.net.URLRequest;
 #if FLX_TOUCH
 import flixel.input.touch.FlxTouchManager;
 #end
@@ -319,6 +320,8 @@ class FlxG
 	 * Contains a list of all plugins and the functions required to `add()`, `remove()` them etc.
 	 */
 	public static var plugins(default, null):PluginFrontEnd;
+
+	public static var inputText(default, null):InputTextFrontEnd = new InputTextFrontEnd();
 
 	public static var initialWidth(default, null):Int = 0;
 	public static var initialHeight(default, null):Int = 0;
--- a/flixel/FlxG.hx
+++ b/flixel/FlxG.hx
@@ -321,6 +321,9 @@ class FlxG
 	 */
 	public static var plugins(default, null):PluginFrontEnd;
 
+	/**
+	 * Used for detecting text input and dispatching commands on input text fields.
+	 */
 	public static var inputText(default, null):InputTextFrontEnd = new InputTextFrontEnd();
 
 	public static var initialWidth(default, null):Int = 0;
--- a/flixel/FlxG.hx
+++ b/flixel/FlxG.hx
@@ -12,7 +12,6 @@ import flixel.system.frontEnds.CameraFrontEnd;
 import flixel.system.frontEnds.ConsoleFrontEnd;
 import flixel.system.frontEnds.DebuggerFrontEnd;
 import flixel.system.frontEnds.InputFrontEnd;
-import flixel.system.frontEnds.InputTextFrontEnd;
 import flixel.system.frontEnds.LogFrontEnd;
 import flixel.system.frontEnds.PluginFrontEnd;
 import flixel.system.frontEnds.SignalFrontEnd;
@@ -320,11 +319,6 @@ class FlxG
 	 * Contains a list of all plugins and the functions required to `add()`, `remove()` them etc.
 	 */
 	public static var plugins(default, null):PluginFrontEnd;
-
-	/**
-	 * Used for detecting text input and dispatching commands on input text fields.
-	 */
-	public static var inputText(default, null):InputTextFrontEnd = new InputTextFrontEnd();
 
 	public static var initialWidth(default, null):Int = 0;
 	public static var initialHeight(default, null):Int = 0;
--- a/flixel/text/FlxInputText.hx
+++ b/flixel/text/FlxInputText.hx
@@ -104,7 +104,7 @@ class FlxInputText extends FlxText implements IFlxInputText
 	/**
 	 * Whether or not the text field is the current active one on the screen.
 	 */
-	public var hasFocus(default, set):Bool = false;
+	public var hasFocus(default, null):Bool = false;
 	
 	/**
 	 * Set the maximum length for the text field. 0 means unlimited.
@@ -389,6 +389,57 @@ class FlxInputText extends FlxText implements IFlxInputText
 		}
 		
 		this.manager = manager;
+	}
+	
+	public function startFocus()
+	{
+		if (!hasFocus)
+		{
+			// set first to avoid infinite loop
+			hasFocus = true;
+			
+			// Ensure that the text field isn't hidden by a keyboard overlay
+			final bounds = getLimeBounds(_pointerCamera);
+			FlxG.stage.window.setTextInputRect(bounds);
+			
+			manager.setFocus(this);
+			
+			if (_caretIndex < 0)
+			{
+				_caretIndex = text.length;
+				_selectionIndex = _caretIndex;
+				updateSelection(true);
+			}
+			
+			restartCaretTimer();
+			
+			_justGainedFocus = true;
+			onFocusChange.dispatch(hasFocus);
+		}
+	}
+	
+	public function endFocus()
+	{
+		if (hasFocus)
+		{
+			// set first to avoid infinite loop
+			hasFocus = false;
+			
+			// make sure we have not already switched to a new focus (probably not needed, but may in the future)
+			if (manager.focus == this)
+			{
+				manager.setFocus(null);
+			}
+			
+			if (_selectionIndex != _caretIndex)
+			{
+				_selectionIndex = _caretIndex;
+				updateSelection(true);
+			}
+			
+			stopCaretTimer();
+			onFocusChange.dispatch(hasFocus);
+		}
 	}
 	
 	override function update(elapsed:Float):Void
@@ -1513,7 +1564,7 @@ class FlxInputText extends FlxText implements IFlxInputText
 		}
 		else if (FlxG.mouse.justPressed && !_justGainedFocus)
 		{
-			hasFocus = false;
+			endFocus();
 		}
 		#end
 		return overlap;
@@ -1569,7 +1620,7 @@ class FlxInputText extends FlxText implements IFlxInputText
 		}
 		if (pressedElsewhere && _currentTouch == null && !_justGainedFocus)
 		{
-			hasFocus = false;
+			endFocus();
 		}
 		#end
 		return overlap;
@@ -1604,7 +1655,7 @@ class FlxInputText extends FlxText implements IFlxInputText
 	 */
 	function updatePointerPress(pointer:FlxPointer):Void
 	{
-		hasFocus = true;
+		startFocus();
 		
 		var relativePos = getRelativePosition(pointer);
 		_caretIndex = getCharAtPosition(relativePos.x + scrollH, relativePos.y + getScrollVOffset());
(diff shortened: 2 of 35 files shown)


Geokureli:
do me a favor and backmerge dev into this branch

If you have a little test case used to try out the new input text class can you share it here? also consider in the near future that there should be a flixel-demo for this.

It may take a couple days to look over everything but consider this greenlit. This is a huge endeavor!

Starmapo:
> If you have a little test case used to try out the new input text class can you share it here?

```haxe
package;

import flixel.FlxG;
import flixel.system.FlxAssets;
import flixel.text.FlxInputText;
import flixel.util.FlxColor;

class FlxInputTextTest extends flixel.FlxState
{
	override function create()
	{
		super.create();
		
		bgColor = 0xFF808080;
		
		final textInput = new FlxInputText(5, 5, 0, "This is a default text input.", 16);
		add(textInput);
		
		final textInput = new FlxInputText(textInput.x, textInput.y + textInput.height + 5, FlxG.width - 20,
			"This is a multiline text input with fieldWidth and fieldHeight set.", 16);
		textInput.fieldHeight = 44;
		textInput.multiline = true;
		add(textInput);
		
		final textInput = new FlxInputText(textInput.x, textInput.y + textInput.height + 5, FlxG.width - 20,
			"This is a multiline text input with word wrapping.", 16);
		textInput.fieldHeight = 44;
		textInput.wordWrap = textInput.multiline = true;
		add(textInput);
		
		final textInput = new FlxInputText(textInput.x, textInput.y + textInput.height + 5, FlxG.width - 20, "This is an uneditable text input.", 16);
		textInput.editable = false;
		add(textInput);
		
		final textInput = new FlxInputText(textInput.x, textInput.y + textInput.height + 5, FlxG.width - 20, "This is an unselectable text input.", 16);
		textInput.selectable = false;
		add(textInput);
		
		final textInput = new FlxInputText(textInput.x, textInput.y + textInput.height + 5, FlxG.width - 20, "This is a formatted text input.");
		textInput.setFormat(FlxAssets.FONT_DEBUGGER, 16, FlxColor.LIME);
		textInput.italic = textInput.bold = true;
		add(textInput);
		
		final textInput = new FlxInputText(textInput.x, textInput.y + textInput.height + 5, FlxG.width - 20, "This text input is aligned to the center.", 16);
		textInput.alignment = CENTER;
		add(textInput);
		
		final textInput = new FlxInputText(textInput.x, textInput.y + textInput.height + 5, FlxG.width - 20, "This text input is aligned to the right.", 16);
		textInput.alignment = RIGHT;
		add(textInput);
		
		final textInput = new FlxInputText(textInput.x, textInput.y + textInput.height + 5, FlxG.width - 20, "Maximum characters: 30", 16);
		textInput.maxLength = 30;
		add(textInput);
		
		final textInput = new FlxInputText(textInput.x, textInput.y + textInput.height + 5, FlxG.width - 20, "only lowercase", 16);
		textInput.forceCase = LOWER_CASE;
		add(textInput);
		
		final textInput = new FlxInputText(textInput.x, textInput.y + textInput.height + 5, FlxG.width - 20, "ONLY UPPERCASE", 16);
		textInput.forceCase = UPPER_CASE;
		add(textInput);
		
		final textInput = new FlxInputText(textInput.x, textInput.y + textInput.height + 5, FlxG.width - 20, "OnlyAlphanumeric123", 16);
		textInput.filterMode = ONLY_ALPHANUMERIC;
		add(textInput);
		
		final textInput = new FlxInputText(textInput.x, textInput.y + textInput.height + 5, FlxG.width - 20, "password", 16);
		textInput.passwordMode = true;
		add(textInput);
		
		final textInput = new FlxInputText(textInput.x, textInput.y + textInput.height + 5, FlxG.width - 20, "This is a customized text input.", 16);
		textInput.caretWidth = 2;
		textInput.caretColor = FlxColor.BLUE;
		textInput.selectionColor = FlxColor.BLUE;
		textInput.selectedTextColor = FlxColor.WHITE;
		add(textInput);
		
		final textInput = new FlxInputText(textInput.x, textInput.y + textInput.height + 5, FlxG.width - 20, "This is a text input without a background.", 16,
			FlxColor.WHITE, FlxColor.TRANSPARENT);
		add(textInput);
	}
}
```

Geokureli (changes requested):
From a functionality standpoint, this is solid, no issues in my tests, so far and it behaves and reacts very well. From a structure standpoint, it's not exactly what I expected but no changes needed. I expected the util part of this to simple dispatch events for anything that was listening, but I like that the manager also manages focus, so this makes sense.

I still need to test on a couple more targets, and give the code a full scan, but I'm not foreseeing many other changes other than the name/package change i mentioned

> * Demos should be added to showcase the new input text, but I'm not good at coming up with ideas for those.

Honestly the above code is a good starting point for a demo, I would just change "This is a customized text input" to "This is a text input with custom selection colors." No rush on that

flixel/FlxG.hx
```diff
@@ -320,6 +321,11 @@ class FlxG
 	 */
 	public static var plugins(default, null):PluginFrontEnd;
 
+	/**
+	 * Used for detecting text input and dispatching commands on input text fields.
+	 */
+	public static var inputText(default, null):InputTextFrontEnd = new InputTextFrontEnd();
+
```
Geokureli, line 328: Oh also, I recommend handling this, in the same manner as [FlxTweens and FlxTimer](https://github.com/HaxeFlixel/flixel/blob/ac488bf401880dbd5248b04f51a9c3492c8a39b8/flixel/system/frontEnds/PluginFrontEnd.hx#L138-L140):

Starmapo: Where should the `globalManager` variable go though? The manager is intended to support more than just FlxInputText (e.g. an FlxBitmapInputText down the line), so putting it in that class doesn't seem like the best option.

Geokureli: For now lets put it in FlxInputText, in the future when FlxBitmapInputText is made we could always have a static `globalManager` getter that references FlxInputText.globalManager. Alternatively, we might move this to a more centralized location, later

DetectiveBaldi:
is it possible that fields like `FlxInputText.callback` could be migrated to `FlxSignal`'s instead of single functions?

Starmapo:
> is it possible that fields like `FlxInputText.callback` could be migrated to `FlxSignal`'s instead of single functions?

I would've done that, but I was trying to preserve compatibility with the original input text from flixel-ui.

Geokureli:
> > is it possible that fields like `FlxInputText.callback` could be migrated to `FlxSignal`'s instead of single functions?
> 
> I would've done that, but I was trying to preserve compatibility with the original input text from flixel-ui.

Let's talk about this! I wonder if we can get away with doing whatever we want in this new class. My initial thoughts:
- People can still use the old class even after this new one is added
- We want people to be able to easily migrate from the old to the new
- We'll, eventually, have to deprecate these legacy fields anyway

Would you mind listing some of these legacy features you migrated over? Specifically ones that you think are outdated

Geokureli (approved):

Starmapo:
> Would you mind listing some of these legacy features you migrated over? Specifically ones that you think are outdated

Sure:
- `callback`: Like DetectiveBaldi said and like I implemented in my library, I would prefer to have these events as separate FlxSignals. We could have both of these options at the same time, with `callback` being there for compatibility purposes.
- `customFilterPattern`: If we had a normal enum for the `filterMode` variable, we could include the filter as an argument for `CUSTOM_FILTER` instead of keeping a separate variable for it. However the original input text used Ints as the filter mode options, so I had to use an enum abstract for compatibility.
- `filterMode`: Just my opinion but maybe we should do like OpenFL and make the user put in their own regex instead of choosing from an enum, as the current options might be lacking in what the coder expects (e.g. the ONLY_ALPHA filter doesn't allow spaces, and the ONLY_NUMERIC filter doesn't allow a minus sign or a decimal point).
- `focusGained` and `focusLost`: Like the other callback, I would prefer to have these as FlxSignals.

Starmapo:
> personally I'm leaning towards 3 unless you think people will struggle with migration (I never use input text so I unno the experience, here)

I don't think it'll be complicated for people to migrate, so I agree with option 3 as well.

I also agree with your version of the filter mode enum, I'll work on the changes later.

Starmapo:
For the migration guide:
- `callback`: Use the `onTextChange` signal instead.
- `customFilterPattern`: Use the `REG(pattern)` enum value in `filterMode` instead.
- `filterMode`: Rename the enum values accordingly:
  - `NO_FILTER` -> `NONE`
  - `ONLY_ALPHA` -> `ALPHABET`
  - `ONLY_NUMERIC` -> `NUMERIC`
  - `ONLY_ALPHANUMERIC` -> `ALPHANUMERIC`
  - `CUSTOM_FILTER` -> `REG(pattern)` (replace `pattern` with what you put in `customFilterPattern`)
- `focusGained` and `focusLost`: Use the `onFocusChange` signal and check the parameter to see if focus has been gained or lost.
- `lines`: Use `multiline` to define whether the text field allows new lines or not.

flixel/text/FlxInputText.hx
```diff
@@ -1055,7 +1061,7 @@ class FlxInputText extends FlxText implements IFlxInputText
 				{
 					restartCaretTimer();
 				}
-				onTextChange.dispatch(text, ENTER_ACTION);
+				onEnter.dispatch(text);
```
Geokureli, line 1064: why wouldn't people want text change events to include enters?

Starmapo: Because its dispatched even if the text isn't multiline and thus doesn't create a new line, so it's not really a "change" event. If the text is multiline, then the new line gets dispatched in `INPUT_ACTION`.

Geokureli: Thats a decent argument for why people wouldn't want enters to dispatch text change events each time, but I still think hitting enter in a multiline text should dispatch a text change event

Starmapo: > I still think hitting enter in a multiline text should dispatch a text change event

Yes, thats why I said the signal still gets dispatched with `INPUT_ACTION` whenever you press enter to add a new line, as the same `addText` function is called for when you input text normally.

Geokureli: tried this for a reference, it doesn't dispatch when multiline if false, but does when true. which makes sense
```hx
text = new TextField();
text.text = "The quick brown fox jumps over the lazy dog";
text.type = INPUT;
text.multiline = false;
text.addEventListener(openfl.events.Event.CHANGE, (e)->trace(text.text));
```

Geokureli: > then the new line gets dispatched in `INPUT_ACTION`.

Wait are you saying you took out ENTER_ACTION becuase it's covered by INPUT_ACTION?

Starmapo: > > then the new line gets dispatched in `INPUT_ACTION`.
> 
> Wait are you saying you took out ENTER_ACTION becuase it's covered by INPUT_ACTION?

No, I said that to specify that adding a new line does dispatch an INPUT_ACTION event, which only happens when the text is multiline. ENTER_ACTION only checks for if enter was pressed regardless of if the text is multiline or not, so its not guaranteed to be caused from a change in the text.

Geokureli: sorry for late reply, this week has been a shit show for me, if I understand correctly, then this sounds fine

Geokureli: Thinking out loud here: At first I thought an enter event seemed unneeded, but I suppose if the dev wants to, say send a chat message on enter or switch focus to the next field, then it's important that the enter action is handled via the action order rather than `FlxG.keys.justPressed` because that might cause actions to happen out of order

good thinking!

Geokureli:
Haven't had time to fully look this over, I've had distracting family matters (everything is okay. Hopefully, I'll be able to look over the weekend

flixel/text/FlxInputText.hx
Starmapo: Honestly, I only added it here because the original flixel-ui input text had it 😅 But you're right, it does seem useful to keep

Starmapo:
> Haven't had time to fully look this over, I've had distracting family matters (everything is okay. Hopefully, I'll be able to look over the weekend

No pressure at all 👍

Geokureli:
there should be a destroy(), btw should implement `IFlxDestroyable`

Starmapo:
> there should be a destroy(), btw

You mean for the FlxInputTextManager?

Geokureli:
> > there should be a destroy(), btw
> 
> You mean for the FlxInputTextManager?

yes

DetectiveBaldi:
@Starmapo iirc your text library is used in `haxeui-flixel` for better text management. once 5.9.0 releases/when this is merged, willy ou be able to migrate haxeui to using the now standard flixel input text instead?

Starmapo:
> @Starmapo iirc your text library is used in `haxeui-flixel` for better text management. once 5.9.0 releases/when this is merged, will you be able to migrate haxeui to using the now standard flixel input text instead?

Yes, I already have the necessary modifications on my local repository, once 5.9.0 is released I will open the PR for haxeui-flixel.

Geokureli:
Some changes:
- added `onTypingAction` to the manager, so anything can listen for events
- allow custom input managers and `inputText.setManager(mng)`
- remove `set_focus` for `setFocus`. (not a fan of setters that do more than set some converted value)
- make `_registeredInputTexts` final, simplifying init/destroy, improving null safety
- fix error with removing inputs while looping through (causing missed removals)
- change `case FOO: if (condition) {foo;}` to `case FOO if (condition): foo;`
- recreate mac cursor movement behavior via modifier keys
- rename HOME and END move commands to TOP and BOTTOM (make distinct from key names, which behave differently on mac)
- fix the issue where select dragging from right to left causes the selection to be left to right

Geokureli:
I'm seeing more heavy overuse of setters for handling huge portions of code, in my experience this is a recipe for fragile code, especially when different setters set each other recursively.

I see more instances of unneeded complexity that I'd like to deal with now

Geokureli:
> I see more instances of unneeded complexity that I'd like to deal with now

I realize that many of these things are copied from the old FlxInputText, and don't need to be corrected now

Geokureli:
Sorry this took so long, I think I was a little nervous since it's fuckin huge, also dealing with some family stuff atm

Thanks a ton!

Starmapo:
> Sorry this took so long, I think I was a little nervous since it's fuckin huge, also dealing with some family stuff atm

Don't worry about it, I wasn't in a rush or anything, and I agree its pretty huge. Thank you for accepting it ^^

PR #3222, opened 2024-07-20
Little changes to `FlxDebugger` code completion.

DetectiveBaldi:
Currently, typing out a simple line in the console, (e.g. `FlxG.state.exists = false;`) and pressing enter will result in code completion automatically adding a phrase to your command, instead of just running it. (resulting in e.g. `FlxG.state.exists = false;clearBitmapLog`)

To fix this I simply added a case to `CompletionHandler.hx` that closes the code completion if the current command ends with `;`.

This pull requests also makes a change to `CompletionList.hx` that allows you to press TAB (along with ENTER) in order to fill in the currently selected completion. This is a common bind in a lot of other programs.


Geokureli:
Thanks!

PR #3225, opened 2024-08-09
fix console completion selection on HTML5

ninjamuffin99:
On HTML5, `FlxG.stage.addEventListener` seems to not properly do the keyboard events for the CompletionList. 

This PR changes it so that `onKeyDown` is passed through from the debug consoles input event listeners, so using the arrow keys to move through the completions now works on HTML5 properly

### DIFF
--- a/flixel/system/debug/completion/CompletionList.hx
+++ b/flixel/system/debug/completion/CompletionList.hx
@@ -13,11 +13,11 @@ class CompletionList extends Sprite
 	public var completed:String->Void;
 	public var selectionChanged:String->Void;
 	public var closed:Void->Void;
-
+	
 	public var filter(default, set):String;
-
+	
 	public var items(default, null):Array<String>;
-
+	
 	var entries:Array<CompletionListEntry> = [];
 	var originalItems:Array<String>;
 	var selectedIndex:Int = 0;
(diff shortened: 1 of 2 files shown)


flixel/system/debug/completion/CompletionList.hx
```diff
@@ -37,8 +36,6 @@ class CompletionList extends Sprite
 		createPopupEntries(capacity);
 		createScrollBar();
 		updateSelectedItem();
-
-		FlxG.stage.addEventListener(KeyboardEvent.KEY_DOWN, onKeyDown);
```
Geokureli, line 41: Do you know why this wasn't working? Is this a `preventDefault` issue? or is the console blocking these events somehow?

flixel/system/debug/console/Console.hx
```diff
@@ -17,7 +17,6 @@ import openfl.events.FocusEvent;
 import openfl.events.KeyboardEvent;
 import openfl.text.TextFieldType;
 import openfl.ui.Keyboard;
-
```
Geokureli, line 20: could you add `"editor.formatOnSaveMode": "modifications"`  to your settings.json and `"trailingWhitespace": true` to your hxformat.json, for future flixel PRs?

flixel/system/debug/completion/CompletionList.hx
ninjamuffin99: I'm unsure, I believe it's not console or `FlxG.stage` specific, since `Lib.current.addEventListener` also doesn't register tab key, and also the autofocus event listener doesn't work either ("Press TAB to focus the console!")

flixel/system/debug/console/Console.hx
ninjamuffin99: note, I had to put `"trailingWhitespace": false` or else it was inputting tabs as whitespace after each line with whitespace! but all whitespacing related accidental changes should be fixed

flixel/system/debug/console/Console.hx
```diff
@@ -196,7 +196,10 @@ class Console extends Window
 	function onKeyDown(e:KeyboardEvent)
 	{
 		if (completionList.visible)
+		{
+			completionList.onKeyDown(e);
```
Geokureli, line 200:
```suggestion
			// Fixes issue with listening for key down events - https://github.com/HaxeFlixel/flixel/pull/3225
			completionList.onKeyDown(e);
```

Geokureli:
Thanks, if we figure out what's blocking events we can revisit this

PR #3226, opened 2024-08-10
Fix text clipping when applying a shadow to it

Vortex2Oblivion:
Currently there seems to be an issue where applying a shadow to a FlxText, the text clips off by whatever the shadow offset is.
image
Here the shadow offset is 8 pixels and the text clips off by 8 pixels without the fix.

With the fix the text appears as it is supposed to be, the text does not clip off
image
Ive tested this on HashLink and C++ targets, it seems to just work(?) but I am unsure if there are any side effects.

I made a quick little test project [here](https://github.com/Vortex2Oblivion/flxtext-shadow-fixes)

### DIFF
--- a/flixel/text/FlxText.hx
+++ b/flixel/text/FlxText.hx
@@ -839,7 +839,7 @@ class FlxText extends FlxSprite
 	override function get_height():Float
 	{
 		regenGraphic();
-		return super.get_height() + shadowOffset.y;
+		return super.get_height();
 	}
 
 	override function updateColorTransform():Void
--- a/flixel/text/FlxText.hx
+++ b/flixel/text/FlxText.hx
@@ -839,7 +839,7 @@ class FlxText extends FlxSprite
 	override function get_height():Float
 	{
 		regenGraphic();
-		return super.get_height();
+		return super.get_height() + shadowOffset.y;
 	}
 
 	override function updateColorTransform():Void
@@ -875,11 +875,11 @@ class FlxText extends FlxSprite
 			oldHeight = graphic.height;
 		}
 
-		var newWidth:Int = Math.ceil(textField.width);
+		var newWidth:Int = Math.ceil(textField.width) + Std.int(shadowOffset?.x);
 		var textfieldHeight = _autoHeight ? textField.textHeight : textField.height;
 		var vertGutter = _autoHeight ? VERTICAL_GUTTER : 0;
 		// Account for gutter
-		var newHeight:Int = Math.ceil(textfieldHeight) + vertGutter;
+		var newHeight:Int = Math.ceil(textfieldHeight) + vertGutter + Std.int(shadowOffset?.y);
 
 		// prevent text height from shrinking on flash if text == ""
 		if (textField.textHeight == 0)
@@ -1042,7 +1042,6 @@ class FlxText extends FlxSprite
 				}
 
 				_matrix.translate(-shadowOffset.x * borderSize, -shadowOffset.y * borderSize);
-
 			case OUTLINE:
 				// Render an outline around the text
 				// (do 8 offset draw calls)
--- a/flixel/text/FlxText.hx
+++ b/flixel/text/FlxText.hx
@@ -875,11 +875,17 @@ class FlxText extends FlxSprite
 			oldHeight = graphic.height;
 		}
 
-		var newWidth:Int = Math.ceil(textField.width) + Std.int(shadowOffset?.x);
+		var newWidth:Int = Math.ceil(textField.width);
 		var textfieldHeight = _autoHeight ? textField.textHeight : textField.height;
 		var vertGutter = _autoHeight ? VERTICAL_GUTTER : 0;
 		// Account for gutter
-		var newHeight:Int = Math.ceil(textfieldHeight) + vertGutter + Std.int(shadowOffset?.y);
+		var newHeight:Int = Math.ceil(textfieldHeight) + vertGutter;
+
+		if (shadowOffset != null)
+		{
+			newWidth += Math.ceil(shadowOffset.x);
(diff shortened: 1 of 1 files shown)


Vortex2Oblivion:
image
Actually it seems that it causes the shadow to clip off at times when setting it to a larger number?
Here I set it to 22.

Sword352:
I think the bitmap itself should be bigger (depending on the shadow offset) in order to draw the full text and shadow

Vortex2Oblivion:
> I think the bitmap itself should be bigger (depending on the shadow offset) in order to draw the full text and shadow

What do you mean?

Vortex2Oblivion:
> > I think the bitmap itself should be bigger (depending on the shadow offset) in order to draw the full text and shadow
> 
> What do you mean?

Oh wait i think i know hang on

Vortex2Oblivion:
> > > I think the bitmap itself should be bigger (depending on the shadow offset) in order to draw the full text and shadow
> > 
> > 
> > What do you mean?
> 
> Oh wait i think i know hang on

Actually wait I don't know what you mean.

Vortex2Oblivion:
image
I was able to fix it finally i think

Vortex2Oblivion:
Wait it wont work on haxe 4.2.5 :/

Vortex2Oblivion:
image
It still has issues when setting to a negative number

Vortex2Oblivion:
image
Fixed issues when using a negative value

Geokureli:
Looking into this now, it's worth mentioning that the intended way to use this effect is:
```hx
text.setBorderStyle(SHADOW, 0xFF000000, 8, 0);
```
That works just fine for me, I didn't even know shadowOffset existed, until now, and I believe it's actually meant to serve a different purpose. When I use the above code, shadowOffset is still (1, 1). Side note, but it's also kinda weird that it's (1, 1) instead of 0,0 because at `text.setBorderStyle(SHADOW, 0xFF000000, 0, 0);` the shadow appears directly under the text, regardless of the value of shadowOffset

That said, it appears shadowOffset is not meant to be used for the purpose you expected and I recommend using `setBorderStyle` instead. I still think this change may be valid, for people who want to use `shadowOffset` for its intended purpose, but first, I need to find out what that purpose is.

I'll get back to you on this

Geokureli:
Scratch that, looks like there are issues using the intended way, it just needed a shadow size larger than 8 to see
Screenshot 2024-08-15 at 12 36 28 PM

Here's the test I'm doing, I'm seeing some really strange behavior with shadowOffset
```hx
package states;

import flixel.FlxG;
import flixel.text.FlxText;

class TextShadowTestState extends flixel.FlxState
{
	var text:FlxText;
	var showBorder = true;
	var shadowSize = 16;
	var shadowQuality = 0;
	
	var instructions:FlxText;
	
	{
		super.create();
		bgColor = 0xFFff0000;
		
		text = new FlxText();
		text.text = "FlxText shadow clipping test lol";
		text.size = 32;
		text.screenCenter();
		// text.borderStyle = SHADOW;
		// text.shadowOffset.set(8, 8);
		applyShadow();
		add(text);
		
		// debug watch
		FlxG.watch.addFunction("bmd", function ()
		{
			final bmp = text.graphic.bitmap;
			return '( w: ${bmp.width} | h: ${bmp.height} )';
		});
		FlxG.watch.addFunction("shd.offset", ()->text.shadowOffset);
		FlxG.watch.addFunction("shd.size", ()->text.borderSize);
		FlxG.watch.addFunction("shd.quality", ()->text.borderQuality);
		FlxG.watch.addFunction("shd.style", ()->text.borderStyle.getName());
		FlxG.log.notice("LEFT/RIGHT: change shadowOffset");
		FlxG.log.notice("SPACE: reset shadowOffset");
		FlxG.log.notice("UP/DOWN: change borderSize");
		FlxG.log.notice("SHIFT + ^ v: change borderQuality");
		FlxG.log.notice("CLICK: toggle border");
		FlxG.debugger.visible = true;
	}
	
	function applyShadow()
	{
		if (showBorder)
			text.setBorderStyle(SHADOW, 0xFF000000, shadowSize, shadowQuality);
		else
			text.setBorderStyle(NONE);
		
		@:privateAccess text._regen = true;
		@:privateAccess text.regenGraphic();
		FlxG.bitmapLog.clear();
		FlxG.bitmapLog.add(text.graphic.bitmap);
	}

	{
		super.update(elapsed);
		
		var redraw = false;
		
		if (FlxG.keys.anyJustPressed([SPACE, RIGHT, LEFT]))
		{
			if (FlxG.keys.justPressed.SPACE) text.shadowOffset.set(1, 1);
			if (FlxG.keys.justPressed.RIGHT) text.shadowOffset.add(1, 1);
			if (FlxG.keys.justPressed.LEFT) text.shadowOffset.add(-1, -1);
			
			redraw = true;
		}
		
		if (FlxG.keys.anyJustPressed([UP, DOWN]))
		{
			if (FlxG.keys.pressed.SHIFT)
			{
				if (FlxG.keys.justPressed.UP && shadowQuality < 2) shadowQuality++;
				if (FlxG.keys.justPressed.DOWN && shadowQuality > 0) shadowQuality--;
			}
			else
			{
				if (FlxG.keys.justPressed.UP) shadowSize++;
				if (FlxG.keys.justPressed.DOWN) shadowSize--;
			}
			
			redraw = true;
		}
		
		if (FlxG.mouse.justPressed)
		{
			showBorder = !showBorder;
			redraw = true;
		}
		
		if (redraw)
			applyShadow();
	}
}
```

Vortex2Oblivion:
I cant seem to find out how im going to fix the issue when using setBorderStyle

Geokureli:
I've started a new PR with a different fix:
https://github.com/HaxeFlixel/flixel/pull/3236

PR #3227, opened 2024-08-11
Remove Matrix Math from default shader

NeeEoo:
Since the current matrix math can be simplified to a simple multiply

### DIFF
--- a/flixel/graphics/tile/FlxGraphicsShader.hx
+++ b/flixel/graphics/tile/FlxGraphicsShader.hx
@@ -6,18 +6,18 @@ class FlxGraphicsShader extends GraphicsShader
 {
 	@:glVertexSource("
 		#pragma header
-		
+
 		attribute float alpha;
 		attribute vec4 colorMultiplier;
 		attribute vec4 colorOffset;
 		uniform bool hasColorTransform;
-		
+
 		void main(void)
 		{
 			#pragma body
-			
+
 			openfl_Alphav = openfl_Alpha * alpha;
-			
+
 			if (hasColorTransform)
 			{
 				openfl_ColorOffsetv = colorOffset / 255.0;
@@ -48,13 +48,7 @@ class FlxGraphicsShader extends GraphicsShader
 
 			color = vec4(color.rgb / color.a, color.a);
 
-			mat4 colorMultiplier = mat4(0);
-			colorMultiplier[0][0] = openfl_ColorMultiplierv.x;
-			colorMultiplier[1][1] = openfl_ColorMultiplierv.y;
-			colorMultiplier[2][2] = openfl_ColorMultiplierv.z;
-			colorMultiplier[3][3] = openfl_ColorMultiplierv.w;
-
-			color = clamp(openfl_ColorOffsetv + (color * colorMultiplier), 0.0, 1.0);
+			color = clamp(openfl_ColorOffsetv + (color * openfl_ColorMultiplierv), 0.0, 1.0);
 
 			if (color.a > 0.0)
 			{
@@ -65,7 +59,7 @@ class FlxGraphicsShader extends GraphicsShader
 	")
 	@:glFragmentSource("
 		#pragma header
-		
+
 		void main(void)
 		{
 			gl_FragColor = flixel_texture2D(bitmap, openfl_TextureCoordv);


Geokureli:
For future PRs, please edit your settings.json's formatting to include:
`"editor.formatOnSaveMode": "modifications"`
and add the following to your hxformat.json
```json
"indentation":{
    "trailingWhitespace": true
}
```

flixel/graphics/tile/FlxGraphicsShader.hx
```diff
@@ -48,13 +48,7 @@ class FlxGraphicsShader extends GraphicsShader
...
 			color = vec4(color.rgb / color.a, color.a);
 
-			mat4 colorMultiplier = mat4(0);
-			colorMultiplier[0][0] = openfl_ColorMultiplierv.x;
-			colorMultiplier[1][1] = openfl_ColorMultiplierv.y;
-			colorMultiplier[2][2] = openfl_ColorMultiplierv.z;
-			colorMultiplier[3][3] = openfl_ColorMultiplierv.w;
-
-			color = clamp(openfl_ColorOffsetv + (color * colorMultiplier), 0.0, 1.0);
+			color = clamp(openfl_ColorOffsetv + (color * openfl_ColorMultiplierv), 0.0, 1.0);
```
Geokureli, line 51: Where is openfl_ColorMultiplierv set? I tried all the following and found no difference with the old or new code:
```hx
sprite.shader.colorMultiplier.value = [1.0, 0.0, 1.0, 1.0];
sprite.color = 0xFFff00ff;
sprite.setColorTransform(1, 0, 1, 1);
```

NeeEoo: Its handled in openfl and drawquaditem

Its the same usage as before

Geokureli: is there a way I can verify that whatever purpose openfl_ColorMultiplierv served before, is still working after this change?

NeeEoo: Removing the * openfl_ColorMultiplierv would just make it not react to .color

Geokureli: it doesn't react to .color, now before and after this change

NeeEoo: On flxsprite?

NeeEoo: This commit is supposed to be a simplification of the math. To replicate the same behavior.

SrtHero278: > is there a way I can verify that whatever purpose openfl_ColorMultiplierv served before, is still working after this change?

matrix multiplication actually.
this originally did
```
r, 0, 0, 0
0, g, 0, 0
0, 0, b, 0
0, 0, 0, a
(rgba is equal to xyzw, its swizzling)
```
image
matrices multiplied with a vector behave like this.
since all of the other values are 0, it just becomes `[x * m11, y * m22, z * m33, w * m44]`
and m11 is r, m22 is g, m33 is b, and m44 is a, the exact same values.
the matrix is just nothing but a middleman.

Geokureli: I still don't know what purpose `openfl_ColorMultiplierv` has because I can't seem to write any code that effects this property. Until I understand what purpose it has I can't approve any changes to its beahvior

MaybeMaru:
| 0xFFffffff  | 0xFFff0000 |
| ------------- | ------------- |
| image | image  |

nebulazorua:
The intent is for the same behaviour but now its faster
the matrix code was an unnecessary step in the shader and just served to be a bit slower

If you notice no change pre and post change with changing sprite.color, that means its working as intended

flixel/graphics/tile/FlxGraphicsShader.hx
NeeEoo:
```hx
var hue = 0.0;

{
	super.update(elapsed);
	sprite.color = FlxColor.fromHSB((hue += elapsed * 180) % 360, 0.5, 1.0);
}
```

Before my change:
https://github.com/user-attachments/assets/1ae37d92-32e9-4fc0-a94b-0bcce7ae1905


After my change:
https://github.com/user-attachments/assets/b3bb82d7-25b3-4cbb-a5a6-95a88a5b1380

PR #3228, opened 2024-08-11
Remove unused stuffs

NeeEoo:
This removes some old stuff which was no longer used.

### DIFF
--- a/flixel/FlxCamera.hx
+++ b/flixel/FlxCamera.hx
@@ -605,14 +605,12 @@ class FlxCamera extends FlxBasic
 		return startTrianglesBatch(graphic, smooth, colored, blend);
 		#else
 		var itemToReturn = null;
-		var blendInt:Int = FlxDrawBaseItem.blendToInt(blend);
 
 		if (_currentDrawItem != null
 			&& _currentDrawItem.type == FlxDrawItemType.TILES
 			&& _headTiles.graphics == graphic
 			&& _headTiles.colored == colored
 			&& _headTiles.hasColorOffsets == hasColorOffsets
-			&& _headTiles.blending == blendInt
 			&& _headTiles.blend == blend
 			&& _headTiles.antialiasing == smooth
 			&& _headTiles.shader == shader)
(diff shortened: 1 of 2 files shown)


flixel/FlxCamera.hx
```diff
@@ -594,9 +594,9 @@ class FlxCamera extends FlxBasic
 	 * Internal variables, used in blit render mode to draw trianglesSprite on camera's buffer.
 	 * Added for less garbage creation.
 	 */
-	static var renderPoint:FlxPoint = FlxPoint.get();
+	static var renderPoint:FlxPoint = new FlxPoint();
 
-	static var renderRect:FlxRect = FlxRect.get();
+	static var renderRect:FlxRect = new FlxRect();
```
Geokureli, line 599: why change this?

NeeEoo: Its incorrect usage, since its a static it is only ever gotten once. And never .put(). And if someone were to use .put on it it would break

Geokureli: I disagree that it's "incorrect", it's makes absolutely no difference

NeeEoo: Oh i didnt know it was changed. It used to make a difference before the reworked in commit e721fca. Should i revert the commit?

Geokureli: What difference did it make? I don't understand what you're referring to

NeeEoo: Eh just forget it. Do i need to do anything?

Geokureli:
I'm not interested in these changes at the moment, at some point I'd like to assess whether we can add blendmode support back in, until then I'd leave it as is

NeeEoo:
> I'm not interested in these changes at the moment, at some point I'd like to assess whether we can add blendmode support back in, until then I'd leave it as is

This was used when FLX_DRAW_TILES still existed.
But now it doesn't exist anymore, so the code isnt used.

And quads still uses blend modes with overrideBlendMode() in flxdrawquaditem

Geokureli:
I'm aware, as I said, I think before we gut it we should see if it can be reimplemented

PR #3243, opened 2024-08-21
Prevent crash in FlxInputText when switching states

### LINKED ISSUE #3242: Crash when switching state while FlxInputText field is active
- **Haxe version:** 4.3.4
- **Flixel version:** 5.8.0+ (dev)
- **OpenFL version:** 9.3.3
- **Lime version:** 8.1.2
- **Affected targets:** Desktop + Web

Switching to another `FlxState` while a `FlxInputText` field is active causes a `FlxInputTextManager` to remain active and continue dispatching events when switching back to the original state. The code location of the error varies depending on how it's triggered, but it causes the app to access a null `textField`.

I have included a test app to reproduce the issue: [Demo.zip](https://github.com/user-attachments/files/16692674/Demo.zip)

**Steps**

1. Run the app and click in the text field.
2. Press the "S" key to change to the next state while the text field is still active.
3. Press the "S" key to change back to the original state.
4. Upon pressing the Up or Down keys, or even clicking in the field again, the game will crash with a "Null access" error.

If we force the field to lose focus when destroying it, this issue no longer occurs.

DigiEggz:
Prevents null `textField` access in `FlxInputText` when switching between states. Resolves #3242.

### DIFF
--- a/flixel/text/FlxInputText.hx
+++ b/flixel/text/FlxInputText.hx
@@ -482,6 +482,7 @@ class FlxInputText extends FlxText implements IFlxInputText
 	 */
 	override function destroy():Void
 	{
+		endFocus();
 		manager.unregisterInputText(this);
 
 		FlxDestroyUtil.destroy(onEnter);


Geokureli:
Thanks

PR #3244, opened 2024-08-22
Fix mouse wheel scrolling not working on FlxInputText

Starmapo:
Like the title says, this fixes the input text not scrolling when using the mouse wheel on it. This can be seen with just a regular text input object with multiline enabled:

```haxe
var text = new FlxInputText(5, 5, FlxG.width - 20, "Hello world!", 16);
text.fieldHeight = 44;
text.multiline = true;
add(text);
```

I also added a comment with information on the fix.


Geokureli:
Thanks!

Issue #2715, opened 2023-01-17
Change draw to take some kind of DrawContext data


Geokureli:
Been thinking on this for a while. We could remove FlxSpriteGroup if the draw() call stack took some kind of matrix or rendering context from their parent group or w/e.

## The basic idea
create a DrawContext class with a matrix, and any rendering/color info that you want to bubble down to members. So, calling draw on any container/group will cause it to copy the incoming draw context and augment it with it's own orientation and rendering properties, which it will pass down to it's members. Since FlxGroups do not have orientation or color info, they simply pass down the context they received, verbatim, but FlxNests (Or whatever replaces FlxSpriteGroups) will add their info to the context before passing it down.

## Advantages
Not only will this simplify "nesting" over FlxSpriteGroups, which bubble changes down every time they are changed (creating a ton of caveats and confusion), FlxNexts don't need to be limited to sprite members, you can add basic FlxGroups.

We could also give camera's a DrawContext, or its own orientation and color rendering information, that is applied to all sprites rendered to it, allowing cameras to have rotation (related to #2679)

## Drawbacks
The main drawback of this is that you can't perform collision checks between objects in different nests and expect them to collide in regards to their rendered positions, as it will collide them in their respective local positions

Geokureli:
One huge drawback I'm not sure I can overcome is mouse events. if the button is placed locally at, say (0, 0) but the parent is at (100, 100), how do we check for mouse overlaps at it's global position rather than it's nested postion? All of my ideas have issues:
- check for mouse overlap in the draw phase 
    - breaks convention, draw phase should be used for rendering only
    - doesn't address the fundamental issue with getting an object's screen position
- store the DrawContext used on the previous draw
    - hacky
    - doesn't work with sprites drawn to multiple cameras
    - won't know if the parent has moved until the next draw call
- store a ref to the parent 
    - openfl does this, but displayObjects can only have a single parent, flxBasics can be in many groups
    - Even if we allowed a sprite to be in multiple groups but only one "nest", this would mean devs can't utilize draw context in their own containers (can provide examples)

note this isn't just an issue with mouse overlaps, it generally causes issues in knowing the screen position of anything

Geokureli:
I think an easy way to add a refs to each members parents is to create a FlxContainer that extends FlxGroup, add a `parent` field to FlxBasic and only allow them to be in one container, by removing them from their previous parent on add. FlxState should extend FlxContainer instead of FlxGroup

[...]

PR #3245, opened 2024-08-25
add world-space FlxCamera rotation

cyn0x8:
implementation of world-space `FlxCamera` rotation via a new `rotation:Float` property (rotates about the center)

uses a new `_rotationCanvas:Sprite` to not interfere with current `canvas` behavior or whatever people might be doing with it

https://github.com/user-attachments/assets/5518267a-85e0-4a7e-b735-94d83702e52c

> [!note]
> this is my first pr here!! criticism and suggestions and changes are appreciated...
> (especially on the margin calculation since the math was kinda figured out by trial & error)

### DIFF
--- a/flixel/FlxCamera.hx
+++ b/flixel/FlxCamera.hx
@@ -39,8 +39,9 @@ private typedef FlxDrawItem = flixel.graphics.tile.FlxDrawQuadsItem;
  *     |-> `_scrollRect:Sprite` (which is used for cropping camera's graphic, mostly in tile render mode)
  *         |-> `_flashBitmap:Bitmap`  (its bitmapData property is buffer BitmapData, this var is used in blit render mode.
  *                                    Everything is rendered on buffer in blit render mode)
- *         |-> `canvas:Sprite`        (its graphics is used for rendering objects in tile render mode)
- *         |-> `debugLayer:Sprite`    (this sprite is used in tile render mode for rendering debug info, like bounding boxes)
+ *         |-> `_rotationCanvas:Sprite` (used for rendering rotated camera in tile render mode)
+ *             |-> `canvas:Sprite`        (its graphics is used for rendering objects in tile render mode)
+ *             |-> `debugLayer:Sprite`    (this sprite is used in tile render mode for rendering debug info, like bounding boxes)
  */
 class FlxCamera extends FlxBasic
 {
@@ -363,6 +364,11 @@ class FlxCamera extends FlxBasic
 	public var angle(default, set):Float = 0;
 
 	/**
+	 * The camera's world rotation in degrees.
+	 */
+	 public var rotation(default, set):Float = 0;
+
+	/**
 	 * The color tint of the camera display.
 	 */
 	public var color(default, set):FlxColor = FlxColor.WHITE;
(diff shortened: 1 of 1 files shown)


richTrash21:
is there a reason why `Sprite.__transform` is used for rotation instead of `Sprite.rotation`? it also seems to break the flash target

DetectiveBaldi:
i think there was an attempt to do this in #2679, but from what i could tell there was an issue where objects would havebeen visible at a certain rotation weren't
does this pr address this?

cyn0x8:
> i think there was an attempt to do this in #2679, but from what i could tell there was an issue where objects that would have been visible at a certain rotation, weren't does this pr address this?

yes, with the margin changes

cyn0x8:
> is there a reason why `Sprite.__transform` is used for rotation instead of `Sprite.rotation`? it also seems to break the flash target

`Sprite.rotation` seemed to be rotating it about the top left corner rather than the center when i was changing it? unsure if that was just a quirk in my test or not... but regardless im guessing itd be more performant than the matrix math so tomorrow ill try using that + translating the sprite to re-center the rotation

also i saw (before the workflow disappeared??) that `Sprite.__transform` doesnt seem to be present on flash target! changing to `Sprite.rotation` would fix that as well

richTrash21:
> `Sprite.rotation` seemed to be rotating it about the top left corner rather than the center when i was changing it?

tested this myself and this problem was present. in that case it might be useful to look into `Sprite.transform`. using it's `matrix` variable i successfully recreated the rotation effect. this variable is also present in the flash target so theoretically it might also fix the flash target (can't say for sure because i didn't test it on flash).

one more thing: while testing i encountered a bug with resizing the window (affects fullscreen too). it seems to misplace the camera if `rotation` set to any value except `0`

Geokureli:
FYI, we have a proposed way to introduce world-rotation, already, here: https://github.com/HaxeFlixel/flixel/issues/2715
the idea is that we will be able to nest objects within objects that when rotated, rotate all the objects nested inside, as a bonus, this will allow us to easily apply the same transformations to the camera as well. Note in the section labelled "Advantages":
> We could also give camera's a DrawContext, or its own orientation and color rendering information, that is applied to all sprites rendered to it, allowing cameras to have rotation (related to https://github.com/HaxeFlixel/flixel/pull/2679)

It'll take longer to implement, but I'd rather go this route than make some adhoc field, like this.

I have to assume this is kind of a cheat that doesn't honor properties like `sprite.isOnScreen` and therefore wouldn't work with things like FlxBackdrop

cyn0x8:
> FYI, we have a proposed way to introduce world-rotation, already, here: #2715 the idea is that we will be able to nest objects within objects that when rotated, rotate all the objects nested inside, as a bonus, this will allow us to easily apply the same transformations to the camera as well. Note in the section labelled "Advantages":
> 
> > We could also give camera's a DrawContext, or its own orientation and color rendering information, that is applied to all sprites rendered to it, allowing cameras to have rotation (related to #2679)
> 
> It'll take longer to implement, but I'd rather go this route than make some adhoc field, like this.
> 
> I have to assume this is kind of a cheat that doesn't honor properties like `sprite.isOnScreen` and therefore wouldn't work with things like FlxBackdrop

awesome! ill close the pr then

Issue #3233, opened 2024-08-14
Console key events getting blocked


Geokureli:
More info here: https://github.com/HaxeFlixel/flixel/pull/3225

PR #3247, opened 2024-08-28
fix for html5 arrow keys not working on the console

### LINKED ISSUE #2190: HaxeFlixel not detecting arrow key presses
- **Haxe version:** 3.4.7
- **Flixel version:** 4.5.1
- **OpenFL version:** 8.6.2
- **Lime version:** 7.1.1
- **Affected targets:** HTML5

```haxe
package;

import openfl.display.Sprite;
import flixel.FlxG;
import flixel.FlxGame;
import flixel.FlxState;
import flash.events.KeyboardEvent;

enum Mode {
	Broken;  // Most keys detected, but NOT arrow keys
	Working; // All keys detected, including arrow keys
}

class Main extends Sprite
{
	public static var mode : Mode = Mode.Broken;

	public function new()
	{
		super();

		if(mode == Mode.Working) {
			addListener();
		} else {
			addChild(new FlxGame(0, 0, PlayState));
		}
	}

	public static function addListener()
	{
		FlxG.stage.addEventListener(KeyboardEvent.KEY_DOWN, Main.keydown);
	}

	public static function keydown(event:KeyboardEvent)
	{
		if([37, 38, 39, 40].indexOf(event.keyCode) >= 0) {
			trace("Arrow key pressed");
		} else {
			trace("Other key pressed");
		}
	}
}

{
	{
		super.create();
		Main.addListener();
	}
}
```

**Observed behavior:** Key presses like A, B, C, Shift, etc. will trigger a message to console, but arrow keys will not. No `KeyboardEvent` fires for arrow keys.

**Expected behavior:**

Arrow key presses should trigger a console message, just like the other keys do.

ninjamuffin99:
fixes #2190 and #3233 

on HTML5, `preventDefaultKeys` in `FlxKeyboard` also calls `preventDefault()` to the keyboard event. This is helpful for not tabbing or scrolling the page when arrow keys or tab is pressed, but side effect is that it cancels the event propegation from reaching our event listeners fully.

This is a bit of a workaround for that, we directly call the event listener ourselves

### DIFF
--- a/flixel/system/debug/console/Console.hx
+++ b/flixel/system/debug/console/Console.hx
@@ -218,6 +218,24 @@ class Console extends Window
 				if (!history.isEmpty)
 					setText(history.getPreviousCommand());
 
+			#if html5
+			// FlxKeyboard.preventDefaultKeys adds "preventDefault" on HTML5
+			// so it ends up not fully propegating our inputs to the stage/event listeners
+			// we do this small work around so we don't need to mess around with lime/openfl events
+			// todo: support the modifier keys
+			case Keyboard.RIGHT:
+				if (FlxG.keys.preventDefaultKeys.contains(Keyboard.RIGHT))
+				{
+					@:privateAccess
+					input.window_onKeyDown(RIGHT, 0);
+				}
+			case Keyboard.LEFT:
+				if (FlxG.keys.preventDefaultKeys.contains(Keyboard.LEFT))
+				{
+					@:privateAccess
+					input.window_onKeyDown(LEFT, 0);
+				}
+			#end
 			case Keyboard.DOWN:
 				if (!history.isEmpty)
 					setText(history.getNextCommand());


Geokureli:
could we bypass preventDefualt if we add the listener to [the stage rather then the input](https://github.com/HaxeFlixel/flixel/blob/f8874a0af18cb0fa3ede891acdbee3e43ba74895/flixel/system/debug/console/Console.hx#L114-L123)?

Since prevent default seems to be popping up a lot, should we consider some common back-end tool for "unprevented" key-events? (doesn't have to happen now)

ninjamuffin99:
I believe not, that snippet you highlighted *attempts* to focus the console if TAB is pressed, but even if it's added to the stage the event listener doesn't follow through. 

*disclaimer: may perhaps work with arrow keys, I just know tab has some special `preventDefault()` functionality in lime i thinky*

Geokureli:
Sorry for the delay, I wanted to test that LEFT and RIGHT didn't scroll the page horizontally when the console was selected, i.e.: bypass the intent of preventDefault entirely. This seems to work fine in [my test](https://geokureli.itch.io/test-1) (pw: flixel)

PR #3248, opened 2024-09-09
logarithmic volume control for global soundtray / volume

ninjamuffin99:
the global volume math is linear, however actual sound and dB stuff is NOT linear. 
-10dB is NOT 10x quieter than -1dB for example

This PR changes specifically the GLOBAL volume control when using the soundtray. Changing via `FlxG.sound.volume = 0.5` for example would scale it linearly as normal, however pressing the + and - keys will properly change the volume logarithmically to sound more "proper". Pressing - until it's about halfway will actually sound like the volume is set to half.


CyndaquilDAC:
THANK YOU. i KNEW something was up with the volume scale in the soundtray!!!

Vortex2Oblivion:
Could linearToLog and logToLinear maybe go in FlxMath instead of are they probably not going to get used

CyndaquilDAC:
> Could linearToLog and logToLinear maybe go in FlxMath instead of are they probably not going to get used

I'm pretty sure this is just for DB purposes. If they were to go in FlxMath, they'd probably need more descriptive names.

Geokureli:
Do we have any examples of how other systems handle volume? I'm looking at [wikipedia](https://en.wikipedia.org/wiki/Decibel) and I see that 10db is 10x the power 0db, and 20db is 100x, but do people expect that behavior with volume controls? 

> -10dB is NOT 10x quieter than -1dB for example

Can we talk specifically about a 1db sound played at certain volume levels? What currently happens and what happens with this change. the above sentence confuses me because you only talk about decibels, not volume levels, and -10db is 10x quieter than 0db so I'm not sure what point you're trying to make.

Also, I think this implementation is flawed, for a few reasons. Consider the following:
```hx
FlxG.sound.volume = 0;
FlxG.sound.changeVolume(0.5);
assert(FlxG.sound.volume == 0.5);// fails volume is 0.0316227766016838
```
there's inconsistent behavior, now, between setting volume directly and changing the volume relatively.

If I'm smelling what you're stepping in, it sounds like, currently, a volume of 0.5 will play a 1db sound at 0.5db, which is not half the "power" and with this change, a volume of 0.5 will be half the power which is actually like -3db?

Geokureli:
found [this](https://learn.microsoft.com/en-us/windows/win32/coreaudio/audio-tapered-volume-controls)
image
Am I crazy or does windows behave the same way flixel does currently?

CyndaquilDAC:
> found [this](https://learn.microsoft.com/en-us/windows/win32/coreaudio/audio-tapered-volume-controls) image Am I crazy or does windows behave the same way flixel does currently? not gonna lie it's hard for me to wrap my head around this stuff

if this is a breaking issue, maybe fully change how flxsound volume works as a breaking change in a major version update instead? the current linear scale is really annyoing (you have to change master volume SUPER low for it to be decently low) but inconsistencies like this arent great either.

Geokureli:
it might be a breaking change, but I figured we could could add a compile flag FLX_LOG_VOLUME and FLX_LINEAR_VOLUME, or even just have a global `FlxG.sound.volumeTapering` or something. im not worried about that, I'm just trying to suggest a different implementation

ninjamuffin99:
what are thoughts on a variable like `scaledVolume`, which if set, it will set to a properly logarithm calculated volume, while still maintaining the `volume` variable if someone wants to change the volume in the linear way. 

So `sound.scaledVolume = 0.5;` will change `volume` and vice versa, so it can be changed without going through a math function directly, and you'd have easy access to both a scaled "auditory perceptive" volume, but also the actual linear volume value.

Geokureli:
1. I'm still not clear on the actual goal and problem, and 2. I still don't like where the conversion is being done.

on the first point:
> Can we talk specifically about a 1db sound played at certain volume levels? What currently happens and what happens with this change. the above sentence confuses me because you only talk about decibels, not volume levels, and -10db is 10x quieter than 0db so I'm not sure what point you're trying to make.

either that or give me some kind of description of the experience you're currently having with sound and what you think should happen, like: "i think .5 volume is too loud, it doesn't sound half as loud as 1.0"

For the second point, I still say the conversion should not be done in `FlxG.sound.volume`, but rather [here](https://github.com/HaxeFlixel/flixel/blob/dev/flixel/sound/FlxSound.hx#L603). this way we are applying the same curve for all FlxSound.volume, FlxG.sound.volume, mySoundGroup.volume and proximity sounds

Lastly, I tried out this branch, I honestly believe 0.5 is now far too quiet with this curve, most sounds I tried seem too quiet to hear at 0.3 volume or less. I graphed the curve of linearToLog and it looks like this:

Screenshot 2024-09-16 at 12 47 05 PM

The opposite of the curve that I see used elsewhere:

Screenshot 2024-09-16 at 1 04 52 PM

What makes intuitive sense to me, is that flixel volume curve should match our PCs' volume curves, so I'm going to record sounds with various speaker volumes and compare the db to playing a sound in flixel at various volume levels

Geokureli:
I'm taking all non-breaking changes off 6.0.0, but I promise i'll get to this on the next release

Geokureli:
untested, for now, but this should account for group volume, sound tray volume, fade volume and individual sound volume

Geokureli:
I'll be honest, I still don't get this logarithmic sound curve. I messed with a few apps and yeah, it seems like spotify and mac use logarithmic, but youtube doesn't. In my tests, I fucking hate the log ones and completely disagree with that article's assessments (assuming this follows that). For example

> This means that we are much more sensitive to small variations in amplitude for silent sounds than for loud sounds

Using your log curve vs the old linear curve made no difference in whether I could distinguish a difference in volume, every increment sounded distinct, but lower volumes sounded far, far less distinct using the log curve, where linear, to me, seemed like each increment was as distinct as the others.

> The perceived maximum volume level is reached around the middle of the slider, making the upper half useless

IMO the left side of the slider is useless in logarithmic, my spotify has never gone below 50% and in the example given in the article I can't hear the sound until I'm at 50% despite my system volume being at max.

Maybe this is just a personal preference, So I made it easy to customize the global sound curve, but I'm keeping the default as it was

Geokureli:
I'm down to revisit this, later. I still feel like there's something I'm not grasping. but hopefully this lets everyone do what they want, for now

PR #3251, opened 2024-09-16
Resize the debug watch window whenever something gets added / removed

ninjamuffin99:
The debugger Watch window would overflow the contents and show offscreen until the user clicks+drags the window around (which then automatically resizes it properly) 

This PR makes it so that whenever something is added to the watch window, it will resize itself.

Using FNF (apologies for the .mov files!)
Before

https://github.com/user-attachments/assets/93446463-9837-4ddb-9dc5-668c16b38e99

After

https://github.com/user-attachments/assets/e766c59b-3cf0-4863-83f1-9d35414ef32f

### DIFF
--- a/flixel/system/debug/watch/Watch.hx
+++ b/flixel/system/debug/watch/Watch.hx
@@ -30,7 +30,7 @@ class Watch extends Window
 		entriesContainer.y = entriesContainerOffset.y;
 		addChild(entriesContainer);
 
-		FlxG.signals.preStateSwitch.add(removeAll);
+		FlxG.signals.preStateSwitch.add(clear);
 	}
 
 	public function add(displayName:String, data:WatchEntryData):Void
@@ -105,15 +105,23 @@ class Watch extends Window
 		updateSize();
 	}
 
-	public function removeAll():Void
+	/**
+	 * internal method to remove all without calling updateSize
+	 */
+	function clear():Void
 	{
-		for (i in 0...entries.length)
+		for (entry in entries)
 		{
-			var entry = entries[i];
 			entriesContainer.removeChild(entry);
 			entry.destroy();
 		}
-		entries.splice(0, entries.length);
+		
+		entries.resize(0);
+	}
+
+	public function removeAll():Void
+	{
+		clear();
 		updateSize();
 	}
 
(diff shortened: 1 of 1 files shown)


Geokureli:
I was ready to merge this but, ngl, these seg faults on unit tests are worrying me, dev branch doesn't have this issue, so it's possibly a related to the change

flixel/system/debug/watch/Watch.hx
```diff
@@ -136,7 +136,6 @@ class Watch extends Window
 		resetEntries();
 		minSize.setTo(getMaxMinWidth() + entriesContainerOffset.x, entriesContainer.height + entriesContainerOffset.y);
 		super.updateSize();
-		reposition(x, y);
```
Geokureli: I didn't see any difference in behavior coming from this line, is there something i'm missing?

Geokureli:
gonna merge later today, lmk if you have issues which my changes, otherwise we can fix in post

flixel/system/debug/watch/Watch.hx
ninjamuffin99: the reposition() comes from the parent `Window.hx` class, which updates it's internal x/y position, as well as repositions the location to be within the window bounds. It's much more noticable if you have a lot of stuff in `FlxG.watch.add()` or whatever that'd overflow the screen. Calling `reposition()` will properly keep the watch window in-bounds, rather than overflowing off the screen where you need to drag it up to see all of the stuff

ninjamuffin99:
here was the demo WITHOUT `reposition()`, as I add more things to the watch window, it doesn't move up until I begin dragging the window with my mouse

https://github.com/user-attachments/assets/ec78f60f-d1c2-4de0-9ce2-caeacb571b51

this one is WITH `reposition()`, as I add things to it, it will move itself up when it would otherwise potentially become offscreen.

https://github.com/user-attachments/assets/95cf6ca7-39f4-4f9d-ade2-28bf184a7553

Geokureli:
maybe it's becuase i tested on html5, but i saw no difference. will try again, what target is that?

Geokureli:
Yup, my mistake, added back

Geokureli:
Thanks!

ninjamuffin99:
woohoo :)

PR #3253, opened 2024-09-18
Fixed a small issue with android config on lime 8.2.0.

MAJigsaw77:
This PR fixes a small were it adds `if="android"` on the outputed `AndroidManifest.xml` on `lime` (`8.2.0`).
asd

### DIFF
--- a/include.xml
+++ b/include.xml
@@ -47,9 +47,9 @@
 		<sample path="${haxelib:flixel-demos}/User Interface" />
 	</section>
 
-	<section if="android">
+	<section if="${lime &ge; 8.2.0}">
 		<config:android>
-			<application android:appCategory="game" if="${lime &ge; 8.2.0}" />
+			<application android:appCategory="game" />
 		</config:android>
 	</section>
 </project>
--- a/include.xml
+++ b/include.xml
@@ -47,7 +47,7 @@
 		<sample path="${haxelib:flixel-demos}/User Interface" />
 	</section>
 
-	<section if="${lime &ge; 8.2.0}">
+	<section if="${lime >= 8.2.0}">
 		<config:android>
 			<application android:appCategory="game" />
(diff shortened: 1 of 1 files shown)


Geokureli:
I'm not familiar with `<config:android>` what does it do, and why `if="android"` is bad here, please elaborate

MAJigsaw77:
> I'm not familiar with what `<config:android>` does, and why `if="android"` is bad here, please elaborate

lime doesn't check for conditions when doing

```xml
<config:android>
	<application android:appCategory="game" if="${lime &ge; 8.2.0}" />
</config:android>
```

It adds up everything within the `config: android` section to the android manifest.

MAJigsaw77:
When i did that pr 8 months ago i haven't thought about that until i tried lime 8.2.0 myself ;--;

Geokureli:
Did you verify that this change works with lime versions earlier than 8.2.0? it seems like it would behave the same as before, but I'd like to double check, since flixel supports lime 8.0.2 at the minimum, and we don't test android in CI (or minimum versions)

MAJigsaw77:
> Did you verify that this change works with lime versions earlier than 8.2.0? it seems like it would behave the same as before, but I'd like to double check, since flixel supports lime 8.0.2 at the minimum, and we don't test android in CI (or minimum versions, yet)

This section can only accure if lime is `8.2.0` or later.
```xml
<section if="${lime >= 8.2.0}">
	<config:android>
		<application android:appCategory="game" />
	</config:android>
</section>
```

These changes have been tested in here https://github.com/FunkinDroidTeam/flixel/blob/logVolume/include.xml and this is the compiled build were the changes have been tested on https://github.com/FunkinDroidTeam/Funkin/actions/runs/10947527345.

Geokureli:
Thanks!

Issue #3252, opened 2024-09-18
`FlxG.camera.bgColor` isn't reset on state switch


DetectiveBaldi:
I'm not sure if this is intentional, but it seems [this getter](https://github.com/HaxeFlixel/flixel/blob/7e7eb5cfe31e2ccbdc49561ca90b7d9108d3d2c3/flixel/system/frontEnds/CameraFrontEnd.hx#L328) for `FlxG.cameras.bgColor` is the cause.
When `FlxG.cameras.reset` is called when switching states, a new camera is created, but because of this getter, it ends up [taking the `bgColor` from the last `FlxG.camera`](https://github.com/HaxeFlixel/flixel/blob/7e7eb5cfe31e2ccbdc49561ca90b7d9108d3d2c3/flixel/FlxCamera.hx#L1083C3-L1083C34).
I'm not sure why this getter exists, in my opinion it makes more sense for `FlxG.cameras.bgColor` to just be a setter that isn't automatically determined by the `bgColor` of `FlxG.camera`.
If there is a reason for this or if it is actually a bug, I would like to know. In my opinion it seems like strange behavior.

Geokureli:
yeah this seems unintentional, but I do think this is *kind of* a breaking change, so I'm adding it to the 6.0.0 release, so we can add it to the migration guide

DetectiveBaldi:
Yeah I understand merging something like this into 6.0.0. I honestly would vouch removing the writing `bgColor` does to cameras in the same pull request, I think it's related enough.
Either way I can have a pull request fixing this issue in a couple hours.

Geokureli:
Looks like "Closes: [issue number]" doesn't actually close them. In the future try, "Fixes", instead. i tried editing it but that didn't work because only happens when it's merged, I presume

Geokureli:
Apparently this broke [MinimalistTD](https://haxeflixel.com/demos/MinimalistTD/) and we never noticed because I wasn't republishing demos, each release. As of now the demos are 5.9.0

So I was correct in that this was a breaking change, but it's still a good change, it just wasn't well communicated, and we need to do more testing on releases

PR #3254, opened 2024-09-18
FlxG.camera.bgColor isn't reset on state switch

DetectiveBaldi:
Closes: #3252

We need to reset `FlxG.camera` to `null` when `FlxG.cameras.reset` is called, otherwise when resetting the state, the newly generated camera will have the previous `FlxG.camera`'s `bgColor`.

This is a breaking change: [more information here](https://github.com/HaxeFlixel/flixel/issues/3252#issuecomment-2358954359).


Geokureli:
thx homie

PR #3255, opened 2024-09-20
Add Angle Getter

CobaltBar:
adds a blank getter to angle in FlxObject for other classes to override.
This could be used for an angle offset, for instance.


Geokureli:
tbh i'm not a fan of these getter/setters and the complexity and obfuscation they cause, I'm trying to move away from them. simply add a `getSpecialAngle` function to your FlxObject/sprite extending class and use that

CobaltBar:
I was trying to avoid using a custom variable, so i wrote a macro to add a getter for angle in my use case.
I thought this might be useful in case someone else needed to use `get_angle`

Should I close the PR? (I have no idea how to respond to this)

PR #3259, opened 2024-09-27
Fix FlxTextInput field border not being clipped correctly

### LINKED ISSUE #3258: FlxInputText - fieldBorderThickness isn't working
@Starmapo was this working last you tried it? The default `fieldBorderThickness` is 1 and the default color is black

Example:
```hx
override function create()
{
    super.create();
    
    final text = new flixel.text.FlxInputText(10, 10, FlxG.width - 20, "Enter Username... (Max 40 Characters)", 32);
    text.font = "Arial";
    add(text);
}
```
Result:

Screenshot 2024-09-26 at 3 50 07 PM

Starmapo:
The FlxTextInput field border was not appearing at all, which turns out to have been from the clipping system not accounting for it being outside the bounds of the actual text field, and thus always making it invisible. This PR fixes that issue.

Closes #3258

### DIFF
--- a/flixel/text/FlxInputText.hx
+++ b/flixel/text/FlxInputText.hx
@@ -615,7 +615,7 @@ class FlxInputText extends FlxText implements IFlxInputText
 	 * Clips the sprite inside the bounds of the text field, taking
 	 * `clipRect` into account.
 	 */
-	function clipSprite(sprite:FlxSprite)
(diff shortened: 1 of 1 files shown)


Geokureli:
Thanks! CI had a hiccup, but seems fine now.

PR #3260, opened 2024-09-29
Attach `DEACTIVATE`/`ACTIVATE` events to `stage.nativeWindow`.

MAJigsaw77:
As of the recent changes to OpenFL's `NativeWindow` (https://github.com/openfl/openfl/commit/31e6f2e59b0faf0e10a4f4f229771b41e39c0aea and https://github.com/openfl/openfl/commit/48e150e584336f107ccc12f46ec988760af812ff), an issue on mobile has been fixed where `DEACTIVATE` and `ACTIVATE` events were not being called properly when user presses the power button, putting the phone into sleep mode.

This PR ensures that the events are now attached to `stage.nativeWindow` instead of the `stage` alone. The change doesn't affect other platforms (the events will continue to function as before). However, until the next OpenFL release, this fix won't take effect.

### DIFF
--- a/flixel/FlxGame.hx
+++ b/flixel/FlxGame.hx
@@ -359,8 +359,8 @@ class FlxGame extends Sprite
 		stage.addEventListener(FocusEvent.FOCUS_OUT, onFocusLost);
 		stage.addEventListener(FocusEvent.FOCUS_IN, onFocus);
 		#else
-		stage.addEventListener(Event.DEACTIVATE, onFocusLost);
-		stage.addEventListener(Event.ACTIVATE, onFocus);
+		stage.nativeWindow.addEventListener(Event.DEACTIVATE, onFocusLost);
+		stage.nativeWindow.addEventListener(Event.ACTIVATE, onFocus);
 		#end
 
 		// Instantiate the initial state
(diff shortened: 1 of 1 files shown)


Geokureli:
Is there a reason why the event is no longer dispatched from the stage, with this change?

MAJigsaw77:
> Is there a reason why the event is no longer dispatched from the stage, with this change? Or rather, why didn't the fix make it so that the event is dispatched through the stage, as expected?

Tbh, i was confused as well but they don't wanna fix the stage to dispatch these events but instead they fixed the nativeWindow so `¯⁠\⁠_⁠(⁠ツ⁠)⁠_⁠/⁠¯`.

Geokureli:
My specific concerns are:
- Will this code work in both the newest and oldest allowed versions of openfl/lime and for all targets?
- What changed and on what version?
- Do we need to check openfl version here to determine where to listen for these events

Currently, our [minimum supported](https://github.com/HaxeFlixel/flixel/blob/eb842489f541ced88084ea6c491699272e7a3a8a/flixel/system/macros/FlxDefines.hx#L131-L141) openfl version is 9.2.2 and minimum lime is 8.0.2

Geokureli:
Thanks, and thanks for looking into my concerns

Dimensionscape:
We didn't change the behavior of Stage, I just fixed an issue with NativeWindow. We may apply a similar fix to Stage if it is appropriate as well.

PR #3265, opened 2024-10-13
Fix memory overflow in debugger for select targets

ACrazyTown:
Because `openfl.System.totalMemory` is an Int, after reaching ~2.14GB (32bit int limit) the memory usage will overflow and no longer display accurate values:
Screenshot 2024-10-13 192013
On Hashlink and the C++ targets we can get a more precise 64bit memory usage value, which allows us to display *way* higher values (and this isn't even close to the limit):
Screenshot 2024-10-13 193424

Unrelated: I wonder if it'd be worth it to use `FlxStringUtil.formatBytes()` here

### DIFF
--- a/flixel/system/debug/stats/Stats.hx
+++ b/flixel/system/debug/stats/Stats.hx
@@ -333,7 +333,14 @@ class Stats extends Window
 	 */
 	public inline function currentMem():Float
 	{
-		return (System.totalMemory / 1024) / 1000;
+		#if cpp
+		var usage:Float = cpp.vm.Gc.memInfo64(cpp.vm.Gc.MEM_INFO_USAGE);
+		#elseif hl
+		var usage:Float = hl.Gc.stats().currentMemory;
+		#else
+		var usage:Float = System.totalMemory;
+		#end
+		return (usage / 1024) / 1000;
 	}
 
 	/**


ACrazyTown:
should `flash.system.System.privateMemory` be used instead on Flash? I'm doing a test where I clone a bunch of bitmaps to get the memory usage up, `totalMemory` keeps hovering around 150mb but `privateMemory` seems to provide an an accurate number, the same one like in task manager

Geokureli:
> should `flash.system.System.privateMemory` be used instead on Flash? I'm doing a test where I clone a bunch of bitmaps to get the memory usage up, `totalMemory` keeps hovering around 150mb but `privateMemory` seems to provide an an accurate number, the same one like in task manager

https://airsdk.dev/reference/actionscript/3.0/flash/system/System.html

[privateMemory](https://airsdk.dev/reference/actionscript/3.0/flash/system/System.html#privateMemory) : Number
The entire amount of memory (in bytes) used by an application.

[totalMemory](https://airsdk.dev/reference/actionscript/3.0/flash/system/System.html#totalMemory) : uint
The amount of memory (in bytes) currently in use that has been directly allocated by Flash Player or AIR.

At a glance, totalMemory seems like the right choice, as in it seems like the true memory usage allocated to run the program.

though privateMemory is a number

ACrazyTown:
> At a glance, totalMemory seems like the right choice, as in it seems like the true memory usage allocated to run the program.

What confused me is that in my tests where I just clone a bitmap a bunch of times, `totalMemory` reports something around 150mb that never seems to grow at all:
image
While in task manager I had ~5gb of usage:
image

Maybe I just don't understand what "directly allocated by Flash Player or AIR" means

> though privateMemory is a number

There's also `totalMemoryNumber`, [haxe api says it requires flash 10.1](https://github.com/HaxeFoundation/haxe/blob/4.3.5/std/flash/system/System.hx#L9), I assume this is compatible with Flixel?

Now that I think about it maybe I should just PR these changes to OpenFL and add `totalMemoryNumber` there?

Geokureli:
> Now that I think about it maybe I should just PR these changes to OpenFL and add `totalMemoryNumber` there?

Not a bad idea

ACrazyTown:
My PR in OpenFL was merged, I'll make a new one here once a new version of OpenFL releases so I can figure out the conditionals

Geokureli:
Please link the OpenFL issue heere as well as in the new PR, thanks!

ACrazyTown:
> Please link the OpenFL issue heere as well as in the new PR, thanks!

https://github.com/openfl/openfl/pull/2744

PR #3266, opened 2024-10-14
Use System.totalMemoryNumber in debugger

ACrazyTown:
This is a redo of #3265 that uses the same fix but implemented directly into OpenFL (https://github.com/openfl/openfl/pull/2744)

The debugger will now use the more precise memory usage value when available (OpenFL version higher or equal to 9.4.0) and fallback to the old one if it's not. This conditional should be removed eventually when Flixel bumps up the minimum OpenFL version.


Geokureli:
Thanks!

PR #3270, opened 2024-10-18
Implement error handling for save data unserialization.

### LINKED ISSUE #3269: Save data can get completely overridden if unserialization fails
https://github.com/HaxeFlixel/flixel/blob/f5087ea69e42750893ac4ec29dc4367714ab815b/flixel/util/FlxSave.hx#L459-L466

Relevant issue: FunkinCrew/Funkin#3679

The v0.5.2 update for Funkin' introduced new properties to the user's save data. One of these properties was an Enum value, and as a result older versions of the game will fail to parse the save data.

In the case that save data fails to parse at this stage, `sharedObject.data` will remain null, and the game will act as though it is a fresh installation with no save data, and write to the save data file.

Without modifying this behavior, there is no chance for backup or recovery of the save data.

EliteMasterEric:
`FlxSave.status` will now be set to `ERROR(e)` in the event that `haxe.Unserializer` encounters an error.

Partially resolves #3269 (still requires the application to query for an error and resolve the issue itself and I'm not sure of any good way to fix issues with the Unserializer).

### DIFF
--- a/flixel/util/FlxSave.hx
+++ b/flixel/util/FlxSave.hx
@@ -167,6 +167,17 @@ class FlxSave implements IFlxDestroyable
 			_sharedObject = FlxSharedObject.getLocal(name, path);
 			status = BOUND(name, path);
 		}
+		catch (s:FlxSaveStatus) {
+			this.status = s;
+			switch (this.status) {
+				case EMPTY:
+					return false;
+				case ERROR(_):
+					return false;
+				case BOUND(_, _):
+					return true;
+			}
+		}
 		catch (e:Error)
 		{
 			FlxG.log.error('Error:${e.message} name:"$name", path:"$path".');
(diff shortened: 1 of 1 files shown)


Geokureli:
can i get an example of how to trigger this issue

Beelijah:
geokureli you can download versions v0.5.2 and v0.5.0 of friday night funkin (game based in haxeflixel), and you can run v0.5.2.

then, complete a song in 0.5.2. now launch 0.5.0.

boom, save data wiped. gone. forever. overwritten. destroyed.

Geokureli:
Is there a simpler way to reproduce this?
1. I'm going to have to test this dozens of times
2. I'd like to make a unit test that checks for it
3. I want to better understand what causes it

Beelijah:
oh idk then, my brain smaller than a lays potato chip

AbnormalPoof:
> Is there a simpler way to reproduce this? Preferably a small test state, that manually saves outdated, raw serialized data, and then loads it via FlxSave:
> 
> 1. I'm going to have to test this dozens of times
> 2. I want to test this via breakpoints and debugging tools, not by watching a game crash
> 3. I'd like to make a unit test that checks for it
> 4. I want to better understand what causes it

AFAIK The issue Friday Night Funkin’ had in https://github.com/FunkinCrew/Funkin/issues/3679 was due to an enum value that wasn’t present in older versions. So I *think* it’s simple as making a save with an enum value, and then launching a variant of the project without said enum value. (The save data will likely fail to parse since the project can’t find the enum that’s referred to in the save data)

Geokureli:
Closing, merged #3286

Thanks for laying the foundation, here

PR #3271, opened 2024-10-19
Stop dispatching onFocus event when closing debugger

ACrazyTown:
This PR solves an issue where closing the debugger would dispatch an `onFocus` event.

One of the issues caused by this issue is sounds playing after the debugger gets closed. 
If you play a sound, pause it, open the debugger and then close it, the sound will get unpaused and continue playing.

This code seems to have been added for a reason but I can't reproduce what the comment is saying. setting `FlxG.stage.focus = null;` doesn't seem to dispatch any focus events. Maybe its an old OpenFL version thing? Or target specific?

### DIFF
--- a/flixel/system/frontEnds/DebuggerFrontEnd.hx
+++ b/flixel/system/frontEnds/DebuggerFrontEnd.hx
@@ -173,8 +173,6 @@ class DebuggerFrontEnd
 		if (!Value)
 		{
 			FlxG.stage.focus = null;
-			// setting focus to null will trigger a focus lost event, let's undo that
-			FlxG.game.onFocus(null);
 
 			#if FLX_MOUSE
 			FlxG.mouse.enabled = true;


Geokureli:
Can you clarify the repro steps? I need to play a sound, pause the sound, then press "~" twice and the sound will resume?

ACrazyTown:
> Can you clarify the repro steps? I need to play a sound, pause the sound, then press "~" twice and the sound will resume?

Yup! Play and pause the sound, open the debugger and close it. It'll dispatch a onFocus event which will then call this:
https://github.com/HaxeFlixel/flixel/blob/049e9edb219762a9f6221881448339ebab6eca9a/flixel/FlxGame.hx#L428
which will cause paused sounds to unpause and continue playing.

https://github.com/user-attachments/assets/ced1bd72-3465-49fc-b1cf-531283a45f72

Geokureli:
I'm a little worried about this change but when I tried to looks for alternatives, they all seemed scarier than this. More info here: https://github.com/HaxeFlixel/flixel/issues/3298

For now, this looks good!

PR #3272, opened 2024-10-23
Make `FlxSound#time` more precise

Sword352:
It seems that directly seeking for `channel.position` is less prone to errors when seeking for `sound.time`.
Not sure where the issue actually comes from yet, but this fixed bunch of issues related to music syncing for me.

```haxe
import flixel.FlxG;
import flixel.FlxState;
import flixel.sound.FlxSound;
import flixel.text.FlxText;

{
	var _instrumental:FlxSound;
	var _voices:Array<FlxSound> = [];

	var _resyncText:FlxText;
	var _resyncs:Int = 0;

	{
		super.create();

		FlxG.fixedTimestep = false;

		_instrumental = FlxG.sound.load(AssetPaths.Inst__ogg).play();
		_voices.push(FlxG.sound.load(AssetPaths.Voices_Opponent__ogg).play());
		_voices.push(FlxG.sound.load(AssetPaths.Voices_Player__ogg).play());

		_resyncText = new FlxText(10, 10, 0, "0", 16);
		add(_resyncText);
	}

	{
		for (voice in _voices)
		{
			if (Math.abs(voice.time - _instrumental.time) > 5)
			{
				voice.time = _instrumental.time;
				_resyncText.text = Std.string(++_resyncs);
			}
		}
	}
}
```
Here, the voices would resync often, causing audio stutters. The changes made by this PR seems to fix it, as well as other issues in my other project
Pr in draft as it may be better to fix the root issue instead

NOTE: this PR introduces behaviour change, originally sounds that aren't updated won't have their `time` value changing during playback. Here it changes regardless.

### DIFF
--- a/flixel/sound/FlxSound.hx
+++ b/flixel/sound/FlxSound.hx
@@ -31,54 +31,54 @@ class FlxSound extends FlxBasic
 	 * Only really matters if you are doing proximity/panning stuff.
 	 */
 	public var x:Float;
-	
+
 	/**
 	 * The y position of this sound in world coordinates.
 	 * Only really matters if you are doing proximity/panning stuff.
 	 */
 	public var y:Float;
-	
+
 	/**
 	 * Whether or not this sound should be automatically destroyed when you switch states.
 	 */
 	public var persist:Bool;
-	
+
 	/**
 	 * The ID3 song name. Defaults to null. Currently only works for streamed sounds.
 	 */
 	public var name(default, null):String;
-	
+
 	/**
 	 * The ID3 artist name. Defaults to null. Currently only works for streamed sounds.
 	 */
 	public var artist(default, null):String;
-	
+
 	/**
 	 * Stores the average wave amplitude of both stereo channels
 	 */
 	public var amplitude(default, null):Float;
-	
+
 	/**
 	 * Just the amplitude of the left stereo channel
 	 */
 	public var amplitudeLeft(default, null):Float;
-	
+
 	/**
 	 * Just the amplitude of the right stereo channel
 	 */
 	public var amplitudeRight(default, null):Float;
-	
+
 	/**
 	 * Whether to call `destroy()` when the sound has finished playing.
 	 */
 	public var autoDestroy:Bool;
-	
+
 	/**
 	 * Tracker for sound complete callback. If assigned, will be called
 	 * each time when sound reaches its end.
 	 */
 	public var onComplete:Void->Void;
-	
+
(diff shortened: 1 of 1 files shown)


Geokureli:
I have issues wit this. I feel like this is a part of a larger issue with FlxSound AND I think the behavior change is not insignificant.

My concern, here: Is there a benefit to setting _time in update? the only case I can imagine is where we do something like

```hx
doSomething(mySound.time); // `time` has some value
someProcessThatTakesAFewMilliseconds();
doSomethingElse(mySound.time); // May not be the same value as before
```

In the above case it may be expected, or important that mySound.time has the same value at both function calls, it will in the old code, but potentially not in the new code. The reason this is (sorta) important is that flixel is entirely based on the principle that "game time" stands still until the start of the next update, when game time advances by FlxG.elapsed. But even then sounds are a little different, they may not advance the same as everything else, and they are not effected by FlxG.timeScale, so I'm torn

I'm wondering if it's better to create a new field in FlxSound called position, eventually phase out time, and see where else we can remove the need for sounds to be updated in order to work.

Maybe something like:
```hx
inline function get_position():Float
{
	if (playing)
		_position  = _channel.position;
	
	return _position;
}
```
But with this would we also need to set _position on pause() calls or do paused sounds still have a _channel we can access?

Sword352:
> I feel like this is a part of a larger issue with FlxSound AND I think the behavior change is not insignificant.
yeah that's why the pr used to be in draft

Sword352:
> But even then sounds are a little different, they may not advance the same as everything else

Yeah it would make sense to make sounds not follow this rule since sound playback is entirely independant from the game loop

Raltyro:
I don't think this would change much since getting source time from OpenAL have a delay of somewhere 30ms

Sword352:
The goal isn't to workaround the gaps between source time updates, but this change makes things more stable in my own use case although it could likely be a deeper issue in `FlxSound`

Raltyro:
but it still won't do much, updating time per frame is enough, constantly calling internal function just to get the same value is not okay

Geokureli:
Looking into this again for the 6.2.0 release. I think as long as the documentation states that sound.time is not guaranteed to be update periodically with the rest of flixel's update loops, we can do this. I would not be surprised if many people assumed this behavior already. This may be considered a breaking change, but only in a hyper-specific cases.

Sword352:
Looking back at this, I think this should be investigated further first since the original issue is pretty strange regardless

PR #3279, opened 2024-10-30
Stricter checks for FlxG.bitmap.maxTextureSize

ACrazyTown:
Checks if any of the Lime OpenGL defines are defined & also checks if `FlxG.stage.window` is hardware accelerated

### DIFF
--- a/flixel/graphics/FlxGraphic.hx
+++ b/flixel/graphics/FlxGraphic.hx
@@ -634,6 +634,15 @@ class FlxGraphic implements IFlxDestroyable
 			bitmap = value;
 			width = bitmap.width;
 			height = bitmap.height;
+
+			#if FLX_OPENGL_AVAILABLE
+			var max:Int = FlxG.bitmap.maxTextureSize;
+			if (max != -1)
+			{
+				if (width > max || height > max)
+					FlxG.log.warn('Graphic dimensions (${width}x${height}) exceed the maximum allowed size (${max}x${max}), which may cause rendering issues.');
+			}
+			#end
 		}
 
 		return value;
--- a/flixel/system/macros/FlxDefines.hx
+++ b/flixel/system/macros/FlxDefines.hx
@@ -254,7 +254,9 @@ class FlxDefines
 			define(FLX_TRACK_GRAPHICS);
 
 		#if (lime_opengl || lime_opengles || lime_webgl)
-		define(FLX_OPENGL_AVAILABLE);
+		// FlxG.stage.window.context.attributes.hardware is not always defined during unit tests
+		if (defined(FLX_NO_UNIT_TEST))
+			define(FLX_OPENGL_AVAILABLE);
 		#end
 		
 		defineInversion(FLX_TRACK_GRAPHICS, FLX_NO_TRACK_GRAPHICS);
--- a/flixel/system/macros/FlxDefines.hx
+++ b/flixel/system/macros/FlxDefines.hx
@@ -86,6 +86,7 @@ private enum HelperDefines
 	FLX_HEALTH;
 	FLX_NO_TRACK_POOLS;
 	FLX_NO_TRACK_GRAPHICS;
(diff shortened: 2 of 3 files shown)


Geokureli:
Can we add a new define to [FlxDefines.HelperDefines](https://github.com/HaxeFlixel/flixel/blob/dev/flixel/system/macros/FlxDefines.hx#L53), instead? like `FLX_HAS_MAX_TEXTURE_SIZE` or something?

ACrazyTown:
> Can we add a new define to [FlxDefines.HelperDefines](https://github.com/HaxeFlixel/flixel/blob/dev/flixel/system/macros/FlxDefines.hx#L53), instead? like `FLX_HAS_MAX_TEXTURE_SIZE` or something?

I was thinking about that too, would make things a lot cleaner to manage.

Also, would it be better to use `Null<Int>` instead of returning -1 when it's not possible to get the max texture size?

Geokureli:
> Also, would it be better to use `Null<Int>` instead of returning -1 when it's not possible to get the max texture size?

I know -1 isn't a great solution, but I rarely think Null<Int> is much better, compared to say, enums. either way, lets stick with -1 rather than introducing a potentially breaking change

I'll test this PR soon

ACrazyTown:
> I know -1 isn't a great solution, but I rarely think Null is much better, compared to say, enums. either way, lets stick with -1 rather than introducing a potentially breaking change

Is it a breaking change if right now trying to access `FlxG.bitmap.maxTextureSize` on software rendered/non-OpenGL builds will result in a crash because the GL context isn't available

ACrazyTown:
Also, I was thinking of adding a check in `FlxGraphic` to warn users if they were making graphics that were larger than `maxTextureSize`, could I sneak that in here or should that be a seperate PR?

Geokureli:
go for it

flixel/graphics/FlxGraphic.hx
```diff
@@ -634,6 +634,15 @@ class FlxGraphic implements IFlxDestroyable
 			bitmap = value;
 			width = bitmap.width;
 			height = bitmap.height;
+
+			#if FLX_OPENGL_AVAILABLE
+			var max:Int = FlxG.bitmap.maxTextureSize;
+			if (max != -1)
+			{
+				if (width > max || height > max)
+					FlxG.log.warn('Graphic dimensions (${width}x${height}) exceed the maximum allowed size (${max}x${max}), which may cause rendering issues.');
```
Geokureli, line 643: What kind of rendering issues will happen? I'm debating whether this should be an error or not

ACrazyTown: From my understanding OpenGL won't upload the texture to the GPU at all and any attempts to render the texture will be replaced with a fallback texture (usually just black or whatever)

ACrazyTown:
Idk why CI failed

Geokureli:
The CI failure seems unrelated, I usually just rerun them

Geokureli:
The remaining CI failures seem less unrelated. I'll have to try unit tests locally

flixel/system/macros/FlxDefines.hx
```diff
@@ -254,7 +254,9 @@ class FlxDefines
 			define(FLX_TRACK_GRAPHICS);
 
 		#if (lime_opengl || lime_opengles || lime_webgl)
-		define(FLX_OPENGL_AVAILABLE);
+		// FlxG.stage.window.context.attributes.hardware is not always defined during unit tests
```
ACrazyTown, line 257: Would it be better to check if `FlxG.stage.window.context.attributes.hardware` (or whatever part of it that fails) is defined rather than outright skipping it?

Geokureli: I don't think so. If it's not defined in any other context I want it to fail loudly, and I don't think we can really unit test gl stuff right now

Geokureli:
Thanks!

PR #3283, opened 2024-11-12
Fix some documentation in AssetFrontEnd

ACrazyTown:


Geokureli:
good catch, I did a lot of copy-pasting for these

PR #3284, opened 2024-11-12
Add more functions and variables to FlxSoundGroup

Vortex2Oblivion:
Adds `pitch` `time` `length` `playing` variables and `stop` and `play` functions.
Also changes to extend `FlxBasic` too for `update` `destory` `kill` etc functions

My use case for this is that I am currently working on adding split vocal files to a funkin modding tool (who would have guessed) and I would like to use a `FlxSoundGroup` for easily pausing/playing/stopping (etc) all vocal tracks at once rather than just using an array or a bunch of different `FlxSound` variables

### DIFF
--- a/flixel/sound/FlxSoundGroup.hx
+++ b/flixel/sound/FlxSoundGroup.hx
@@ -61,7 +61,7 @@ class FlxSoundGroup extends FlxBasic
 	{
 		super.kill();
 		for (sound in sounds)
-			@:privateAccess sound.cleanup(false);
+			sound.kill();
 	}
 	
 	override public function update(elapsed:Float):Void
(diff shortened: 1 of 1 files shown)


Geokureli:
I kinda disagree with most of these. the entire purpose of sound groups is to update sounds and allow easy ways to mute or set the volume of music vs sounds. I don't recommend using them for organization in any other way. Many of these features seem more in line with `FlxSpriteGroup` than `FlxGroup`, which is a model Flixel is moving away from. I would just recommend building the features you need in a proprietary class/util

Geokureli:
oops didn't mean to close until we discussed this.

Geokureli:
> I suppose thats fair but it seems strange that there is no play/stop functions when there are pause/resume functions

I believe pause and resume are just ways to stop or start updating the sounds

PR #3293, opened 2024-11-24
call setFilters from set_filters

Geokureli:
alternative/addition to https://github.com/HaxeFlixel/flixel-demos/pull/367


Vortex2Oblivion:
i feel like this could probably be closed since setFilters doesnt exist anymore

PR #3295, opened 2024-11-28
Check if maxTextureSize >0 rather than !=-1

ACrazyTown:
In some strange circumstances `FlxG.bitmap.maxTextureSize` can return 0 or a number lower than 0. This is a bit safer way to try to make sure that the returned value is valid


Geokureli:
Good catch!

PR #3299, opened 2024-12-01
Remove OpenFl `KeyboardEvent`s in `FlxKeyManager` on destroy to avoid crashes

DetectiveBaldi:
Can be easily replicated by destroying an `FlxKeyboard` / `FlxKeyManager`.

### DIFF
--- a/flixel/input/FlxKeyManager.hx
+++ b/flixel/input/FlxKeyManager.hx
@@ -228,6 +228,9 @@ class FlxKeyManager<Key:Int, KeyList:FlxBaseKeyList> implements IFlxInputManager
 	 */
 	public function destroy():Void
 	{
+		FlxG.stage.removeEventListener(KeyboardEvent.KEY_DOWN, onKeyDown);
+		FlxG.stage.removeEventListener(KeyboardEvent.KEY_UP, onKeyUp);
+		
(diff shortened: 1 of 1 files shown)


Geokureli:
I need an actual use case for why you would need to destroy a FlxKeyboard

DetectiveBaldi:
> I need an actual use case for why you would need to manually destroy a FlxKeyboard.

i'm creating a keyboard manager for an options menu and am removing it from the input list and destroying it when exiting the state

Geokureli:
> > I need an actual use case for why you would need to manually destroy a FlxKeyboard.
> 
> i'm creating a keyboard manager for an options menu and am removing it from the input list and destroying it when exiting the state also i'm not sure how a use case for manually destroying the object is relevant here, the destroy method is meant to clean up objects and references, yet here it clearly isn't

I don't see how it's not relevant, you're making this change because you're manually destroying FlxKeyManager, and you're apparently the first person to do ever do that and see undesired behavior. I always ask for a practical repo, because tons of PRs are from people who don't actually have a use for their change and not including an example is a red flag.

That said, I asked because I read this on my phone, I'm home now, looking through the code and yeah, this makes sense

Geokureli:
Thanks!

PR #3304, opened 2024-12-05
Update FlxAssets.hx

Jarrio:
May improve completion response with projects that have a lot of assets

### DIFF
--- a/flixel/system/FlxAssets.hx
+++ b/flixel/system/FlxAssets.hx
@@ -186,8 +186,12 @@ class FlxAssets
 		#if doc_gen
 		return [];
 		#else
+		#if !display
 		final buildRefs = flixel.system.macros.FlxAssetPaths.buildFileReferences;
 		return buildRefs(directory, subDirectories, exprToRegex(include), exprToRegex(exclude), rename, listField);
+		#else
+		return [];
+		#end
 		#end
 	}
 


moxie-coder:
> May improve completion response with projects that have a lot of assets

I don't get it, display is for VSC, how does this help exactly?

moxie-coder:
well actually, nevermind it's used for code completion, but wouldn't this break projects or something?? Like I still don't get it unless I'm just stupid

Geokureli:
Won't this make AssetPaths not work at all for code completion? whats the point of AssetPaths then? plus I think yu have slowdown because you have a lot of assets, maybe we should try and cache these refs, better?

Geokureli:
> > Won't this make AssetPaths not work at all for code completion? whats the point of AssetPaths then? plus I think yu have slowdown because you have a lot of assets, maybe we should try and cache these refs, better?
> 
> yeah so I was right about my assumption then, it pretty much would just break it then

I haven't tested anything, but I am making the same assumption

moxie-coder:
> > > Won't this make AssetPaths not work at all for code completion? whats the point of AssetPaths then? plus I think yu have slowdown because you have a lot of assets, maybe we should try and cache these refs, better?
> > 
> > 
> > yeah so I was right about my assumption then, it pretty much would just break it then
> 
> I haven't tested anything, I'm just making the same assumption

fair enough though

Jarrio:
yea, you guys are correct. I had zoomed in on solving completion and failed to actually remember the point of FlxAssets lol

PR #3310, opened 2024-12-11
Deprecate screenCenter(), add gameCenter() and viewCenter()

### LINKED ISSUE #3309: remove `screenCenter`
Deprecate it for `gameCenter` which will behave exactly the same, and add `viewCenter` or something that respects camera position and zoom

richTrash21:
Closes #3309

### DIFF
--- a/flixel/FlxObject.hx
+++ b/flixel/FlxObject.hx
@@ -1256,24 +1256,59 @@ class FlxObject extends FlxBasic
 			kill();
 	}
 	#end
-
+	
 	/**
 	 * Centers this `FlxObject` on the screen, either by the x axis, y axis, or both.
 	 *
-	 * @param   axes   On what axes to center the object (e.g. `X`, `Y`, `XY`) - default is both. 
+	 * @param   axes   On what axes to center the object (e.g. `X`, `Y`, `XY`) - default is both.
 	 * @return  This FlxObject for chaining
 	 */
+	@:deprecated("screenCenter is deprecated, use gameCenter")
 	public inline function screenCenter(axes:FlxAxes = XY):FlxObject
+	{
+		return gameCenter(axes);
+	}
+	
+	/**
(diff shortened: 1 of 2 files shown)


Geokureli:
this is all good but I'm not gonna add this to the 5.9.0 release as we need to update all the documentation, demos and snippets, first

Geokureli:
Actually we need to update the unit tests, first, too.
https://github.com/HaxeFlixel/flixel/blob/dev/tests/unit/src/flixel/FlxObjectTest.hx#L344-L371

richTrash21:
> this is all good but I'm not gonna add this to the 5.9.0 release as we need to update all the documentation, demos and snippets, first

oh ok, so i guess it will be merged in 6.0.0?

Geokureli:
changed my mind, I need to edit a bunch of demos for 5.9.0 already, so this is now on 5.9.0

Geokureli:
Sorry, changed my mind again, I'm not sure I'm happy with these names, and I think there's other issues, involved here.
`gameCenter` isn't very intuitive, and I feel like `viewCenter` should use the graphical midpoint, not the hitbox midpoint, or maybe there should be two, and maybe this is also true for `gameCenter`?

I'm leaning now, towards `FlxG.center(sprite)` and `FlxG.centerHitBox(obj)` and maybe `myCamera.center(sprite)` or something. I may shelf this for a bit and think about it

NeeEoo:
I would also kinda like a sprite.centerTo(otherSprite) or different function names

to center a sprite on top of another sprite

Geokureli:
> I would also kinda like a sprite.centerTo(otherSprite) or different function names
> 
> to center a sprite on top of another sprite

I've been thinking about that too, but that might be better as a static extension, and it has the same issue of confusing collider width/height with graphic width/height

richTrash21:
redoing this rn, will make a pr with the new implementation soon

PR #3314, opened 2024-12-11
Automatically append file extension to sound asset ids

Geokureli:
Define flag FLX_DEFAULT_SOUND_EXT to allow sound asset ids to omit the extension. Useful when targeting multiple platforms that use different sound files. Can also use the flag to determine the desired sound extension, setting the flags value to "mp3" "wav" or "ogg" will use that extension. The default extension can be read via `FlxG.assets.defaultSoundExtension`

### DIFF
--- a/flixel/sound/FlxSound.hx
+++ b/flixel/sound/FlxSound.hx
@@ -337,6 +337,8 @@ class FlxSound extends FlxBasic
 	/**
 	 * One of the main setup functions for sounds, this function loads a sound from an embedded MP3.
 	 *
+	 * **Note:** If the `FLX_SOUND_ADD_EXT` flag is enabled, you may omit the file extension
+	 *
 	 * @param	EmbeddedSound	An embedded Class object representing an MP3 file.
 	 * @param	Looped			Whether or not this sound should loop endlessly.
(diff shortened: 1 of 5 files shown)


ACrazyTown:
Does it make sense to limit to only mp3, wav and ogg? On web, audio format support is limited by the browser so it supports a lot more than just those 3

Geokureli:
> Does it make sense to limit to only mp3, wav and ogg? On web, audio format support is limited by the browser so it supports a lot more than just those 3

do you have examples?

ACrazyTown:
> do you have examples?

For file formats? Off the top of my head m4a/aac, flac and ogg/opus

EliteMasterEric:
@ACrazyTown The main reason this exists is because OGG is not supported on all browsers (*glares intently at Safari*) and because MP3 is not supported on desktop (for some reason).

This is a very useful PR as is, considering I manually do something similar on multiple projects.

Geokureli:
> @ACrazyTown The main reason this exists is because OGG is not supported on all browsers (_glares intently at Safari_) and because MP3 is not supported on desktop (for some reason).
> 
> This is a very useful PR as is, considering I manually do something similar on multiple projects.

I think you misread ACrazyTowns post

PR #3322, opened 2024-12-25
Make sure Git SHA is for the Flixel repository

ACrazyTown:
`git rev-parse HEAD` (which is what FlxGitSHA.hx uses to get the current commit SHA) will get the current commit of a git repostory, EVEN in subdirectories. 
So if you have a project with a git repository and a .haxelib folder with a haxelib release of Flixel, FlxGitSHA will get the project's current git commit thinking that it's Flixel's current commit

This PR adds a `--show-toplevel` argument to the command which returns the path of the git repository and compares it with the path from `getLibraryPath()`

Merry Christmas!

### DIFF
--- a/flixel/system/macros/FlxGitSHA.hx
+++ b/flixel/system/macros/FlxGitSHA.hx
@@ -1,5 +1,6 @@
 package flixel.system.macros;
 
+import haxe.io.Path;
 import haxe.macro.Context;
 import haxe.macro.Expr;
 import sys.io.Process;
@@ -58,7 +59,7 @@ class FlxGitSHA
 			}
 		}
 
-		return result;
+		return Path.normalize(result);
 	}
 
 	public static function getGitSHA(path:String):String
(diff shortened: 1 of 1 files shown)


Geokureli (changes requested):
FlxGitSHA was deprecated, [very recently](https://github.com/HaxeFlixel/flixel/commit/64a1970940c4ea1e05c538ba28b49ea30e2ba422#diff-ee2fafeb516d96242b8c2e7be33044aab55302a58b5c16301e434b548f13060cR12). Use the new [FlxVersionMacro](https://github.com/HaxeFlixel/flixel/blob/dev/flixel/system/macros/FlxVersionMacro.hx) tool, instead

ACrazyTown:
I wasn't using it directly. It's used by FlxVersion and by extension the debugger to show the version in the top left. I'm not sure if it still happens with the new versioning tool, I'll have to check. Either way I'll close this

PR #3327, opened 2024-12-30
Simplify FlxGraphicShader

Geokureli:
Addition to https://github.com/HaxeFlixel/flixel/pull/3326

Another attempt at the other part of https://github.com/HaxeFlixel/flixel/pull/3321

Simplifies FlxGraphicsShader and removes ifs for mix calls. Let's hope that's a good thing

## To Dos
- [ ] See if thats a good thing
- [ ] Figure out how to profile shaders
- [ ] Learn more about shader attributes and flixel's batch rendering

### DIFF
--- a/flixel/graphics/tile/FlxDrawQuadsItem.hx
+++ b/flixel/graphics/tile/FlxDrawQuadsItem.hx
@@ -130,7 +130,6 @@ class FlxDrawQuadsItem extends FlxDrawBaseItem<FlxDrawQuadsItem>
 			shader.colorOffset.value = colorOffsets;
 		}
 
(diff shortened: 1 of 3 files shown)


Geokureli:
gonna shelf this until I learn why hasTransform is being set to false

Raltyro:
Why change if statements to mix?

Raltyro:
Any plan to do this the same to openfl GraphicsShader too?

Geokureli:
> Any plan to do this the same to openfl GraphicsShader too?

None of this will happen if I don't find a good way to profile these changes. As for openfl, I dunno, the gains here are pretty small, and changing it wont impact us all that much as we already overwrite theirs

PR #3336, opened 2025-01-10
Replace `FlxSound.onComplete` with `FlxSound.onFinish`

DetectiveBaldi:
Adjusts FlxSound to comply with the new changes in flixel revolving change a bunch of single callbacks to use FlxSignals instead.

### DIFF
--- a/flixel/sound/FlxSound.hx
+++ b/flixel/sound/FlxSound.hx
@@ -6,6 +6,7 @@ import flixel.math.FlxMath;
 import flixel.math.FlxPoint;
 import flixel.system.FlxAssets.FlxSoundAsset;
 import flixel.tweens.FlxTween;
+import flixel.util.FlxSignal;
 import flixel.util.FlxStringUtil;
 import openfl.events.Event;
 import openfl.events.IEventDispatcher;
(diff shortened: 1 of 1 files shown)


flixel/sound/FlxSound.hx
```diff
@@ -69,10 +70,16 @@ class FlxSound extends FlxBasic
 	 */
 	public var autoDestroy:Bool;
 	
+	/**
+	 * Signal that is dispatched on sound complete.
+	 */
+	public var onFinish:FlxSignal;
```
Geokureli, line 76: needs to be read only, lets use `final`

```suggestion
	public final onFinish:FlxSignal;
```

and add `onFinish.removeAll();` to `destroy()` to clear up memory

Geokureli (changes requested):
I also wanna make sure init() can't be called multiple times, I.E. via recycling or something. if it can then we would need to find a way to remove the listener added there, but not any added externally

DetectiveBaldi:
> I also wanna make sure init() can't be called multiple times, I.E. via recycling or something. if it can then we would need to find a way to remove the listener added there, but not any added externally

I did some more testing with recycling. I noticed that the callbacks would not be cleared if you killed a sound, and it got recycled. That is fixed with commits I'm pushing now, however I do want to note that some of the line ordering may look a bit inconsistent due to when FlxSound calls certain functions, which might be a problem

Also regarding making onFinish a final: It requires me to assign the variable in the constructor compared to in the `reset` method like all other variables in the class, might not be important, just something to considered

Geokureli:
Thanks, won't be able to test this for a short while, though

PR #3337, opened 2025-01-19
AssetFrontEnd misinformation!?!?!?!?

CrusherNotDrip:
I was setting up FlxG.assets for a HaxeFlixel project and I'm pretty sure this multi-line comment is wrong for the usage of `FLX_CUSTOM_ASSETS_DIRECTORY` define as it's supposed to be just `assets` and not `../../../assets`. When I used `../../../assets`, it just straight up gave me an error.

In the [pull request](https://github.com/HaxeFlixel/flixel/pull/2982) for FlxG.assets, it also says to use just `assets` for the `FLX_CUSTOM_ASSETS_DIRECTORY` define.


Geokureli:
Thanks! 
It used to be relative to the export path but I changed it at the last second, for various reasons, some more obvious than others. I changed the doc in a few places but forgot this one, and probably others

PR #3357, opened 2025-02-04
feat: add `luminance` variable to FlxColor

ninjamuffin99:
Luminance is the "perceived" brightness of a color. Where computers treat/calculate RGB values the same, the human eye is more sensitive to green light, so our `get_luminance` getter weighs green stronger than red or blue in a brightness calculation.

rgb -> luminance from here: https://www.w3.org/TR/AERT/#color-contrast


Geokureli:
sweet

PR #3360, opened 2025-02-08
Precise volume change callbacks

Blue2359:
I've noticed that both `onVolumeChange` and `volumeHandler` seem to return the newly set value of `FlxG.sound.volume` while `volume` itself still stays as the same previous value until the return of the setter function gets called, this PR sets `volume` to the new value first and then the callbacks get dispatched.

This lead to causing problems with my code, would like to see this get pulled.


Geokureli:
Thanks!

Issue #3162, opened 2024-06-03
Detect changes to clipRect without needing to reset the field in order to take effect


DetectiveBaldi:
currently, using `FlxRect` for fields such as `FlxSprite.clipRect` requires reassigning in order for changes to take effect. i've seen multiple people question this, and i've wondered if there's a simpler way around it.

Some ideas:
- Create a class similar to `FlxCallbackPoint` (potentially `FlxCallbackRect`) that acts similarly to the callback point class.
- Implement this functionality in `FlxRect` itself, possibly a `callback` function that gets called when setting x, y, width, height.

If we went with the latter, we could potentially remove `FlxCallbackPoint` and add similar functionality into the standard `FlxPoint` class.

Geokureli:
As I said in chat, I'm not a fan of FlxCallbackPoints for memory reasons and I'm even trying to do away with that in the long run. but I think there may be another way to detect changes to the clipRect field, i'll do some test when I have time

Geokureli:
@Sword352 had an idea of polling the current cliprect in draw() (and all other places) rather than triggering some refresh on change or detecting changes. I think this would work for actually rendering the sprite but fear it may make other affected getters more costly. Its worth consideration, nonetheless

DetectiveBaldi:
> As I said in chat, I'm not a fan of FlxCallbackPoints for memory reasons and I'm even trying to do away with that in the long run. but I think there may be another way to detect changes to the clipRect field, i'll do some test when I have time

a "bit" late but do you mind sharing the idea you had in mind? depending on what it is i could attempt implementation and appropriate testing

Geokureli:
i'll make a quick mock-up of what I have in mind, thanks for reminding me

Issue #2898, opened 2023-08-23
Add frameClipRect and worldClipRect to FlxSprite


Geokureli:
Currently, clipRect is defined in the frame's coordinate space, where `clipRect.set(0, 0, frameWidth, frameHeight)` will show the entire frame. but what if we honor clipRect in `FlxBackdrop` or `FlxTiledSprite`, should clipRect tile the clipped frame it, or clip the tiled frame? We almost had a [PR](https://github.com/HaxeFlixel/flixel/pull/2643) to clip post-scaled sprites. I think what we're getting at is there should be two clipRects on sprite, one applied to the frame and another applied when finally drawing to the destination.

Hence frameClipRect and worldClipRect.

To add this we should deprecate clipRect when adding both and, internally, make have a check to use clipRect if it is non-null and frameClipRect is null.

PR #3363, opened 2025-02-11
Sprite `clipRect` improvements

Sword352:
- Don't round `clipRect` to ensure smooth clipping no matter the scale
- Put `clipRect` back into the pool when the sprite is destroyed

Here is a sample demonstrating the visual changes of this pull request:
```haxe
package;

import flixel.FlxG;
import flixel.FlxSprite;
import flixel.FlxState;

{
	var sprite:FlxSprite;
	var speed:Float = 1 / 4;
	var total:Float = 0;

	{
		super.create();

		sprite = new FlxSprite();
		sprite.loadGraphic("flixel/images/logo/default.png");
		sprite.scale.set(12, 12);
		sprite.updateHitbox();
		sprite.screenCenter();
		add(sprite);

		sprite.clipRect = flixel.math.FlxRect.get(0, 0, sprite.frameWidth, sprite.frameHeight);
	}

	function updateClipRect(increment:Float):Void
	{
		total += increment;
		sprite.clipRect.y = total;
		sprite.clipRect = sprite.clipRect;
	}

	{
		super.update(elapsed);

		if (FlxG.keys.pressed.DOWN)
			updateClipRect(sprite.frameHeight * speed * elapsed);

		if (FlxG.keys.pressed.UP)
			updateClipRect(sprite.frameHeight * -speed * elapsed);
	}
}
```

Without this change, the sprite from the sample clips in "chuncks"
https://github.com/user-attachments/assets/36343c09-b20e-4c6a-80f0-d219225f2620

With this change, the sprite now clips much smoother
https://github.com/user-attachments/assets/304f8b45-75c4-4cfe-bd1b-3f0a01b07c20

### DIFF
--- a/flixel/FlxSprite.hx
+++ b/flixel/FlxSprite.hx
@@ -436,6 +436,8 @@ class FlxSprite extends FlxObject
 		_frame = FlxDestroyUtil.destroy(_frame);
 		_frameGraphic = FlxDestroyUtil.destroy(_frameGraphic);
 
+		clipRect = FlxDestroyUtil.put(clipRect);
+
 		shader = null;
 	}
 
@@ -1562,10 +1564,7 @@ class FlxSprite extends FlxObject
 	@:noCompletion
 	function set_clipRect(rect:FlxRect):FlxRect
 	{
-		if (rect != null)
-			clipRect = rect.round();
-		else
-			clipRect = null;
+		clipRect = rect;
 
 		if (frames != null)
 			frame = frames.frames[animation.frameIndex];


Geokureli:
1. Removing the rounding is a breaking change. Either we add a flag to toggle the behavior (which is kinda ugly), or we have people override set_clipRect:
```hx

class ClipSprite extends flixel.FlxSprite
{
    override function set_clipRect(rect:FlxRect):FlxRect
    {
        this.clipRect = rect;
        
        if (frames != null)
            frame = frames.frames[animation.frameIndex];
        return rect;
    }
}
```
2. Sprite's calling `put` on it's own cliprect is also sort of a breaking change. Other rects/points in FlxSprite (like `velocity` or `_rect`) are created internally by FlxSprite or FlxObject, and are therefore "owned" and fully "managed" by these classes, unlike clipRect which is set externally, and therefore expected to be managed externally. For instance, if you had a series of sprites sharing the same clipRect instance or use that rect elsewhere, this change would break that

Geokureli:
Also we have a few changes coming to clipRect, we want to introduce a second `sourceRect` which is rounded and refers to the section of the frame to draw, where clipRect will change to use world coordinates and not get rounded. That should happen in flixel 7.0.0, or sooner if we just make a new field for it and deprecate clipRect

https://github.com/HaxeFlixel/flixel/issues/2898
https://github.com/HaxeFlixel/flixel/issues/3162 <-tangentially related

Geokureli:
Closing this, both of these are not bug-fixes, but preferential changes, as in, changing the intended behavior of clipRect, which may cause unintended changes or even runtime crashes to existing projects

If you would like to contribute to #2898 or help come up with a plan for implementing that, it would be appreciated

PR #3365, opened 2025-02-14
Fix distanceTo regression in FlxPoint

oscarcederberg:
Fix regression introduced in 6.0.0 where a multiply in `distanceSquaredTo` was changed to an add, meaning all `distanceTo`/`dist`-functions got broken.

Added tests for all the affected functions as well.

In case a minimal setup to test the issue is needed:
```haxe
package;

import flixel.math.FlxPoint;
import flixel.FlxG;
import flixel.FlxSprite;
import flixel.FlxState;
import flixel.text.FlxText;
import flixel.util.FlxColor;

{
    var player = new FlxSprite(0, 128);
    var enemy = new FlxSprite (0, 0);
    var text = new FlxText(0, 0, 0, "", 16);

    {
        super.create();
        player.makeGraphic(32, 32, FlxColor.WHITE);
        enemy.makeGraphic(32, 32, FlxColor.RED);

        player.screenCenter(X);
        enemy.screenCenter(XY);

        add(text);
        add(player);
        add(enemy);
    }

    {
        var up:Bool = FlxG.keys.pressed.UP;
        var down:Bool = FlxG.keys.pressed.DOWN;
        var left:Bool = FlxG.keys.pressed.LEFT;
        var right:Bool = FlxG.keys.pressed.RIGHT;

        if (up && down) {
            up = down = false;
        }
        if (left && right) {
            left = right = false;
        }

        var dx = 0, dy = 0;
        dx = left ? -1 : dx;
        dx = right ? 1 : dx;
        dy = up ? -1 : dy;
        dy = down ? 1 : dy;

        var velocity = 250 * new FlxPoint(dx, dy).normalize();
        player.velocity.set(velocity.x, velocity.y);

        text.text = 'distanceTo: ${player.getPosition().distanceTo(enemy.getPosition())}
        distanceSquaredTo: ${player.getPosition().distanceSquaredTo(enemy.getPosition())}
        dist: ${player.getPosition().dist(enemy.getPosition())}
        distSquared: ${player.getPosition().distSquared(enemy.getPosition())}
        ';

        super.update(elapsed);
    }
}
```
Note how `distanceTo`/`dist` reports `NaN` or weird values depending on if the player is above or below the enemy and that `distanceSquaredTo`/`distSquared` can get negative values without the fix.

### DIFF
--- a/flixel/math/FlxPoint.hx
+++ b/flixel/math/FlxPoint.hx
@@ -752,7 +752,7 @@ import openfl.geom.Point;
 	 */
 	public overload inline extern function distanceSquaredTo(x:Float, y:Float):Float
 	{
-		return (this.x - x) * (this.x - x) + (this.y - y) + (this.y - y);
+		return (this.x - x) * (this.x - x) + (this.y - y) * (this.y - y);
 	}
 
 	/**
--- a/tests/unit/src/flixel/math/FlxPointTest.hx
+++ b/tests/unit/src/flixel/math/FlxPointTest.hx
@@ -45,6 +45,45 @@ class FlxPointTest extends FlxTest
 	}
 
 	@Test
+	function testDistanceTo():Void
+	{
+		assertDistanceTo(0, 0, 0, 0, 0);
+		assertDistanceTo(0, 0, 1, 0, 1);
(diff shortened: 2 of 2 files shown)


Geokureli:
Thanks!

PR #3368, opened 2025-02-18
Query max texture size only once instead of every time it's needed.

EliteMasterEric:

PR #3369, opened 2025-02-18
Query max texture size only once instead of every time it's needed.

EliteMasterEric:
Redo of #3368 to allow edits by maintainers.

### DIFF
--- a/flixel/system/frontEnds/BitmapFrontEnd.hx
+++ b/flixel/system/frontEnds/BitmapFrontEnd.hx
@@ -342,10 +342,14 @@ class BitmapFrontEnd
 	}
 
 	#if FLX_OPENGL_AVAILABLE
+	static var _maxTextureSize:Int = -1;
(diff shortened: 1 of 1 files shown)


Geokureli:
- Remember to uncuddle brackets, in all flixel code
- I used a static local var, which is only accessible in this method, example:
```hx
class Test
{
    static public var isReady(default, null):Bool;
    static function main()
    {
        trace(maxTextureSize); // -1
        isReady = true;
        final size = maxTextureSize;
        trace(size == maxTextureSize); // true
        trace(size == maxTextureSize); // true
        trace(size == maxTextureSize); // true
    }
    
    static public var maxTextureSize(get, never):Int;
    static public function get_maxTextureSize()
    {
        static var size = -1;
        if (size < 0 && isReady)
            size = Std.int(Math.random() * 10);
      
        return size;
    }
}```

PR #3372, opened 2025-02-20
FlxTween: Add framerate option

### LINKED ISSUE #3367: FlxTween framerate feature
I've been working on some projects using ``FlxTween`` for in-game animations, but they always update and render at the same frame rate as the game, for example, 60 FPS. This makes them look way too smooth compared to the rest of the animations.

It would be reuseful to have a ``framerate`` option for tweens so they can achieve a more cartoony look independent of the game's actual framerate, depending on the developer's preference, perhaps rendering at 24 FPS.

It could work something like this:

```haxe
var sprite = new FlxSprite();
FlxTween.tween(sprite.scale, {x: 4, y: 4}, 1, {
    framerate: 24
    // As an idea, maybe it could work similarly to FlxG.drawFramerate and FlxG.updateFramerate?
    // drawFramerate: 24
    // updateFramerate: 60
}
```

lunacynne:
implements / closes #3367

### DIFF
--- a/flixel/tweens/FlxTween.hx
+++ b/flixel/tweens/FlxTween.hx
@@ -505,6 +505,7 @@ class FlxTween implements IFlxDestroyable
 	public var active(default, set):Bool = false;
 	public var duration:Float = 0;
 	public var ease:EaseFunction;
+	public var framerate:Float;
 	public var onStart:TweenCallback;
 	public var onUpdate:TweenCallback;
 	public var onComplete:TweenCallback;
@@ -562,6 +563,7 @@ class FlxTween implements IFlxDestroyable
 		onUpdate = Options.onUpdate;
 		onComplete = Options.onComplete;
 		ease = Options.ease;
+		framerate = Options.framerate != null ? Options.framerate : 0;
 		setDelays(Options.startDelay, Options.loopDelay);
 		this.manager = manager != null ? manager : globalManager;
 	}
@@ -619,13 +621,23 @@ class FlxTween implements IFlxDestroyable
 
 	function update(elapsed:Float):Void
 	{
+		var preTick:Float = _secondsSinceStart;
 		_secondsSinceStart += elapsed;
+		var postTick:Float = _secondsSinceStart;
+
 		var delay:Float = (executions > 0) ? loopDelay : startDelay;
 		if (_secondsSinceStart < delay)
 		{
 			return;
 		}
-		scale = Math.max((_secondsSinceStart - delay), 0) / duration;
+
+		if (framerate > 0)
+		{
+			preTick = Math.fround(preTick * framerate) / framerate;
+			postTick = Math.fround(postTick * framerate) / framerate;
+		}
+
+		scale = Math.max((postTick - delay), 0) / duration;
 		if (ease != null)
 		{
 			scale = ease(scale);
@@ -647,7 +659,7 @@ class FlxTween implements IFlxDestroyable
 		}
 		else
 		{
-			if (onUpdate != null)
+			if (postTick > preTick && onUpdate != null)
 				onUpdate(this);
 		}
 	}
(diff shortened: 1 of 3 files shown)


Geokureli:
> I think you could change the capitilzation to "frameRate" in favor of FlxAnimation

good to be consistent, but it seems flixel is already inconsistent on this
https://github.com/HaxeFlixel/flixel/blob/dev/flixel/FlxG.hx#L127-L134

https://github.com/search?q=repo%3AHaxeFlixel%2Fflixel%20framerate&type=code
flash/openfl use capital R, I personally like lowercase, when we come to a decision we should rename the existing ones in a different PR

Raltyro:
fair enough, that works too

flixel/tweens/FlxTween.hx
```diff
@@ -619,13 +621,22 @@ class FlxTween implements IFlxDestroyable
 
 	function update(elapsed:Float):Void
 	{
-		_secondsSinceStart += elapsed;
+		var preTick:Float = _secondsSinceStart;
+		var postTick:Float = (_secondsSinceStart += elapsed);
```
Geokureli, line 625:
```suggestion
		var preTick:Float = _secondsSinceStart;
		_secondsSinceStart += elapsed
		var postTick:Float = _secondsSinceStart;
```

flixel/tweens/FlxTween.hx
```diff
@@ -919,6 +930,12 @@ typedef TweenOptions =
 	 */
 	@:optional var ease:EaseFunction;
 
+	/**
+	 * Optional set framerate for this tween to update at.
+	 * This also affects how often `onUpdate` is called.
+	 */
+	@:optional var framerate:Float;
```
Geokureli, line 937:
```suggestion
	@:optional var framerate:Null<Float>;
```

tests/unit/src/flixel/tweens/FlxTweenTest.hx
```diff
@@ -432,6 +432,26 @@ class FlxTweenTest extends FlxTest
...
+		// various tween framerate values
+		var testFramerates:Array<Float> = [5, 10, 10.1, 24, 29.9, 30, 30.1, 31, 59.9, 60, 100];
+		for (framerate in testFramerates)
+		{
+			var tweenUpdates:Int = 0;
+			var tween:FlxTween = FlxTween.tween(this, {value: 1000}, 2, {startDelay: 0.54321, framerate: framerate, onUpdate: function(_) tweenUpdates++});
+			step(FlxG.updateFramerate * 3); // 3 full seconds
+			Assert.areEqual(Std.int(Math.ceil(Math.min(framerate, FlxG.updateFramerate) * 2)), tweenUpdates);
+		}
+	}
```
Geokureli, line 453: it's a little difficult to understand what this is testing, and generally, we should assert against literal values rather than an equation. Also try to avoid "throwing the kitchen sink" at every test, make it very obvious what each test covers and why it's needed

Example:
```hx
function assertFpsTween(duration:Float, ?fps:Float, ?options:TweenOptions, expected:Int)
{
	var updates = 0;
	
	if (options)
		options = {};
	
	if (fps != null)
		options.fps = fps;
	
	options.onUpdate = function(_) updates++;
	options.onComplete = function(_) Assert.areEqual(expected, updates);
	FlxTween.tween({value: 0.0}, {value: 1}, 2.0, options);
}

assertFpsTween(2.0, null, null, 60);
assertFpsTween(2.0, 0, null, 60);
assertFpsTween(2.0, 5, null, 10);
assertFpsTween(2.0, 10, { startDelay: 0.5 }, 20);
assertFpsTween(2.0, 10.1, null, 21);
assertFpsTween(2.0, 29.9, null, 60);
assertFpsTween(2.0, 100, null, 60);
//... other edge cases

step(FlxG.updateFramerate * 3); // 3 full seconds
```

Geokureli (changes requested):

cyn0x8:
the number of updates appears to be `floor(max(duration * tweenfps, duration * gamefps - 1))`

Geokureli (approved):
No issues, will merge once 6.1.1 is released (hopefully soon)

PR #3378, opened 2025-02-25
Add custom text option to FlxSoundTray

slameron:
Add functionality to FlxSoundTray / `FlxG.game.soundTray.show()` to allow custom labels.
Background will change size to fit the new text or bars, whichever is wider.
This disables word wrap on the text field to allow larger widths on the text.

Adds new public var `defaultLabel` which is used when changing the volume with the plus and minus keys, and when using `FlxG.game.soundTray.show()` without providing a new label.
Example usage:

```haxe
FlxG.game.soundTray.defaultLabel = 'MASTER VOLUME';

// change sound volume settings
FlxG.game.soundTray.show(amountChanged > 0, 'SOUND VOLUME', 0.5);
```

ezgif-30d29502155da6

### DIFF
--- a/flixel/system/frontEnds/SoundFrontEnd.hx
+++ b/flixel/system/frontEnds/SoundFrontEnd.hx
@@ -372,7 +372,10 @@ class SoundFrontEnd
 		#if FLX_SOUND_TRAY
 		if (FlxG.game.soundTray != null && soundTrayEnabled)
 		{
-			FlxG.game.soundTray.show(up);
+			if (up)
+				FlxG.game.soundTray.showIncrement();
+			else
+				FlxG.game.soundTray.showDecrement();
 		}
 		#end
 	}
--- a/flixel/system/ui/FlxSoundTray.hx
+++ b/flixel/system/ui/FlxSoundTray.hx
@@ -173,7 +173,7 @@ class FlxSoundTray extends Sprite
 		y = 0;
 		visible = true;
 		active = true;
-		var globalVolume:Int = Math.round(newVolume != null ? newVolume : FlxG.sound.volume * 10);
+		var globalVolume:Int = Math.round((newVolume != null ? newVolume : FlxG.sound.volume) * 10);
 
 		if (FlxG.sound.muted)
 		{
--- a/flixel/system/ui/FlxSoundTray.hx
+++ b/flixel/system/ui/FlxSoundTray.hx
@@ -20,6 +20,7 @@ import openfl.text.GridFitType;
  * The flixel sound tray, the little volume meter that pops down sometimes.
  * Accessed via `FlxG.game.soundTray` or `FlxG.sound.soundTray`.
  */
+@:allow(flixel.system.frontEnds.SoundFrontEnd)
 class FlxSoundTray extends Sprite
 {
 	/**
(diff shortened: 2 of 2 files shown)


moxie-coder:
nice, I’ve had issues with custom text when extending the class so this helps with that

slameron:
Forgot to use the newVolume parameter, it would just default to FlxG.sound.volume.
The newVolume parameter lets you set a number separate from FlxG.sound.volume such as a custom FlxG.save.data.musicVolume to make separate sound channels to show on the sound tray.

Geokureli:
I'm confused by this change, I don't get why you would ever need to set this text from external code. In your example you're changing what field is changed by the volume tray, but wouldn't you do this by extending the volume tray and using that? if you extend the volume try, why not just change the label from there?

moxie-coder:
> I'm confused by this change, I don't get why you would ever need to set this text from external code. In your example you're changing what field is changed by the volume tray, but wouldn't you do this by extending the volume tray and using that? if you extend the volume try, why not just change the label from there?

I have tried this before in an project I worked on, example [here](https://github.com/Erizur/FNF-DaveEngine/blob/main/source/flixel/system/ui/FlxSoundTray.hx) I tried to extend it and change the text so it wouldn't be shadowing the class, but it just made the two texts overlay each other instead, no matter what I did (unless I wasn't supposed to create two texts, but then I recall it would say that "text" did not exist)

moxie-coder:
> > I have tried this before in an project I worked on, example [here](https://github.com/Erizur/FNF-DaveEngine/blob/main/source/flixel/system/ui/FlxSoundTray.hx) I tried to extend it and change the text so it wouldn't be shadowing the class, but it just made the two texts overlay each other instead, no matter what I did (unless I wasn't supposed to create two texts, but then I recall it would say that "text" did not exist)
> 
> I can understand why you would want to make it easier to do this in an extending class, but I do not understand why you would set this text from anywhere else. OP's example Is the part I'm really struggling with:
> 
> ```hx
> FlxG.game.soundTray.defaultLabel = 'MASTER VOLUME';
> 
> // change sound volume settings
> FlxG.game.soundTray.show(amountChanged > 0, 'SOUND VOLUME', 0.5);
> ```

they probably wanted an easier way to set custom labels presumably, but I'll wait for OPs response

Geokureli:
if there was a change to make bg and text private instance vars rather than local function vars. I would accept that change

This is a change that allows you to change the volume via the `show()` method, which is hacky.

it seems like you're using the sound tray to show changes to various volume controls, and I don't think I'm a fan of this. sounds like you should extend the sound tray for what you're doing

slameron:
My bad on the poor example.

Geokureli:
yeah, rather than changing the behavior of show() to fit the new need you want from it, I recommend extending sound tray and building that functionality, as for this PR, I'm only insterested in parts

slameron:
I've gone ahead and made a new PR #3384 to make the text and bg instance vars.

Geokureli:
1 sec

slameron:
> also please don't make PRs from slameron:dev, make a feature branch, eg: "slameron:sound-tray", it's like 1000x less work for me to make edits

will do in the future

Geokureli:
- Added `tray.showAnim(volume, sound, duration, label)` 
- Deprecated `show()`
- Made SoundFrontEnd use `showIncrement` and `showDecrement`
- Renamed text to `_label` and bg to `_bg`

Geokureli:
Any comments/concerns with this?

slameron:
Nope, looks good

Geokureli:
> my bad i didnt check all the code it doesnt update the text with the new label 1 sec

Oh good catch

PR #3380, opened 2025-02-25
Add PS5 / DualSense controller model support

slameron:
Currently PS5/PS4 controllers seem to work, but the `model` shows up as UNKNOWN.
This adds `'dualshock 4'` to the model check for ps4 controllers, and adds ps5 controllers to the list as well to fix the model name issue.


Geokureli:
FWIW my ps4 controllers seem to work [here](https://haxeflixel.com/demos/GamepadTest/). they show up as:

Screenshot 2025-04-28 at 4 55 08 PM

but I assume there's no downside to adding more

PR #3388, opened 2025-03-06
fix cliprect on scaled members in FlxSpriteGroup

slameron:
setting cliprect to a flxsprite group that has scaled members results in weird clipping

this video shows the clipping prior to the change, notice the profile picture becomes visible further to the left than other members of the group, as well as being clipped due to the height of the clipRect being 102 and the original size of the profile picture before being scaled was 184 - after scaling down the picture is 72 pixels. 

https://github.com/user-attachments/assets/841808ec-57e2-4302-a729-95370a747220

with the change, FlxSpriteGroup changes the height of the members' clipRect based on their scale.
after the change:

https://github.com/user-attachments/assets/474aeb47-c16b-4330-8f67-6f6e4baf4898

note that in the videos, the crown is not a part of the sprite group, so it being visible before the rest of the group is not relevant

### DIFF
--- a/flixel/group/FlxSpriteGroup.hx
+++ b/flixel/group/FlxSpriteGroup.hx
@@ -1086,7 +1086,8 @@ class FlxTypedSpriteGroup<T:FlxSprite> extends FlxSprite
 		if (ClipRect == null)
 			Sprite.clipRect = null;
 		else
-			Sprite.clipRect = FlxRect.get(ClipRect.x - Sprite.x + x, ClipRect.y - Sprite.y + y, ClipRect.width, ClipRect.height);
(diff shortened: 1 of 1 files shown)


Geokureli:
I should mention that there is some previous discussions on cliprect + group + scaling
https://github.com/HaxeFlixel/flixel/pull/2643
https://github.com/HaxeFlixel/flixel/issues/2301
https://github.com/HaxeFlixel/flixel/issues/2898

I assume this is making a similar change to what was proposed in the first link, making clipRect refer to a rect in the world rather than a rect in the source. If so, thats a breaking change and I think we should go towards  the solution proposed in 2898

Issue #2710, opened 2022-12-31
large FlxBitmapText draws slowly even when only a small portion is on screen


Geokureli:
note to self, see if you can cull chars not on screen for better performance

PR #3395, opened 2025-03-11
Improve FlxBitmapText rendering when using renderTile with clipRect

Geokureli:
Fixes: https://github.com/HaxeFlixel/flixel/issues/2710

Less drastic version of https://github.com/HaxeFlixel/flixel/pull/3394

`FlxSprite`: fix issue where setting cliprect completely off the frame prevents it from working later
`FlxBitmapText`: Improve performance, memory, and allow more control via extension
`FlxRect`: Add `contains` method
`FlxFrame`: Add `overlaps`, `contains` and `isContained`  methods
`FlxFrame`: Make constructor public
`FlxColorTransformUtil`: Add `reset`, `set`, multiple `scaleMultipliers` overloads, overloaded `setMultipliers` and made `alpha` args  optional in most functions
`FlxColor`: Made `alpha` args  optional in most functions

### DIFF
--- a/tests/unit/src/flixel/util/FlxColorTransformUtilTest.hx
+++ b/tests/unit/src/flixel/util/FlxColorTransformUtilTest.hx
@@ -60,12 +60,6 @@ class FlxColorTransformUtilTest extends FlxTest
 		assertCTEquals(0.4, 0.5, 0.6, 1.5, 0, 0, 0, 0);
 		color.scaleMultipliers(0.4, 0.5, 0.6);
 		assertCTEquals(0.16, 0.25, 0.36, 1.5, 0, 0, 0, 0);
-		color.scaleMultipliers(0.4, 0.5);
-		assertCTEquals(0.064, 0.125, 0.36, 1.5, 0, 0, 0, 0);
-		color.scaleMultipliers(0.4);
-		assertCTEquals(0.0256, 0.125, 0.36, 1.5, 0, 0, 0, 0);
-		color.scaleMultipliers();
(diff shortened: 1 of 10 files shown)


moxie-coder:
wouldn’t thus be a duplicate PR??

moxie-coder:
oh never mind I see it

DetectiveBaldi:
I noticed in the other pr about this topic it mentions how this could contribute to per-character text effects. Is there work that still needs to be done there? Or are there any examples already available

Geokureli:
> I noticed in the other pr about this topic it mentions how this could contribute to per-character text effects. Is there work that still needs to be done there? Or are there any examples already available

I'm planning on making examples using #3394 as it's not as feasible using this PR. Even then, since this is only usable with renderTile, it may end up being an unofficial external tool, rather than a flixel-supported feature of flixel/addons

tests/unit/src/flixel/util/FlxColorTransformUtilTest.hx
```diff
@@ -0,0 +1,225 @@
...
+		color.set(0x0).set(c);
+		assertCTEquals(c.redFloat, c.greenFloat, c.blueFloat, c.alphaFloat, 0, 0, 0, 0);
+	}
+	
+	@Test
+	function testScaleMultipliersRGBA()
+	{
+		color.scaleMultipliers(0.4, 0.5, 0.6, 1.5);
+		assertCTEquals(0.4, 0.5, 0.6, 1.5, 0, 0, 0, 0);
+		color.scaleMultipliers(0.4, 0.5, 0.6);
```
Geokureli, line 61: i need to test  more optional args

Issue #3334, opened 2025-01-08
Problem with rendering to proper sprite coordinates


kfjustis:
- **Haxe version:** 4.3.6
- **Flixel version:** 5.9.0
- **OpenFL version:** 9.4.1
- **Lime version:** 8.2.2
- **Affected targets:** HL and Native (macOS arm and linux, couldn't test windows)

________________________________

**Code snippet reproducing the issue**:

```haxe
package;

import flixel.FlxG;
import flixel.FlxSprite;
import flixel.FlxState;
import flixel.util.FlxColor;

class PlayState extends FlxState
{
	override public function create()
	{
		super.create();

		bgColor = 0xFFE40066;

		var sprite = new FlxSprite();
		sprite.makeGraphic(200, 200, 0xFF345995);
		sprite.x = FlxG.width - sprite.width;
		sprite.y = 200;
		add(sprite);
	}
	override public function update(elapsed:Float)
	{
		super.update(elapsed);
	}
}
```
Minimum project if you want to clone and run quickly: https://github.com/kfjustis/SpriteOffsetBug
________________________________

**Observed behavior:** The right side of the sprite rect is not exactly flush with the edge of the play area after maximizing the game window. By maximize, I mean both resizing the window and true fullscreen.

After launching the game via `lime test hl`:
<img width="752" alt="image" src="https://github.com/user-attachments/assets/10ee1be2-84d7-4d5d-88b0-6c4c1396d11b" />

After maximizing the window via double-clicking the title bar:
<img width="1624" alt="image" src="https://github.com/user-attachments/assets/1fe65fdc-27e9-4d5a-a9cc-db2502756925" />

Cropped and zoomed in to show the extremely tiny offset issue (some pink comes through between the blue sprite and the black bar on the right):
<img width="522" alt="image" src="https://github.com/user-attachments/assets/7d55f3de-33bc-4201-9851-47ccc128eb5a" />

**Expected behavior:** There should be no gap between the right side of the sprite and the black bar.

kfjustis:
Totally possible that this if from a lower-level library like OpenFL, but I found this during my prototyping for a Haxe Winter Jam project and haven't seen anyone else mention a problem like this one. My HashLink version is a little outdated since I could only get the version that comes with lime to work on my Mac, but it happens on my Linux desktop and through the native target as well, so I think I can rule that out. Any info would be great!

kfjustis:
[...]

PR #3397, opened 2025-03-12
Small FlxBGSprite size fix

MaybeMaru:
Due to some issue with openfl size render rounding using ``FlxBGSprite`` can create a small one pixel gap when fullscreened.
| Before| After|
| -------- | ------- |
| image | image|

Heres my test code:
```haxe
{
	{
		super.create();
		FlxG.camera.bgColor = FlxColor.GREEN;
		var bgSprite = new FlxBGSprite();
		bgSprite.color = FlxColor.RED;
		add(bgSprite);
	}
}

```

### DIFF
--- a/flixel/system/FlxBGSprite.hx
+++ b/flixel/system/FlxBGSprite.hx
@@ -29,7 +29,7 @@ class FlxBGSprite extends FlxSprite
 			}
 
 			_matrix.identity();
-			_matrix.scale(camera.viewWidth, camera.viewHeight);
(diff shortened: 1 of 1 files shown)


ACrazyTown:
I wonder if the underlying issue is related to https://github.com/HaxeFlixel/flixel/issues/3334

Geokureli:
> I wonder if the underlying issue is related to #3334

Yeah seems like the same symptom of the same issue, I'm not sure there's anything we can do on our end other than making it a little bigger or using a different bg color

PR #3400, opened 2025-03-16
Fix `FlxAtlas` not persisting between states with persist flag set to true

Sword352:
I think an oversight was made in the check so the atlas would get cleared from memory even if you had `persist` set to true.


Geokureli:
Definitely an oversight, good find

PR #3402, opened 2025-03-25
Don't clear the asset key when destroying an FlxGraphic

EliteMasterEric:
I've found that I sometimes have to debug issues with accidentally disposing of FlxGraphics I don't mean to. In these cases, I generally try to diagnose the issue by setting a breakpoint and inspecting the problematic graphic, but since the key gets set to null, I can't see which graphic is causing problems.


Geokureli:
Do we really ever need to null out strings?

PR #3406, opened 2025-03-30
[FlxSound] Remove Flash conditional in loadByteArray()

TheGaloXx:
When trying to find a way to load a sound from bytes, I came across `loadByteArray()` in `FlxSound`. The function is currently inside a `flash11` conditional. However, I tested it on both Windows and HashLink targets and it worked just fine.

I'd say that this conditional should be removed (unless it causes other problems ofc) since it sounds pretty useful and the only way for me to find it was to scroll through `FlxSound`.

### DIFF
--- a/flixel/sound/FlxSound.hx
+++ b/flixel/sound/FlxSound.hx
@@ -13,9 +13,7 @@ import openfl.media.Sound;
 import openfl.media.SoundChannel;
 import openfl.media.SoundTransform;
 import openfl.net.URLRequest;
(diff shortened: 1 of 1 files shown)


Geokureli:
seems like this #if is an outdated relic, have you successfully used this method on a non-flash target?

TheGaloXx:
> seems like this #if is an outdated relic, have you successfully used this method on a non-flash target?

I've only tried it on Windows and HashLink so far

PR #3413, opened 2025-04-08
Improve `FlxSprite` color transform handling, deprecate `FlxColor.to24Bit()`

richTrash21:
Changes:
- `FlxSprite.setColorTransform()` now bypasses `color` and `alpha` setters.
- `FlxSprite.useColorTransform` now accounts for alpha offset.
- `FlxColor.to24Bit()` is now deprecated in favor of `FlxColor.rgb` field.

### DIFF
--- a/flixel/FlxSprite.hx
+++ b/flixel/FlxSprite.hx
@@ -255,14 +255,14 @@ class FlxSprite extends FlxObject
 	 * Blending modes, just like Photoshop or whatever, e.g. "multiply", "screen", etc.
 	 */
 	public var blend(default, set):BlendMode;
-
+	
 	/**
 	 * Tints the whole sprite to a color (`0xRRGGBB` format) - similar to OpenGL vertex colors. You can use
 	 * `0xAARRGGBB` colors, but the alpha value will simply be ignored. To change the opacity use `alpha`.
 	 * @see https://snippets.haxeflixel.com/sprites/color/
 	 */
-	public var color(default, set):FlxColor = 0xffffff;
-
+	public var color(default, set):FlxColor = FlxColor.WHITE;
+	
 	public var colorTransform(default, null):ColorTransform;
 
 	/**
@@ -1008,7 +1008,7 @@ class FlxSprite extends FlxObject
 			dirty = true;
 		return positions;
 	}
-
+	
 	/**
 	 * Sets the sprite's color transformation with control over color offsets.
 	 * With `FlxG.renderTile`, offsets are only supported on OpenFL Next version 3.6.0 or higher.
@@ -1025,13 +1025,13 @@ class FlxSprite extends FlxObject
 	public function setColorTransform(redMultiplier = 1.0, greenMultiplier = 1.0, blueMultiplier = 1.0, alphaMultiplier = 1.0,
 			redOffset = 0.0, greenOffset = 0.0, blueOffset = 0.0, alphaOffset = 0.0):Void
 	{
-		color = FlxColor.fromRGBFloat(redMultiplier, greenMultiplier, blueMultiplier).rgb;
-		alpha = alphaMultiplier;
+		@:bypassAccessor color = FlxColor.fromRGBFloat(redMultiplier, greenMultiplier, blueMultiplier, 0);
+		@:bypassAccessor alpha = alphaMultiplier;
 
 		colorTransform.setMultipliers(redMultiplier, greenMultiplier, blueMultiplier, alphaMultiplier);
 		colorTransform.setOffsets(redOffset, greenOffset, blueOffset, alphaOffset);
 
-		useColorTransform = alpha != 1 || color != 0xffffff || colorTransform.hasRGBOffsets();
+		useColorTransform = alpha != 1 || color != 0xffffff || colorTransform.hasRGBAOffsets();
 		dirty = true;
 	}
 	
@@ -1040,15 +1040,16 @@ class FlxSprite extends FlxObject
 		if (colorTransform == null)
 			return;
 
-		useColorTransform = alpha != 1 || color != 0xffffff;
+		useColorTransform = alpha != 1 || color.rgb != 0xffffff;
 		if (useColorTransform)
 			colorTransform.setMultipliers(color.redFloat, color.greenFloat, color.blueFloat, alpha);
 		else
 			colorTransform.setMultipliers(1, 1, 1, 1);
 
+		useColorTransform = useColorTransform || colorTransform.hasRGBAOffsets();
 		dirty = true;
 	}
-
+	
 	/**
 	 * Checks to see if a point in 2D world space overlaps this `FlxSprite` object's
 	 * current displayed pixels. This check is ALWAYS made in screen space, and
--- a/flixel/FlxSprite.hx
+++ b/flixel/FlxSprite.hx
@@ -263,7 +263,10 @@ class FlxSprite extends FlxObject
 	 */
 	public var color(default, set):FlxColor = FlxColor.WHITE;
 	
-	public var colorTransform(default, null):ColorTransform;
+	/**
+	 * The color effects of this sprite, changes to `color` or `alplha` will be reflected here
+	 */
+	public final colorTransform = new ColorTransform();
 
 	/**
 	 * Whether or not to use a `ColorTransform` set via `setColorTransform()`.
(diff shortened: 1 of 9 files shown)


Geokureli:
In general, splitting each change into a separate PR is preferred. Multi-issue PRs tend to take much longer and are less likely to get approved, but let's see what we got here

> * `FlxSprite.setColorTransform()` now bypasses `color` and `alpha` setters.

What was the issue, previously?

> * `FlxSprite.useColorTransform` now accounts for alpha offset.

Can you elaborate on this?

> * `FlxColor.to24Bit()` is now deprecated in favor of `FlxColor.rgb` field.

cool

flixel/FlxSprite.hx
```diff
@@ -1506,11 +1507,12 @@ class FlxSprite extends FlxObject
 	@:noCompletion
 	function set_alpha(Alpha:Float):Float
```
Geokureli, line 1508: side note, I really hate these setters, and may remove them in the future

richTrash21:
> > * `FlxSprite.setColorTransform()` now bypasses `color` and `alpha` setters.
> 
> What was the issue, previously?

Calling `setColorTransform()` would set `color` and `alpha` via setters. If values of these fields were changed, they would call `updateColorTransform()`, which is pointless since `setColorTransform()` does the same thing. Adding `@:bypassAccessor` metadata makes so fields are set directly, reducing unnececeary `updateColorTransform()` calls.

> 
> > * `FlxSprite.useColorTransform` now accounts for alpha offset.
> 
> Can you elaborate on this?

After calling `setColorTransform()`, `useColorTransform` represented if the sprite had any transparency, coloring or RGB offset, but ignored `alphaOffset`. This was fixed by using RGBA offset check, instead of RGB.
There was a similar problem with `updateColorTransform()`, after calling which, `useColorTransform` ignored any offset.

Note: while testing `useColorTransform` i also discovered a bug where using `useFramePixels` would result in incorrect color transformation on animated sprites. Writing an issue about that.

Geokureli:
I don't see any evidence that alphaOffset actually works, enough to support checking for it

flixel/FlxSprite.hx
```diff
@@ -1025,13 +1025,13 @@ class FlxSprite extends FlxObject
...
-		color = FlxColor.fromRGBFloat(redMultiplier, greenMultiplier, blueMultiplier).to24Bit();
-		alpha = alphaMultiplier;
+		@:bypassAccessor color = FlxColor.fromRGBFloat(redMultiplier, greenMultiplier, blueMultiplier, 0);
+		@:bypassAccessor alpha = alphaMultiplier;
 
 		colorTransform.setMultipliers(redMultiplier, greenMultiplier, blueMultiplier, alphaMultiplier);
 		colorTransform.setOffsets(redOffset, greenOffset, blueOffset, alphaOffset);
 
-		useColorTransform = alpha != 1 || color != 0xffffff || colorTransform.hasRGBOffsets();
+		useColorTransform = alpha != 1 || color != 0xffffff || colorTransform.hasRGBAOffsets();
```
Geokureli, line 1034: should be color.rgb

richTrash21:
> I don't see any evidence that alphaOffset actually works, enough to support checking for it

Here's a test project for that.

```haxe
package;

import flixel.FlxG;
import flixel.FlxSprite;
import flixel.math.FlxMath;
import flixel.text.FlxText;

{
	var sprite:FlxSprite;
	var alphaText:FlxText;
	var alphaOffsetText:FlxText;

	var alpha = 1.0;
	var alphaOffset = 0.0;

	{
		super.create();

		// FlxG.camera.bgColor = 0xFF999999;

		sprite = new FlxSprite(60, 60).makeGraphic(200, 200);
		add(sprite);

		alphaText = new FlxText(60, 280, 0, "", 16);
		add(alphaText);

		alphaOffsetText = new FlxText(60, 320, 0, "", 16);
		add(alphaOffsetText);
	}

	{
		super.update(elapsed);

		final mult = 3 * (FlxG.keys.pressed.SHIFT ? 10 : 1);

		if (FlxG.keys.pressed.Q)
			setAlpha(alpha - 0.01 * elapsed * mult);
		if (FlxG.keys.pressed.E)
			setAlpha(alpha + 0.01 * elapsed * mult);
		if (FlxG.keys.justPressed.W)
			setAlpha(1);

		if (FlxG.keys.pressed.A)
			setAlphaOffset(alphaOffset - 1 * elapsed * mult);
		if (FlxG.keys.pressed.D)
			setAlphaOffset(alphaOffset + 1 * elapsed * mult);
		if (FlxG.keys.justPressed.S)
			setAlphaOffset(0);

		sprite.setColorTransform(
			sprite.colorTransform.redMultiplier, sprite.colorTransform.greenMultiplier, sprite.colorTransform.blueMultiplier, alpha,
			sprite.colorTransform.redOffset,     sprite.colorTransform.greenOffset,     sprite.colorTransform.blueOffset,     alphaOffset
		);

		alphaText.text = "alpha: " + FlxMath.roundDecimal(alpha, 3);
		alphaOffsetText.text = "alphaOffset: " + FlxMath.roundDecimal(alphaOffset, 3);
	}

	function setAlpha(value:Float)
		alpha = FlxMath.bound(value, 0, 1);

	function setAlphaOffset(value:Float)
		alphaOffset = FlxMath.bound(value, -255, 255);
}
```

Note: only negative `alphaOffset` values have visible effect, but doc states that positive work too? I also tested openfl sprite to compare behaviour (not a part of this test project) and it's the same, so i guess this is intended.

> I also don't understand why useColorTransform is needed, in most cases

`useColorTransform` is used to determine if `updateFramePixels()` should color `framePixels`. This also causes #3414, so we may just deprecate it.

Geokureli:
updateFramePixels and useFramePixels only effect renderBlit targets, btw, are you using these in a project

richTrash21:
> updateFramePixels and useFramePixels only effect renderBlit targets

useFramePixels actually works perfectly on renderTile, have no idea what you ment by that.

> btw, are you using these in a project

Yes, but now as an experiment, disabled caching is a bit concerning

Geokureli:
> > updateFramePixels and useFramePixels only effect renderBlit targets
> 
> useFramePixels actually works perfectly on renderTile, have no idea what you ment by that.

Yeah sorry, this works on renderTile, but it seems like the purpose of these is to enable blitting functionality when flixel is in renderTile mode

> > btw, are you using these in a project
> 
> Yes, but now as an experiment, disabled caching is a bit concerning

What I mean is, are you only using these in academic experiments, or do you have a practical use for this?

I recently found out that updateFramePixels was [used in getPixelAt()](https://github.com/HaxeFlixel/flixel/pull/3419/files#diff-76ff6e513fb29977162eae873bfd4df47ff2666cdacd89548ef4b4ab2c3598fcL1090) when it didn't need to and I'm wondering if it should just be avoided in renderTile, altogether

flixel/FlxSprite.hx
```diff
@@ -1040,15 +1040,16 @@ class FlxSprite extends FlxObject
...
 			return;
 
-		useColorTransform = alpha != 1 || color != 0xffffff;
+		useColorTransform = alpha != 1 || color.rgb != 0xffffff;
 		if (useColorTransform)
 			colorTransform.setMultipliers(color.redFloat, color.greenFloat, color.blueFloat, alpha);
 		else
 			colorTransform.setMultipliers(1, 1, 1, 1);
 
+		useColorTransform = useColorTransform || colorTransform.hasRGBAOffsets();
```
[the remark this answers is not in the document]
Geokureli: I still don't understand a case where this will behave differently than:
```hx
useColorTransform = alpha != 1 || color.rgb != 0xffffff || colorTransform.hasRGBAOffsets();
```

furthermore why can't this just be:
```hx
useColorTransform = alpha != 1 || color.rgb != 0xffffff || colorTransform.hasRGBAOffsets();
colorTransform.setMultipliers(color.redFloat, color.greenFloat, color.blueFloat, alpha);
```

> Note: on second thought, maybe it will be better to just set color offset to 0 on updateColorTransform()?

I agree, but this is could cause unintended side effects to existing projects

Geokureli:
1. I noticed that bypassing the setters affected FlxText which overrides `set_color` and `updateColorTransform`
2. ~~I made colorTransform final~~ I removed the part where colorTransform is destroyed, openfl rectangles aren't pooled so nulling them in destroy won't make a significant different. eventually this field will be final, when flixel-addons is updated
3. I deprecated `useColorTransform`, and added a simple `hasColorTransform()`
4. I added the code from https://github.com/HaxeFlixel/flixel/pull/3416, just so I can test these changes together

let me know if you see any issues, here. also in 3416 you mention:
> Running this code will result in this: applied color appear darker or as a mix of previous and current colors.

I do not see that, and I'm not sure why I would, am I missing something?

Geokureli:
Merging now for 6.1.0, Thanks!

PR #3417, opened 2025-04-13
Improve thrown error message when attempting to draw a destroyed FlxGraphic

VioletSnowLeopard:
The error message that is thrown when attempting to draw a destroyed graphic currently is vague. It doesn't say what it tried to access, but was destroyed:

Screenshot 2025-04-13 075124

It is harder to troubleshoot the issue if you don't know which graphic is getting destroyed when its not supposed to. Thanks to [this PR](https://github.com/HaxeFlixel/flixel/pull/3402), the key of a FlxGraphic isn't set to null when the graphic is destroyed. This means that it can be used in the thrown error message:

Screenshot 2025-04-12 230131

This aids in the process of figuring out which graphic was destroyed so you can start working on finding where it got destroyed.

### DIFF
--- a/flixel/FlxCamera.hx
+++ b/flixel/FlxCamera.hx
@@ -608,7 +608,7 @@ class FlxCamera extends FlxBasic
 		
 		// TODO: catch this error when the dev actually messes up, not in the draw phase
 		if (graphic.isDestroyed)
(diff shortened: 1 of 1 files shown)


Geokureli:
Thanks!

PR #3419, opened 2025-04-17
Add clipToWorldBounds and converters between: frame, world, view and game

### LINKED ISSUE #2301: Oddities when using clipRect + scale.

Geokureli:
Closes #2301 and #2898

- `FlxCamera`:  Adds `worldToViewPosition`, `viewToWorldPosition`, `gameToViewPosition`, `viewToGamePosition` as well as x and y specific methods for each
- `FlxObject`: Adds `getViewPosition`, `getViewX` and `getViewY`
- `FlxSprite`:
  - Helpers methods for `clipRect`: `clipToWorldRect`, `clipToViewRect`, `clipToWorldBounds` and `clipToViewBounds`
  - Converters: `worldToFramePosition`, worldToFramePositionSimple  and `viewToFramePosition`
- `FlxFrame`: Adds `getPixelAt` and `toSourcePosition`
- `FlxRect`: Adds the `abs` method
- `IFlxPooled` and `FlxDestroyUtil`: Adds `putWeak` methods

## To Do
- [x] Retry CI when stuff works
- [x] Find out why this fails on hardware="false"
- [x] Use different x/y values in tests. this would have caught the previous error
- [x] Make getPixelAt this work with sprite.color and colorTransform


rodney528 (commented):
The only thing I have to say is... does overload really need to be used that much?

Geokureli:
> The only thing I have to say is... does overload really need to be used that much?

Is there a disadvantage to it?

rodney528:
Idk much about overload besides being able to use the same function name multiple times. Just feels like too many versions of the same function.

Geokureli:
> Idk much about overload besides being able to use the same function name multiple times. Just feels like too many versions of the same function.

Why is it "too many"? What is the downside of more vs less?

rodney528:
> > Idk much about overload besides being able to use the same function name multiple times. Just feels like too many versions of the same function.
> 
> Why is it "too many"? What is the downside of more vs less?

Just feels weird to me ig.

Geokureli:
> Just feels weird to me ig.

Well, let me know when you find a reason, or if you have a suggestion

PR #3421, opened 2025-04-23
Add muted property to FlxSoundGroup

ACrazyTown:
A convenience property like `FlxG.sound.muted` that allows us to mute sounds on a group basis.

In my use case I wanted to allow muting sound effects and music seperately and this seems like the simplest option to do so as it will retain the volume of the group

### DIFF
--- a/flixel/sound/FlxSound.hx
+++ b/flixel/sound/FlxSound.hx
@@ -605,7 +605,7 @@ class FlxSound extends FlxBasic
 		final volume = (group != null ? group.volume : 1.0) * _volume * _volumeAdjust;
 		
 		#if FLX_SOUND_SYSTEM
-		if (FlxG.sound.muted)
+		if (FlxG.sound.muted || (group != null && group.muted))
 			return 0.0;
 		
 		return FlxG.sound.applySoundCurve(FlxG.sound.volume * volume);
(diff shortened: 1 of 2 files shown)


flixel/sound/FlxSound.hx
```diff
@@ -605,7 +605,7 @@ class FlxSound extends FlxBasic
 		final volume = (group != null ? group.volume : 1.0) * _volume * _volumeAdjust;
 		
 		#if FLX_SOUND_SYSTEM
-		if (FlxG.sound.muted)
+		if (FlxG.sound.muted || (group != null && group.muted))
```
Geokureli, line 608: the `#if FLX_SOUND_SYSTEM` is to check is FlxG.sound exists, groups and soudns can exist without that, so this should not be checked here. group volume is check in line 605. also we should consider having group.getVolume() that checks muted and returns 0

Geokureli (changes requested):

Geokureli (approved):

Geokureli (changes requested):
Just realized that `volume` has a setter which calls updateTransform on every sound. This should behave similarly

ACrazyTown:
> Just realized that `volume` has a setter which calls updateTransform on every sound. This should behave similarly

Just to clarify, should I make a setter for muted that does the same thing as volume's setter?

Geokureli:
> > Just realized that `volume` has a setter which calls updateTransform on every sound. This should behave similarly
> 
> Just to clarify, should I make a setter for muted that does the same thing as volume's setter?

yes, please

Geokureli (approved):

Geokureli:
Thanks!

PR #3428, opened 2025-04-30
Implement `FlxSignal.cancel`/`canceled`

DetectiveBaldi:
FlxSignals have lacked the ability to cancel propagation for a long time now, even though it is found in related OpenFl and Lime classes. I have recently had the need to cancel FlxSignal dispatch during propagation, hence this pull request

Also reorganized the order of some class fields to fit with the interface

May need more advanced testing (not sure how unit tests etc will work out) but this simple test cases compiles and functions as expected ("Progress Three" is never traced):

```haxe
package;

import flixel.FlxState;

import flixel.util.FlxSignal;

{
	public var processes:FlxSignal;

	{
		super.create();

		processes = new FlxSignal();

		processes.add(processOne);

		processes.add(processTwo);

		processes.add(processThree);

		processes.dispatch();
	}

	{
		super.update(elapsed);
	}

	public function processOne():Void
	{
		trace("Process One");
	}

	public function processTwo():Void
	{
		trace("Process Two");

		processes.cancel();
	}

	public function processThree():Void
	{
		trace("Progress Three");
	}
}
```

This is my first time interacting with abstracts, so I might be missing something basic

### DIFF
--- a/flixel/util/FlxSignal.hx
+++ b/flixel/util/FlxSignal.hx
@@ -129,9 +129,9 @@ private class FlxBaseSignal<T> implements IFlxSignal<T>
 
 	public function new()
 	{
+		canceled = false;
 		handlers = [];
 		pendingRemove = [];
-		canceled = false;
 	}
 
 	public function add(listener:T)
@@ -322,10 +322,7 @@ private class Macro
 			for (handler in handlers)
 			{
 				if (canceled)
-				{
-					canceled = false;
 					break;
-				}
 
 				handler.listener($a{exprs});
 
@@ -333,6 +330,7 @@ private class Macro
 					removeHandler(handler);
 			}
 
+			canceled = false;
 			processingListeners = false;
 
 			for (handler in pendingRemove)
--- a/flixel/util/FlxSignal.hx
+++ b/flixel/util/FlxSignal.hx
@@ -12,6 +12,8 @@ abstract FlxTypedSignal<T>(IFlxSignal<T>)
 {
 	public var dispatch(get, never):T;
 
+	public var canceled(get, set):Bool;
+
 	public function new();
 
 	public inline function add(listener:T):Void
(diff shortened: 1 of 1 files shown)


Geokureli:
FlxSignals are not openfl events, they are a different, simpler and IMO a better system

All signal listeners should be agnostic to one another, they do not depend on each other and they do not affect each other, this is a good thing!

If you have a case where listener 3 depends on what listener 2 is doing, that's something your system should handle when adding those listeners because clearly they are part of some connected system

PR Denied, this is needless complexity that is more likely to cause debugging issues and confusion. If you need this I urge you to reconsider what it is you're doing, even if you implement this yourself in a personal project I think it's just generally a terrible idea

Geokureli:
Could you share what problem you were trying to solve with this, I'd be down to discuss a better alternative

DetectiveBaldi:
I have a state with multiple interactible buttons, with said buttons containing an FlxSignal that is called when they're clicked. I add two listeners to these signals. The first signal in short runs a bunch of checks to decide whether to continue advancing in the menu, with the second signal fades music and switches the state. In some cases the state switch needs to be canceled, therefore I need to suspend the second signal. However, these buttons can be interacted with multiple times, so I can't simply just remove the second listener if the check fails the first time.

Also, how could signal canceling cause debugging issues? I feel like the usage is extremely straightforward. The implementation is also extremely simple, nothing very complex here.

Geokureli:
> I have a state with multiple interactible buttons, with said buttons containing an FlxSignal that is called when they're clicked. I add two listeners to these signals. The first signal in short runs a bunch of checks to decide whether to continue advancing in the menu, with the second signal fades music and switches the state. In some cases the state switch needs to be canceled, therefore I need to suspend the second signal.

In this case I would have two signals, one when the button is pressed and another for when the menu decides it can perform the action associated with the button. For instance `onBackPressed` and `onNavigateBack`

> Also, how could signal canceling cause debugging issues? I feel like the usage is extremely straightforward. The implementation is also extremely simple, nothing very complex here.

Imagine adding a listener to `FlxG.update` (or whatever signal), and it simply doesn't seem to work every time, and you're not sure why or exactly when, so you step in the code to try and reproduce it, following each listener, across various dispatches, only to find out some tool you're using hijacked the signal in the exact way you are describing.

This is also just a core principal of signals, they are there to let you know something happened, knowing that nothing else will prevent that notification. if you want something to know when a button is pressed, listen for that button, if you want to know when the menu decides to navigate to another page, listen for that. The moment you have to listeners that depend on each other, you've made a terrible mistake

Issue #3430, opened 2025-05-06
CollisionFrontEnd


Geokureli:
Spinoff from https://github.com/HaxeFlixel/flixel/pull/3418

[FlxObject.separate](https://github.com/HaxeFlixel/flixel/blob/dev/flixel/FlxObject.hx#L160) is weirdly laid out and could use a redo.

Operations like computeOverlapX should just compute the overlap, and not set touching flags. We can refactor this system completely and avoid breaking backwards compatibility if we move all of this to `FlxG.collision`.

I think `FlxG.collision.separate` should compute the overlap of each axis and separate on the smaller axis before the other

MondayHopscotch:
I think point 3 will have other implications on backwards compatibility with how `FlxTilemaps` currently work. The current implementation of doing the axes in a fixed order results in tilemaps being very forgiving as far as tiles' `allowCollisions` is concerned.

Namely, having the tiles all set to `allowCollisions = ALL` works fine in this scenario (moving character in blue, tilemap in orange):

![Image](https://github.com/user-attachments/assets/9a523853-9033-4bf3-a29f-9e1e8de82ee1)

As the player approaches the seams between tiles, it doesn't matter which axis is resolved first as long as it's consistent.

With the 'smaller axis first' resolution, having proper `allowCollisions` on the tiles becomes critical as this scenario:

![Image](https://github.com/user-attachments/assets/f5d03942-53d7-464f-9b1e-529dc1efaba3)

The outcome depends on which order the tile collisions are resolved. if the left ground tile is resolved first, the player snags (as the x-overlap is the smallest axis). If the right ground tile is resolved first, they do not snag.

I only bring this up as `FlxTilemap` doesn't try to do any smart collisions adjustments based on tile adjacency (at least not that I have spotted). It feels like the current implementation is better in the AABB world that flixel currently lives in due to simplicity and compatibility. I'm not fully against moving toward more of a SAT approach, but we may want to just collect other things that will be impacted by moving to that style of resolution.

----

I know autoTile + proper tile properties makes this a non-issue (when used properly), but I would assume most people aren't using these features. I know I historically haven't.

Geokureli:
> I know autoTile + proper tile properties makes this a non-issue (when used properly), but I would assume most people aren't using these features. I know I historically haven't.

Agreed, it's pretty much assumed that nearly all tilemaps will have all solid tiles' `allowCollisions` of `ANY`, any system that cannot handle this is not fit for release.

[...]

PR #3418, opened 2025-04-14
Piecewise axis separate

MondayHopscotch:
The existing separation logic does not correctly detect/resolve the following scenario:

1. Object A's starting position (red dotted box) does not overlap with Object B (black box) on either the X or Y axis
2. Object A moves to now be overlapping on _BOTH_ axes of Object B (shown by red solid box)
image

This is caused because the "swept hull's" look at each axis individually before updating Object A's X and Y components.

### Current implementation
image

1. Shows Object A's last (dotted red) and current (solid red) positions
2. Shows Object A's intermediate position in Yellow
3. Blue box shows swept hull along X-axis based on Object A's `last` coordinates
4. No collision found, so Object A's intermediate x-position is updated
5. Blue box shows swept hull along Y-axis based on Object A's `last` coordinates
6. No collision found, so Object A's intermediate y-position is updated

### New implementation
image
1. Shows Object A's last (dotted red) and current (solid red) positions
2. Shows Object A's intermediate position in Yellow
3. Blue box shows swept hull along X-axis based on Object A's `last` coordinates
4. No collision found, so Object A's intermediate x-position is updated
  * Not pictured: After this step, we temporarily update Object A's last x-position so that new swept hull captures the correct space
5. Blue box shows swept hull along Y-axis based on Object A's newly updated `last` coordinates
6. No collision found, so Object A's intermediate y-position is updated

Issue discussed in discord [here](https://discord.com/channels/162395145352904705/165234904815239168/1361086759315968012)

Reference material that I remembered existed when I started running into this from [Leilani's island devlog](https://forums.tigsource.com/index.php?topic=46289.msg1386874#msg1386874)

Issue #2716, opened 2023-01-17
Expand FlxObject colliders


Geokureli:
Obviously a huge endeavor, but I think I have a way to implement this with backwards compatibility. Putting in 6.0.0 for now, but it's likely to get pushed back.

## FlxCollider
This can either be an enum, or done with inheritance. but the basic types (for now) would be:
- **Rect/AABB** `(x, y, w, h)`: Similar to the existing collider, but the main difference is that the shape can be placed relative to the sprite's x and y. Previously you would use sprite.x ,y, width and height to define the collider and use offset.x/y to align the graphic to the hit shape. this contradicts what most devs intuit. Now you can use x and y to set the top-left of the collider relative to the object's x/y (you can also still use offset, for backwards compat, or for any reason you want)
- **Circle** `(x, y, r)`: I've worked out circle-circle collision using FlxQuadTrees in a jam game, so it's almost ready, we need some non rect shape just to show that other shapes are possible, but we don't need to launch this with a huge polygonal collision library. a circle (and all shapes) will still have a `width` and `height`, which is used for quadtree partitioning

## FlxObjects
With this change, FlxObjects will just be FlxBasics with a position and a collider. for backwards compatibility, we can make obj.width a getter for obj.collider.width, this should prevent confusion between a sprite's graphic size and collision size. In the long run, I want flixel to have more of a decoration pattern, where instead of FlxObjects and FlxSprites, there are just FlxBasics with collider and graphic components.

PR #3431, opened 2025-05-08
Add FlxG.collision and FlxColliders

Geokureli:
https://github.com/HaxeFlixel/flixel/issues/3430

While FlxG.collision is a good thing to have and helpful in general, the main purpose of this PR is to better handle collisions described in https://github.com/HaxeFlixel/flixel/pull/3418

Still isn't perfect, in the EZ platformer demo when you jump up against a vertical wall you'll hit the bottom of a tile, rather than slide up it. One way to fix this is to check the object's deltaY and check tiles in reverse when going up, that should fix this, but I'm wondering if this any less ad-hoc and confusing than #3418.

A couple other features in this PR, are:
- Overlap and collide don't use Dynamic, and have type params
- Can compute overlap without setting touching flags
- Paves the way to discerning a "collision overlap" with a simpler, normal "overlap" which ignores objects' `last` field
- `FlxCollision`: Static extension methods `obj.getDeltaRect()` and `objA.overlapsDelta(objectB)`

TODO:
- [x] Check tiles in reverse when moving up
- [ ] Come up with a shorter/better name than `FlxG.collision.collides`/`overlaps`
- [ ] Test the new overlap and collide with type params
- [ ] See if anything from `FlxCollision` should go in `FlxG.collision`
- [ ] Add new separator method for non-orthogonal overlaps, I.E.: reflect colliders' velocity along the normal of the collision overlap, rather than reflecting X and Y axis individually
- [ ] Fix CI
- [ ] Add circle collisions
- [ ] Fix bugs and features in [test state](https://gist.github.com/Geokureli/1758a0178afc615da9e7a863bbc8a0a9)
  - [ ] implement FlxTilemapExt's `slopeSlowDownFactor` feature
  - [x] Fix finicky collisions (example image below)

Screenshot 2025-05-29 at 11 16 25 AM

### DIFF
--- a/flixel/FlxG.hx
+++ b/flixel/FlxG.hx
@@ -1,5 +1,6 @@
 package flixel;
 
+import flixel.group.FlxGroup;
 import flixel.math.FlxMath;
 import flixel.math.FlxRandom;
 import flixel.math.FlxRect;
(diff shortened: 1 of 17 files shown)


Geokureli:
The reverse iteration worked for EZPlatformer, but I noticed that `FlxTilemapExt` doesn't work with `FlxG.collision.collides`, because it uses `objectOverlapsTiles`, it should either use `forEachCollidingTile`, or I should make `objectCollidesTiles` and use that. I should note that I think FlxTilemapExt is outdated and it's not *extremely* important to make it work with the new system unless FlxG.collides/overlaps is deprecated before a replacement is made for `FlxTilemapExt`

flixel/system/frontEnds/CollisionFrontEnd.hx
```diff
@@ -0,0 +1,475 @@
...
+			if (abs(overlapX) < abs(overlapY))
+			{
+				updateTouchingFlagsXHelper(object1, object2);
+				separateXHelper(object1, object2, overlapX);
+			}
+			else
+			{
+				updateTouchingFlagsYHelper(object1, object2);
+				separateYHelper(object1, object2, overlapY);
+			}
```
MondayHopscotch, line 188: This is effectively just a "poor man's" SAT implementation, as it's not checking the projection axes of the shapes involved, and instead just implicitly looks at x and y projections). I haven't convinced myself it's better than the existing "first X, then Y" process other than this avoids my reuse of the objects' `last` to make it work correctly.

The main benefit of this to me feels like it may open up flixel to (one day) support non rectangular shapes. Not sure if there's desire to support n-gon shapes. Circles don't strictly need something like SAT to work.
The downside to this is now internal seems between shapes will cause snagging in a somewhat unpredictable manner as it depends on the dimensions of the overlap with the seam.

flixel/system/frontEnds/CollisionFrontEnd.hx
```diff
@@ -0,0 +1,475 @@
...
+	 * NOTE: this takes the entire area of `FlxTilemap`s into account (including "empty" tiles).
+	 * Use `FlxTilemap#overlaps()` if you don't want that.
+	 *
+	 * @param   a        The first object or group you want to check.
+	 * @param   b        The second object or group you want to check. Can be the same group as the first.
+	 * @param   notify   Called on every object that overlaps another.
+	 * @param   process  Called on every object that overlaps another, determines whether to call the `notify`.
+	 * @return  Whether any overlaps were detected.
+	 */
+	overload public inline extern function overlap<TA:FlxObject, TB:FlxObject>(a, ?b, ?notify:(TA, TB)->Void, ?process:(TA, TB)->Bool)
```
MondayHopscotch, line 41: I like this change as I almost always write the first two lines of my notify/process func as
```
var player:Player = cast a;
var other:Enemy = cast e;
```

Will this blow up if someone collides/overlaps a group that has the wrong types in it? More concretely if they do something like `overlap<Player, Coin>(player, allCollidableObjectsInGame, (p, c) -> { ... });`

I am not in a spot where I can just run the code, but wanted to ask while it's on my mind so I don't lose the thought.

flixel/FlxG.hx
```diff
@@ -334,6 +335,12 @@ class FlxG
 	 * @since 5.9.0
 	 */
 	public static var assets(default, null):AssetFrontEnd = new AssetFrontEnd();
+
+	/**
+	 * Contains helper functions relating to retrieving assets
```
TechnikTil, line 340:
```suggestion
	 * Contains helper functions relating to collision
```

sorry, i shouldve done this from the get-go

flixel/system/frontEnds/CollisionFrontEnd.hx
Geokureli: I still need some testing, but i dont think it will any worse than using wrong types with the current system, which just uses Dynamics.  One benefit of this is that it ensures the notify and process use the same types

flixel/system/frontEnds/CollisionFrontEnd.hx
```diff
@@ -0,0 +1,475 @@
...
+	{
+		return overlapHelper(a, b, notify, process);
+	}
```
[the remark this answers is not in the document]
Geokureli: I've also been meaning to make a macro that would  decide these types based on the types of args passed in

flixel/system/frontEnds/CollisionFrontEnd.hx
Geokureli: Can you link me to an explanation of SAT? It's hard to google this

MondayHopscotch: This is the tutorial/guide I used back when I implemented this long ago:
https://www.metanetsoftware.com/technique/tutorialA.html
https://www.metanetsoftware.com/technique/tutorialB.html

Unfortunately, all the flash demos are dead, but it did a really good explanation for my brain.

The gist of SAT is that you project two collision shapes onto each the shapes' normals in isolation. The smallest overlapping axis is the one you move the shapes along. Incoming attempt at oversimplifying it just to give a starting point:

SATDemoAttempt

1. First image shows two shapes we've identified as needing resolution (I'm glossing over the broad phase) and all of the normals of said shapes. Of which there are 4, as direction doesn't matter (2 from the square, and 2 from the triangle as it's bottom edge is not unique)
2. Project both shapes along all 4 of these axes and find the one of shortest overlap
3. Resolve the shapes along that axis of shortest overlap.

This works for convex polygons.

I had commented that that flixel does a "poor man's" version of this as it just assumes we are only needing to "project" along the X and Y axes. So the idea is there, but there is no consideration for more complicated shapes other than axis-aligned rectangles.

MondayHopscotch: Looks like you can install the Ruffle extension from the chrome store and it will let you view the embedded flash player elements. Highly recommend as it they are a bunch of interactive demos to let you play with the concepts they talk about.

Geokureli:
Made a ton of changes. Now object.computeCollisionOverlap returns how the objects should be separated, before it returned how they should be separated on each axis, and the caller would ignore the larger one. This should be better for non-AABB objects, thanks for the tip @MondayHopscotch, I loved those metanet tutorials but completely forgot they existed.

I also started work on making slope tools for tilemaps, it all works somewhat well, except in my new version of FlxTIlemapExt glue-down is not great, and i haven't implemented slow down when going u slopes. I also made FlxTIlemapExt more compatible with `FlxG.collision.collide`

I'm considering making an ICollider interface, and change all refs to FlxObject to IColliders (but only in new code), this way you can make custom colliders without having to extend FlxObject, this is in line with [this goal](https://github.com/HaxeFlixel/flixel/issues/2716). colliders would only need a `getBounds(resultRect)` and a `computeCollisionOverlap` The issue here is that FlxQuadTrees would need to be built from the ground up. Though, FlxQuadTrees are [in need of an update](https://github.com/HaxeFlixel/flixel/pull/2891) so maybe having another class that will one day replace the existing one is an option (a very ugly option). The simple option is to just use FlxObjects, but then changing all this later is a huge pain

PR #3436, opened 2025-05-27
allow destroying flxsignals during dispatch

DetectiveBaldi:
At first glance this does seem like a bit of a strange request, don't have high hopes for getting this merged

This issue has come up a lot when trying to `destroy` sprites when an animation completes. I would much prefer to be able just destroy during dispatch rather than `kill` the sprite, or use a longer if check under `update`

### DIFF
--- a/flixel/util/FlxSignal.hx
+++ b/flixel/util/FlxSignal.hx
@@ -302,12 +302,15 @@ private class Macro
 
 			processingListeners = false;
 
-			for (handler in pendingRemove)
+			if (pendingRemove != null)
 			{
-				removeHandler(handler);
+				for (handler in pendingRemove)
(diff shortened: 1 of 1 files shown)


DetectiveBaldi:
> Can I get an example of what fails, one that should work after this change

```haxe
package;

import flixel.util.FlxSignal;
import flixel.FlxState;

{
	public var mySignal:FlxSignal;

	{
		super.create();

		mySignal = new FlxSignal();

		mySignal.add(dispatchOne);

		mySignal.dispatch();
	}

	public function dispatchOne():Void
	{
		// uncommenting causes crash
		// mySignal.destroy();
		trace("Dispatched one");
	}
}
```

Issue #3361, opened 2025-02-10
Removing Neko support/NO_HEALTH Define in Flixel 6.0.0+


moxie-coder:
Neko is in short, an dead brother of Hashlink (which has superseded Neko) and it's pretty outdated lately, it's last updated being in mid 2024 (when it was deprecated, it's not long ago but still, Neko is discontinued). And as such, since Flixel 6.0.0 has removed a bunch of deprecated stuff and given how Hashlink is more better then Neko, I'd figured we could finally deprecate then remove Neko in upcoming Flixel 6.0.0 patches, including the `FLX_NO_HEALTH` define that I had made/suggested since it was for backwards compatibility with older flixel projects and seemed no longer necessary since people could extend FlxBasic/FlxObject & add it themselves, I think it would be wise to remove it soon in an upcoming Flixel 6.0.0 patch since it's deprecated and soon to be removed much like how Neko should be

moxie-coder:
EDIT: turns out Neko was deprecated in 2021, my bad, but that shows how old Neko is and it's not actively maintained anymore, as stated in the readme, so I don't see why it should be sticking around if it's not used anymore and causes issues in favor of Hashlink

Geokureli:
https://github.com/HaxeFlixel/flixel/pull/3407
object.health will now warn people if it is used without defining FLX_HEALTH. this is to prepare for when FLX_NO_HEALTH is defined by default, unless FLX_HEALTH is defined (the opposite of how it is now) in 7.0.0

PR #3437, opened 2025-05-28
Remove Neko Target

CyndaReal:
Addresses #3361 

As stated in the original issue that brought this to my attention, Neko is no longer officially supported, and has been deprecated since 2021. Very, very few people use it (or seemingly have in a long time), and it has been all but entirely superseded by Hashlink.

This PR removes the Neko target and all mentions of it.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       matrix:
         haxe-version: ["4.2.5", "4.3.6"]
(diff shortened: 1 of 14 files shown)


CyndaReal:
> would like to note that there is probably additional code in related flixel repositories that would probably be affected by this idea it also might be worth labeling the target as deprecated instead of outright removing it (this is mentioned in the linked issue)

Yeah.

PR #3445, opened 2025-07-06
Use set in FlxPoint

gimmesoda:
Previously when using the set command in FlxCallbackPoint, callback was called twice, but now it is called only once


Geokureli:
@gimmesoda I was planning on merging this now that 6.1.1 was released

gimmesoda:
Oh, sorry, I'll create the repo again and make a new pr

PR #3447, opened 2025-07-09
implement overloaded methods on FlxColor

### LINKED ISSUE #3444: Implement overloaded methods in FlxColor
Being able to pass in an array directly into `FlxColor.from` functions, `FlxColor.fromRGB()` for example would be really nice to have 
It would be fairly easy to implement just create an overload on the functions to accept an array.

Vortex2Oblivion:
Allows for `Array` values to be used on `from` functions (eg: `fromRGB`)
Closes #3444

### DIFF
--- a/flixel/util/FlxColor.hx
+++ b/flixel/util/FlxColor.hx
@@ -1,8 +1,8 @@
 package flixel.util;
 
-import flixel.tweens.FlxEase;
 import flixel.math.FlxMath;
 import flixel.system.macros.FlxMacroUtil;
+import flixel.tweens.FlxEase;
 
 /**
  * Class representing a color, based on Int. Provides a variety of methods for creating and converting colors.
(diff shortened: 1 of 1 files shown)


Geokureli:
1. Make sure your settings doesn't modify the entire file's formatting, via: `"editor.formatOnSaveMode": "modifications"` in .vscode/settings.json
2. This seems like a weird choice to put in Flixel proper, rather than in your game engine's utils, either with static extension or abstract extension. I'm mainly hesitant because different people might want different implementations of this. I do this in a project, now, where the array comes from a parsed json file, and in my implementation I throw an error on debug modes when the values are invalid or the array is not the right length, while on production it silently assumes 0, wherever needed. 

I'm really avoiding adding tiny little ad-hoc features like this, in favor of people making their own, especially when they may work differently than how people expect

Vortex2Oblivion:
>  I'm mainly hesitant because different people might want different implementations of this.

What "different implementation" would someone want?

> I'm really avoiding adding tiny little ad-hoc features like this, in favor of people making their own, especially when they may work differently than how people expect. 

It returns a color from an array of values, what would someone expect instead?

Vortex2Oblivion:
Mainly I just want to be able to pass an array in directly instead of doing `FlxColor.fromRGB(array[0], array[1], array[2])`

Geokureli:
> Mainly I just want to be able to pass an array in directly instead of doing `FlxColor.fromRGB(array[0], array[1], array[2])`

Why are they in an array to begin with? The idea of storing colors as [255, 127, 255] seems unlikely

Vortex2Oblivion:
RGB values for healthbar stuff in JSON files

Geokureli:
Why would the json store colors as an array of integers, rather than say an int or string of 0xFFffffff?

Vortex2Oblivion:
> Why would the json store colors as an array of integer bytes, rather than say a single 4byte int or hex string?

old system, too late to change it

PR #3449, opened 2025-07-12
Improve zoomed debug hitboxes performance

MaybeMaru:
This PR clamps the rendered hitbox just outside the bounds of the camera. This has to be done since, by the way openfl works, the rects get cached as bitmaps and when zoomed can easily generate absurdly large bitmaps, which cause the framedrops.
It also fixes the hitbox being slightly round around the edges.

This PR has been tested on both native and web targets and seem to perform with the same consistent framerate.
This solution is only necessary for non-flash targets, as flash doesn't have this issue.

It is important to note, however, that this fix is only temporal. Even though it fixes the issue, a shader solution (such as the one proposed in https://github.com/HaxeFlixel/flixel/pull/3168) would be the way to go to completely skip all the bitmap caching openfl has to handle in the background.

| Before  | After |
| ------------- | ------------- |
|  before | after |


Geokureli:
I see no immediate reasons not to merge this, but I have questions:
1. Can you share your testing environment?
2. I like the MITER, does that improve the performance or was it purely aesthetic? I imagine rounded corners are harder to draw
> ... the rects get cached as bitmaps...
3. This doesn't make any sense to me, I don't see why they would be cached as a bitmap, can you back up this claim?
> Even though it fixes the issue, a shader solution (such as the one proposed in https://github.com/HaxeFlixel/flixel/pull/3168)
4. At best this alleviates specific cases of bad debug draw performance, there are many that have nothing to do with zoom or off-screen objects

Geokureli:
> > 4. At best this alleviates specific cases of bad debug draw performance, there are many that have nothing to do with zoom or off-screen objects
> 
> I mentioned zoom because it’s the most common reason for these leaks to happen in the Flixel projects I've worked on. But they're also possible in other common cases, such as trying to render an object that is scaled beyond the bounds of a camera (like an overlay or background sprite).

The point I was trying to make but failed to mention was that I would remove the "Fixes 3164" as this is not a total fix... At least I think it's not but I haven't tested.

MaybeMaru:
> The point I was trying to make but failed to mention was that I would remove the "Fixes 3164" as this is not a total fix... At least I think it's not but I haven't tested. My guess is that the issue will need to stay open, even after this

Ah I see what you mean. Yeah that's why I mentioned this could be considered more of a "temporal" solution while the method Cheems suggested gets worked on further. I'll remove the "fixes" part of the comment.

Geokureli:
The bounds aren't well aligned on HL. I think we should extend it slightly beyond the bounds, in fact I wouldn't mind if it was

Screenshot 2025-07-29 at 10 42 10 AM

Here's my [TestState](https://gist.github.com/Geokureli/c507f71baa037bbc66a3c388ad86b500)

> Another change that I've been thinking about is keeping the 1 pixel thickness consistent between camera zooms, I can push that too if its a feature that could be wanted here.

How would you do this without adding a new arg to `drawDebugBoundingBox` (a breaking change) or assuming which camera it's being drawn to?

Geokureli:
> Sorry for the delay, I've been busy with some things and semi forgot about this PR.

No rush

> > I think we should extend it slightly beyond the bounds
> 
> Im not entirely sure if this would be a good idea because when testing for collisions in something like a platformer it could present some visual overlap between hitboxes, but I can experiment if you want.

This doesn't seem terrible, but it's worth checking in some practical situations

> > How would you do this without adding a new arg to `drawDebugBoundingBox` (a breaking change) or assuming which camera it's being drawn to?
> 
> I thought of implementing it as a breaking change, but on second thought maybe it's not a good idea.

Feel free to make a separate issue for it, and it can be added to the roadmap

Geokureli:
Moving this off the 6.1.1 release, I just noticed this doesn't work in blit targets, using my test state above, try testing with `hardware="false"` in your project.xml's `<window/>` node. It seems to apply zoom twice when calculating the camera's view. You'll also note that the bounding boxes are offset a little, that's unrelated (Just made: https://github.com/HaxeFlixel/flixel/issues/3494)

Screenshot 2025-09-25 at 1 35 00 PM

Geokureli:
Fixed the issue on blit targets, changed my mind of the following:
> I think we should extend it slightly beyond the bounds, in fact I wouldn't mind if it was noticeably bigger, so as to standout from the game's graphics

I think we are good to go, gonna merge in the morning as I'm too tired to wait for CI, right now

PR #3453, opened 2025-07-23
Better pixel font support for FlxText

ACrazyTown:
OpenFL's text renderer always applies smoothing to the text UNLESS the following conditions are met:
- `textField.antiAliasType` is set to `ADVANCED`
- `textField.sharpness` is `400`

https://github.com/openfl/openfl/blob/31910509ebd49d1320792e66d761ade9039de5fb/src/openfl/display/_internal/CairoTextField.hx#L164-L169
https://github.com/openfl/openfl/blob/31910509ebd49d1320792e66d761ade9039de5fb/src/openfl/display/_internal/CanvasTextField.hx#L165-L168

This fix applies those properties to the text field if the text's anti-aliasing is set to false, causing a noticeable improvement when rendering pixel fonts. This only seems to have a significant effect on native targets, HTML5 doesn't seem affected.

Before:
image

After:
image


CyndaReal:
THANK YOU

Geokureli:
This is huge, btw!

Geokureli:
Yeah, I had no idea about sharpness of 400, but looking back in the discord it came up a few times

ACrazyTown:
> I'm still seeing weird artifacting on FlxText, is it because I need to use the git version of openfl?

Could you share the code for this? I can't seem to reproduce this both on OpenFL 9.4.1 and git.

> This is likely unrelated, but I also noticed FlxText doesn't work with FLX_RENDER_TRIANGLES

Yeah, this shouldn't have an impact on something like that

ACrazyTown:
Weird. Maybe it's an OS difference? I'm on Windows. Either way this seems to be outside of Flixel's scope

ACrazyTown:
> @ACrazyTown, I'm actually not able to detect any difference on any target with this change. I coulda swore I tested this personally, but mac//windows/hl/html5 shows no difference, regardless of whether I comment out the set_antialiasing override. Do I need to use git lime/openfl or something?

You'll need a git version of openfl with this commit https://github.com/openfl/openfl/commit/215ec5b3445af773c23db1f6be34f3e7492cf0f3, if you don't have it already. Without it the text rendering is bugged and it'll apply the anti-aliasing of the first text drawn to all text instances. There should be a noticable difference only on desktop targets. Lmk if this works out or not so I can take a deeper look

Geokureli:
> There should be a noticable difference only on desktop targets. Lmk if this works out or not so I can take a deeper look

Even with that commit, I'm seeing no difference on any target. It renders the same, regardless of whether the sharpness is 100 or 400.

Windows/HL:
win-dev

Mac/HL:
mac


Here's my test state:
```hx
package;

import flixel.FlxG;
import flixel.FlxSprite;
import flixel.text.FlxText;

{
	final textStr = "The quick brown fox jumps over the lazy dog";
	var text:FlxText;
	
	{
		super.create();
		bgColor = 0xFF808080;
		
		// final text1 = new FlxText(0, 0, FlxG.width, "Hello World!");
		final text1 = new FlxText(0, 0, 0, "Hello, world!", 16);
		text1.alignment = CENTER;
		text1.screenCenter();
		text1.y += 20;
		add(text1);
		
		text = new FlxText(0, text1.y + text1.height + 2, FlxG.width, textStr);
		text.alignment = CENTER;
		text.screenCenter(X);
		add(text);
	}
	
	{
		super.update(elapsed);
		
		{
			text.textField.antiAliasType = text.textField.antiAliasType == NORMAL ? ADVANCED : NORMAL;
			text.textField.sharpness = text.textField.sharpness == 400 ? 100 : 400;
			@:privateAccess
			text._regen = true;
		}
		
		if (FlxG.keys.justPressed.RIGHT && text.text.length < textStr.length)
			text.text = textStr.substr(0, text.text.length + 1);
		
		if (FlxG.keys.justPressed.LEFT && text.text.length > 0)
			text.text = textStr.substr(0, text.text.length - 1);
	}
}
```

Ignore that the images say "Hello, world!"

ACrazyTown:
@Geokureli I noticed an issue in your test case. It was setting the text field's `antiAliasType` to `NORMAL` instead of `ADVANCED` when sharpness is `400`. 
```haxe
{
	text.textField.antiAliasType = text.textField.antiAliasType == ADVANCED ? NORMAL : ADVANCED;
	text.textField.sharpness = text.textField.sharpness == 400 ? 100 : 400;
	@:privateAccess
	text._regen = true;
}
```

I also made it possible to change the font size in order to make it the effect more visible:
```hx
if (FlxG.keys.justPressed.UP)
	text.size += 10;

if (FlxG.keys.justPressed.DOWN)
	text.size -= 10;
```

With these change I can see a visible difference when hitting space. Like Vortex mentioned, it's more noticeable with a smaller/bigger font size. 

For example, pressing UP once and full-screening the window on a 1080p monitor produces the infamous FlxText smoothing effect. By pressing SPACE, the smoothing effect disappears and the text becomes as pixel-perfect as possible. (There's a small slice in the middle but that's because 640x480 can't scale up to 1920x1080 perfectly).

image
image

It's also noticeable with bigger font sizes. Here's the text at size 58px and the default window size of 640x480:

(Pixel-perfect off)
image

(Pixel-perfect on)
image

PR #3454, opened 2025-07-28
`FlxFrame`: Fix rotated frames not rendering on Hashlink

### LINKED ISSUE #3433: Rotated frames not rendering on Hashlink
After updating HaxeFlixel and all of it's backend libraries to their latest versions, I've noticed sprites specifically won't render rotated frames. This issue only seems prevalent when targeting Hashlink and it does not seem to affect the C++ target on my end. 

Using the debugger console and setting the frame's `angle` parameter to 0 made the affected frames render once again (minus the rotation).

Image

Sample code:
```haxe
{
	{
		super.create();

		var sprite:FlxSprite = new FlxSprite();
		sprite.frames = FlxAtlasFrames.fromSparrow("assets/images/TestImage.png", "assets/images/TestImage.xml");
		sprite.animation.addByPrefix("static", "TestAnimation");
		sprite.animation.play("static");
		add(sprite);
	}
}
```

Sample sparrow atlas with rotated frame:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<TextureAtlas imagePath="TestImage.png">
    <SubTexture name="TestAnimation0001" x="0" y="0" width="150" height="150" rotated="true"/>
</TextureAtlas>
```

rifxii:
Closes #3433 

This seems to solve the previously mentioned issues I had with rotated frames not rendering on Hashlink.

Judging by what I saw while testing, it seems that the HL compiler is inlining `MatrixVector.set` by using pointers of the given variable. So instead of the set function in `MatrixVector.copyFrom` performing `this[0] = -b; this[1] = a;`, it's instead doing `this[0] = -this[1]; this[1] = this[0];` and messing up the whole matrix. 

I also found setting a reference to the variables before `set()` also fixes the issue WITHOUT the removal of the inline but I'm not sure which one is better to deal with and if there's any sort of performance loss from removing the inline.

Example:
```haxe
#if hl
final a:Float = a; 
final b:Float = b;
#end

set(-b, a, ...);
```

### DIFF
--- a/flixel/graphics/frames/FlxFrame.hx
+++ b/flixel/graphics/frames/FlxFrame.hx
@@ -842,7 +842,7 @@ abstract MatrixVector(Vector<Float>)
 		ty = 0;
 	}
 	
-	public inline function set(a = 1.0, b = 0.0, c = 0.0, d = 1.0, tx = 0.0, ty = 0.0)
+	public #if !hl inline #end function set(a = 1.0, b = 0.0, c = 0.0, d = 1.0, tx = 0.0, ty = 0.0)
 	{
 		set_a(a);
 		set_b(b);


MaybeMaru:
If you haven't already this should probably be reported to the Hashlink or Haxe repo too. Seems like a very important issue with the target.

Geokureli:
Also, I've been extremely busy and forgot about this PR, if this prevents the error, I'm okay with merging this, since the haxe change won't fix haxeflixel's minimum supported haxe versions.

PR #3455, opened 2025-07-30
Fix alpha channel

NickMGC:
In HaxeFlixel 5.9.0, the drawFX function was overhauled, but alpha handling was unaccounted for. This overhaul broke how alpha was displayed when using `FlxCamera.flash()`- when assigning a color with opacity, it would ignore the alpha value and display the color at full opacity. This pull request fixes that issue.


Geokureli:
Thanks!

PR #3457, opened 2025-08-03
Fix function description in FlxMath

glintfishGH:
The comment describing FlxMath.wrap says it wraps the value between 0 and max, when it doesn't do that.


Geokureli:
Thanks!

PR #3459, opened 2025-08-06
FlxCamera: Fix filter memory leak

lunacynne:
currently the game rapidly leaks memory if 5 conditions are met:
- using `renderTile`
- some camera `bgColor` alpha is `0`
- no sprites or anything being rendered to this camera (nothing in draw stack)
- at least one `BitmapFilter` applied to this camera
- no flash fx or color fade fx active on this camera

unsure of exactly what the lower-level issue is causing this bug, but my guess is that after the `canvas.graphics` is cleared in `CameraFrontEnd.lock`, due to the above conditions its never drawn to again before the next frame, and so the filters are put in a weird state and somehow memory is leaked

this pr fixes the issue by adding a `force` option to `FlxCamera.fill` to bypass the alpha early-return for `renderTile`, and forcing a `bgColor` fill for it if any filters are currently active on the camera regardless of the alpha value

---

to reproduce the issue:
```haxe
package;

import openfl.filters.BitmapFilter;

import flixel.FlxCamera;
import flixel.FlxG;
import flixel.FlxSprite;
import flixel.FlxState;

class TestState extends FlxState {
	override function create():Void {
		super.create();
		
		add(new FlxSprite().makeGraphic(1, 1, 0x0));
		// ^ remove this line and the game will leak memory like crazy
		
		var cam:FlxCamera = new FlxCamera();
		cam.bgColor = 0x0;
		FlxG.cameras.add(cam, true);
		cam.filters = [new BitmapFilter()];
	}
}
```

before and after fix:

Screenshot 2025-08-06 153823

image

### DIFF
--- a/flixel/FlxCamera.hx
+++ b/flixel/FlxCamera.hx
@@ -1660,7 +1660,7 @@ class FlxCamera extends FlxBasic
 	 * @param   Color        The color to fill with in `0xAARRGGBB` hex format.
 	 * @param   BlendAlpha   Whether to blend the alpha value or just wipe the previous contents. Default is `true`.
 	 */
-	public function fill(Color:FlxColor, BlendAlpha:Bool = true, FxAlpha:Float = 1.0, ?graphics:Graphics):Void
+	public function fill(Color:FlxColor, BlendAlpha:Bool = true, FxAlpha:Float = 1.0, ?graphics:Graphics, force:Bool = false):Void
 	{
 		if (FlxG.renderBlit)
 		{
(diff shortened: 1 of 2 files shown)


cyn0x8:
extra arg is causing issues... ill figure out a different way soon
trying to avoid just putting the filter check at the alpha early-return because `fill` isnt only used for `bgColor`; idk if the check there would be acceptable or not

cyn0x8:
i know for sure the leak happens on cpp windows; i didnt see it happen on hl windows or html5

and as for the `FxAlpha == 0` i believe its intended to be an optimization? an early return skipping fully transparent cam fills
i didnt remove it in case it had to be kept, but if its acceptable i can remove the check entirely

Geokureli:
Thanks!

ACrazyTown:
I believe the alpha check was there to avoid an OpenGL draw call. The rectangle, albeit invisible, is still submitted to be drawn. I don't know if there are any performance implications to it but it is a draw call nonetheless

It might be worth looking into this deeper, since it probably isn't a Flixel issue.

PR #3469, opened 2025-08-30
Deprecate FlxGraphicSource in favor of FlxGraphicAsset

rifxii:
FlxGraphicSource from my understanding serves the same purpose as FlxGraphicAsset, so I merged FlxGraphicSource's resolveBitmapData function and BitmapData class support into FlxGraphicAsset, deprecated it, and updated any usage of it inside flixel. 

My main issue with FlxGraphicSource is the lack of support for FlxGraphic. When using FlxAtlas, one of the very few classes that support it, having to write `FlxAtlas.addNode(graphic.bitmap);` instead of `FlxAtlas.addNode(graphic);` feels like an unnecessary workaround.


Geokureli:
Thanks!

Issue #3267, opened 2024-10-15
[ENHANCEMENT] Implement null safety


EliteMasterEric:
[Null safety](https://haxe.org/manual/cr-null-safety.html) is an optional feature of the Haxe compiler which enforces compile-time checking of nullable values.

By implementing null safety throughout HaxeFlixel, we can greatly reduce the frequency of crashes and other errors. This can be done using the `@:nullSafety` annotation at the top of each class, and can be done incrementally until every class in the library is compliant.

This would not break existing code which has null safety enabled, but it would:

- Improve documentation by informing users which variables are `Null`-able (such as `FlxG.sound.music`).
- Reduce the incidence of certain crashes within the engine by guaranteeing they are caught and handled.
- Improve the compatibility with other code where null safety is enabled.

As an example of the last case, I have started to enable null safety in Funkin', and have had to create blocks of code where null safety is not enabled in several places, simply because Flixel itself regularly neglects to use `Null<T>` for nullable values (again, trying to set `FlxG.sound.music = null` makes the compiler complain until you specifically exclude that line).

One major obstacle to implementing null safety is that null safety is necessarily final (see [this issue](https://github.com/HaxeFoundation/haxe/issues/11304) or [any of these issues](https://github.com/HaxeFoundation/haxe/labels/feature-null-safety)). Overall though, I would like to see it become a long-term goal of the engine, especially since it can be done in a non-breaking manner in most cases.

Geokureli:
I've been considering null safety when adding new features and modifying existing ones, and I'm all for it. For stuff like this I tend to prefer a bunch of specific changes rather than a giant sweeping change across the entire repo, as it's usually much easier to test.

For instance, your recommended change to SoundFrontEnd would be great, along with anything else in that class. In general whenever I think about null safety, I first see if I can make it so that the value is never null, rather than nulling out values in destroy, can they instead be a final var that gets cleared. I've also trying to reduce nullable fields by using enums, where the field is only defined in the cases that use it. I think there's many more cases that can be improved using these methods.

There are cases that really baffle me, though. For instance `FlxG.state`, theres definitely a period in every flixel game where this can be null, but the idea of null checking this every time you use, or doing `FlxG.state.sure()` every time seems like an absolute pain especially since it's very easy to avoid null refs with this particular field.

Null-safety is something enabled per-module, right? We could make a checklist of modules that are already null-safe or modules that can easily become null-safe and enable/finalize them

[...]

Issue #2803, opened 2023-05-15
Raise minimum haxe version from 4.2.5 to 4.3.X


Geokureli:
Continuing off of #2774. It's still far too early to switch to 4.3 as a minimum target, it hasn't even been a month and adoption for 4.3 will be notably slower than 4.2 since there are so many deprecation warnings and odd errors that have turned up, notably #2780

Can use self in abstracts that reference themself, example: https://github.com/HaxeFlixel/flixel/pull/2802

- Nice, but has no actual benefit to users

TheHENOOB:
wait until Haxe 4.3.x releases to Ubuntu PPA.
https://launchpad.net/~haxe/+archive/ubuntu/releases

Geokureli:
there's a bug in Haxe 4.2.5, 4.3.1 and nightly, preventing us from making FlxBaseTilemap and FlxTypedTilemap an abstract class. I do want to  make that change but it will have to wait

Geokureli:
Going to officially drop haxe 4.2 in flixel 7

PR #3477, opened 2025-09-13
Refactor the `FlxG.random.color` function

VioletSnowLeopard:
The `color` function in FlxRandom returns a random FlxColor. There are optional arguments for min, max, and alpha values. Because they are optional, they need to be checked if they are null before using them. The current code does this with an if / else if / else branch. This ends up repeating code with slight variations. Also, if you plan to implement null safety for this class, the compiler does not like this method.

This PR greatly simplifies the code inside this function by using `?.` and `??` operators instead of the if statements for null checks. 33 lines of code have been reduced to 4 without any functional difference. It will also not cause any issues when compiling with null safety enabled.

(There are still other issues in FlxRandom for compiling with null safety. This just fixes the issues from the `color` function.)

### DIFF
--- a/flixel/math/FlxRandom.hx
+++ b/flixel/math/FlxRandom.hx
@@ -326,39 +326,10 @@ class FlxRandom
 	 */
 	public function color(?Min:FlxColor, ?Max:FlxColor, ?Alpha:Int, GreyScale:Bool = false):FlxColor
 	{
-		var red:Int;
-		var green:Int;
-		var blue:Int;
-		var alpha:Int;
-
-		if (Min == null && Max == null)
-		{
-			red = int(0, 255);
-			green = int(0, 255);
-			blue = int(0, 255);
-			alpha = Alpha == null ? int(0, 255) : Alpha;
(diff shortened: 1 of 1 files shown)


VioletSnowLeopard:
I believe the 4.2.5 builds are failing because haxe didn't have the null safety operators `?.` and`??` in that version. You don't have to accept this PR if you want to keep support for haxe 4.2.5.

Although I will say that trying to implement null safety without the `?.` and `??` operators could be difficult.

Geokureli:
Related:
- https://github.com/HaxeFlixel/flixel/issues/3267
- https://github.com/HaxeFlixel/flixel/issues/2803

We will eventually raise the minimum haxe version and start, but I find this particular usage of ?. to be frivolous and overly complicated. I also don't think this function is very useful, in general, and if it was useful, it doesn't even need to use `Null<Int>` for min and max.

Personally, I recommend using `FlxColor.fromHSL`, `fromHSB` or `fromRGB` with random args

PR #3478, opened 2025-09-14
Add loops to FlxSound

glowsoony:
Adding a method to have a certain amount of times a FlxSound can loop before cleaning up


glowsoony:
close this in favor of [#3546](https://github.com/HaxeFlixel/flixel/pull/3546)

PR #3479, opened 2025-09-14
FlxSound: Rework load method names

### LINKED ISSUE #3466: Allow FlxSoundAsset from ByteArray
`new FlxSound().loadByteArray` exists but there's no way to pass a byte array into `FlxG.sound.play/load`. If [FlxSoundAsset](https://github.com/HaxeFlixel/flixel/blob/dev/flixel/system/FlxAssets.hx#L38) took a ByteArray those methods would work and we could have a better name convention for `loadEmbedded/loadByteArray`

ACrazyTown:
Closes #3466.

- Deprecates `loadEmbedded()`,  `loadStream()` and `loadByteArray()` and replaces them with `load()` and `loadFromURL()`.
- Deprecate `FlxG.sound.stream()` for `FlxG.sound.loadFromURL()` to stay consistent
- ~~Add `loadStreamed()` to allow loading streamed sounds. Additionally adds required functionality to `FlxG.assets`~~ https://github.com/HaxeFlixel/flixel/pull/3518
- Modifies `FlxSoundAsset` to allow passing in a ByteArray.

### DIFF
--- a/flixel/sound/FlxSound.hx
+++ b/flixel/sound/FlxSound.hx
@@ -382,35 +382,6 @@ class FlxSound extends FlxBasic
 	}
 
 	/**
-	 * Loads a streamed sound from the provided file path.
-	 * This does not load sounds from web locations. Use `loadFromURL()` for that, instead.
-	 * 
-	 * If sound streaming is not supported, a normal sound will be returned.
-	 * 
-	 * **Note:** If the `FLX_DEFAULT_SOUND_EXT` flag is enabled, you may omit the file extension
-	 * 
-	 * @param path The path to the sound asset.
-	 * @param looped Whether or not this sound should loop endlessly.
-	 * @param autoDestroy Whether or not this FlxSound instance should be destroyed when the sound finishes playing.
-	 * @param onComplete Called when the sound finishes playing.
-	 * @return This FlxSound instance (nice for chaining stuff together, if you're into that).
-	 * 
-	 * @since 6.2.0
-	 */
-	public function loadStreamed(path:String, looped:Bool = false, autoDestroy:Bool = false, ?onComplete:Void->Void):FlxSound
-	{
-		cleanup(true);
-
-		if (FlxG.assets.exists(path, MUSIC))
-			_sound = FlxG.assets.getMusicUnsafe(path);
-		else
-			FlxG.log.error('Could not find a Sound asset with an ID of \'$path\'.');
-		
-		// NOTE: can't pull ID3 info from embedded sound currently
-		return init(looped, autoDestroy, onComplete);
-	}
-
-	/**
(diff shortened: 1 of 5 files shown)


Geokureli:
Is this something that can be unit tested?

Geokureli:
NVM, [FlxSoundTest](https://github.com/HaxeFlixel/flixel/blob/dev/tests/unit/src/flixel/sound/FlxSoundTest.hx#L3) doesn't actually test any of the existing args, so let's look into that in the future

ACrazyTown:
> Just gonna say it, a byte array isn't "embedded", so it's odd to use this to load it... Maybe we should add a new `load()` method and deprecate loadEmbedded (maybe in a later version)?
> 
> Thoughts?

Sounds good. While we're at it, I think it might be worth renaming [`loadStream()`](https://api.haxeflixel.com/flixel/sound/FlxSound.html#loadStream) as well. From what I can tell the main purpose of it seems to be to load sounds from web locations, but there's a different concept of "streamed sounds" that would be nice to have built in.

Streamed sounds are helpful to use less memory when dealing with large music tracks for example. Instead of loading the entire file and playing it, they load (and unload) smaller chunks of data as they play.

We could have something like `load()`, `loadFromURL()` and `loadStreamed()` (but maybe that could be an argument in `load()` instead)

Geokureli:
`loadFromURL` seems best to me

Also Openfl does have the ability to "stream" audio, but I don't know if flixel leverages this

Geokureli:
> > Also Openfl does have the ability to "stream" audio, but I don't know if flixel leverages this
> 
> I believe it doesn't. I think we just have to use `Assets.getMusic()` (which internally sets up the sound to be streamed) instead of `Assets.getSound()`

I forgot how I did it, but there's a way on html5 to stream a sound from the web

ACrazyTown:
Should I repurpose this and also include the load() changes here or should that be done in a seperate PR?

Geokureli:
> Should I repurpose this and also include the load() changes here or should that be done in a seperate PR?

Please include that change, here

ACrazyTown:
There's still some quirks I need to figure related to audio streaming. To not bloat this PR I'll strip out parts related to audio streaming and re-add them in a separate PR later.

ACrazyTown:
> Maybe don't restrict MUSIC asset type to lime_vorbis in the future? since who knows if lime is going to add another format being able to streamed beside vorbis

Yeah, that's something that could be addressed in a patch release with a compiler conditional. I have to stick to checking `lime_vorbis` and messing with the VorbisFile API currently because Lime doesn't really have a proper cross-platform way to check if audio streaming is supported.

Geokureli:
Thanks!

PR #3480, opened 2025-09-15
Fix inaccurate game filter sizes

### LINKED ISSUE #2258: Game filters: wrong size?
`PlayState.hx`

```haxe
package;

import flixel.text.FlxText;
import flixel.system.FlxAssets.FlxShader;
import openfl.filters.ShaderFilter;
import flixel.FlxG;
import flixel.FlxState;

{
	{
		super.create();
		FlxG.game.setFilters([new ShaderFilter(new BorderShader())]);
		FlxG.game.filtersEnabled = true;

		var text = new FlxText();
		text.text = "Hello World!";
		text.screenCenter();
		add(text);
	}

	{
		super.update(elapsed);
		if (FlxG.mouse.pressed) FlxG.camera.shake(0.01, 0.1);
		if (FlxG.keys.anyJustPressed([ONE])) FlxG.game.filtersEnabled = !FlxG.game.filtersEnabled;
	}
}



class BorderShader extends FlxShader 
{
	
	@:glFragmentSource("
        #pragma header
		
		void main()
		{
			vec2 uv = openfl_TextureCoordv;
			vec3 col = texture2D(bitmap, vec2(uv)).rgb;
			
			if ((uv.x > 0.0 && uv.x < 0.2) || (uv.x < 1.0 && uv.x > 0.8) || (uv.y > 0.0 && uv.y < 0.2) || (uv.y < 1.0 && uv.y > 0.8)) {
				gl_FragColor = vec4(1., 0., 0., 1.);
				return;
			}
			gl_FragColor = vec4(col, 1.);
		}
	")
	
	public function new() 
	{
        super();
	}
}

```

**Observed behavior:**
Game filters appear to detect wrong game size.
GameFilter 15_06_2020 17_48_07
This is filter applied to whole game

GameFilter 15_06_2020 17_48_26
This is filter applied to whole game after a camera shake

GameFilter 15_06_2020 17_48_49
This is filter applied to camera

**Expected behavior:**
Game filter to look like camera filter

In last image bottom border != top border because screenshot cut lowest part to keep window toolbar (dunno why).

ACrazyTown:
Should close #2258!

In short, Flixel used to add the cameras' OpenFL Sprites to the FlxGame instance along with the mouse, debugger, sound tray, etc. Because of this, game filters did not report accurate sizes because they included things outside of the game view.

To resolve this, I added a `_cameraContainer` Sprite to FlxGame which exclusively holds all of the cameras' Sprites. Game filters are now applied to this container and will no longer affect the debugger, mouse and other OpenFL level Sprites in FlxGame. I guess this is technically a breaking change, but I don't think game filters ever worked properly to begin with.

I tried splitting up each change in a different commit to hopefully make the reviewing process easier

### DIFF
--- a/flixel/system/frontEnds/CameraFrontEnd.hx
+++ b/flixel/system/frontEnds/CameraFrontEnd.hx
@@ -57,22 +57,22 @@ class CameraFrontEnd
 	 * Handy for PiP, split-screen, etc.
 	 * @see flixel.FlxBasic.cameras
 	 *
-	 * @param	NewCamera         The camera you want to add.
-	 * @param	DefaultDrawTarget Whether to add the camera to the list of default draw targets. If false, 
+	 * @param	newCamera         The camera you want to add.
+	 * @param	defaultDrawTarget Whether to add the camera to the list of default draw targets. If false, 
 	 *                            `FlxBasics` will not render to it unless you add it to their `cameras` list.
 	 * @return	This FlxCamera instance.
 	 */
-	public function add<T:FlxCamera>(NewCamera:T, DefaultDrawTarget:Bool = true):T
+	public function add<T:FlxCamera>(newCamera:T, defaultDrawTarget:Bool = true):T
 	{
 		FlxG.game._cameraContainer.addChild(newCamera.flashSprite);
 
-		list.push(NewCamera);
-		if (DefaultDrawTarget)
-			defaults.push(NewCamera);
+		list.push(newCamera);
+		if (defaultDrawTarget)
+			defaults.push(newCamera);
 		
-		NewCamera.ID = list.length - 1;
-		cameraAdded.dispatch(NewCamera);
-		return NewCamera;
+		newCamera.ID = list.length - 1;
+		cameraAdded.dispatch(newCamera);
+		return newCamera;
 	}
 	
 	/**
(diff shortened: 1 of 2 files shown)


Geokureli:
What causes this to be a breaking change?

flixel/system/frontEnds/CameraFrontEnd.hx
```diff
@@ -57,22 +57,22 @@ class CameraFrontEnd
...
-		FlxG.game.addChildAt(NewCamera.flashSprite, FlxG.game.getChildIndex(FlxG.game._inputContainer));
-		
-		list.push(NewCamera);
-		if (DefaultDrawTarget)
-			defaults.push(NewCamera);
+		FlxG.game._cameraContainer.addChild(newCamera.flashSprite);
+
+		list.push(newCamera);
+		if (defaultDrawTarget)
+			defaults.push(newCamera);
```
Geokureli, line 71: Why the change from `addChildAt` to `addChild`? wouldn't this add new cameras to show behind old ones, rather than in front?

Geokureli:
> What I was trying to say is that game filters used to apply to the whole screen (debugger, mouse, etc.). With this, they'll just apply to all the cameras. This is technically a difference with how they work (breaking change?).

I'm also unsure as to whether this is "breaking", but I do think we should continue to apply the filters to the mouse. I know I personally rely on this behavior for pixel-perfect behavior.

> However in this case I think it can slide through. I don't think game filters ever worked properly and this fixes them and also makes them work more like intended (by leaving the debugger and other OpenFL sprites unaffected)

Saying it didn't "ever work properly" is a bit misleading, most of the time it worked entirely as expected, and in a few small cases it behaved weird if you needed the size to be exact.

Getting back to the original problem:
> In short, Flixel used to add the cameras' OpenFL Sprites to the FlxGame instance along with the mouse, debugger, sound tray, etc. Because of this, game filters did not report accurate sizes because they included things outside of the game view.

Is there a way to force the width and height of the game? is that what https://github.com/openfl/openfl/pull/2797 is trying to do?

I tried adding this to the sample in #2258 and it seems to work
```hx
FlxG.game.scrollRect = new openfl.geom.Rectangle(0, 0, FlxG.stage.stageWidth, FlxG.stage.stageHeight);
```

Geokureli:
I tested the scrollRect fix on the [FlxBunnyMark demo](https://haxeflixel.com/demos/FlxBunnymark/). I didn't notice any difference, but not sure if more FlxSprites would ever impact this. If you have any ideas of how I should test this, let me know, otherwise I'll make a new PR with that fix, soon.

PR #3481, opened 2025-09-16
FlxCamera: Honor filter changes when `active = false`

### LINKED ISSUE #3470: Camera filters not updating if camera has active set to false
Quick description, if you push a filter to camera.filters while said camera isn't active, the filter just won't register

DetectiveBaldi:
Closes #3470

The simplest solution, the issue with the alternative proposal is that `filters.push` would no longer function, which would break user code


Geokureli:
> The simplest solution, the issue with the alternative proposal is that filters.push would no longer function, which would break user code

This is not the case, the change I recommended would allow .push and is not a breaking change. But this is still the simpleast solution.

PR #3485, opened 2025-09-17
add unit tests for #3436

Geokureli:
Test calling destroy and removeAll in a dispatched signal


Geokureli:
@DetectiveBaldi, here's how I run unit tests in my my personal project: https://github.com/Geokureli/flixel-tests/blob/main/.vscode/settings.json#L10-L20

Comments about the "og book" and "new book" are referring to my macbooks, which have different relative paths to the flixel source

PR #3488, opened 2025-09-18
Breaking Change: Rendering cleanup

MaybeMaru:
For a while, HaxeFlixel has had remnants in it's rendering code of different approaches to rendering, deprecated openfl behaviour or features, among other things. This PR is meant as a cleanup of all of that unused/broken behaviour to have clean and straight to the point rendering code.

Its important to note that this PR contains breaking changes, therefore it should only be merged in a major version change, along with all the necessary fixes to external libraries like flixel-addons.

Here's a list of all the changes done:

* Replaced buffer clearing instances of ``splice`` with ``resize``.
* Removed unused ``blending`` and ``blendToInt``, as they're remnants from before openfl supported blends in the graphics api.
* BREAKING CHANGE: Removed ``hasColorOffsets`` from draw items. Even though this function was passed around it was never actually used with the original purpose of reducing the number of variables needed to push to the graphics vertex shader. I tried implementing it myself but it ended up being overcomplicated and only slowed down performance with all the extra checks it requires. Now its unified with the ``colored`` variable.
* Implemented FlxStrip's ``repeat`` value to tile rendering. If it's turned off it'll use the same wrap method as normal quad rendering.
* Brought some FlxDrawTrianglesItem code up to date with the methods used in FlxDrawQuadsItem.
* Reduced the number of checks in the graphics shader by changing slightly around the order of code.
* Reduced the number of checks in colored render items by first checking if the transform exists before starting the for loop. So it doesnt need to get checked for each vertex of the draw call. This also allows Haxe to inline the array pushes in FlxDrawQuadsItem, skipping the for loop completely.

Making this PR a draft first to make sure it works on all targets, so far I've tested desktop and web.
Feedback is appreciated!

### DIFF
--- a/flixel/FlxStrip.hx
+++ b/flixel/FlxStrip.hx
@@ -32,7 +32,7 @@ class FlxStrip extends FlxSprite
 
 	public var colors:DrawData<Int> = new DrawData<Int>();
 
-	public var repeat:Bool = false;
+	public var repeat:Bool = true;
 
 	override public function destroy():Void
 	{
--- a/flixel/graphics/tile/FlxDrawQuadsItem.hx
+++ b/flixel/graphics/tile/FlxDrawQuadsItem.hx
@@ -35,11 +35,11 @@ class FlxDrawQuadsItem extends FlxDrawBaseItem<FlxDrawQuadsItem>
 		super.reset();
 		rects.length = 0;
 		transforms.length = 0;
-		alphas.splice(0, alphas.length);
+		alphas.resize(0);
 		if (colorMultipliers != null)
-			colorMultipliers.splice(0, colorMultipliers.length);
+			colorMultipliers.resize(0);
 		if (colorOffsets != null)
-			colorOffsets.splice(0, colorOffsets.length);
+			colorOffsets.resize(0);
 	}
 
 	override public function dispose():Void
@@ -71,7 +71,7 @@ class FlxDrawQuadsItem extends FlxDrawBaseItem<FlxDrawQuadsItem>
 		for (i in 0...VERTICES_PER_QUAD)
 			alphas.push(alphaMultiplier);
 
-		if (colored || hasColorOffsets)
+		if (colored)
 		{
 			if (colorMultipliers == null)
 				colorMultipliers = [];
@@ -79,32 +79,28 @@ class FlxDrawQuadsItem extends FlxDrawBaseItem<FlxDrawQuadsItem>
 			if (colorOffsets == null)
 				colorOffsets = [];
 
-			for (i in 0...VERTICES_PER_QUAD)
+			if (transform != null)
 			{
-				if (transform != null)
+				for (i in 0...VERTICES_PER_QUAD)
 				{
 					colorMultipliers.push(transform.redMultiplier);
 					colorMultipliers.push(transform.greenMultiplier);
 					colorMultipliers.push(transform.blueMultiplier);
+					colorMultipliers.push(1);
 
 					colorOffsets.push(transform.redOffset);
 					colorOffsets.push(transform.greenOffset);
 					colorOffsets.push(transform.blueOffset);
 					colorOffsets.push(transform.alphaOffset);
 				}
-				else
+			}
+			else
+			{
+				for (i in 0...(VERTICES_PER_QUAD * 4))
 				{
 					colorMultipliers.push(1);
-					colorMultipliers.push(1);
-					colorMultipliers.push(1);
-
-					colorOffsets.push(0);
-					colorOffsets.push(0);
-					colorOffsets.push(0);
 					colorOffsets.push(0);
 				}
-
-				colorMultipliers.push(1);
 			}
 		}
 	}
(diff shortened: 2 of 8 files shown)


flixel/FlxStrip.hx
```diff
@@ -32,7 +32,7 @@ class FlxStrip extends FlxSprite
 
 	public var colors:DrawData<Int> = new DrawData<Int>();
 
-	public var repeat:Bool = false;
+	public var repeat:Bool = true;
```
Geokureli, line 35: I don't like change the default value of args, even if it is in a major release there's no good way to broadcast this to devs and they end up having either unknown bugs, or hard to track down bugs in their game.

flixel/graphics/tile/FlxDrawQuadsItem.hx
```diff
@@ -114,24 +110,28 @@ class FlxDrawQuadsItem extends FlxDrawBaseItem<FlxDrawQuadsItem>
...
 			throw 'Attempted to render an invalid FlxDrawItem, did you destroy a cached sprite?';
-		
+
 		final shader = shader != null ? shader : graphics.shader;
 		shader.bitmap.input = graphics.bitmap;
 		shader.bitmap.filter = (camera.antialiasing || antialiasing) ? LINEAR : NEAREST;
 		shader.alpha.value = alphas;
 
-		if (colored || hasColorOffsets)
+		if (colored)
```
Geokureli, line 123: I wanna say I tried this before and got unexpected issues in some cases. I'll try and find it.

flixel/graphics/tile/FlxGraphicsShader.hx
```diff
@@ -28,30 +28,20 @@ class FlxGraphicsShader extends GraphicsShader
...
 		{
 			vec4 color = texture2D(bitmap, coord);
-			if (!(hasTransform || openfl_HasColorTransform))
+            if (color.a == 0.0)
 				return color;
 			
-			if (color.a == 0.0)
-				return vec4(0.0, 0.0, 0.0, 0.0);
+            if (!(hasColorTransform || openfl_HasColorTransform))
+				return color * openfl_Alphav;
```
Geokureli, line 39: Why?

Geokureli (commented):
I skimmed over and found a couple concerns, so far. In general I'd really prefer separate changes in separate PRs, so I can run each one through the gauntlet, especially the innocuous little style changes.

There's a lot that can be improved with flixel's rendering and if it takes breaking changes or a bunch of sweeping changes in orchestration, so be it, but It's important that for every individual change to flixel, we think about unit testing, or creating ways to visually verify that this and future changes work in all expected edge cases.

flixel/graphics/tile/FlxGraphicsShader.hx
MaybeMaru: This part simplifies the number of checks in the shader. The old ``if (!(hasTransform || openfl_HasColorTransform))`` part never had any effect because ``hasTransform`` was always set to true. Plus the alpha check should be done early since if its fully transparent it would never get affected by the color transform either way.

flixel/graphics/tile/FlxDrawQuadsItem.hx
MaybeMaru: Please do let me know if it causes any issues, I've tried in many of my projects and its not causing any problems on my part.

MaybeMaru:
> I skimmed over and found a couple concerns, so far. In general I'd really prefer separate changes in separate PRs, so I can run each one through the gauntlet, especially the innocuous little style changes.
> 
> There's a lot that can be improved with flixel's rendering and if it takes breaking changes or a bunch of sweeping changes in orchestration, so be it, but It's important that for every individual change to flixel, we think about unit testing, or creating ways to visually verify that this and future changes work in all expected edge cases.

I've been thinking about it for a bit and I do think I could divide some of the changes to seperate PRs to be easier to merge and pin point. I also rethought the idea of the hasColorOffsets breaking change, while internally I think it should be removed as it causes no issues, I may add it back to startQuadBatch and startTrianglesBatch just for compatibility sake.

flixel/graphics/tile/FlxGraphicsShader.hx
MaybeMaru: Yep, using NVIDIA Nsight Graphics I can get some detailed information about the draw time required, both on CPU and GPU of openfl/flixel draw calls. For my testing I used the FlxBunnyMark demo for my testing with 20000 bunnies, which fill the screen up with a lot of pixels to be passed through the shader.
image
On my testing, with the graphics shader code, the bunnies take around 2.98 GPU ms to be rendered on screen.
image

In comparison, with the new graphics shader code, the bunnies take around 2.74 GPU ms to be rendered on screen.
image

Its not a major performance improvement since its just the removal of one if check, but its an improvement nonetheless.

ACrazyTown:
Just wanna add that I made #3467 a bit ago, which would check off one of the things from this list (remove unused blending variable and blendToInt() method)

PR #3497, opened 2025-09-28
Patch Sprite Color Batching and Default Sprite Color

RapperGF:
Currently, in HaxeFlixel, attempting to batch multiple FlxSprites with color transforms causes batching to break.
If any sprite has a color that is not pure white, it will batch. This causes sprites to be drawn in separate colored batches, even if the color is set.

This PR changes the default color of sprites to `FlxColor.TRANSPARENT` (effectively zero tinting) while still respecting all colors, including `FlxColor.WHITE`

With this change:
- Sprites set to `FlxColor.WHITE` will be batched together without altering their original colors.
- Sprites without a color (or with `FlxColor.TRANSPARENT`) will render with no tint, restoring previous behavior.
---
### Previous Behavior: 
image

### New Behavior: 
image

---
Example project: 

```hx
package;

import flixel.FlxG;
import flixel.FlxSprite;
import flixel.util.FlxColor;

class ColoringTest extends flixel.FlxState {

    var grp:FlxSpriteGroup;

    override public function create() {
        grp = new FlxSpriteGroup();

        var sprW = 16;
        var sprScale = 5;
        var spacing = sprW * sprScale;

        var screenW = FlxG.width; 
        var spritesPerRow = Math.floor(screenW / spacing);

        var totalSprites = 60;

        for (i in 0...totalSprites) {
            var row = Math.floor(i / spritesPerRow);
            var col = i % spritesPerRow;

            var spr = new FlxSprite(col * spacing, 20 + row * spacing);
            spr.scale.set(sprScale, sprScale);

            if (i == 0) {
                // First sprite is always white
                spr.color = FlxColor.WHITE;
            } else {
                // 20% chance to be white, otherwise random color
                if (Math.random() < 0.2) {
                    spr.color = FlxColor.WHITE;
                } else {
                    spr.color = FlxColor.fromRGB(
                        Std.int(Math.random() * 256),
                        Std.int(Math.random() * 256),
                        Std.int(Math.random() * 256)
                    );
                }
            }

            grp.add(spr);
        }

        grp.screenCenter();
        add(grp);
    }
}
```

### DIFF
--- a/flixel/FlxCamera.hx
+++ b/flixel/FlxCamera.hx
@@ -745,7 +745,7 @@ class FlxCamera extends FlxBasic
 		}
 	}
 
-	public function drawPixels(?frame:FlxFrame, ?pixels:BitmapData, matrix:FlxMatrix, ?transform:ColorTransform, ?blend:BlendMode, ?smoothing:Bool = false,
+	public function drawPixels(?frame:FlxFrame, ?pixels:BitmapData, matrix:FlxMatrix, ?transform:ColorTransform, ?colored:Bool, ?blend:BlendMode, ?smoothing:Bool = false,
 			?shader:FlxShader):Void
 	{
 		if (FlxG.renderBlit)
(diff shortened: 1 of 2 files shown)


LeonGamerPS1:
fire 🔥

Geokureli:
Moving off of the 6.1.2 release. changing the default color is a breaking change and a somewhat hacky fix, imo

rodney528:
Wouldn't this potentially break alpha related stuffs when a sprite is black?

PR #3499, opened 2025-10-02
FlxBasic, FlxGroup: Render Order Support

Electr0Gunner:
Haxeflixel has lacked an easy way to layer objects in the groups when drawing.
The current way was to use `add()` if you never moved it or `insert()` but this required knowing what object was where in the group.

By using a `renderOrder` variable, now added into FlxBasic, you can sort the draw call order.
In `FlxGroup.draw()` i created an array of indices for each member, and sort them based on the member's `renderObject` value.
This is the fastest and least expensive method i could think of, i made tje index array local which might've been a bad idea.
It also has a new `renderOrderMode` public variable, that uses bitwise operators to determine what type of sorting should happen. (Currently only renderOrder based is implemented)

Example:

```haxe
var sprite1 = new FlxSprite(0,0).makeGraphic(30, 30);
sprite1.renderOrder = 3;
var sprite2 = new FlxSprite(0,0).makeGraphic(30, 30);
sprite2.renderOrder = 1;
add(sprite1);
add(sprite2);
```

In this example, `sprite1` will always draw over `sprite2`

I also added a FlxDefine called `FLX_NO_RENDER_ORDER` to disable the render ordering incase projects don't need it.

I don't expect this to be approved, but i hope this feature would be put on some todo list for the future, as this is a pretty helpful feature.

## TESTS
I have tested the code on Windows 11 cpp target.

### DIFF
--- a/flixel/FlxBasic.hx
+++ b/flixel/FlxBasic.hx
@@ -38,13 +38,13 @@ class FlxBasic implements IFlxDestroyable
 	 * Controls whether `draw()` is automatically called by `FlxState`/`FlxGroup`.
 	 */
 	public var visible(default, set):Bool = true;
-
+	#if !FLX_NO_RENDER_ORDER
 	/**
 	 * An index that sorts the FlxBasic when `draw()` is called by `FlxState`/`FlxGroup`.
 	 * This produces easy layer sorting, not being needed to rely on `add()` or `insert()`.
 	 */
 	public var renderOrder:Int = 0;
-
+	#end
 	/**
 	 * Useful state for many game objects - "dead" (`!alive`) vs `alive`. `kill()` and
 	 * `revive()` both flip this switch (along with `exists`, but you can override that).
--- a/flixel/FlxBasic.hx
+++ b/flixel/FlxBasic.hx
@@ -40,6 +40,12 @@ class FlxBasic implements IFlxDestroyable
 	public var visible(default, set):Bool = true;
 
 	/**
+	 * An index that sorts the FlxBasic when `draw()` is called by `FlxState`/`FlxGroup`.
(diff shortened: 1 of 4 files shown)


moxie-coder:
Absolutely love this, the fact that HaxeFlixel had no Z index and relied on this wasn’t ideal, because you had no good way to change the draw order besides `insert` as far as I’m aware

Geokureli:
Even with these changes, I'm still in camp "no".

I don't think we should gate features like this behind compiler-flags either. This will increase complexity of testing and maintaining the codebase, and it will make it harder to communicate issues to one another. If you think its handy to hide a feature behind a compile flag, thats a good sign we probably shouldn't add it at all.

> i don't expect this to be approved at all. But i do want a feature like this to be considered as extending classes, adding macros, are not such easy ways to add sorting like this. With a define like i added this can be disabled like other features in flixel.

Make an issue stating the problem you are having, give actual snippets of code from actual projects, describing what you think is cumbersome, and why and once we identify an actual problem can start talking about solutions.

Electr0Gunner:
Compiler flags features are mostly an accidental fall through from my c++ methods, sorry bout that

For the extending part
Well
I am usually using alot of classes from base haxeflixel

Considering renderOrder is applied to flxbasic
This would allow all objects to be ordered with little to no issues

I guess using an interface could work


Thank you for looking into this.

Geokureli:
I feel like you're not fully reading my replies

> Make an issue stating the problem you are having, give actual snippets of code from actual projects, describing what you think is cumbersome, and why and once we identify an actual problem can start talking about solutions. Side note, but extending classes is actually a decent solution here, and I would like to know why you don't think so.

PR #3504, opened 2025-10-06
`FlxTilemap`: Fix debug draw when zoomed

### LINKED ISSUE #3502: Tilemap debug draw stops drawing rects too early
Happens on dev branch. It seems to only happen to tilemaps, as far as I can tell. Debug rects will stop rendering despite the fact that they're on screen, and it gets worse the more the camera is zoomed in.

https://github.com/user-attachments/assets/eed17f27-9280-4a20-822d-f1861e4971ee

Geokureli:
Fixes #3502 

@ACrazyTown does this fix your issue?


ACrazyTown:
> @ACrazyTown does this fix your issue?

It does, thanks!

PR #3511, opened 2025-10-12
FlxG.Assets: Check any sound extension before appending in getSoundAddExt

ACrazyTown:
Otherwise, passing in something like `sound.m4a` would result in `sound.m4a.ogg` despite the fact that `m4a` is a valid sound extension.

### DIFF
--- a/flixel/system/FlxAssets.hx
+++ b/flixel/system/FlxAssets.hx
@@ -1,5 +1,6 @@
 package flixel.system;
 
+import haxe.io.Path;
 import haxe.macro.Expr;
 #if !macro
 import flixel.FlxG;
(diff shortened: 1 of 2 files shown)


moxie-coder:
I’m pretty sure this broke something in the past, but I don’t remember. Also this seems useful for if (and when) Lime/OpenFL ever supports MP3 on desktop (since the patents expired) and other audio formats that SDL2 supports which I see is quite an handful

ACrazyTown:
> I’m pretty sure this broke something in the past, but I don’t remember. Also this seems useful for if (and when) Lime/OpenFL ever supports MP3 on desktop (since the patents expired) and other audio formats that SDL2 supports which I see is quite an handful

This won't have any difference when loading an MP3. What the code does is add a sound extension, if you haven't already added it yourself. The issue is that it assumes the extension will be a mp3/ogg/wav, so trying to pass in anything other than that (eg. m4a or opus) results in a duplicate file extension

Geokureli:
Thanks!

Geokureli:
> I’m pretty sure this broke something in the past, but I don’t remember. Also this seems useful for if (and when) Lime/OpenFL ever supports MP3 on desktop (since the patents expired) and other audio formats that SDL2 supports which I see is quite an handful

Not entirely sure, I'm down to find out, especially now that there's more tools for auto-extensions in FlxSounds

PR #3512, opened 2025-10-12
`FlxGraphicsShader`: Remove redundant body glsl

MaybeMaru:
This issue stems from the code changes found when the implementation of ``camera.color`` was fixed. (https://github.com/HaxeFlixel/flixel/pull/3326)
For whatever reason, the original code for the fragment source was replaced for a fragment body insert, most likely due to an oversight. ``glFragmentBody`` does not replace the fragment body code, it only adds onto it, so instead of directly using ``flixel_texture2D`` the shader was running first the openfl fragment body code and then the flixel fragment body code. As seen here in this picture taken from a debugger of a flixel game compiled build:

image

Here's how the shader looks like after these changes:

image

This fix has been tested and ``camera.color`` keeps working correctly.

### DIFF
--- a/flixel/graphics/tile/FlxGraphicsShader.hx
+++ b/flixel/graphics/tile/FlxGraphicsShader.hx
@@ -61,7 +61,7 @@ class FlxGraphicsShader extends GraphicsShader
 		{
 			gl_FragColor = flixel_texture2D(bitmap, openfl_TextureCoordv);
 		}
-	")
+	", true)
 	public function new()
 	{
 		super();
--- a/flixel/graphics/tile/FlxGraphicsShader.hx
+++ b/flixel/graphics/tile/FlxGraphicsShader.hx
@@ -54,9 +54,14 @@ class FlxGraphicsShader extends GraphicsShader
 			return color * openfl_Alphav;
 		}
 	", true)
-	@:glFragmentBody("
-		gl_FragColor = flixel_texture2D(bitmap, openfl_TextureCoordv);
-	", true)
+	@:glFragmentSource("
+		#pragma header
+
+		void main(void)
+		{
+			gl_FragColor = flixel_texture2D(bitmap, openfl_TextureCoordv);
+		}
+	")
 	public function new()
 	{
 		super();


Geokureli:
Please never use [screenshots of text or error messages](https://meta.stackoverflow.com/questions/285551/why-should-i-not-upload-images-of-code-data-errors)

SrtHero278:
> Please never use [screenshots of text or error messages](https://meta.stackoverflow.com/questions/285551/why-should-i-not-upload-images-of-code-data-errors)

i can understand your concern with that a little bit, but a portion of those points are argueably dismissable i think its to emphasize that its on nsight.
here have a video to further prove that.

https://github.com/user-attachments/assets/80d809b5-00d0-4b28-ae35-cec1f6e24a9c

SrtHero278:
my apologies that it doesnt embed right

Geokureli:
> my apologies that it doesnt embed right

Note: you need to put an empty line before video embeds or they won't show up, you can see my edit to your message.

> I can understand your concern with that a little bit, but a portion of those points are argueably dismissable i think its to emphasize that its on nsight.

I don't see how that is relevant to my request

superpowers04:
> This issue stems from the code changes found when the impleme...

+1

flixel/graphics/tile/FlxGraphicsShader.hx
```diff
@@ -54,9 +54,14 @@ class FlxGraphicsShader extends GraphicsShader
...
-		gl_FragColor = flixel_texture2D(bitmap, openfl_TextureCoordv);
-	", true)
+	@:glFragmentSource("
+		#pragma header
+
+		void main(void)
+		{
+			gl_FragColor = flixel_texture2D(bitmap, openfl_TextureCoordv);
+		}
+	")
```
ACrazyTown: Shouldn't this be `", true)` as to not revert the changes made in https://github.com/HaxeFlixel/flixel/pull/3347?

MaybeMaru: fixed

Vortex2Oblivion:
I tested on FlxBunnyMark and there's definitely performance improvements

Geokureli:
> For whatever reason, the original code for the fragment source was replaced for a fragment body insert, most likely due to an oversight.

> `glFragmentBody` does not replace the fragment body code, it only adds onto it

Yeah, I assumed otherwise. My goal was to allow people to use `@glFragmentBody` in their FlxGraphicShader extending classes. Is that something that can be done here? If we added [#pragma body](https://github.com/Geokureli/openfl/blob/pretty-shader-info/src/openfl/display/GraphicsShader.hx#L99) to this glFragmentSource will it add openfl's body as well? If not, should we account for emscripten [the way openfl does](https://github.com/Geokureli/openfl/blob/pretty-shader-info/src/openfl/display/GraphicsShader.hx#L91)? I'm not sure whether flixel supports emscripten (If it does it would have been broken by the implementation prior to 3326, so it's unlikely)

In a future PR we may want to consider bypassing openfl.display.GraphicsShader and extending [openfl.display.Shader](https://github.com/Geokureli/openfl/blob/pretty-shader-info/src/openfl/display/Shader.hx) directly. For now, this solution seems fine, unless we CAN add `#pragma body`

Geokureli:
Thanks!

If we can add a #pragma body here, we can do that in a future pr, or go with the long term fix mentioned above

PR #3514, opened 2025-10-18
Use set in FlxPoint (Reopen PR #3445)

dtwotwo:
This PR (#3445) has been reopened at the request of the original contributor, since it was accidentally closed earlier.
@gimmesoda personally asked me to reopen it and can confirm this action.

No further changes were made — this is purely to restore the original discussion and review context.

### DIFF
--- a/flixel/math/FlxPoint.hx
+++ b/flixel/math/FlxPoint.hx
@@ -7,31 +7,31 @@ import openfl.geom.Point;
 
 /**
  * 2-dimensional point class
- * 
+ *
  * ## Pooling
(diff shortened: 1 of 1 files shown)


Geokureli:
The XY callback is not being called twice because of redundant getters, but because of a typo, here:https://github.com/HaxeFlixel/flixel/blob/dev/flixel/math/FlxPoint.hx#L1744

The callbackY arg is never used, causing the xy callback to be called whenever the y callback should have been called.

I'm unable to make changes to your branches so I made a new one: https://github.com/HaxeFlixel/flixel/pull/3524 (also added various other things)

PR #3521, opened 2025-11-15
Make sure onFocusLost gets called on substates.

### LINKED ISSUE #3520: FlxSubState.onFocus and FlxSubState.onFocusLost don't get called
As the title says, only the main active state receives a call to `onFocus` and `onFocusLost`, and any open substates of that state don't get this function called.

Should be a quick fix, will submit a PR in a moment.

EliteMasterEric:
It took me longer to make a commit that wasn't filled with whitespace changes automatically made by HXFormat than it did to write the actual PR.

Resolves #3520.

### DIFF
--- a/flixel/FlxState.hx
+++ b/flixel/FlxState.hx
@@ -202,13 +202,17 @@ class FlxState extends FlxContainer
 	 * This method is called after the game loses focus.
 	 * Can be useful for third party libraries, such as tweening engines.
 	 */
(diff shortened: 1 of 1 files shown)


Geokureli:
I've never been able to observe a focus event, this is relevant  to some previous issues, let me find and link them. 

Also this doesn't fit the HF format requirements, remove the cuddled brackets and split these particular ifs into two lines

PR #3522, opened 2025-11-16
FlxG: query maxTextureSize when window is created

ACrazyTown:
This fixes a crash in a project where threaded caching of assets is done immediately upon load. Because it's on a thread, OpenGL freaks out and crashes, though it's interesting to note that the crash only happened on macOS.

I think it's best to just let Flixel query this once when the window is created and then it'll be cached

### DIFF
--- a/flixel/FlxG.hx
+++ b/flixel/FlxG.hx
@@ -539,6 +539,10 @@ class FlxG
 		FlxG.height = height;
 
 		initRenderMethod();
+		#if FLX_OPENGL_AVAILABLE
+		// Query once when window is created and cache for later
+		bitmap.get_maxTextureSize();
+		#end
 
 		FlxG.initialWidth = width;
 		FlxG.initialHeight = height;
--- a/flixel/system/frontEnds/BitmapFrontEnd.hx
+++ b/flixel/system/frontEnds/BitmapFrontEnd.hx
@@ -343,6 +343,8 @@ class BitmapFrontEnd
 
 	#if FLX_OPENGL_AVAILABLE
 	static var _maxTextureSize = -1;
+
+	@:allow(flixel.FlxG)
 	function get_maxTextureSize():Int
 	{
 		if (_maxTextureSize < 0 && FlxG.stage.window.context.attributes.hardware)


Geokureli:
Can you share a test case that causes the error?

ACrazyTown:
> Can you share a test case that causes the error?

```Haxe
{
	{
		super.create();

		// by uncommenting the trace here the crash no longer happens
		// the PR does something similar by querying maxTextureSize asap
		// when the window is created

		// trace(FlxG.bitmap.maxTextureSize);

		var f = new lime.app.Future(() -> {
			flixel.FlxG.bitmap.create(1000, 1000, 0);
		}, true);
	}
}
```

```
SIGNAL 11
lime._internal.backend.native.NativeOpenGLRenderContext.getInteger(lime/_internal/backend/native/NativeOpenGLRenderContext.hx:1705)
lime._internal.backend.native.NativeOpenGLRenderContext.getParameter(lime/_internal/backend/native/NativeOpenGLRenderContext.hx:1825)
flixel.system.frontEnds.BitmapFrontEnd.get_maxTextureSize(flixel/system/frontEnds/BitmapFrontEnd.hx:349)
flixel.graphics.FlxGraphic.set_bitmap(flixel/graphics/FlxGraphic.hx:612)
flixel.graphics.$FlxGraphic.__constructor__(flixel/graphics/FlxGraphic.hx:410)
flixel.graphics.$FlxGraphic.createGraphic(flixel/graphics/FlxGraphic.hx:252)
flixel.graphics.$FlxGraphic.fromRectangle(flixel/graphics/FlxGraphic.hx:220)
flixel.system.frontEnds.BitmapFrontEnd.create(flixel/system/frontEnds/BitmapFrontEnd.hx:86)
PlayState.~create.0(PlayState.hx:16)
lime.app.$FutureWork.threadPool_doWork(lime/app/Future.hx:349)
lime.app.$FutureWork.~run.0(lime/app/Module.hx:0)
lime.system.$ThreadPool.__executeThread(lime/system/WorkOutput.hx:251)
sys.thread._Thread.$HaxeThread.~create.0(/usr/local/lib/haxe/std/hl/_std/sys/thread/Thread.hx:152)
```

Like I mentioned above I only started getting this crash when testing on a Mac. Either way I think this falls into "undefined behavior" as OpenGL is meant to be accessed from a single thread only, the Windows OpenGL implementation might just be getting lucky

Geokureli:
Thanks!

PR #3524, opened 2025-12-01
FlxCallbackPoint: Fix duplicate calls and call xy when x or y is set directly

Geokureli:
Alternative to https://github.com/HaxeFlixel/flixel/pull/3514 (Most code was copied from this)

Changing a point's x should trigger the x and xy callback, once (if non-null), same with y, any method that changes both should call all 3 callbacks, once.

- The y callback arg was never being used, so setting xy called the xy callback twice.
- Every operation function now calls set() rather than setting x and y individually.
- Setting x and y individually now calls the xy callback as well.
- Added a unit test to verify the number of sets for each method with each number of callbacks

### DIFF
--- a/flixel/math/FlxPoint.hx
+++ b/flixel/math/FlxPoint.hx
@@ -539,8 +539,8 @@ import openfl.geom.Point;
 	/**
 	 * Helper function, just copies the values from this point to the specified point.
 	 *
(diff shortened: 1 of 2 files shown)


Geokureli:
CI is failing because of https://github.com/HaxeFoundation/hscript/issues/155

PR #3525, opened 2025-12-01
Revert "FlxG: query maxTextureSize when window is created"

Geokureli:
Reverts HaxeFlixel/flixel#3522

I'm getting errors on my html5 tests

### DIFF
--- a/flixel/FlxG.hx
+++ b/flixel/FlxG.hx
@@ -539,10 +539,6 @@ class FlxG
 		FlxG.height = height;
 
 		initRenderMethod();
-		#if FLX_OPENGL_AVAILABLE
-		// Query once when window is created and cache for later
-		bitmap.get_maxTextureSize();
-		#end
 
 		FlxG.initialWidth = width;
 		FlxG.initialHeight = height;
--- a/flixel/system/frontEnds/BitmapFrontEnd.hx
+++ b/flixel/system/frontEnds/BitmapFrontEnd.hx
@@ -343,8 +343,6 @@ class BitmapFrontEnd
 
 	#if FLX_OPENGL_AVAILABLE
 	static var _maxTextureSize = -1;
-
-	@:allow(flixel.FlxG)
 	function get_maxTextureSize():Int
 	{
 		if (_maxTextureSize < 0 && FlxG.stage.window.context.attributes.hardware)


ACrazyTown:
What error were you getting?

Geokureli:
GL is null, [here](http://github.com/HaxeFlixel/flixel/blob/dev/flixel/system/frontEnds/BitmapFrontEnd.hx#L351).

```
Uncaught TypeError TypeError: Cannot read properties of null (reading 'getParameter')
    at get_maxTextureSize (/Users/booboo/Documents/haxe/lib/flixel/flixel/flixel/system/frontEnds/BitmapFrontEnd.hx:351:27)
    at flixel_FlxG.init (/Users/booboo/Documents/haxe/lib/flixel/flixel/flixel/FlxG.hx:544:3)
    at flixel_FlxGame (/Users/booboo/Documents/haxe/lib/flixel/flixel/flixel/FlxGame.hx:260:3)
    at createTest (/Users/booboo/Documents/haxe/lib/flixel/flixel-tests/Source/Main.hx:247:22)
    at Main (/Users/booboo/Documents/haxe/lib/flixel/flixel-tests/Source/Main.hx:23:13)
    at DocumentClass (/Users/booboo/Documents/haxe/lib/flixel/flixel-tests/export/html5/haxe/ApplicationMain.hx:343:6)
    at ApplicationMain.start (/Users/booboo/Documents/haxe/lib/flixel/flixel-tests/export/html5/haxe/ApplicationMain.hx:221:7)
    at tmp (/Users/booboo/Documents/haxe/lib/flixel/flixel-tests/export/html5/haxe/ApplicationMain.hx:121:28)
    at dispatch (/usr/local/lib/haxe/lib/lime/8,3,0/src/lime/_internal/macros/EventMacro.hx:91:17)
    at display_onUnload (/Users/booboo/Documents/haxe/lib/openfl/openfl/src/openfl/display/Preloader.hx:101:5)
    at __dispatchEvent (/Users/booboo/Documents/haxe/lib/openfl/openfl/src/openfl/events/EventDispatcher.hx:458:6)
    at __dispatchEvent (/Users/booboo/Documents/haxe/lib/openfl/openfl/src/openfl/display/DisplayObject.hx:1516:16)
    at __dispatchWithCapture (/Users/booboo/Documents/haxe/lib/openfl/openfl/src/openfl/display/DisplayObject.hx:1577:10)
    at dispatchEvent (/Users/booboo/Documents/haxe/lib/openfl/openfl/src/openfl/display/DisplayObject.hx:1212:10)
    at onLoaded (/Users/booboo/Documents/haxe/lib/flixel/flixel/flixel/system/FlxBasePreloader.hx:446:3)
```
Could've sworn I tested your branch on html5 but it seems I didn't. Not sure if there's a unit test that can catch this.

ACrazyTown:
I will take a proper look when I'm able to, but is it possible this was running in a browser that doesn't support WebGL? It might also need a check to see if the render context is hardware accelerated

Geokureli:
After more sleuthing I just noticed that my html5 builds seem to use renderBlit, I have no idea why, I'm not using `<window hardware="false"/>`, but the stage's context is "canvas"

ACrazyTown:
Ah yep checks out. Can I redo the PR including this fix?

I’d also check if your browser has hardware acceleration enabled. WebGL
won’t work without it

Geokureli:
After reverting I'm still seeing the issue, but this immediately started happening after pulling this change, so I'm not sure whats going on

Edit: Tried this on a different project and it works fine, even when I manually enable render blit

ACrazyTown:
Seems like something strange is going on. The if statement I mentioned actually shouldn't be necessary because it's already there, in `get_maxTextureSize()`:
https://github.com/HaxeFlixel/flixel/blob/687c1ccb631a75cb35dec743db917fcb186a654f/flixel/system/frontEnds/BitmapFrontEnd.hx#L348-L349

So it seems that `FlxG.stage.window.context.attributes.hardware` is falsely reporting true, even though the context is using Canvas and not WebGL. This might be an issue in OpenFL/Lime

EDIT: https://github.com/openfl/lime/pull/2002 should fix the if statement falsely passing. I think we could revert the revert now 😅

Geokureli:
Thanks for that! Is there something we can do in flixel to avoid this crash. Lime releases are pretty infrequent so I'd like to have something working for the upcoming flixel 6.1.2 release

ACrazyTown:
> Thanks for that! Is there something we can do in flixel to avoid this crash. Lime releases are pretty infrequent so I'd like to have something working for the upcoming flixel 6.1.2 release

The if check in `get_maxTextureSize()` could be changed to `if (FlxG.stage.window.context.attributes.hardware #if html5 && FlxG.stage.window.context.type != CANVAS #end)` temporarily. The patch will likely be included with Lime 8.3.1, whenever that releases.

Geokureli:
I made this change, locally, it worked. I can use yours, if you think one is better than the other. 
```hx
	#if FLX_OPENGL_AVAILABLE
	static var _maxTextureSize = -1;
	function get_maxTextureSize():Int
	{
		if (_maxTextureSize < 0)
		{
			final hardware = FlxG.stage.window.context.attributes.hardware
				// Prevents lime error (fixed here: https://github.com/openfl/lime/pull/2002)
				&& #if (html5 && lime <= "8.3.0") FlxG.renderTile #else true #end;
			
			if (hardware)
				_maxTextureSize = cast GL.getParameter(GL.MAX_TEXTURE_SIZE);
			else
				_maxTextureSize = 0;
		}
		
		return _maxTextureSize;
	}
	#end
```

Small note: setting `_maxTextureSize = 0;` prevents it from checking every time

I'll try the lime git, too, though sometimes using git lime on my mac is a hassle

ACrazyTown:
Oh, now that you mention it, I think it could just boil down to `if (FlxG.renderTile)` as renderTile is only true when OpenGL is used

Geokureli:
> Oh, now that you mention it, I think it could just boil down to `if (FlxG.renderTile)` as renderTile is only true when OpenGL is used

Meaning this?
```hx
	#if FLX_OPENGL_AVAILABLE
	static var _maxTextureSize = -1;
	function get_maxTextureSize():Int
	{
		if (_maxTextureSize < 0 && FlxG.renderTile)
			_maxTextureSize = cast GL.getParameter(GL.MAX_TEXTURE_SIZE);
		
		return _maxTextureSize;
	}
	#end
```

In the future I plan to make the global sub-object `FlxG.rendering.method` instead of `FlxG.renderMethod` and put stuff like `isGl`or something in there.

ACrazyTown:
> Meaning this?

Yep, seems good!

> In the future I plan to make the global sub-object `FlxG.rendering.method` instead of `FlxG.renderMethod` and put stuff like `isGl`or something in there.

Is there an issue for something like this where it could be discussed? Recently I was thinking about how Flixel handles rendering, and imo all the main rendering bits should be moved outside of FlxCamera and into a separate renderer class.

Geokureli:
> Is there an issue for something like this where it could be discussed? Recently I was thinking about how Flixel handles rendering, and imo all the main rendering bits should be moved outside of FlxCamera and into a separate renderer class.

https://github.com/HaxeFlixel/flixel/issues/3527

PR #3530, opened 2025-12-05
Optimization to avoid redraws when setting text to the same value.

EliteMasterEric:
By checking if `this.text == value` before redrawing the text, you save on redraws and improve performance a bit.

I did this for several other values on FlxText too.


Geokureli:
Please use the setting`"editor.formatOnSaveMode":"modifications"` to avoid full file reformats, or just disable your personal autoformatting style on flixel directories

Geokureli:
Fixed via #3532

PR #3531, opened 2025-12-07
FlxG: Add macro to determine VERSION

Geokureli:
I hate doing this manually

made using: https://gist.github.com/Geokureli/9d0ac4598aa595b79f43919850601701


Cheemsandfriends:
I have some concerns with this

Does this mean we have to compile with internet access at all times if we're downloading flixel from Haxelib?

Geokureli:
> I have some concerns with this
> 
> Does this mean we have to compile with internet access at all times if we're downloading flixel from Haxelib?

No, this works the same way as `#if (flixel < version("5.3.0"))` because the value of the flixel flag is its version. as for the git sha, that already existed in the old code

Issue #3527, opened 2025-12-02
Create RenderFrontEnd


Geokureli:
Move FlxG.renderMethod, renderBlit and renderTile there. Add other helpers like `isGL` or `drawMode` and various default rendering modes like pixel-perfect and smoothing

Issue #2915, opened 2023-09-20
Flixel rendering overhaul


ninjamuffin99:
https://haxeflixel.com/blog/08-openfl-5-compatibility-status-update/

Around 2017, Beeblrox posted some progress he made in redoing the way HaxeFlixel does things like shaders, and FlxStrip rework. By the title of that blog, that was on OpenFL 5. We are now on OpenFL 9! 

https://github.com/AustinEast/flixel/tree/Beeblerox-FlxMaterial-and-camera-scaling

Sometime around a year ago, I was getting Austin East to work on getting beeble's work running in modern flixel/haxe/openfl/lime

According to messages between me and austin:

> 2022/10/20
> ninjamuffin99: so where was the things with all the polygons / tris or whatev, thats with all the openGL rendering right?
> that includes like, making FlxStrip play nicer and all that?
> 
> austin: yeah all that high level API stuff is in :) like the stuff from the famous beeble blog post on the flixel website lol

and another conversation with austin about the project

> 
> okay so ive gone over everything and i want to break down all the big items and where theyre at. this will also be helpful for me since there is a lot going on here lol
> i originally started this as a port of beeble's branch, but things sorta evolved from there, based on conversations i had w him and looking at other renderers out there
> austineast — 05/29/2022 10:35 PM
> beeble's branch featured that fancy opengl renderer, but it was heavy tied to openfl, which has changed quite a bit since the branch was worked on and everything is basically broken. lol. at first i tried to fix it as-is, but after talking to beeble about it, he recommended rewriting it in a way that would be a bit closer to the metal (ie more opengl code through lime vs openfl) and pointed me to haxepunk's renderer, which ended up being a good resource
> austineast — 05/29/2022 10:42 PM
> i noticed haxepunk had a nice backend abstraction setup, which really helped keep the haxepunk codebase/renderer clean and easy to parse. its written in a way where any lime/openfl/nme specific code is separated out to a backends/ directory. and since i was in the process of planning out the flixel renderer rewrite, it seemed like a good idea to start doing that with flixel's codebase to help me really learn the internals and really future proof this new renderer
> austineast — 05/29/2022 11:06 PM
> with all that being said, heres the general statuses of the different pieces:
> 
> 1. merge beeble's branch into the latest flixel and get it compiling : complete (once it was compiling, i discovered that while the renderer abstraction was theoretically working well, openfl changes had really broken all of them hard. that lead me to the next task)
> 
[...]

Issue #1073, opened 2014-05-11
Consider separating concepts of screen from camera


JoeCreates:
Presently, FlxCamera represents both the position at which a world is being observed (FlxCamera.scroll x/y) and the rectangular screen on which the world is drawn. It might be worth considering splitting these up, having both a FlxScreen and FlxCamera.

The obvious drawback is that it adds a little more complexity, although in practice most of this complexity would be hidden.

Benefits would include:-
- Easily swap cameras while maintaining screen size and position, as cameras are simply attached to a screen.
- View the same camera from different screens.
- Camera position would actually be the camera position (more intuitive than having a "scroll" position
- Encourage better implementations of existing features such as shake and zoom, which currently manipulate the entire screen position and size, whereas it would be better to implement these in terms of the screens view of the world.

I think most stuff you already do with FlxG.camera(s) would stay the same. There would need to be some way of tracking the screens, though.

Additional considerations:-
- Screen would have to be responsible for actually using some properties of the camera, such as the cameras bounds, as the effect is dependent on the size of the screen.
- Attach multiple cameras to the same screen? Possibly useful for HUD layers?
##

PR #3539, opened 2025-12-23
Rework & separate renderers

ACrazyTown:
Here we go, first step towards a renderer overhaul. Looks like a pretty big and scary change but 99% of this is just moving stuff around.

Shout out goes to Beeblerox & Austin East, a lot of this is based on the [original renderer overhaul branch](https://github.com/AustinEast/flixel/tree/Beeblerox-FlxMaterial-and-camera-scaling), I'm just porting bits and pieces to modern Flixel. 

### General restructure
- Adds the abstract `FlxRenderer` and `FlxCameraView` classes, which are extended by different renderer implementations in `flixel.system.render`
  - `FlxCameraView` implements all per-camera functionality, such as the `draw[...]()` methods. All cameras now own a view instance, and drawing methods like `camera.drawPixels()` are now simply shortcuts to `camera.view.drawPixels()`
  - `FlxRenderer` implements and abstracts away some global rendering logic, such as adding/removing cameras and the start/end of the rendering loop. Access it via `FlxG.renderer`
- Moved some renderer-specific classes like the FlxDrawItems into their respective package (`flixel.system.render.quad`)

and more...? (More changes are being worked on in the background, but I'm not sure whether they'll end up part of this PR or another one, see: [FlxTexture](https://github.com/ACrazyTown/flixel/pull/2), [custom OpenGL renderer branch](https://github.com/ACrazyTown/flixel/tree/feat/opengl-renderer-3))


ACrazyTown:
I decided to deprecate all the internals and make them point to where they were moved, to avoid breaking anything.

Addons and UI will need to be updated to avoid warnings, I'll get to that in a bit. Addons specifically seems to have an issue because `FlxShaderMaskCamera` overrides `camera.fill()`, it'll need to be updated to also add `overload extern`, hopefully that's not a blocker

I think this is in a good enough of a state now for a review, so I'll undraft

flixel/FlxCamera.hx
```diff
@@ -1638,105 +1398,60 @@ class FlxCamera extends FlxBasic
...
+		view.fill(color, blendAlpha);
+	}
+
+	/**
+	 * Fill the camera with the specified color.
+	 *
+	 * @param   color        The color to fill with in `0xAARRGGBB` hex format.
+	 * @param   blendAlpha   Whether to blend the alpha value or just wipe the previous contents. Default is `true`.
+	 */
+	overload public inline extern function fill(color:FlxColor, blendAlpha:Bool = true):Void
```
Geokureli: Rather than overloading maybe we should define a new fillView() or use view.fill directly?

ACrazyTown: I guess that could work. I'm leaning more towards `fillView()`, I think having a convenience method in `FlxCamera` would be nice. What about all the other overloaded methods? Ideally I wanted to keep everything working as is, but I completely forgot to account that people might be overriding these methods

flixel/FlxCamera.hx
```diff
@@ -196,7 +220,17 @@ class FlxCamera extends FlxBasic
...
+	public var flashSprite(get, set):Sprite;
+	inline function set_flashSprite(value:Sprite):Sprite
+	{
+		var sprite = FlxG.renderTile ? viewQuad.flashSprite : viewBlit.flashSprite;
+		return sprite = value;
+	}
+	inline function get_flashSprite():Sprite
+	{
+		return FlxG.renderTile ? viewQuad.flashSprite : viewBlit.flashSprite;
+	}
```
ACrazyTown: Note to self, this can probably be simplified to just `view.display`

ACrazyTown:
Some more notes and thoughts. I'm currently playing around trying to implement an OpenGL renderer to really test the flexibility of this system.

### Overloading methods in FlxCameraView won't work:
For stuff like FlxMaterial and FlxTrianglesData, the function signatures of the core rendering functions need to be changed. To avoid breaking changes, I was just going to do this with overloads, but that didn't work out because overloads need to be inlined and you can't override inline functions. 

I got around this by deprecating drawPixels() and copyPixels() for draw() and copy() in FlxCameraView, but I couldn't do the same in FlxCamera because it inherits FlxBasic.draw(). I ended up overloading the camera's drawPixels() and copyPixels() to call the new methods in camera view. Pretty gross solution IMO but I can't think of a better way without a breaking change.

### Unifying renderBlit with other renderers
I did some work related to this in https://github.com/HaxeFlixel/flixel/pull/3539/changes/3cd97d9b4ea0ef5d4eb36849cf8809eacc02c80e, but there's still a couple places where we have to do things differently depending on the renderer, notably `drawPixels()` and `copyPixels()`:

https://github.com/HaxeFlixel/flixel/blob/c32ab91babccfff4515665db3358168af70496c6/flixel/FlxCamera.hx#L743-L744

When the blitting renderer is used, `pixels` will be used for the rendering, otherwise `frame` is used. I wonder if it's somehow possible to avoid this special behavior and just pass in one thing, without having to know what renderer is used. The user should just have to call `camera.drawPixels(...);` once, and the underlying implementation should take care of any special quirks.

### Isolating direct access to renderer
~~Flixel shouldn't access any renderer implementations directly from common code, it should be done through the renderer abstraction. First thing that comes to mind is that we need to abstract way [`maxTextureSize`](https://github.com/HaxeFlixel/flixel/blob/c32ab91babccfff4515665db3358168af70496c6/flixel/system/frontEnds/BitmapFrontEnd.hx#L344-L355)~~

~~EDIT: This could be done via the [RenderFrontEnd](https://github.com/HaxeFlixel/flixel/issues/3527) via some method like `FlxG.render.getMaxTextureSize()`~~

EDIT 2: This is no longer an issue with recent changes, see next comment

ACrazyTown:
Originally this was just a port of the renderer abstraction by Beeblerox that I ported over to modern Flixel. As I was playing around with it, I found that there were certain things I wasn't too fond of, specifically the fact that access to the renderer was only possible through cameras, so here goes an attempt at a V2.

`FlxCameraView` has been demoted, and no longer handles talking to the renderer. This is now done by  the global `FlxRenderer` (Accessible via `FlxG.renderer`) and its implementations. `FlxCameraView` is still around, though it now mainly just stores the various objects needed per-camera for rendering (e.g. the flash sprites and whatever).

Considering that the renderer is now global, I've also integrated some of the helpers mentioned in #3527 directly into `FlxRenderer`.

I also wonder if it'd be worth deprecating the drawing methods in `FlxCamera`, and pointing users who need advanced control over the renderer to use `FlxG.renderer` instead.

ACrazyTown:
I think this is now in a reviewable state. No clue why CI is failing tho

Geokureli:
> No clue why CI is failing tho

[flixel-addons FlxShaderMaskCamera.hx:110: characters 20-26](https://github.com/HaxeFlixel/flixel-addons/blob/dev/flixel/addons/display/FlxShaderMaskCamera.hx#L110): Field render is declared 'override' but doesn't override any field
[system.render.quad.FlxQuadRenderer.hx:96: characters 18-38](https://github.com/HaxeFlixel/flixel/pull/3539/files#diff-49903f9f4bf1dc25cce837910578c72846b7165465e9d4e9eb7aae4b2b598b40R96): Type not found : FlxDrawTrianglesItem
[system.render.quad.FlxQuadRenderer.hx:113: characters 18-38](https://github.com/HaxeFlixel/flixel/pull/3539/files#diff-49903f9f4bf1dc25cce837910578c72846b7165465e9d4e9eb7aae4b2b598b40R113): Type not found : FlxDrawTrianglesItem

Geokureli:
Do we need to remove the render method from FlxCamera?

Seems like we could call FlxG.render.begin(this) from FlxCamera.render

ACrazyTown:
> Do we need to remove the render method from FlxCamera?

Oops, this is my bad! I accidentally got rid of it while refactoring some of my changes. Brought it back now.

The CI confused me because I saw all the snippets compiling fine yet the action was failing. Turns out I missed that it also compiles the demos... Would it be possible to make the action stop as soon as there's a single failure?

Geokureli:
> Would it be possible to make the action stop as soon as there's a single failure?

Maybe that's what [this](https://github.com/HaxeFlixel/flixel/blob/dev/.github/workflows/main.yml#L17) does? 

We're still getting a hxSehException on cpp unit tests. Not sure what that means. Are you able to run/debug unit tests locally? May get a stack trace, that way

I'll start looking over this PR today, I've only been skimming, so far

ACrazyTown:
> We're still getting a hxSehException on cpp unit tests. Not sure what that means. Are you able to run/debug unit tests locally? May get a stack trace, that way

```
Uncaught exception: Null access .method
Called from flixel.graphics.frames.$FlxFrame.__constructor__(flixel/graphics/frames/FlxFrame.hx:162)
Called from flixel.text._FlxBitmapText.$ReusableFrame.__constructor__(flixel/text/FlxBitmapText.hx:1797)
Called from .init(flixel/text/FlxBitmapText.hx:329)
```

Seems like `FlxG.renderer` is null during unit tests... which is weird since it should be created immediately along with the `FlxGame`. Even weirder, commenting this single line makes the unit test boot up without crashes:
https://github.com/HaxeFlixel/flixel/blob/5f4dc29ea8e59278feb33245065e343c80bfcbf1/tests/unit/src/TestMain.hx#L27

I'm not familiar with the unit tests, so I'm not really sure what could be going on. ~~Interesting that just pushing the `TestSuite` class into the array causes `FlxG.renderer` to be null.~~

EDIT: I'm fairly certain it's because [`ReusableFrame` is static](<https://github.com/HaxeFlixel/flixel/blob/dev/flixel/text/FlxBitmapText.hx#L329>), therefore the render method check is done BEFORE we have a renderer

EDIT 2: Made a change to create the `ReusableFrame` once, at runtime. Hopefully this is an acceptable workaround. C++ CI now passes.

ACrazyTown:
@Geokureli How do you feel about this way of preventing breaking changes:

```haxe
	// The old deprecated signature, uses a helper material so that it can call the internal method
	@:deprecated	
	overload extern public inline function drawPixels(?frame:FlxFrame, ?pixels:BitmapData, matrix:FlxMatrix, ?transform:ColorTransform, ?blend:BlendMode,
			smoothing:Bool = false, ?shader:FlxShader):Void
	{
		_helperMaterial.blendMode = blend;
		_helperMaterial.smoothing = smoothing;
		_helperMaterial.shader = shader;
		
		drawPixelsInternal(frame, pixels, _helperMaterial, matrix, transform);
	}
	
	// The new signature, just call the internal method
	overload extern public inline function drawPixels(?frame:FlxFrame, ?pixels:BitmapData, material:FlxMaterial, matrix:FlxMatrix,
			?transform:ColorTransform):Void
	{
		drawPixelsInternal(frame, pixels, material, matrix, transform);
	}
	
	// For internal use only
	// For example, was reworked to use the new FlxMaterial API
	function drawPixelsInternal(?frame:FlxFrame, ?pixels:BitmapData, material:FlxMaterial, matrix:FlxMatrix, ?transform:ColorTransform) {}
```

`drawPixels()` is just a call to `drawPixelsInternal()`, which does the actual rendering. We could document the internal draw methods as "use at your own risk" and note that its signature could change at any time. Then we'd just overload the public method like I did above. This way there's a breaking change in only the renderer implementation, which would just be Flixel's, or someone else's who's crazy enough to build their own. Either way I think it's a very rare case so IMO that'd be fine.

I might have ideas for more changes soon, as I'm battletesting this implementation yet again!

ACrazyTown:
Also while I'm at it, I might as well ask if you know what the deal is with the batching functions in `FlxCamera` (e.g. `startQuadBatch()`). I see they're public, but hidden from autocomplete. Are they meant to be a public API? Do we also have to avoid breaking changes with them?

I thought they were only meant to be used by the draw functions, but it seems they're also used by `FlxBitmapText`, `FlxTilemap` and some other addons classes, for some reason

theoo-h:
hell yeah

flixel/system/frontEnds/CameraFrontEnd.hx
```diff
@@ -253,87 +257,52 @@ class CameraFrontEnd
...
 	inline function lock():Void
+	{
+		clear();
+	}
+
+	/**
+	 * Called by the game object to clear all the camera buffers for the next draw pass.
+	 */
+	@:allow(flixel.FlxGame)
+	inline function clear():Void
```
Geokureli: clear is too vague, I would have assumed it meant to clear the list of cameras, rather than to clear each camera buffer. I suggest clearBuffers

Geokureli:
Hello! I'm not dead, I've finally started my deep dive into this daunting thing. I'll probably have a million questions about the order in which things are happening and why things are done a certain way because I'm not very well-versed in rendering (yet). I'm gonna try to deploy this branch to https://dev.haxeflixel.com/demos/ and manually verify every demo, as its just too easy to miss something big in some one-off file.

I really wish there was a way to automate tests of the rendering systems, without capturing screen grabs and comparing to expected images.

At a glance, I'm not sure I like how FlxG.renderer is used, where a camera is set at the current target via `FlxG.renderer.begin()` followed by methods like clear() or render(). 

1. These seem like under the hood methods used by flixel systems, rather than something most devs will need global access to (like the other FrontEnds in FlxG).
2. I would really like to avoid this kind of modal global system, where a target camera is stored internally, and then has operations performed on it
Is there a reason that, [this code](https://github.com/HaxeFlixel/flixel/pull/3539/changes#diff-f146bf6b13231fec9a84ea391ae618c49743b0adcdab4382bbc8df0323e4d9f3R292-R293) can't simply be `FlxG.renderer.render(myTargetCamera);`? Specifically, can methods in FlxRenderer take the target camera as arg, and perform operations on that instance without needing to store it internally first? My assumption is that other things down the road, likely in the draw tree, call FlxG.renderer methods, asserting or assuming that it's targeting the desired camera, or something.

My expectation was that this is what FlxCameraView was for. I.E.: internal flixel code would have access to each camera's view for low-level operations, which are separated from FlxCamera's high-level features. I was expecting calls like this:
```hx
if (FlxG.renderBlit)
{
	camera.fill(camera.bgColor, camera.useBgAlphaBlending);
	camera.screen.dirty = true;
}
else
{
	camera.fill(camera.bgColor.rgb, camera.useBgAlphaBlending, camera.bgColor.alphaFloat);
}
```
to become this:
```hx
camera.view.clear();
```
or 
```hx
@:privateAccess camera.renderer.clear();
```
rather than this:
```hx
FlxG.renderer.clear();// which should honestly be called clearCurrent()
```

I assume there's good reasons, but they aren't clear to me, and therefore won't be clear to devs.

This seems like the most broad and structural complaint I can find, so let's talk about this while I go through this with a fine-toothed comb

ACrazyTown:
> 1. These seem like under the hood methods used by flixel systems, rather than something most devs will need global access to (like the other FrontEnds in FlxG).

I agree somewhat, but this is where Flixel's architecture comes into play. `FlxSprite.draw()` is a public method that any extending class may override to change draw behavior, and in order to do that it needs to have access to the rendering methods. Overriding `draw()` and calling drawing methods directly is something that seems to be relatively common for more complex sprites. It's done by [core classes](https://github.com/search?q=repo%3AHaxeFlixel%2Fflixel+camera.draw&type=code), [addons](https://github.com/search?q=repo%3AHaxeFlixel%2Fflixel-addons+camera.draw&type=code) and also by libraries like FlxAnimate.

Compare this to something like OpenFL, where display objects don't have any control over how they're drawn, and their entire draw phase is handled internally. See for example: https://github.com/openfl/openfl/blob/develop/src/openfl/display/OpenGLRenderer.hx#L871-L894

Also, most of the methods in `FlxRenderer` that were ported over were already public in `FlxCamera`, with the exception of like `clear()` and `render()`. Sure, we could make them private. I kept them public mainly to avoid a breaking change

> 2. I would really like to avoid this kind of modal global system, where a target camera is stored internally, and then has operations performed on it

Truth is, Flixel's current render methods, and the way they're tied to the camera are a bit unconventional. Renderers are usually a global-ish concept, whereas cameras would just be render targets/render textures. (In Flixel land at least, generally they'd just be a matrix modifier or something)

> Is there a reason that, [this code](https://github.com/HaxeFlixel/flixel/pull/3539/changes#diff-f146bf6b13231fec9a84ea391ae618c49743b0adcdab4382bbc8df0323e4d9f3R292-R293) can't simply be `FlxG.renderer.render(myTargetCamera);`? Specifically, can methods in FlxRenderer take the target camera as arg, and perform operations on that instance without needing to store it internally first? My assumption is that other things down the road, likely in the draw tree, call FlxG.renderer methods, asserting or assuming that it's targeting the desired camera, or something.

Your assumption is correct, I was hoping it'd simplify the draw tree by not having to constantly keep track of the camera. While `renderer.begin()` is not doing too much right now, in the future, with another renderer implementation it might start doing more work, especially once we throw stuff like render textures into the mix. 

---

Some more thought definitely needs to be put in this. I've been trying to implement changes from #2915 in a separate branch, based on this, to try and stress test and future proof it as much as possible. I too have some mixed feelings about the singleton `FlxG.renderer` approach and how it works. 

I've been trying to avoid breaking changes here, to make the transition easier, but I do wonder if biting the bullet and breaking things would allow us to make the nicer/easier to use. I assume maintaining 2 different branches with major changes between them would be a bit of a hassle, though.

Also, I had some additional concerns/questions, in case you missed them: https://github.com/HaxeFlixel/flixel/pull/3539#issuecomment-3828098549 , https://github.com/HaxeFlixel/flixel/pull/3539#issuecomment-3828126970

Geokureli:
> Truth is, Flixel's current render methods, and the way they're tied to the camera are a bit unconventional. Renderers are usually a global-ish concept, whereas cameras would just be render targets/render textures. (In Flixel land at least, generally they'd just be a matrix modifier or something)

My issue isn't with having global render tools, I think having FlxG.renderer.doStuff() is a necessary part of this, and can allow a lot of cool customizations. 

> I've been trying to avoid breaking changes here, to make the transition easier, but I do wonder if biting the bullet and breaking things would allow us to make the nicer/easier to use. I assume maintaining 2 different branches with major changes between them would be a bit of a hassle, though.

Let's talk about what breaking changes would make this easier. I'm only wary of breaking changes because I didn't think there would be immediate benefits to this refactor, but that doesn't seem to be the case, re: multitexturing.

> Your assumption is correct, I was hoping it'd simplify the draw tree by not having to constantly keep track of the camera. 

Breaking changes would help here, too. We have methods like drawSimple/drawComplex(camera), which you preserved, but I assume our end goal is to have something like a global draw call that handles those distinctions. We talked about adding drawTo(camera), which could be done in a patch with no issues. We could change FlxBasic to:
```hx
function draw()
{
    for (camera in getCamerasLegacy())
        drawTo(camera);
}

function drawTo(camera:FlxCamera)
{
    visibleCount++;
}
```

 I assume we're aligned on all I've said so far, but I prefer the following (rough pseudocode):
 ```hx
override function drawTo(camera:FlxCamera)
{
    super.draw();
    
    FlxG.renderer.drawTo(this, camera);
}
```

compared to:
```hx
override function drawTo(camera:FlxCamera)
{
    super.drawTo(camera);
   
    FlxG.renderer.setRenderTarget(camera);
    FlxG.renderer.drawToCurrent(this);
}
```

In the latter method, things like FlxBitmapText - which draw its frame multiple times per update in order to create a border - wouldn't need to pass the camera around to every function. I much prefer the first, however, I think passing a camera around isn't an issue, compared to the hassle of debugging issues that arise from not setting the render target, or calling a function that unknowingly sets the global render target, or changing a function so that it now sets the target, when the calling function expected it to not do so. In general I expect global methods to be context-free, where calling the same method with the same args should have a similar result regardless of what was called prior.

There's a lot of other planned features that would benefit from breaking changes:
- #3544 
- #2715
- #3164

I also think they fit together into one solution, but that's a big-ass discussion that I just typed 4 paragraphs of and then promptly deleted, to stay in scope.

If we need to make breaking changes and there's no good way to ease devs into the new system we can make alpha branches and personally help migrate funkin and other largely used projects/forks before releasing. 

> While renderer.begin() is not doing too much right now, in the future, with another renderer implementation it might start doing more work, especially once we throw stuff like render textures into the mix.

Got any examples?

ACrazyTown:
> Let's talk about what breaking changes would make this easier. 

I'm not entirely sure, yet. I'll have to sit and think about it for a bit and see if there's anything we can change about the draw phase to make things more convenient.

> I'm only wary of breaking changes because I didn't think there would be immediate benefits to this refactor, but that doesn't seem to be the case, re: multitexturing.

While this specific PR won't bring much benefits by itself, I've been working on implementing all of the bells and whistles from #2915 in parallel to this. Once we figure out this main abstraction part, all of the fancy features should follow relatively soon.

> In the latter method, things like FlxBitmapText - which draw its frame multiple times per update in order to create a border - wouldn't need to pass the camera around to every function. I much prefer the first, however, I think passing a camera around isn't an issue, compared to the hassle of debugging issues that arise from not setting the render target, or calling a function that unknowingly sets the global render target, or changing a function so that it now sets the target, when the calling function expected it to not do so. In general I expect global methods to be context-free, where calling the same method with the same args should have a similar result regardless of what was called prior.

I thought about it some more and I agree. Though, I also think passing the camera in the draw functions might also shoot us in the foot down the road. What happens when we want to render sprites to a texture? Would that mean that render textures would also have to have a camera? Or perhaps you'd have to make a new camera and then draw to its internal texture?

I think cameras are kinda weird, and yet they're such a fundamental part. They act as render targets but also do actual camera stuff.

My idea wouldn't work either, since sprites change the render target when they're drawn. The only way you could draw it to a texture would be to do it manually via `sprite.drawTo()`, but that would take a camera as well...

> There's a lot of other planned features that would benefit from breaking changes:
> ...
> I also think they fit together into one solution, but that's a big-ass discussion that I just typed 4 paragraphs of and then promptly deleted, to stay in scope.

I'd like to hear more about that. Perhaps in a seperate issue? If these changes can compliment each other it might be worth also keeping them in mind while we're overhauling the renderer.

One thing I know is that Beeblerox's branch has improved performance with debug draw, by drawing them using triangles in OpenGL. So that'd be another thing done by #2915 

> > While renderer.begin() is not doing too much right now, in the future, with another renderer implementation it might start doing more work, especially once we throw stuff like render textures into the mix.
> 
> Got any examples?

I haven't implemented render textures yet, but in my current experimental branch, changing the camera requires an update of the projection matrix:
https://github.com/ACrazyTown/flixel/blob/22b30e7b8ba2148a19f387984b8e451d3006a032/flixel/system/render/gl/FlxGLRenderer.hx#L74-L78

Beeblerox's branch also does some additional work when swapping the render target: https://github.com/AustinEast/flixel/blob/Beeblerox-FlxMaterial-and-camera-scaling/flixel/system/render/gll/FlxGLView.hx#L95-L115

Geokureli:
> I thought about it some more and I agree. Though, I also think passing the camera in the draw functions might also shoot us in the foot down the road. What happens when we want to render sprites to a texture? Would that mean that render textures would also have to have a camera? Or perhaps you'd have to make a new camera and then draw to its internal texture?

The passed arg doesn't need to be a camera, that was just a quick example, we can pass in a camera.view, a FlxRenderTarget, an interface, some kind of AnyOfThree, or even overload it to take various types
 
> I think cameras are kinda weird, and yet they're such a fundamental part. They act as render targets but also do actual camera stuff.

Yeah these should be split

> I'd like to hear more about that. Perhaps in a seperate issue? If these changes can compliment each other it might be worth also keeping them in mind while we're overhauling the renderer.

I'll try to type it up soon

> I haven't implemented render textures yet, but in my current experimental branch, changing the camera requires an update of the projection matrix: https://github.com/ACrazyTown/flixel/blob/22b30e7b8ba2148a19f387984b8e451d3006a032/flixel/system/render/gl/FlxGLRenderer.hx#L74-L78
> 
> Beeblerox's branch also does some additional work when swapping the render target: https://github.com/AustinEast/flixel/blob/Beeblerox-FlxMaterial-and-camera-scaling/flixel/system/render/gll/FlxGLView.hx#L95-L115

I only skimmed, but I still think it's doesn't need to be a modal global tool. Rather than FlxG.renderer setting internal vars based on passed in render targets for future calls, the renderer can set up all that within the render target, and future calls with that target passed in also manipulate that render target.

Writing this just before going to bed. Sorry if my tired brain isn't as clear as I think it is

ACrazyTown:
Here's a rough idea, inspired by your suggestions and [Unity's camera system](https://docs.unity3d.com/6000.3/Documentation/ScriptReference/Camera.html). This should be possible to accomplish without any breaking changes.

We keep `FlxCamera` as-is, for now. I kinda can't think of any ideas to split its functionality without breaking literally everything. Also one thing I failed to account is that, while yes, they essentially are just fancy render textures, they also render their texture to the screen! Without them we'd have to have a special `FlxSprite` or something, so I don't think it's worth the hassle at the moment. (EDIT: found https://github.com/HaxeFlixel/flixel/issues/1073, which is relevant)

Regarding render textures, this would mean that you'd have to draw things to an intermediate camera. Though then you might not need to manually manage the texture yourself, and just yoink `camera.texture` or something like that. We could also do something similar to Beeblerox's [`FlxRenderTarget`](https://github.com/AustinEast/flixel/blob/Beeblerox-FlxMaterial-and-camera-scaling/flixel/effects/FlxRenderTarget.hx)

We'll pass in cameras as arguments to the draw functions, as you suggested. To make this work, camera views will now hold some kind of draw command buffer, which just holds info about the sprites to draw. Later in the draw phase, once all of the sprites have "drawn" themselves, the renderer will render all of the added cameras and the sprites in the buffer.

In a way this is similar to what we have now with the FlxDrawItem API, where nothing is really "drawn" until the `graphics.drawQuads()` calls.

ACrazyTown:
Never mind, I think this will have to be a breaking change after all. Abstracting away how data will be sent to the renderer will require reworking some stuff, and I think it'd be far easier to just make a new API than to try and wire up the existing FlxDrawItems to it

What branch should I target?

Geokureli:
I have something I'd like to try, hopefully I can dive in this weekend. If we're starting on 7.0.0 now I should prepare 6.2.0 for release soon

Geokureli:
I believe I see a path forward, give me a few more days

ACrazyTown:
By all means, take your time

Geokureli:
There's some more ideas I wanna try, but [this is what I've done](https://github.com/ACrazyTown/flixel/compare/seperate-renderer...Geokureli:flixel:geo-separate-renderer?expand=1). In the process of writing this I continued to make a bunch of changes so some things may be out of date.

1. Remove FlxG.renderer.begin, all renderer functions take the view as an arg.
2. Added `FlxTypedRenderer<TView:FlxCameraView>` so extensions can deal with the specific view type they were made for rather than the base type upcasted everywhere.
3. Moved `render`, `clear` `fill` and a lot of debug drawing methods from renderer to view. This allows internal methods used by these to be private, rather than `@:noCompletion public`, but also there wasn't much benefit to these being in render. More on this later...
4. FlxObject needs to keep calling the existing local method `drawDebugBoundingBox` and `getDebugBoundingBoxColor`. These methods are not inlined, and could be overridden (In fact I think I done that a couple times). I also made beginDebugDraw return the graphics for simplicity (more ideas for that later).
5. `FlxCameraView` is an abstract class and all empty methods moved there were changed to abstract methods. `FlxCameraView` can never be instantiated and all extending classes must implement the abstract methods. I did the same with `FlxRenderer`
6. View's setters now affect the camera's corresponding field. I assumed this was an accident
7. Moved all deprecated camera methods/getters/setters to the bottom, also change all the getters/setters so they won't crash in any ways that were safe in 6.1, e.g.: refs to `_bounds` would have crashed in render tile mode because viewBlit is null, and it's also more reflection safe so it should work for some (but not all) games that use hscript, or polymod
8. I removed targetGraphics from quadView, `camera.fill` calls with a specific graphics arg don't call view.fill, they simply draw a rect to the target. Similarly, I'm not calling camera.calcMargin from view, camera does that itself, I simplified other tangled things like that.
9. `viewBlit` and `ViewQuad` are private. I may revert that, but I feel like getting into specific renderers should be done behind the scenes, or done by people who know what they're getting into. Note how in Pointer.hx I added a worldToDebug converter in view rather than accessing blit/quad separately. My goal is to expose simple back-end agnostic methods for rendering whenever possible

BTW I tested this branch with Renaine, a game that does a lot of ad-hoc/hacky rendering stuff and it seemed to work fine

## Next steps:
Gonna test a bunch of demos, with the current changes and compare rendering and debug drawing to the last release. This will take a bit, so while that is happening we can discuss solidify the design of all this.

### Decouple FlxCameraView from FlxCamera
`FlxCameraView` requires a camera, but ideally there should be render targets that are not tied to a camera. For example: tilemaps have buffers to prevent redrawing the same stuff every frame, ideally, that could be a render target that the tilemap draws do when it changes, then the buffer render target is drawn to the camera render target each frame. Currently views reference the camera for things like scroll, size, the view rect, and all the getters. I'm thinking I would make a base class `FlxView` or `FlxRenderTarget` that can have these fields without an internal reference to a camera. Many of these methods that take a FlxCameraView would be changed to take the base class.

Also, I like the name "view" for these, but I wonder if that will be confusing in all of the existing things called "view", i.e. `camera. viewToGamePosition()` and worldToView and whatnot. i dunno, maybe they'll still make perfect sense

### No.3 Continued: Moved renderer methods to view
I fee like this new way is a lot cleaner and more customizable. Rather than having everything call
```
FlxG.renderer.begin(camera);
FlxG.renderer.drawThing(); // reads and changes a bunch of stuff on camera.view
```
On top of removing begin altogether, I think it should just be `view.drawThing()`, this way people can extend FlxCamera/FlxCameraView and override each feature. These view methods may call `FlxG.renderer.drawThing(this);` as a way to centralize hefty rendering operations. This also allows customization for rendering to all views by extending the renderer.

### No.4 Continued
Right now, none of FlxCameraView's debug drawing methods are not used, due to no.4 above, except `path.drawNode` calls `view.drawDebugFilledRect`. Once those methods are removed in a major release, the new, non-deprecated functions that call them will be changed to call the view's debug draw methods. For now people who override the new methods are expected to call view's drawing methods directly. We should add build macros that throw a warning when these deprecated methods are overridden, otherwise they won't get a warning unless they call super() (i think).

### Obfuscate openfl.display.Graphics
Much like we talked about with hiding Bitmap with some backend agnostic image editing tools, we should do the same for Graphics. Perhaps an abstract over Graphics with drawRect() methods that call FlxG.renderer.drawRect(this), or something

### "Quad" vs "Tile"
The render method name is `DRAW_TILES`. I hate having one thing say quad and another say tiles, is one more accurate the the other?

Geokureli:
Apparently `@:bypassAccessor` doesn't work with setters> I could have sworn it did. Gonna rethink this a bit

ACrazyTown:
Nice work, thanks for helping out with this. I like most of the changes here, though I do have some concerns/questions.

### Re: No.3

I think moving `render()` and `clear()` make sense as well, though I have some mixed feelings about moving `fill()` and the debug drawing methods as they're rendering commands and they interact with the renderer (eg. `fill()` just draws a colored quad that covers the screen). This will make more sense with the OpenGL renderer where objects are tied to the renderer rather than to cameras. See for example how I was handling `fill()` while messing around with OpenGL: https://github.com/ACrazyTown/flixel/blob/feat/opengl-renderer-2/flixel/system/render/gl/FlxGLRenderer.hx#L112-L121

> "This allows internal methods used by these to be private, rather than @:noCompletion public"

I don't think I completely understand what you mean. If you mean that `view.fill()` will call some private `renderer.fill()` method then that makes sense, I agree, and ignore this entire paragraph lol.

### Re: Decouple FlxCameraView from FlxCamera

I've been thinking about this too. I think it's doable to have some arbitrary buffer draw commands can be executed on (though perhaps this should just be `FlxRenderTexture`?). For drawing sprites and whatnot I think we'd still need a camera though, because of how the draw phase works. 

Food for thought, Beeblerox's [`FlxRenderTarget`](https://github.com/AustinEast/flixel/blob/Beeblerox-FlxMaterial-and-camera-scaling/flixel/effects/FlxRenderTarget.hx)

### Re: Obfuscate openfl.display.Graphics

Hmm, some food for thought, the renderer overhaul had the [`FlxDraw`](https://github.com/AustinEast/flixel/blob/Beeblerox-FlxMaterial-and-camera-scaling/flixel/FlxDraw.hx) class as a way to draw hardware accelerated shapes. It's not a direct equivalent to `Graphics` as it lacks features but performance wise it should be much better. Perhaps we could unify the two in some way, use `FlxDraw` when drawing to the screen and `Graphics` when stamping to a texture, or something.

### Re: "Quad" vs "Tile"

This bothered me as well, but I couldn't decide what to do. `DRAW_QUADS` would be a more accurate name as that's the method used (`graphics.drawQuads()`). `DRAW_TILES` is the older method which was removed in like OpenFL 4.0 (and it's why Flixel wasn't compatible with later OpenFL versions for a few years). I'm not sure why `DRAW_TILES` wasn't renamed before, perhaps it was an oversight?

## Personal concerns

### Breaking core function signatures
This is not completely related to these specific changes, but I thought I'd mention it so that we could future proof stuff before things are set in stone. Ideally I was hoping that the OpenGL renderer, which I'd implement some time after this is merged, wouldn't break anything. That way it could be locked behind a compiler flag in Flixel 6.x while it gets fully polished and tested, and made default in Flixel 7. (and perhaps we drop the `DRAW_QUADS` renderer then?)

One issue with this is that the renderer overhaul comes with some function signature changes. As an example, here's what `drawTriangles()` could look like with the changes:

```haxe
// Old
function drawTriangles(graphic:FlxGraphic, vertices:DrawData<Float>, indices:DrawData<Int>, uvtData:DrawData<Float>,
		?colors:DrawData<Int>, ?position:FlxPoint, ?blend:BlendMode, repeat:Bool = false, smoothing:Bool = false, ?transform:ColorTransform,
		?shader:FlxShader):Void {}

// New
function drawTriangles(graphic:FlxGraphic, data:FlxTrianglesData, material:FlxMaterial, matrix:FlxMatrix, ?transform:ColorTransform):Void {}
```
It's a pretty significant, and IMO welcome change. I've been thinking over how to implement it in a non-breaking way though. Everything I've thought of thus far has some flaw:
- Deprecate old, and introduce new methods
  - This would be the easiest way, but what do we call them? `drawPixels()` and `copyPixels()` are in my opinion kinda strange already, and I'm not really good with naming stuff lol. Maybe something like `drawQuad()`? For triangles perhaps we could consider something like `drawMesh()` but this would create an inconsistency where everything else refers to triangles but the renderer method refers to it as a mesh.
- Overloads
  - I'm not sure if this can even be pulled off due to inheritance. I played around with making the renderer an abstract class a while back in hopes of avoiding this, and while it technically made it possible to overload methods, they'd only be accessible if we had a typed reference to the extending class and not in the base class
- https://github.com/HaxeFlixel/flixel/pull/3539#issuecomment-3828098549
  - A bit ugly, and not really non-breaking, just gives us an excuse to do so.

It's not really crucial that the OpenGL renderer gets into Flixel 6, but I think it'd be really nice, and also would make the transitional period a lot smoother.

Geokureli:
> ### Breaking core function signatures
> One issue with this is that the renderer overhaul comes with some function signature changes. 

The secret is it's always gonna be ugly. I tend to make new methods, in your example I'd add drawMaterialTriangles or something.

> It's not really crucial that the OpenGL renderer gets into Flixel 6, but I think it'd be really nice, and also would make the transitional period a lot smoother.

The sooner we try to utilize the new system the sooner we'll realize we're doing something wrong. The more we can plan now, the better.


> ### Re: No.3
> I think moving `render()` and `clear()` make sense as well, though I have some mixed feelings about moving `fill()` and the debug drawing methods as they're rendering commands and they interact with the renderer (eg. `fill()` just draws a colored quad that covers the screen). This will make more sense with the OpenGL renderer where objects are tied to the renderer rather than to cameras. See for example how I was handling `fill()` while messing around with OpenGL: https://github.com/ACrazyTown/flixel/blob/feat/opengl-renderer-2/flixel/system/render/gl/FlxGLRenderer.hx#L112-L121

This is where I'm really struggling to understand the idea. This isn't really how I imagine other global utils are used, and it seems weird, here. Is the idea behind FlxG.renderer that the "screen" is a global object that things draw to, where the views simply determine how world coordinate translate to screen coordinates? I see it very differently, where the screen has a bunch of render targets, with specific layering, orientation, filters and properties. To me objects draw to a render target, regardless of the backend, even in the case of a GL backend, the Flixel dev is choosing which sprites draw to which render target. Perhaps the backend combines that all into one actual screen, but the notion of these being individual things is still important. Thats why filling a view instance via a global call feels weird. 

In contrast, say I wanted to fill the screen at an arbitrary time during the draw call, say this had nothing do with a specific view, I simply want the entire screen filled, then I could see using `FlxG.renderer.fill(myColor, myMergeAlpha)`, but that's not what we have here, all of the global rendering functions are specific to a view, and it uses the view's bounds, therefore it makes perfect sense to call `view.fill`.

> > "This allows internal methods used by these to be private, rather than @:noCompletion public"
> 
> I don't think I completely understand what you mean. If you mean that `view.fill()` will call some private `renderer.fill()` method then that makes sense, I agree, and ignore this entire paragraph lol.

Kinda, maybe, idunno...? In the example above fill wouldn't be private, and it's public in your gl example, so that confused me. I also see that your glRenderer's fill uses a shared rect, so the setup kinda makes sense, unlike the blit and quad views which each have a separate displayObject they draw to, gl can just use the same one. To me though, it would make more sense to put a global `fillRect(someRect, color, mergeAlpha)` in renderer, and have views call that with the camera's margins. Cameras can also have filters(in the future perhaps there could be blending options, too), so if the Gl renderer just draws everything to the screen rather than separating them into specific layers, would we lose that functionality?

I think regardless of what's going on in the backend, object's specified to draw to render targets should call methods on that render target to draw itself. The global util should not need to be an intermediary between those (at least as the default way of drawing to a view). FlxG.renderer should more be for things like project wide settings/configs, and global tools that interact specifically and directly with the backend renderer.

I hope this makes sense, I did the thing where I type all this right before going to bed. I have no concerns or comments on any of your other points

ACrazyTown:
> This is where I'm really struggling to understand the idea. This isn't really how I imagine other global utils are used, and it seems weird, here. Is the idea behind FlxG.renderer that the "screen" is a global object that things draw to, where the views simply determine how world coordinate translate to screen coordinates? I see it very differently, where the screen has a bunch of render targets, with specific layering, orientation, filters and properties. To me objects draw to a render target, regardless of the backend, even in the case of a GL backend, the Flixel dev is choosing which sprites draw to which render target. Perhaps the backend combines that all into one actual screen, but the notion of these being individual things is still important. Thats why filling a view instance via a global call feels weird.

I'll mostly talk about the potential OpenGL renderer here, since it follows the traditional render setup, though the same logic should apply to the other render backends.

The latter interpretation is correct. There can and will be multiple different render targets that all the drawing commands will be executed on. This is how cameras will work, with each camera having its own render texture. The only interaction with the screen (back buffer) will be at the end of the frame, when the camera textures are drawn.

My goal with the global renderer was to keep all the core backend logic centralized, so that we'd keep all the backend OpenGL calls in one place, rather than scattered across multiple classes. There's also benefits to keeping some stuff, like vertex buffers global on a renderer-level rather than per render-target. For example, to batch sprites, we really only need one vertex and index buffer as any state change (such as a render-target swap) will require us to do a draw call and flush the buffer. If we sort our **draw commands*** to be in order per render target, we can draw everything without a buffer change, which is good because in OpenGL you should keep [state changes to a minimum](https://stackoverflow.com/a/15924917) (for reference in the link, a VBO refers to a vertex buffer).

*Stuff like this is why I was considering introducing some sort of draw command buffer, ideally without any additional overhead. I believe modern graphics APIs (i.e. Vulkan, Metal, WebGPU) also use command buffers, so I guess that's future proofing in a way.

We could add convenience methods per view that are basically just something like
```haxe
// in FlxCameraView
public inline function drawPixels(...)
{
  FlxG.renderer.drawPixels(this, ...);
}
```
but I still think that all the main rendering stuff should be handled by `FlxRenderer` implementations.

> In contrast, say I wanted to fill the screen at an arbitrary time during the draw call, say this had nothing do with a specific view, I simply want the entire screen filled, then I could see using `FlxG.renderer.fill(myColor, myMergeAlpha)`, but that's not what we have here, all of the global rendering functions are specific to a view, and it uses the view's bounds, therefore it makes perfect sense to call `view.fill`.

We could allow direct drawing to the screen (back buffer) if the passed in `view` argument is `null`. This is usually what happens when you pass in `null` to some `setRenderTarget()` method in other higher level graphics libraries. For example, [Context3D.setRenderToTexture()](https://api.openfl.org/openfl/display3D/Context3D.html#setRenderToTexture).

> Kinda, maybe, idunno...? In the example above fill wouldn't be private, and it's public in your gl example, so that confused me. I also see that your glRenderer's fill uses a shared rect, so the setup kinda makes sense, unlike the blit and quad views which each have a separate displayObject they draw to, gl can just use the same one. To me though, it would make more sense to put a global `fillRect(someRect, color, mergeAlpha)` in renderer, and have views call that with the camera's margins. Cameras can also have filters(in the future perhaps there could be blending options, too), so if the Gl renderer just draws everything to the screen rather than separating them into specific layers, would we lose that functionality?
> 
> I think regardless of what's going on in the backend, object's specified to draw to render targets should call methods on that render target to draw itself. The global util should not need to be an intermediary between those (at least as the default way of drawing to a view). 

I think I cleared up the points here in the paragraph above

> FlxG.renderer should more be for things like project wide settings/configs, and global tools that interact specifically and directly with the backend renderer.

Hmm, though `FlxG.renderer` doesn't expose much in terms of additional functionality. It implements all the higher level drawing methods we've previously had in FlxCamera. I think your point would make sense if we had an actual graphics API built on top. Perhaps something like OpenFL's Context3D, or Ceramic/Clay's [GraphicsDriver](https://github.com/ceramic-engine/clay/blob/main/src/clay/spec/GraphicsDriver.hx). This way we could have a single `FlxHardwareRenderer` or something that just deals with our custom API, and we just swap out the graphics driver. Less duplicating stuff at the cost of less case specific optimizations. It's probably overkill though, and it also wouldn't fare well with the current blit and quad renderer

Geokureli:
I think my issue is with enforcing the same API across all backends. I'm okay with low-level helpers being in an extension of renderer, but having all these methods defined in the base renderer, and redefined in extensions is going to make things overly rigid, and not well organized in most cases. The interface with the render targets doesn't need to match the interface with the global renderer and if we break that connection I imagine things will go much more smoothly.

> My goal with the global renderer was to keep all the core backend logic centralized, so that we'd keep all the backend OpenGL calls in one place, rather than scattered across multiple classes

I kinda (mostly) agree with this. I think `FlxGLRenderer` could be the global way to interface with gl directly and call gl specific utils, where `FlxGlView` should be a render target who's public interface is defined in `FlxCameraView`. For example, in FlxGLView.hx:
```hx
var renderer(get, never):FlxGLRenderer;
inline function get_renderer() return cast (FlxG.renderer, FlxGLRenderer);

override function fill(color:FlxColor, blendAlpha:Bool)
{
    renderer.fillRect(getScreenRect(), color, blendAlpha);
}
```

This is somewhat pseudo code, I dunno if renderer.fillRect draws to the default buffer or whatever, but note that `view.fill` is defined in FlxCameraView, and implemented in this extension, however fillRect does not need to be defined in FlxRenderer, it can be defined and implemented entirely in `FlxGLRenderer` GL views have direct access to the gl renderer and can call gl specific things that other backends may not need to define. Likewise, anyone can assert that the current renderer is gl, if their project is set up that way, cast it and directly call gl methods on it.

That said, while all renderers don't need to define the same helpers/tools, we may want to define common tools in FlxG.renderer that people can access without caring about what backend is being used, but those global fields do not need to match view's common methods. Things like [FlxG.renderer.drawMode = PIXEL_PERFECT](https://github.com/HaxeFlixel/flixel/issues/3527)

ACrazyTown:
Yeah I think this could make sense. I don't know how I feel about implementing draw methods into `FlxCameraView` though, cause it might complicate other render targets unrelated to cameras. Perhaps it'd be better to implement them into a base `FlxView` class like you suggested?

I suppose what could really help clear things up is to actually try and implement the GL renderer into the new abstraction, which is what I'll be doing soon. Is your branch up to date? Not gonna merge here yet, but just so I know for reference locally

Geokureli:
Gonna try a couple things, I'll let you know when I'm done, hopefully by sometime tomorrow

Geokureli:
quoted from https://github.com/HaxeFlixel/flixel/pull/3567
 
> I'm a bit worried about tying the render methods to the camera view but I suppose this boils back to my previous sentence

The important distinction I wanna make is that, while we may move the rendering somewhere else, I do think it's important that to draw a sprite to a camera's view, you should call something akin to `camera.view.drawThing(this)` rather than something like `FlxG.renderer.drawThingToView(camera.view, this)`, but viewBlit may end up calling `privateRefToBlitRenderer.drawThing(this, sprite)`. I wanna focus on finding the best way to define these methods, before deciding exactly where they will end up.

> > FlxTexture, new methods that take bitmaps should take textures, instead
> 
> I have a draft (as discussed in #3540) for this typed up in a branch somewhere. It's implemented as a new class rather than an abstract so there is a bit of friction with implementing it with existing systems. Will need to revisit it and clean it up

The reason I make FlxVertexBuffer an abstract, is so that method args could be changed to take that type without breaking existing calls to it (as Graphics and FlxVertexBuffer can be implicitly casted to one another). Once devs migrate, FlxVertexBuffer will be a different type that will allow (or wrap) various underlying types. 1 option is to make it a class wrapper with derived types that use each specific underlying type. Another option is an abstract OneOfMany<T...> that calls the appropriate renderer

ACrazyTown:
> The important distinction I wanna make is that, while we may move the rendering somewhere else, I do think it's important that to draw a sprite to a camera's view, you should call something akin to `camera.view.drawThing(this)` rather than something like `FlxG.renderer.drawThingToView(camera.view, this)`, but viewBlit may end up calling `privateRefToBlitRenderer.drawThing(this, sprite)`. I wanna focus on finding the best way to define these methods, before deciding exactly where they will end up.

Yeah this makes sense. To be clear, I'm not against putting the render methods in camera view, I'm moreso against tying them to cameras specifically, as this would lead us further away from separating cameras and rendering (#1073). 

It'll be easier to reason with this once we have all the missing pieces, so my focus now is to implement `FlxTexture`, `FlxRenderTexture` and the GL renderer (so that we can actually use the render texture). I already have basic quads working, so hopefully it won't take too long to get most things up and running.

In the meantime, I'd appreciate any help on #3566, as I think that's going to be one of the toughest parts to figure out.

> The reason I make FlxVertexBuffer an abstract, is so that method args could be changed to take that type without breaking existing calls to it (as Graphics and FlxVertexBuffer can be implicitly casted to one another). Once devs migrate, FlxVertexBuffer will be a different type that will allow (or wrap) various underlying types. 1 option is to make it a class wrapper with derived types that use each specific underlying type. Another option is an abstract OneOfMany<T...> that calls the appropriate renderer

I don't think I can make `FlxTexture` an abstract because it introduces some new fields, but also works slightly differently. If devs need the actual BitmapData they can either:
  - Use `texture.handle` to get the underlying representation. (Not recommended, if/when Flixel adopts other backends this will be changed to whatever type)
  - or use `texture.getBitmap()`, which returns a `abstract FlxBitmap(BitmapData)` and also undumps the texture if needed (Recommended)

Hooking it up is doable, it's just an extreme hassle because everything expects a `BitmapData`. I'll try to finish and PR it here soon, tho

ACrazyTown:
@Geokureli I've opened up a PR for FlxTexture: https://github.com/ACrazyTown/flixel/pull/2

flixel/system/render/FlxCameraView.hx
```diff
@@ -0,0 +1,335 @@
...
+	 * @param   colors     A vector containing the colors to use per vertex. Currently does not work with any renderer.
+	 * @param   position   A point representing the top-left position to draw to.
+	 * @param   blend      The blend mode to use, optional.
+	 * @param   repeat     Whether the graphic should repeat.
+	 * @param   smoothing  Whether to use smoothing (anti-aliasing) when drawing.
+	 * @param   transform  The color transform to use, optional.
+	 * @param   shader     The shader to use, optional (used only with the DRAW_TILES renderer).
+	 */
+	public function drawTriangles(graphic:FlxGraphic, vertices:FlxVector2d<Float>, indices:FlxVector2d<Int>, uvtData:FlxVector2d<Float>, ?colors:FlxVector2d<Int>,
+		?position:FlxPoint, ?blend:BlendMode, repeat:Bool = false, smoothing:Bool = false, ?transform:ColorTransform, ?shader:FlxShader)
```
ACrazyTown: @Geokureli Is `FlxVector2d` meant to act as an abstraction over openfl's Vector, that would eventually become a new class, or is it just a helper? It makes sense for positions and uvs, which are treated as (x, y) pairs, but it doesn't make sense to use it for indices and colors which have a value per coordinate pair.

Geokureli:
> I've opened up a PR for FlxTexture: [ACrazyTown#2](https://github.com/ACrazyTown/flixel/pull/2)

Exciting! I've been away doing uncle things, gonna look this over today.

flixel/system/debug/watch/Tracker.hx
```diff
@@ -265,26 +265,4 @@ class Tracker extends Watch
...
-	}
-
-	public function toString():String
-	{
-		return FlxStringUtil.getDebugString([
-			LabelValuePair.weak("variables", variables),
-			LabelValuePair.weak("extensions", extensions)
-		]);
-	}
-}
```
ACrazyTown, line 290: @Geokureli Isn't this a breaking change? Importing `flixel.system.debug.watch.Tracker.TrackerProfile` will now fail because it was moved to a seperate class (https://github.com/HaxeFlixel/flixel/pull/3539/commits/59b4346794842fc742bfbc3955e949c90b4a652e)

PR #3542, opened 2026-01-02
FlxStringUtil: Optimize and improve bitmapToCSV and prevent errors when farbling

### LINKED ISSUE #3541: FlxStringUtil: bitmapToCSV fails on brave when shields are up
Can be seen here: https://haxeflixel.com/demos/FlxCollisions/

Brave is manipulating the result of bitmap.getPixel() to prevent fingerprinting on user uploaded images. A process known as "farbling", outlined in [this old issue](https://github.com/HaxeFlixel/flixel/pull/2286).

I don't know how much use this method is getting, but we should warn about this as it may never show up on dev tests. This method in general has a shit implementation, and could be reworked.

Image

Geokureli:
Fixes #3541

- `FlxStringUtil`
    - Optimize `arrayToCSV` and therefore `imageToCSV` and `bitmapToCSV`
    - Detects browsers manipulating getPixel result for safety features and counteracts this in bitmapTo by getting the nearest mapped color
    - Add `bitmap32ToCSV`and `image32ToCSV` which does not ignore alpha
- `FlxTilemap`: Add overloads for `loadMapFromGraphic` deprecate the one that allows both `invert` and `colorMap` args
- `FlxArrayaUtil`: Add `scale` methods with overloads for arrays and 2d arrays
- `FlxColor`: Add `getDistance` and `nearest` methods for comparing differences in colors

### DIFF
--- a/flixel/util/FlxStringUtil.hx
+++ b/flixel/util/FlxStringUtil.hx
@@ -447,7 +447,10 @@ class FlxStringUtil
 
 	/**
 	 * Converts a BitmapData object to a comma-separated string.
-	 *
+	 * 
+	 * **NOTE:** Due to OpenFL premultiplying all pixel colors on set,
+	 * any pixel with a 0 alpha and non-zero color components will be treated as 0x00000000
+	 * 
 	 * @param	bitmap   A Flash BitmapData object, preferably black and white.
 	 * @param	scale    Default is 1. Scale of 2 means each pixel forms a 2x2 block of tiles, and so on.
 	 * @param	colorMap An array of rgba color values in the order they're intended to be assigned as indices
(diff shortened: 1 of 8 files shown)


flixel/util/FlxStringUtil.hx
```diff
@@ -378,193 +378,291 @@ class FlxStringUtil
...
-		var height:Int = Std.int(Data.length / Width);
-		var index:Int;
-		var offset:Int = 0;
+		return bitmapToCSVHelper(bitmap, scale, colorMap, true);
+	}
 
-		while (row < height)
+	/**
+	 * Converts a BitmapData object to a comma-separated string.
+	 *
```
Geokureli:
```suggestion
	 * 
	 * **NOTE:** Due to OpenFL premultiplying all pixel colors on set,
	 * any pixel with a 0 alpha and non-zero color components will be treated as 0x00000000
	 * 
```

PR #3545, opened 2026-01-07
FlxSprite: Fix null camera check on clipToViewBounds.

glowsoony:
I assume this is wrong behavior for the null check?


Geokureli:
lol, whoops. Thanks!

Glad to see people are testing new changes before they are released

PR #3548, opened 2026-01-13
FlxBitmapText: Fix alignment when changing text width

slameron:
CENTER alignment is off for flxbitmaptext after changing the text.

before the change, changing the text resulted in the text to look off-center
see the 'Upgrades' text on the top, this gif is prior to the change
pre-change

change it to call getLineWidth on the line rather than use the pre-stored line widths so the new text can be aligned properly
post-change

it may be better to simply update the `_linesWidth` upon changing the text, but this seems to fix in this case

### DIFF
--- a/flixel/text/FlxBitmapText.hx
+++ b/flixel/text/FlxBitmapText.hx
@@ -1037,7 +1037,7 @@ class FlxBitmapText extends FlxSprite
 		for (i in 0...numLines)
 		{
 			line = _lines[i];
-			lineWidth = _linesWidth[i];
+			lineWidth = getLineWidth(i);
 
 			// LEFT
 			ox = font.minOffsetX;


Geokureli:
Ignore failing CI, I created an issue in hscript, here: https://github.com/HaxeFoundation/hscript/issues/161

flixel/text/FlxBitmapText.hx
```diff
@@ -1037,7 +1037,7 @@ class FlxBitmapText extends FlxSprite
 		for (i in 0...numLines)
 		{
 			line = _lines[i];
-			lineWidth = _linesWidth[i];
+			lineWidth = getLineWidth(i);
```
Geokureli, line 1040: This fix is odd, to me. Should we instead update the _linesWidth cache and continue to use it? or if it's not really useful, should it be removed entirely? If _lineWidth is not updated when text changes, it could affect other systems, or if it is updated later we can cache it earlier

slameron: _linesWidth gets cached in the getter for textWidth, which only is run in computeTextSize when autoSize is on, meaning changing a text with fieldwidth/autosize off doesnt actually update the _linesWidth.

either way the only place _linesWidth is used internally is there so i figure it probably wouldnt hurt to remove it completely and just call getLineWidth.  Not sure if anything could extend FlxBitmapText that uses it though.

Geokureli: Yeah, looks like its still recalculated each time the getter is invoked, so it's not really helping at all. I think some kind of caching would be nice but that is a separate issue

Geokureli:
Thanks! game looks cool, too

PR #3553, opened 2026-02-08
FlxMouse: Fix visibility when starting the game unfocused

### LINKED ISSUE #2237: Mouse is visible after `FlxG.mouse.visible=false` if the game starts unfocused

realvirtu:
Fixes #2237 

This PR fixes an issue with the mouse always being visible when the game starts unfocused. This issue is fixed by setting `_visibleWhenFocusLost` to `visible` upon the game starting.

### DIFF
--- a/flixel/input/mouse/FlxMouse.hx
+++ b/flixel/input/mouse/FlxMouse.hx
@@ -583,6 +583,8 @@ class FlxMouse extends FlxPointer implements IFlxInputManager
 		// Call set_visible with the value visible has been initialized with
 		// (unless set in create() of the initial state)
 		set_visible(visible);
+
+		_visibleWhenFocusLost = visible;
 	}
 
 	/**


TechnikTil (commented):
3 line banger (would approve but the haxeflixel repo doesn’t allow that)

Geokureli:
Thanks!

Geokureli:
> 3 line banger (would approve but the haxeflixel repo doesn’t allow that)

I believe anyone can go to the files tab of a flixel PR and submit a review, lmk. This PR is already merged, so people can only comment, rather than actually reviewing

PR #3554, opened 2026-02-10
FlxAnimationController: fix off-by-one and negative modulo bugs

dtwotwo:
fixes bugs in FixAnimationController

- corrected off-by-one error in add() frame validation
  (while (i-- >= 0) -> while (i-- > 0))
- fixed negative frame index handling in set_frameIndex:
  negative values are now clamped to 0 with a log warning,
  values >= numFrames are wrapped with a log warning

### DIFF
--- a/tests/unit/src/flixel/animation/FlxAnimationControllerTest.hx
+++ b/tests/unit/src/flixel/animation/FlxAnimationControllerTest.hx
@@ -278,6 +278,74 @@ class FlxAnimationControllerTest extends FlxTest
 		Assert.isTrue (sprite.animation.exists("anim4"), 'missing "anim4"');
 	}
 
+	@Test // #3554
+	function testAddAllValidFrames()
+	{
+		var bitmapData = new BitmapData(4, 1);
+		sprite.loadGraphic(bitmapData, true, 1, 1);
+		
+		sprite.animation.add("test", [0, 1, 2, 3]);
+		
+		var anim = sprite.animation.getByName("test");
+		Assert.isNotNull(anim);
+		FlxAssert.arraysEqual([0, 1, 2, 3], anim.frames);
+	}
+	
+	@Test // #3554
+	function testAddSomeInvalidFrames()
+	{
+		var bitmapData = new BitmapData(4, 1);
+		sprite.loadGraphic(bitmapData, true, 1, 1);
+		
+		sprite.animation.add("test", [0, 4, 2, 5, 3]);
+		
+		var anim = sprite.animation.getByName("test");
+		Assert.isNotNull(anim);
+		FlxAssert.arraysEqual([0, 2, 3], anim.frames);
+	}
+	
+	@Test // #3554
+	function testAddSingleValidFrame()
+	{
+		var bitmapData = new BitmapData(4, 1);
+		sprite.loadGraphic(bitmapData, true, 1, 1);
+		
+		sprite.animation.add("test", [0]);
+		
+		var anim = sprite.animation.getByName("test");
+		Assert.isNotNull(anim);
+		FlxAssert.arraysEqual([0], anim.frames);
+	}
+	
+	@Test // #3554
+	function testAddAllInvalidFrames()
+	{
+		loadSpriteSheet();
+		
+		sprite.animation.add("test", [5, 6, 7]);
+		
+		var anim = sprite.animation.getByName("test");
+		Assert.isNull(anim);
+	}
+	
+	@Test // #3554
+	function testFrameIndexNegativeWraps()
+	{
+		#if !js
+		var bitmapData = new BitmapData(4, 1);
+		sprite.loadGraphic(bitmapData, true, 1, 1);
+		
+		sprite.animation.add("test", [0, 1, 2, 3]);
+		sprite.animation.play("test");
+		
+		sprite.animation.frameIndex = -1;
+		Assert.areEqual(3, sprite.animation.frameIndex);
+		
+		sprite.animation.frameIndex = -4;
+		Assert.areEqual(0, sprite.animation.frameIndex);
+		#end
+	}
+
 	function loadSpriteSheet():Void
 	{
 		var bitmapData = new BitmapData(2, 1);
--- a/tests/unit/src/flixel/animation/FlxAnimationControllerTest.hx
+++ b/tests/unit/src/flixel/animation/FlxAnimationControllerTest.hx
@@ -329,9 +329,8 @@ class FlxAnimationControllerTest extends FlxTest
 	}
 	
 	@Test // #3554
-	function testFrameIndexNegativeWraps()
+	function testFrameIndexNegativeClampsToZero()
 	{
-		#if !js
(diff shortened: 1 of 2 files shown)


Geokureli:
For posterity, the off-by-one bug was introduced here: https://github.com/HaxeFlixel/flixel/pull/2762 (D'oh!)

Can you provide examples that would have failed in prior versions? We should add unit tests for them. I'm also not sure I understand how negative frame indices are used in set_frameIndex

dtwotwo:
> For posterity, the off-by-one bug was introduced here: #2762 (D'oh!)
> 
> Can you provide examples that would have failed in prior versions? We should add unit tests for them. I'm also not sure I understand how negative frame indices are used in set_frameIndex

off-by-one in `add()`:

the loop `while (i-- >= 0)` fires an extra iteration on every `add()` call. when `i` is `0`, the condition `0 >= 0` is `true`, `i` becomes `-1`, and the body accesses `framesToAdd[-1]`. on cpp this is an out-of-bounds read that corrupts the array. The same file already uses the correct pattern in `removeInvalidFrames` - `while (i-- > 0)`.

negative modulo in `set_frameIndex`:

`frameIndex` is a public property, and internally `copyFrom` can pass the default value of `-1` into it. in Haxe, `-1 % 4 == -1`, not `3`, so `_sprite.frames.frames[-1]` is an out-of-bounds access. `FlxMath.mod` handles wrapping correctly.

unit tests:

added `testAddAllValidFrames`, `testAddSomeInvalidFrames`, `testAddSingleValidFrame`, `testAddAllInvalidFrames`, `testFrameIndexNegativeWraps` for test possible fallbacks

- CI on old code **without fix** (cpp fails): https://github.com/noegarbage/flixel-prs/actions/runs/21875412703 (tests used wrong sprite sheet size, but failures confirm the same bugs)
- CI on new code **with fix**: https://github.com/noegarbage/flixel-prs/actions/runs/21878869711

tests/unit/src/flixel/animation/FlxAnimationControllerTest.hx
```diff
@@ -278,6 +278,74 @@ class FlxAnimationControllerTest extends FlxTest
...
+		
+		sprite.animation.add("test", [0, 1, 2, 3]);
+		sprite.animation.play("test");
+		
+		sprite.animation.frameIndex = -1;
+		Assert.areEqual(3, sprite.animation.frameIndex);
+		
+		sprite.animation.frameIndex = -4;
+		Assert.areEqual(0, sprite.animation.frameIndex);
+		#end
```
Geokureli: Why !js?

Furthermore, it seems weird that setting `frameIndex` to its default value would actually set it to 3. Granted, prior to this fix it'd be set to 1, which is equally weird. To me, it makes sense to simply clamp negative values

Can you tell me about your specific use case with negative frames that made you want to make this change?

dtwotwo: > Why !js?
> 
> Furthermore, it seems weird that setting `frameIndex` to its default value would actually set it to 3. Granted, prior to this fix it'd be set to 1, which is equally weird.
> 
> Can you tell me about your specific use case with negative frames that made you want to make this change?

`#if !js`: on js targets, array access with a negative index returns `undefined` instead of crashing, and the behavior of `framePixels` differs - existing tests like `testSetFrameIndex` use the same `#if !js` guard for this reason.

about `negative frame indices`: i don't have a specific use case for negative frames. the point of this change isn't to support them - it's to prevent the crash. before this fix, if `set_frameIndex` receives -1 (which is the default field value), the % operator in Haxe preserves the sign: `-1 % 4 == -1`. then `_sprite.frames.frames[-1]` is an out-of-bounds access that crashes on cpp target.

if you'd prefer to clamp instead of wrap, something like:

```haxe
if (Frame < 0) Frame = 0;
```

would also fix the crash. i used `FlxMath.mod` because it felt more consistent, but I'm fine with either approach - the important part is that `-1 % numFrames` no longer produces a negative array index.

Geokureli: > `#if !js`: on js targets, array access with a negative index returns `undefined` instead of crashing, and the behavior of `framePixels` differs - existing tests like `testSetFrameIndex` use the same `#if !js` guard for this reason.

For cases like this I would just run it on all targets, to ensure unified, desired behavior, not just targets that previously failed

> about `negative frame indices`: i don't have a specific use case for negative frames. the point of this change isn't to support them - it's to prevent the crash. 

Honestly, I often get annoyed when tools try to silently continue even when given values that are clearly unintentional. Many times, it's because I'm doing something wrong, and don't find out until something else goes wrong, and if it had just thrown an error instead of "smartly" handling it, it would have saved me a ton of debugging.

That said, this particular setter is already sanitizing the input values, so it should guard against negative numbers, too. Let's use 0 for negative numbers and log warnings for any invalid values

I suggest this, but I'm open to other approaches/wording
```hx
if (Frame < 0)
{
	FlxG.log.warn('frameIndex must be a positive number, got $Frame, using 0, instead');
	Frame = 0;
}
else if (Frame >= numFrames)
{
	FlxG.log.warn('frameIndex must be less than $numFrames, got $Frame, wrapping to ${Frame % numFrames}, instead');
	Frame = Frame % numFrames;
}
```

dtwotwo: > > `#if !js`: on js targets, array access with a negative index returns `undefined` instead of crashing, and the behavior of `framePixels` differs - existing tests like `testSetFrameIndex` use the same `#if !js` guard for this reason.
> 
> For cases like this I would just run it on all targets, to ensure unified, desired behavior, not just targets that previously failed
> 
> > about `negative frame indices`: i don't have a specific use case for negative frames. the point of this change isn't to support them - it's to prevent the crash.
> 
> Honestly, I often get annoyed when tools try to silently continue even when given values that are clearly unintentional. Many times, it's because I'm doing something wrong, and don't find out until something else goes wrong, and if it had just thrown an error instead of "smartly" handling it, it would have saved me a ton of debugging.
> 
> That said, this particular setter is already sanitizing the input values, so it should guard against negative numbers, too. Let's use 0 for negative numbers and log warnings for any invalid values
> 
> I suggest this, but I'm open to other approaches/wording
> 
> ```hx
> if (Frame < 0)
> {
> 	FlxG.log.warn('frameIndex must be a positive number, got $Frame, using 0, instead');
> 	Frame = 0;
> }
> else if (Frame >= numFrames)
> {
> 	FlxG.log.warn('frameIndex must be less than $numFrames, got $Frame, wrapping to ${Frame % numFrames}, instead');
> 	Frame = Frame % numFrames;
> }
> ```

good point, i agree - clamping to 0 with a warning is more helpful than silent wrapping. i applied your suggested approach.

also removed `#if !js` from the test - it was copied from `testSetFrameIndex` which checks pixel values, but this test only checks the integer frameIndex, so it should run on all targets.

Geokureli:
Thanks!

Issue #3581, opened 2026-03-29
Add FlxMatrix pooling


MaybeMaru:
Same as with FlxPoint and FlxRect, FlxMatrix is a very commonly used class in HaxeFlixel for math calculations.
However, strangely, it is the only one of the group that lacks pooling support to reuse instances of FlxMatrix.

Adding the same kind of pooling that FlxPoint and FlxRect have would be extremely useful, both for developers and within HaxeFlixel itself, as it would reduce the number of instances and the memory overhead caused by constantly creating and destroying FlxMatrix objects.

Geokureli:
Global pooling isn't worth it IMHO, all of the [point handling methods with optional result args](https://github.com/search?q=repo%3AHaxeFlixel%2Fflixel%20%22%3Fresult%3AFlxPoint%22&type=code) add a lot of complexity, and bugs due to human error is pretty common (both on flixel's end and the devs utilizing them). They require a few unit tests every time we add one, and we're not good about adding them all. 

My actual goal is to deprecate and remove FlxCallbackPoint (by replacing FlxSpriteGroup), which would allow us to remove point pooling and inline the FlxBasePoint constructor, preventing any actual instantiation of a lot of FlxPoints. For most other cases [raising the minimum supported haxe version to 4.3.7](https://github.com/HaxeFlixel/flixel/issues/2803) will allow us to use a dedicated static local field for every method rather than recycling a new one each call.

Another issue with changing FlxMatrix to be pooled is that nothing is set up to use recycled matrices. So adding pool related fields we're just adding more memory to each matrix. If I were to add something like this, I would need to see benchmark tests proving the performance/memory benefits. My guess is that places that could benefit can likely use something simpler

MaybeMaru:
> Global pooling isn't worth it IMHO, all of the [point handling methods with optional result args](https://github.com/search?q=repo%3AHaxeFlixel%2Fflixel%20%22%3Fresult%3AFlxPoint%22&type=code) add a lot of complexity, and bugs due to human error is pretty common (both on flixel's end and the devs utilizing them). They require a few unit tests every time we add one, and we're not good about adding them all.

Personally I wouldn't consider pooling to be such a big issue for the codebase, or at least in this case in particular. I can understand the problem for the whole "weak pooling" HaxeFlixel implementation since that's a very weird and unintuitive system that most devs and not even HaxeFlixel itself uses, but this isn't exactly what im proposing here.

[...]

PR #3582, opened 2026-03-29
Add FlxMatrix pooling

MaybeMaru:
As mentioned in https://github.com/HaxeFlixel/flixel/issues/3581 this PR implements pooling functionality to FlxMatrix to reduce the number of instances and the memory overhead caused by constantly creating and destroying FlxMatrix objects.

Keeping this as a draft for now since I believe I should implement uses of the pooling inside of the HaxeFlixel codebase first before setting it ready to merge.

### DIFF
--- a/flixel/effects/FlxMatrixSprite.hx
+++ b/flixel/effects/FlxMatrixSprite.hx
@@ -1,5 +1,6 @@
 package flixel.effects;
 
+import flixel.util.FlxDestroyUtil;
 import flixel.FlxSprite;
 import flixel.graphics.frames.FlxFrame;
 import flixel.math.FlxMatrix;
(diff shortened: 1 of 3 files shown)


flixel/effects/FlxMatrixSprite.hx
```diff
@@ -17,15 +18,21 @@ class FlxMatrixSprite extends FlxSprite
 	 */
 	public final renderMatrix:FlxMatrix;
 	
-	public function new (x = 0.0, y = 0.0, ?simpleGraphic)
+	public function new(x:Float = 0.0, y:Float = 0.0, ?simpleGraphic:FlxGraphicAsset)
```
Geokureli, line 21: I haven't changed my mind on this PR, yet, but I wanted to note: please don't add types, here. I keep removing these, as the redundancy adds more visual noise than clarity, IMO. I specifically don't want redundant typing when there are default values and on these core `OneOfMany` asset types where it actually allows multiple types.

PR #3586, opened 2026-04-06
Remove redundant music checks in frontend methods

### LINKED ISSUE #3580: Music doesn't resume after onFocus in FlxSoundFrontEnd
**Haxe version: 4.3.7
Flixel version: _6.2.0 (dev)_
OpenFL version: 9.5.1
Lime version: 8.3.1
Affected targets: Windows (and other platforms)**

**Observed behavior:**
When using FlxG.sound.music together with FlxG.sound.list of sounds:

If game loses focus, FlxG.sound.onFocusLost() correctly pauses all sounds
When focus is regained (onFocus()), sounds in list resume, but music doesn't

It seems that music is handled separately from list but shares most of the same lifecycle methods. This can lead to inconsistent behavior:

music may not resume after pausing
music may be destroyed unexpectedly when list is cleaned up
music and list logic is duplicated in multiple places (pause(), resume(), onFocusLost(), onFocus())

**Expected behavior:**
FlxG.sound.music should behave consistently with sounds in list resume() after onFocus() should resume music as well as other sounds

Geokureli:
Fixes #3580 

Introduced in #3558

Now that music is added to `FlxG.sound.list` (an unintended change, but not a bad thing) `FlxG.sound.onFocusLost` calls music.onFocusLost twice. the second would set _resumeOnFocus to false since it was already paused

I'm not sure why music was not added to the list, I'll check for any other unintended side effects of this change before merging. That may have been be a breaking change, as if anyone checked the if the list contains the music, it would have been false, but is now true, but that seems like an odd thing to do.

~~I don't think I can unit test this, but i'll look and see~~ Added tests

### DIFF
--- a/flixel/system/frontEnds/SoundFrontEnd.hx
+++ b/flixel/system/frontEnds/SoundFrontEnd.hx
@@ -413,11 +413,6 @@ class SoundFrontEnd
 	 */
 	public function pause():Void
 	{
-		if (music != null && music.exists && music.active)
-		{
-			music.pause();
-		}
-
 		for (sound in list.members)
 		{
 			if (sound != null && sound.exists && sound.active)
@@ -432,11 +427,6 @@ class SoundFrontEnd
 	 */
 	public function resume():Void
 	{
-		if (music != null && music.exists)
-		{
-			music.resume();
-		}
-
 		for (sound in list.members)
 		{
 			if (sound != null && sound.exists)
@@ -453,12 +443,6 @@ class SoundFrontEnd
 	 */
 	public function destroy(forceDestroy = false):Void
 	{
-		if (music != null && (forceDestroy || !music.persist))
-		{
-			music.destroy();
-			music = null;
-		}
-
 		for (sound in list.members)
 		{
 			if (sound != null && (forceDestroy || !sound.persist))
@@ -554,9 +538,6 @@ class SoundFrontEnd
 	@:allow(flixel.FlxGame)
 	function update(elapsed:Float):Void
 	{
-		if (music != null && music.active)
-			music.update(elapsed);
-
 		if (list != null && list.active)
 			list.update(elapsed);
 
@@ -576,11 +557,6 @@ class SoundFrontEnd
 	@:allow(flixel.FlxGame)
 	function onFocusLost():Void
 	{
-		if (music != null)
-		{
-			music.onFocusLost();
-		}
-
 		for (sound in list.members)
 		{
 			if (sound != null)
@@ -593,11 +569,6 @@ class SoundFrontEnd
 	@:allow(flixel.FlxGame)
 	function onFocus():Void
 	{
-		if (music != null)
-		{
-			music.onFocus();
-		}
-
 		for (sound in list.members)
 		{
 			if (sound != null)
--- a/tests/unit/src/flixel/system/frontEnds/SoundFrontEndTest.hx
+++ b/tests/unit/src/flixel/system/frontEnds/SoundFrontEndTest.hx
@@ -98,7 +98,6 @@ class SoundFrontEndTest
 		assertSoundProps(1.0, false, false);
 		#end
 	}
-	#end
 	
 	@Test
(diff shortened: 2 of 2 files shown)


ACrazyTown:
@Geokureli This is a breaking change. It breaks any project that manually replaces `FlxG.sound.music`, most notably Funkin'.

From what I've gathered, `FlxG.sound.music` used to be updated independently of the sound list. If you wanted to replace it with a custom sound you could simply reassign it (and remove it from the sound list, if created via `FlxG.sound`). ([example](https://github.com/FunkinCrew/Funkin/blob/main/source/funkin/audio/FunkinSound.hx#L387-L397))

Since it's no longer updated independently, projects that do the aforementioned now have music that doesn't update, which is especially problematic in FNF's case as the sound time no longer updates, and therefore the gameplay gets stuck

Geokureli:
Setting music directly is actually fairly common, I'm surprised I didn't think of that. I may revert this, but first I'll try a new PR that fixes the original issue without adding music to the list, I suppose

SimonArs:
> @Geokureli This is a breaking change. It breaks any project that manually replaces `FlxG.sound.music`, most notably Funkin'.
> 
> From what I've gathered, `FlxG.sound.music` used to be updated independently of the sound list. If you wanted to replace it with a custom sound you could simply reassign it (and remove it from the sound list, if created via `FlxG.sound`). ([example](https://github.com/FunkinCrew/Funkin/blob/main/source/funkin/audio/FunkinSound.hx#L387-L397))
> 
> Since it's no longer updated independently, projects that do the aforementioned now have music that doesn't update, which is especially problematic in FNF's case as the sound time no longer updates, and therefore the gameplay gets stuck

Honestly didn't notice any problems with this. At most, music stopped playing after switching state, but this was solved simply by doing `music.persist` to true

Geokureli:
> Honestly didn't notice any problems with this. At most, music stopped playing after switching state, but this was solved simply by doing `music.persist` to true

Can you share your code for playing music?

SimonArs:
> > Honestly didn't notice any problems with this. At most, music stopped playing after switching state, but this was solved simply by doing `music.persist` to true
> 
> Can you share your code for playing music?

Sure. I use wrapper around FlxG.sound.playMusic():

```haxe
public static function mysic(?options:SoundOptions, ?n:String):Null<FlxSound> {
	if (n != null)
		musico = n;

	final o:SoundOptions = options;

	final lib:String = o?.library ?? 'sodaluv';
	final m:Sound = Paths.sound(lib == 'sodaluv' ? 'music/$musico' : musico, lib);
	if (m == null)
		return null;

	final msc:FlxSound = FlxG.sound.playMusic(m, o?.group, o?.volume ?? Fei.prf.musicVol, o?.loop ?? true, o?.onComplete);
	msc.persist = true; //fixes mentioned issue

	if (o?.time != null)
		msc.time = o.time;

	final prox:ProxData = o?.proximity;
	if (prox != null) {
		msc.proximity(prox.x, prox.y, prox.target, prox.radius, prox.pan ?? true);

		if (prox.pitch != null)
			msc.pitch = prox.pitch;
	}

	return msc;
}
```

Geokureli:
> > Can you share your code for playing music?
> 
> Sure. I use wrapper around FlxG.sound.playMusic():

Thanks! It sounds like you're saying that, before this change, playMusic would always persist across state switches, and now it no longer does unless you specifically make it persist, is that correct?

If so, I don't see why that would have changed, and I should look into it.

Edit: I looked into it, it used to set persist to true, so I added that back. Thanks, again!

SimonArs:
> > > Can you share your code for playing music?
> > 
> > Sure. I use wrapper around FlxG.sound.playMusic():
> 
> Thanks! It sounds like you're saying that, before this change, playMusic would always persist across state switches, and now it no longer does unless you specifically make it persist, is that correct?
> 
> If so, I don't see why that would have changed, and I should look into it.
> 
> Edit: I looked into it, it used to set persist to true, so I added that back. Thanks, again!

Yes, you got it right. I was glad to help with it!

PR #3593, opened 2026-04-18
FlxBitmapText and FlxText: Add OUTLINE_CARDINAL

TracedInPurple:
Works like the fast outline method, but uses cardinal directions instead of diagonals. This generally looks better for pixelated fonts.

You could also treat this as a variant of `OUTLINE_FAST`. Renaming `OUTLINE_FAST` to `OUTLINE_DIAGONAL` might make things clearer, but it could break existing projects, so it’s probably not worth changing.

# OUTLINE_FAST (diagonal)
image_2026-04-18_182950061

# OUTLINE_CARDINAL
image

# OUTLINE (the normal one)
image

### DIFF
--- a/flixel/text/FlxBitmapText.hx
+++ b/flixel/text/FlxBitmapText.hx
@@ -1291,6 +1291,20 @@ class FlxBitmapText extends FlxSprite
 					func( 0,  i); // lower-middle
 					func( i,  i); // lower-right
 				}
+			case OUTLINE_CARDINAL:
+				// Render an outline around the text (4 draws | LEFT, RIGHT, UP, DOWN)
+				var iterations:Int = Std.int(borderSize * borderQuality);
+				iterations = (iterations <= 0) ? 1 : iterations;
+				final delta = Std.int(borderSize / iterations);
+				for (iter in 0...iterations)
+				{
+					final i = delta * (iter + 1);
+					func(-i, 0); // middle-left
+					func(i * 2, 0); // middle-right
+					func(-i, -i); // upper-middle
+					func(0, i * 2); // lower-middle
+				}
+
 			case OUTLINE_FAST:
 				// Render an outline around the text in each corner (4 draws)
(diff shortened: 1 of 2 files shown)


EkicalPog:
God this is so fucking hot.

toffeecaramel:
WE NEED THIS!!!!!

Jaduplex (commented):
God yes.

flixel/text/FlxBitmapText.hx
```diff
@@ -1291,6 +1291,20 @@ class FlxBitmapText extends FlxSprite
...
+				var iterations:Int = Std.int(borderSize * borderQuality);
+				iterations = (iterations <= 0) ? 1 : iterations;
+				final delta = Std.int(borderSize / iterations);
+				for (iter in 0...iterations)
+				{
+					final i = delta * (iter + 1);
+					func(-i, 0); // middle-left
+					func(i * 2, 0); // middle-right
+					func(-i, -i); // upper-middle
+					func(0, i * 2); // lower-middle
```
Geokureli: Still need to test this locally, but I'm surprised by i*2 and upper middle being -i, -i. Looking at OUTLINE_FAST I would have expected the following

```suggestion
					final i = delta * (iter + 1);
					func(-i,  0); // left
					func( i,  0); // right
					func( 0, -i); // up
					func( 0,  i); // down
```

TracedInPurple: That's definitely a mistake on my end
i did not realize that FlxBitmapText seems to absolute instead of incremental like FlxText

Geokureli (changes requested):
Fix issues listed above and there will need to be a PR made to flixel-ui to prevent CI failing

Geokureli:
Moving this to 6.3.0 since CI is still failing

Edit: Oh it's failing because of flixel-ui. I'll take a look

Geokureli (approved):

Geokureli:
Thanks!

PR #3605, opened 2026-05-31
FlxG.cameras: Honor defaultDrawTarget arg in insert method

glowsoony:
When inserting at a position higher or equal to the list length, it never included the argument for add.


Geokureli:
Thanks!

PR #3609, opened 2026-06-04
FlxFrame: replace implicit 'to' cast with @:to function for Haxe 4.3 compatibility

HomuHomu833:
haxe 4.3 introduced stricter validation for abstract `from/to` declarations, requiring us to do this change, the implicit `to flixel.util.FlxPool.IFlxPooled` on FlxUVRect fails this check since FlxRect's underlying type isn't directly IFlxPooled.

```bash
 ERROR  C:/Users/Homura/GitHub/FNF-Shadow-Engine/.haxelib/flixel/git/flixel/graphics/frames/FlxFrame.hx:764: characters 45-75

 764 | abstract FlxUVRect(FlxRect) from FlxRect to flixel.util.FlxPool.IFlxPooled
     |                                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     | You can only declare from/to with compatible types
```

### DIFF
--- a/flixel/graphics/frames/FlxFrame.hx
+++ b/flixel/graphics/frames/FlxFrame.hx
@@ -842,8 +842,13 @@ enum abstract FlxFrameAngle(Int) from Int to Int
  * `bottom`. This is for optimization reasons, to reduce arithmetic when drawing vertices
  */
 @:forward(put)
-abstract FlxUVRect(FlxRect) from FlxRect to flixel.util.FlxPool.IFlxPooled
+abstract FlxUVRect(FlxRect) from FlxRect
 {
+	@:to public inline function toIFlxPooled():flixel.util.FlxPool.IFlxPooled
+	{
+		return this;
+	}
+
 	public var left(get, set):Float;
 	inline function get_left():Float { return this.x; }
 	inline function set_left(value):Float { return this.x = value; }


Geokureli:
I'm using 4.3 and I've never seen this issue, our 4.3 CI does't fail either. I see you're using FNF-Shadow-Engine, a lot of those fnf libs shadow classes causing issues like this, can you verify that this issue happens on a flixel project that does not use this shadow engine? If so, please let us know your build command and haxelib versions for Flixel, Openfl and Lime

Edit: 
> FlxUVRect fails this check since FlxRect's underlying type isn't directly IFlxPooled

Yes it is: https://github.com/HaxeFlixel/flixel/blob/dev/flixel/math/FlxRect.hx#L10

HomuHomu833:
> I'm using 4.3 and I've never seen this issue, our 4.3 CI does't fail either. I see you're using FNF-Shadow-Engine, a lot of those fnf libs shadow classes causing issues like this, can you verify that this issue happens on a flixel project that does not use this shadow engine?

it can be reproduced with [hscript-improved](https://github.com/CodenameCrew/hscript-improved/tree/codename-dev)'s custom classes support; include this haxelib, enable `CUSTOM_CLASSES` (or `ALLOWED_ABSTRACT_AND_ENUM` ?) and include [flixel in hscript-improved's Config.hx file](https://github.com/ShadowEngineTeam/FNF-Shadow-Engine/blob/bf6aff2e68e0ae5a62e2d74ca528e931da8037a8/source/library_shadows/compatibility/hscript/Config.hx#L6) and this exact issue starts to come.

Geokureli:
I'm not against changes to support these libs, but how is this not an issue with these libs? if FlxRect is a IFlxPooled, why do these libs need special logic? I fear this issue will pop up again and again.

Can you clarify the issue, I'm not sure why this is failing, and you mention that something changed in 4.3, what exactly changed? does that change need to be revisited?

Geokureli:
> it can be reproduced with [hscript-improved](https://github.com/CodenameCrew/hscript-improved/tree/codename-dev)'s custom classes support; include this haxelib, enable `CUSTOM_CLASSES` and include [flixel folder in hscript-improved's Config.hx file](https://github.com/ShadowEngineTeam/FNF-Shadow-Engine/blob/bf6aff2e68e0ae5a62e2d74ca528e931da8037a8/source/library_shadows/compatibility/hscript/Config.hx#L6) and this exact issue starts to come.

Also are we sure this issue shows up in a new project that uses hscrpt-improved the way you outline? are you sure this isn't partly due to a shadowed class in the shadow engine?

HomuHomu833:
> Also are we sure this issue shows up in a new project that uses hscrpt-improved the way you outline? are you sure this isn't partly due to a shadowed class in the shadow engine?

this issue do pop up outside SE (Shadow Engine) and with flixel git I've tested.

> I'm not against changes to support these libs, but how is this not an issue with these libs?

I can't contact anybody in CodenameCrew to discuss this and they blocked me on GitHub.

HomuHomu833:
I'll probably just close this PR as I told here CodenameCrew blocked me long ago and I have no contact with them and [maybe their abstract macro is broken?](https://github.com/CodenameCrew/hscript-improved/blob/codename-dev/hscript/macros/AbstractHandler.hx), I'm really unsure. I'm just sorry for bothering.

Geokureli:
If this PR is the only course of action you have, then let's just do it. especially if it happens without shadow engine

Geokureli:
I created an issue in hscript-improved that (I assume) you can track, despite not being able to post, maybe they will reconsider the block, as you seem like a nice, rational person

Geokureli:
Added some comments to the new method, and made it private
https://github.com/HaxeFlixel/flixel/pull/3610

PR #3616, opened 2026-06-17
Fix clipToView* methods for spriteContainers

Geokureli:
Calling clipToViewRect on a spriteContainer would cause a crash, this prevents it. Eventually we may override this method in SpriteGroup to avoid animation and may just call clipToViewRect on all members rather than going through clipRectTransform


Geokureli:
this doesn't work, go with the other solution, which works in Renaine's spritecont extension

Geokureli:
Closing for #3629

PR #3618, opened 2026-06-22
FlxSpriteUtil: add putWeak to the points in drawPolygon

CodeAnomaly08:
very simple change, was using drawPolygon with weak points, and realized it wasn't using putWeak on it to automatically put those weak points, figured it probably should be doing that.


Geokureli:
Thanks!

Issue #3611, opened 2026-06-07
Add centering on object


SimonArs:
Simple helper function that centers one object on another object
Based on the principle of working as `screenCenter`
I suggest adding it to `FlxObject`

```haxe
public inline function centerOnObject(obj:FlxObject, axes:FlxAxes = XY):FlxObject {
	switch (axes) {
		case X: x = obj.x + (obj.width - width) / 2;
		case Y: y = obj.y + (obj.height - height) / 2;
		case XY: setPosition(obj.x + (obj.width - width) / 2, obj.y + (obj.height - height) / 2);
		case NONE:
	}
	return this;
}
```

Geokureli:
I've got plans for a broader alignment system rather than adding a trillion little methods like this, I likely won't take PRs for one-offs like this as it's easy to have these in your own engine via [static extension](https://haxe.org/manual/lf-static-extension.html)

SimonArs:
> I've got plans for a broader alignment system rather than adding a trillion little methods like this, I likely won't take PRs for one-offs like this as it's easy to have these in your own engine via [static extension](https://haxe.org/manual/lf-static-extension.html)

Yeah it makes sense

PR #3625, opened 2026-07-11
Alignment

SimonArs:
Adds `FlxObject.alignTo()` utility method for aligning `FlxObject` relative to another `FlxObject` or game window

Supports horizontal (`LEFT`, `CENTER`, `RIGHT`) and vertical (`TOP`, `CENTER`, `BOTTOM`) alignment independently, with optional offsets for both axes. If no target object is provided, alignment is performed relative to game window (`FlxG.width` / `FlxG.height`)

This provides more convenient alternative to manually calculating positions every time object needs to be aligned, reducing boilerplate and improving readability

Continuation of idea from https://github.com/HaxeFlixel/flixel/issues/3611

### DIFF
--- a/flixel/FlxObject.hx
+++ b/flixel/FlxObject.hx
@@ -1241,6 +1241,7 @@ class FlxObject extends FlxBasic
 	 * @param   axes   On what axes to center the object (e.g. `X`, `Y`, `XY`) - default is both. 
 	 * @return  This FlxObject for chaining
 	 */
+	@:deprecated("screenCenter is deprecated, use alignTo(CENTER, CENTER) or alignTo(null, CENTER) or alignTo(CENTER)")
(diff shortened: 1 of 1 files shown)


Geokureli:
This still really isn't what I had in mind. If you want to use this feature in your games or release this as a tool for other I recommend doing so with static extension in a public library. I likely won't be taking PRs that add alignment methods directly to FlxObject.

I linked to static extension in 3611, it's a great way to add features to foundational classes like FlxObject without making it hard to back-track or redesign features down the road. I think if we added this particular implementation, now, it would definitely fit some people's needs but not all and it would be hard to expand it later, or replace it.

SimonArs:
Thanks for clarification, yes, that makes sense. I hadn't considered API stability aspect for FlxObject. I'll close PR then

PR #3629, opened 2026-07-13
FlxSpriteGroup: Call clipToViewBounds on members

Geokureli:
- Fixes a crash when `clipToViewRect`, `clipToViewBounds`, `clipToScreenRect` and `clipToScreenBounds` are called on sprite containers and groups.
- Calls the above methods on all members rather than setting clipRect and having that propegate


Geokureli:
I would have released this in 6.2 if I remembered that clipToViewBounds was added in 6.2...

PR #3636, opened 2026-07-16
Revert "Remove redundant music checks in frontend methods"

Geokureli:
Reverts HaxeFlixel/flixel#3586

> This is a breaking change. It breaks any project that manually replaces `FlxG.sound.music`, most notably Funkin'.
> 
> From what I've gathered, `FlxG.sound.music` used to be updated independently of the sound list. If you wanted to replace it with a custom sound you could simply reassign it (and remove it from the sound list, if created via `FlxG.sound`). ([example](https://github.com/FunkinCrew/Funkin/blob/main/source/funkin/audio/FunkinSound.hx#L387-L397))
> 
> Since it's no longer updated independently, projects that do the aforementioned now have music that doesn't update, which is especially problematic in FNF's case as the sound time no longer updates, and therefore the gameplay gets stuck


moxie-coder:
I'm guessing this broke something?

ACrazyTown:
> I'm guessing this broke something?

https://github.com/HaxeFlixel/flixel/pull/3586#issuecomment-4982179670

PR #3637, opened 2026-07-16
FlxTileFrames: Cleanup points

### LINKED ISSUE #3607: Use FlxPoint.weak() instead of FlxPoint.get() for frame size points
While I was exploring places in my project where I have point leak I found this place in **FlxSprite loadGraphic()**:
```
                if (animated)
                        frames = FlxTileFrames.fromGraphic(graph, FlxPoint.get(frameWidth, frameHeight));
                else
                        frames = graph.imageFrame;
```
Imo this is the perfect place to use a **weak point**. Moreover, point isnt put anywhere after using

Upd: the same leak in **loadRotatedGraphic()**

Geokureli:
Fixes #3607

Added overloads with ints instead of points, to use weak points. Eventually we should ditch points wherever possible, since we have to round them whenever use them

### DIFF
--- a/flixel/FlxCamera.hx
+++ b/flixel/FlxCamera.hx
@@ -1093,6 +1093,11 @@ class FlxCamera extends FlxBasic
 		scroll = FlxDestroyUtil.put(scroll);
 		targetOffset = FlxDestroyUtil.put(targetOffset);
 		deadzone = FlxDestroyUtil.put(deadzone);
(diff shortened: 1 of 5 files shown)


Geokureli:
Moving to 6.3.0. seems safe and innocuous but I'm about to release, and these are always the things that end up getting rushed and causing problems

PR #3642, opened 2026-07-29
add `_heldKeyList` to FlxKeyManager to prevent unexpected key presses

### LINKED ISSUE #3641: FlxG.keys.justPressed returns true at odd times
With the following example:
```haxe
package;

import flixel.FlxG;
import flixel.util.FlxColor;
import flixel.FlxState;
import flixel.FlxSubState;

{
	{
		super.create();
	}

	{
		super.update(elapsed);

		if (FlxG.keys.justPressed.ESCAPE)
			openSubState(new RedPlace());
	}
}


class RedPlace extends FlxSubState
{
	public function new():Void
	{
		super(FlxColor.RED);
	}

	{
		super.update(elapsed);

		if (FlxG.keys.justPressed.ESCAPE)
			close();
	}
}
```
Pressing escape in PlayState, and then holding it, will briefly open RedPlace, and then close it shortly after, but this doesn't make much sense, because the condition to close RedPlace is using justPressed, not pressed

From what I can tell, what is happening is that, the keyboard inputs are all forced to be released, (through FlxG.inputs.onStateSwitch) when the sub state opens. This flips justPressed to false, which is expected. However, because Flixel relies on [OpenFL keyboard events](https://github.com/HaxeFlixel/flixel/blob/a8ef79f7667f2b1da4657357b7974d34264fd2e3/flixel/input/FlxKeyManager.hx#L324) to flip these bools, justPressed unexpectedly gets set back to true in this case, due to the OpenFL events getting triggered by keyboard repeat

This was an issue I found out while working on a pr for FlxControls, I'm not currently sure of any solutions or if this is an issue with other input devices

DetectiveBaldi:
Fixes #3641, open to better ways of doing this

### DIFF
--- a/flixel/input/FlxKeyManager.hx
+++ b/flixel/input/FlxKeyManager.hx
@@ -58,6 +58,11 @@ class FlxKeyManager<Key:Int, KeyList:FlxBaseKeyList> implements IFlxInputManager
 	var _keyListMap:Map<Int, FlxInput<Key>> = new Map<Int, FlxInput<Key>>();
 
 	/**
+	 * Interal storage of held input keys, populated by OpenFl events.
+	 */
+	var _heldKeyList:Array<Bool> = [];
+
+	/**
 	 * Check to see if at least one key from an array of keys is pressed.
 	 *
 	 * @param	KeyArray 	An array of key names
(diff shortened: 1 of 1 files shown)


MTUnixic:
bump

Geokureli:
> bump

Please don't post here if you have nothing to add, I haven't forgotten about this. I'm not 100% satisfied with this solution and I'll get to it when I have time

PR #3644, opened 2026-08-05
FlxVersionMacro: Fix build error by null checking some values 

ACrazyTown:
In some rare cases I can't seem to reproduce (but people have experienced, see [relevant convo in Haxe discord](https://discord.com/channels/162395145352904705/165234904815239168/1534580726287044759)) summoning a process would fail which would in turn cause a build error as FlxVersionMacro does string operations without null-checking them first.

Very simple fix is to just null check the process outputs before doing anything with them

### DIFF
--- a/flixel/system/macros/FlxVersionMacro.hx
+++ b/flixel/system/macros/FlxVersionMacro.hx
@@ -76,7 +76,9 @@ class FlxVersionMacro
 				final libPath = getLibraryPath(id);
 				final sha = getGitSHA(libPath);
 				final branchRaw = getGitBranch(libPath);
-				final defaultBranch = getGitDefaultBranch(libPath).split("/").pop();
+				var defaultBranch = getGitDefaultBranch(libPath);
+				if (defaultBranch != null)
+					defaultBranch = defaultBranch.split("/").pop();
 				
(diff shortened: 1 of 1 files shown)


flixel/system/macros/FlxVersionMacro.hx
```diff
@@ -76,7 +76,9 @@ class FlxVersionMacro
 				final libPath = getLibraryPath(id);
 				final sha = getGitSHA(libPath);
 				final branchRaw = getGitBranch(libPath);
-				final defaultBranch = getGitDefaultBranch(libPath).split("/").pop();
+				var defaultBranch = getGitDefaultBranch(libPath);
+				if (defaultBranch != null)
+					defaultBranch = defaultBranch.split("/").pop();
 				
 				final branch = branchRaw == defaultBranch ? null : branchRaw;
```
ACrazyTown, line 83: Also, an interesting thing to note is that these branch values seem to be completely unused.

Geokureli: lol, just checked this, not sure how I missed that

Geokureli:
This seems like a good candidate for null-safety. I remembered to put Null<T> for these methods but didn't null check the result

